PR #10, opened 2017-11-17
Creating terra-navigation-layout

tbiethman:
### Summary
Creating terra-navigation-layout, a package that includes a variety of components and utilities to provide configuration-based, `react-router`-driven navigation to Terra applications.

While the components themselves aren't all that complicated, the overall configuration and whatnot can be. I have updated the site with quite a bit of information and a working example, so please take a look at that and let me know where there is still confusion.

Resolves: #10

https://terra-framework-deployed-pr-10.herokuapp.com/static/#/components/navigation-layout

Dave's going to graciously work on the tests, as they're pretty light right now.


packages/terra-framework-site/src/app/site.scss
```diff
@@ -0,0 +1,91 @@
+// Styles specific to development only
+// stylelint-disable
+:local {
+  .app {
+    height: 100%;
+    position: relative;
+    width: 100%;
+  }
+
+  // .site-nav {
```
emilyrohrbough: Oops. Looks like this got missed in clean up.

packages/terra-framework-site/src/examples/navigation-layout/NavigationLayoutStandard.jsx
```diff
@@ -0,0 +1,126 @@
...
+      component: {
+        tiny: {
+          componentClass: MenuExample,
+        },
+        small: {
+          componentClass: MenuExample,
+        },
+      },
+      children: {
+        '/page1': {
```
mjhenkes: could we be more descriptive with the page names?

mjhenkes: Nm, this is just the example

packages/terra-framework-site/src/examples/navigation-layout/HeaderExample.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Link, withRouter } from 'react-router-dom';
+
+const propTypes = {
+  layoutConfig: PropTypes.object,
+};
+
+const HeaderExample = ({
+  layoutConfig,
```
JakeLaCombe: Nitpick:

If your hashes are small, I find it more readable if it's just one line.

packages/terra-framework-site/src/app/App.scss
```diff
@@ -9,3 +9,7 @@ body {
   position: relative;
   width: 100%;
 }
+
+.base {
```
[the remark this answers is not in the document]
bjankord: This may need to be a boolean prop. Some apps will need this at 100% and others won't (I believe it causes issues for embedded iframes).

packages/terra-navigation-layout/docs/README.md
```diff
@@ -0,0 +1,176 @@
...
+import { BrowserRouter } from 'react-router-dom';
+import NavigationLayout from 'terra-navigation-layout';
+
+const config = {
+  // Configuration should feature routing definitions for header, menu, and
+  // content regions.
+};
+
+const ExampleApp = () => (
+  <BrowserRouter>
```
bjankord: Are we using the BrowserRouter or the HashRouter?

packages/terra-navigation-layout/src/NavigationLayout.jsx
```diff
@@ -0,0 +1,172 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {
+  withRouter,
+} from 'react-router-dom';
+
+import AppDelegate from 'terra-app-delegate';
+import Layout from 'terra-layout';
+import breakpoints from 'terra-responsive-element/lib/breakpoints.scss';
```
bjankord: We should find a better way to share breakpoints, I'm fine with this for now.

packages/terra-framework-site/src/app/components/Components.jsx
```diff
@@ -1,14 +1,35 @@
...
+    if (example) {
+      return (
+        <Route key={example.path} path={example.path} component={example.component} />
+      );
+    }
+    return undefined;
+  })
+);
+
+const generateRedirect = (config) => {
```
SpartaSixZero: On line 25, we use parenthesis but on line 13, we don't use parenthesis around the single parameter. Do we have a styling rule regarding this? Consistent styling throughout is important.

packages/terra-navigation-layout/docs/README.md
tbiethman: The site uses the HashRouter, and the example embedded in the site uses the MemoryRouter, though it shouldn't really matter. The NavigationLayout component doesn't actually create a Router itself (it must be wrapped in one, as stated in the README), allowing us to defer to whatever type of router consumers want to use.

packages/terra-navigation-layout/src/NavigationLayout.jsx
tbiethman: Doing so has been on my TODO this for a while; hopefully I can get around to it soon. This is becoming (even more) unsustainable.

packages/terra-framework-site/src/app/components/Components.jsx
tbiethman: I feel you. This does seem like an arbitrary difference, but these styles around arrow functions are built into our lint rules: https://github.com/airbnb/javascript#arrows--implicit-return

Basically, if an arrow function has a single parameter and the function body is a single returned expression (and therefore wrapped in '()' vs. '{}'), they prefer the parentheses around the parameter to be removed. If there's more than one parameter, the parentheses come back. If the function body contains more than a explicit return statement, parentheses come back, regardless of the parameter count. 

:man_shrugging:

packages/terra-navigation-layout/docs/README.md
```diff
@@ -0,0 +1,176 @@
...
+<Switch>
+  <Route path="/a/b" component={BComponent} />
+  <Route path="/a" component={AComponent} />
+  <Route path="/" component={RootComponent} />
+</Switch>
+```
+
+Because the `Switch` renders the first matching route and ignores the rest, if the configuration was not inverted within it, the path `/a/b` would match against `/` and `RootComponent` would be rendered for every route.
+
+#### Configuration API
```
JakeLaCombe: I think this guide would be a little easier to read if we put the configuration api before the RoutingStack. I get lost as I see the configuration before I know what it is.

packages/terra-navigation-layout/docs/README.md
JakeLaCombe: Looks like you can use either. https://github.com/cerner/terra-framework/pull/10/files#diff-62f42bd66d8165f6c16db68f97afcf80R13

packages/terra-navigation-layout/src/NavigationLayout.jsx
```diff
@@ -0,0 +1,172 @@
...
+        menu={this.decorateElement(menuComponent, config.menu)}
+        menuText={menuText}
+      >
+        {this.decorateElement(contentComponent, config.content)}
+      </Layout>
+    );
+  }
+}
+
+NavigationLayout.propTypes = propTypes;
```
JakeLaCombe: No need for defaults?

packages/terra-navigation-layout/src/NavigationLayoutContent.jsx
```diff
@@ -0,0 +1,55 @@
...
+   */
+  routeConfig: routeConfigPropType,
+  /**
+   * The current responsive size.
+   */
+  navigationLayoutSize: PropTypes.oneOf(supportedComponentBreakpoints),
+  /**
+   * The String path to redirect to should the routes specified by the routeConfig all fail to match.
+   */
+  redirect: PropTypes.string,
```
JakeLaCombe: most API's (mostly the authentication ones) I see usually have this as redirectURI, and I think that sounds better as well, as redirect sounds more like a verb than a noun

packages/terra-navigation-layout/src/NavigationLayoutContent.jsx
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * Flag to enable navigation within the RoutingStack.
+   */
+  stackNavigationIsEnabled: PropTypes.bool,
```
JakeLaCombe: Most of our other props start the is at the beginning (isStackNavigationEnabled). I think this reads easier, as the is immediately indicates the prop is a boolean

packages/terra-navigation-layout/src/configurationPropTypes.js
```diff
@@ -0,0 +1,107 @@
...
+ *       // The same component can be rendered with different prop values at different breakpoints
+ *       componentClass: DefaultComponent,
+ *       props: {
+ *         prop1: 'Huge Value for prop1',
+ *       },
+ *     },
+ *   }
+ */
+const componentConfigPropType = PropTypes.objectOf((propValue, key, componentName, location, propFullName) => {
+  const keyIsValid = key === 'default' || supportedComponentBreakpoints.indexOf(key) >= 0;
```
JakeLaCombe: Nitpick:

validKey and validInput feel like variable names that would make this code easier to read IMO.

packages/terra-navigation-layout/src/routing/RoutingStackDelegate.js
```diff
@@ -0,0 +1,70 @@
...
+  return create(mergedData);
+};
+
+/**
+ * Performs a shallow equality on the supportedAttributes to determine the equality of two
+ * RoutingStackDelegate Object instances.
+ */
+const isEqual = (delegateA, delegateB) => {
+  if (!delegateA || !delegateB) {
+    return false;
```
JakeLaCombe: If both are null, empty, or undefined, wouldn't they technically be equal?

JakeLaCombe:
I noticed there were not any nightwatch tests. Would it be worth it to write one to verify that the component setups the navigation correctly by clicking on links and verify that you get routed correctly?

packages/terra-navigation-layout/docs/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+The configuration object is structured like:
+```js
+{
+  // The routes specified under the `header` key will be generated in the Layout's `header` region.
```
yuderekyu: Just curious, any reason why we're using single line comments here vs block comments?

tbiethman:
Since some concerns were raised with the nested routing config structure, I am going to flatten it out. This will make the configuration API easier to consume, but it will also complicate the processing internally. I'll try and make the changes as minimal as possible, but until those changes are made this should not be merged.

packages/terra-framework-site/src/app/site.scss
tbiethman: Whoops, I'll remove this.

packages/terra-framework-site/src/examples/navigation-layout/HeaderExample.jsx
tbiethman: I agree, I'll update this.

packages/terra-navigation-layout/src/NavigationLayout.jsx
tbiethman: For most, the undefined value of unprovided props is all the default we want. The config prop is marked as required, as that's really the only one that doesn't make sense to ever omit. I can mark the router-provided props as required, though that doesn't change much from a consumer standpoint.

packages/terra-navigation-layout/src/NavigationLayoutContent.jsx
tbiethman: I agree, I'll update this to `redirectPath` to keep the language consistent with other components.

packages/terra-navigation-layout/src/NavigationLayoutContent.jsx
tbiethman: Yeah this is something I go back and forth on. A lot of times when we prefix a prop with `is`, we are describing the Component itself:
<Menu isOpen>
<Input isActive>
<Component is{Adjective}>

In this case, the prop is describing an aspect of the component:
<Component {Aspect}Is{Adjective}>

I feel like this is still in line with the previous usage. If we switch it up to isStackNavigationEnabled, I feel like the prop starts feeling like a question instead of a statement. It feels a little off, but I understand the consistency concerns. Are there better ways to name this prop?

packages/terra-navigation-layout/src/configurationPropTypes.js
tbiethman: I agree, I will update this.

packages/terra-navigation-layout/src/routing/RoutingStackDelegate.js
tbiethman: You're right. I placed the direct equality check ahead of this one to cover that use case.

packages/terra-navigation-layout/docs/README.md
tbiethman: Nah, I can update those if it'll read easier.

packages/terra-framework-site/src/examples/navigation-layout/HeaderExample.jsx
```diff
@@ -0,0 +1,22 @@
...
+const propTypes = {
+  layoutConfig: PropTypes.object,
+};
+
+const HeaderExample = ({ layoutConfig }) => (
+  <div style={{ height: '60px', width: 'calc(100% - 10px)', border: '4px dashed lightgrey', margin: '5px', position: 'relative', display: 'flex' }}>
+    <div style={{ position: 'absolute', top: '50%', left: '50%', color: 'grey', transform: 'translate3d(-50%, -50%, 0)' }}>
+      <h2 style={{ display: 'inline', margin: 0 }}>Header</h2>
+      {layoutConfig.toggleMenu && <button style={{ display: 'inline', marginLeft: '5px', height: '25px', border: '1px dashed lightgrey' }} onClick={layoutConfig.toggleMenu}>Toggle Menu</button>}
+      {layoutConfig.size !== 'small' && layoutConfig.size !== 'tiny' && <Link to="/page1" style={{ paddingLeft: '5px' }}>Page 1</Link>}
```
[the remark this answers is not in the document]
tbiethman: I'd rather come back around and revamp these to use the templates currently in development.

packages/terra-navigation-layout/src/routingUtils.js
```diff
@@ -0,0 +1,105 @@
+import { matchPath } from 'react-router-dom';
+
+/**
+ * Given a route configuration and size, returns an Array of ordered routing data objects
+ * representing the routes available at that size.
+ * @param {Object} routeConfig The route configuration to reduce. The expected shape matches that of the routeConfigPropType.
+ * @param {String} size The breakpoint size used during configuration parsing.
+ * @param {Array[String]} parentPaths The array of String paths that are ancestors to the given routeConfig.
+ */
+const reduceRouteConfig = (routeConfig, size) => {
```
ryanthemanuel: Might be good to have some concrete examples of what is going on here in the comments for posterity.

packages/terra-navigation-layout/src/routingUtils.js
```diff
@@ -0,0 +1,105 @@
...
+ * @param {String} pathname The path to be searched for in the routeConfig.
+ * @param {Array} routes The route data to match against.
+ * @param {String} size The breakpoint size used for configuration parsing.
+ */
+const validateMatchExists = (pathname, routes) => {
+  if (!routes || !pathname) {
+    return false;
+  }
+
+  for (let i = 0, length = routes.length; i < length; i += 1) {
```
ryanthemanuel: Can we use the forEach

tbiethman: If we did, we couldn't abort the loop when we find the first match.

PR #17, opened 2017-12-15
Site Revamp

emilyrohrbough:
### Summary
Its not perfect yet, but here's the mock restructure we would expect.


packages/terra-navigation-layout/examples/IndexExampleTemplate.jsx
```diff
@@ -0,0 +1,70 @@
...
+  handlePanelToggle() {
+    this.setState({ panelIsOpen: !this.state.panelIsOpen });
+  }
+
+  render() {
+    const { title, example, exampleSrc, children } = this.props;
+
+    const indexExample = (
+      <div>
+        <button onClick={this.handlePanelToggle} style={{ display: 'inline-block' }}>Show Source Code</button>
```
JakeLaCombe: Do we have a documented convention on when to use inline styles vs classes? We get a lot of questions from consumers on this, and it'd be great if we have a source of truth to pass to consumers. Otherwise, you'll have different developers both on and off Terra that will give an opinion that may not match the coded examples.

bjankord:
Closing this in favor of moving this code to a new repo.

PR #21, opened 2018-01-09
Update repo to use with theme repo

bjankord:
### Summary
* Adding `src` and `tests` directories to npm packages so we can access them in the theme repos.
* Updated postinstall script to be able to detect in what situation it is being run in. (npm install in monorepo vs installing monorepo into another project)

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,8 @@
     "pretest": "npm run lint",
     "props-table": "lerna run props-table",
     "postinstall": "link-parent-bin && npm run compile && npm run bootstrap",
+    "postinstall:package": "npm run compile && npm run bootstrap",
+    "postinstall:project": "link-parent-bin && npm run compile && npm run bootstrap",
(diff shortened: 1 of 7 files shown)


scripts/post-install/index.js
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
mjhenkes: Eventually this could be a terra-toolkit script

package.json
```diff
@@ -58,6 +58,8 @@
     "pretest": "npm run lint",
     "props-table": "lerna run props-table",
     "postinstall": "link-parent-bin && npm run compile && npm run bootstrap",
```
kschuste: shouldn't you update this line to be the following:

"postinstall": "node scripts/post-install/index.js",

bjankord: Good call, [updated](https://github.com/cerner/terra-framework/pull/21/commits/9f51c3b1127464c0f6a3b1a52638ff714f011866)

PR #22, opened 2018-01-10
Terra Layout Peformance Enhancement

dkasper-was-taken:
### Summary
Add debounce, and move css breakpoint parsing to load time.

@cerner/terra


packages/terra-layout/src/Layout.jsx
```diff
@@ -72,6 +72,17 @@ class Layout extends React.Component {
     window.removeEventListener('resize', this.updateSize);
   }
 
+  debounce(fn, delay) {
```
tbiethman: Can we just use lodash's debounce package instead of rolling our own? https://www.npmjs.com/package/lodash.debounce

PR #23, opened 2018-01-15
Application Name Component

yuderekyu:
### Summary
Initial skeleton for the Application Name component (generated by `generator-terra-module`). This component will be used in the higher order Navigation component to display/switch between applications a consumer has access to.

@cerner/terra


emilyrohrbough:
Is there a tech design associated with this component?

yuderekyu:
^ There is not. Per offline discussion with @mjhenkes, we decided this component is simple enough to start without one/ we are still waiting on UX for a finalized version.

To remain consistent with other components, this package should be `terra-application-name`, not `terra-framework-application-name`. I'll go ahead and change `generator-terra-module` to only append `terra` instead of `terra-framework`, then recreate this skeleton.

PR #25, opened 2018-01-16
Generated boilerplate for package for embedded-component

mhemesath:
### Summary
Generated boilerplate code for EmbeddedComponent. This is a component which will allow mounting external react apps in the same frame. Actual code will be in followup PR.

See #27 for technical details.

@cerner/terra


mjhenkes:
Could you set an ignore for EmbeddedContent like we did with terra-button 2.0 so that master stays releasable? https://github.com/cerner/terra-core/blob/master/lerna.json

Also could you throw together a tech design issue that explains what this component is intended to do?

packages/terra-site/src/examples/embedded-component/Index.jsx
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
yuderekyu: Didn't realize that terra-framework's existing site package is `terra-framework-site`. We should move this example to the `terra-framework-site` package. In the mean time, just created this [issue](https://github.com/cerner/generator-terra-module/issues/78).

bjankord: Yeah, we'll want to move this to `terra-framework-site`. Copy / paste `embedded-component/Index.jsx` into `packages/terra-framework-site/src/examples/` and then this is good to go.

mhemesath:
> Also could you throw together a tech design issue that explains what this component is intended to do?

@mjhenkes  tech design is here #27

packages/terra-embedded-component/package.json
```diff
@@ -0,0 +1,49 @@
...
+  },
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2",
+    "terra-base": "^2.0.0"
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "prop-types": "^15.5.8",
+    "terra-base": "^2.0.0"
```
zhongyn: it seems terra-base is not used in this component. Also it's added as peerDenpendencies above

PR #26, opened 2018-01-16
Theming Repo Setup

emilyrohrbough:
### Summary
To correctly install terra-core within theming repos, we need the follow packages listed as dependencies to correctly run install, clean and compile scripts.


mjhenkes:
@emilyrohrbough I think i get why these are needed, but could you elaborate?

emilyrohrbough:
@mjhenkes Within the theming repos, we are install `terra-framework` as a devDependency. During the install process, the package runs the postInstall script (which uses shelljs) to then run the link-parent-bin script (which uses link-parent-bin). Then it will use lerna to compile each packages which used clean:compile (which uses rimraf) then clean:build (which uses babel-cli -- peer dependency of babel-core) to compile each package.

emilyrohrbough:
Talked with @mjhenkes. Have a few more ideas to try before using this approach

Issue #27, opened 2018-01-16
Create Component to Embed External React Apps


mhemesath:
# Issue Description
We need the ability to mount external react applications into the DOM. This enables us to combine apps through configuration rather than maintaining a massive webpack instance with dynamic imports that must be updated with each app update. 

## Issue Type

- [x] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The component would be consumed like this:
```js
import React, { Component } from 'react';
import EmbeddedComponent from './EmbeddedComponent';

import './App.css';

class App extends Component {
  render() {
    const loading=<div>Loading, please wait</div>;

    return (
      <div className="App">
        <EmbeddedComponent basePath="http://externaldomain.com/myApp" placeholder={loading}/>
      </div>
    );
  }
}

export default App;
```

The `basePath` will point to a base location which is expected to have a file `manifest.json`. The manifest is generated using the [webpack-manifest-plugin](https://github.com/danethurber/webpack-manifest-plugin). Example manifest:
```json
{
  "ar-intl-local.js": "ar-intl-local.8b89e84b9dd407d27826.js",
  "ar-translations.js": "ar-translations.cdad6be775ddcb106b6c.js",
  "de-intl-local.js": "de-intl-local.e7014c301a02e34b135e.js",
  "de-translations.js": "de-translations.3e802213adb19ffad47f.js",
  "en-GB-intl-local.js": "en-GB-intl-local.466fe7ba468e71f59d48.js",
  "en-GB-translations.js": "en-GB-translations.8d392998b43336e205cf.js",
  "en-US-intl-local.js": "en-US-intl-local.329870131ef92acef02d.js",
  "en-US-translations.js": "en-US-translations.e86978e4d3f21af705c6.js",
  "en-intl-local.js": "en-intl-local.a2db357afa1f66cd3e9e.js",
  "en-translations.js": "en-translations.3b5203d3ff7000964dd2.js",
  "es-ES-intl-local.js": "es-ES-intl-local.50de76237c5b4ea61995.js",
  "es-ES-translations.js": "es-ES-translations.f77d13737c97e31e41da.js",
  "es-US-intl-local.js": "es-US-intl-local.99cbfee4baa20a778e8b.js",
  "es-US-translations.js": "es-US-translations.8d7f66af308b5d2bc5f1.js",
  "es-intl-local.js": "es-intl-local.e002521d76f8191b6b6c.js",
[...]

PR #36, opened 2018-01-18
Embedded component

mhemesath:
### Summary
Pull request for embedded component. See #27 

@cerner/terra


mhemesath:
Need to add example/tests yet

packages/terra-embedded-component/src/EmbeddedComponent.jsx
```diff
@@ -1,31 +1,113 @@
...
- */
-  name: PropTypes.string,
+
+ /**
+  * The basePath to load the manifest and manifest asset dependencies from.
+  * */
+  basePath: PropTypes.string.isRequired,
+
+  /**
+   * A placeholder to display while the component is laoding.
```
zhongyn: loading typo

packages/terra-embedded-component/docs/README.md
```diff
@@ -13,9 +15,67 @@
...
 import EmbeddedComponent from 'terra-embedded-component';
 
-<EmbeddedComponent {props...} />
+<EmbeddedComponent basePath="http://mycomponent.com/path" {props...} />
 ```
 
+## Building Embeddable Components
+
+EmbeddedComponent looks for a `manifest.json` at the `basePath` to identify the component dependencies.
+The JavaScript modules are exepcted to be in [AMD](https://en.wikipedia.org/wiki/Asynchronous_module_definition) format.
```
JakeLaCombe: *expected

packages/terra-embedded-component/docs/README.md
```diff
@@ -13,9 +15,67 @@
...
+### Package the component as an AMD module
+In webpack set the following configuration for your `output` config:
+
+```js
+output: {
+  filename: '[name].[chunkhash].js',
+  path: path.join(__dirname, 'dist'),
+
+  // Required if the component is hosted from a path different than the embedding script file.
+  publicPath: 'http://componentdomain.com/',
```
JakeLaCombe: Is it possible to make this match with the example you have on line 18? I think that would be easier for consumers to see how to configure this portion.

packages/terra-embedded-component/src/EmbeddedComponent.jsx
```diff
@@ -1,31 +1,144 @@
...
 import PropTypes from 'prop-types';
-import classNames from 'classnames/bind';
-import 'terra-base/lib/baseStyles';
-import styles from './EmbeddedComponent.scss';
+import 'systemjs/dist/system';
+import { loadStylesheet } from './assets';
 
-const cx = classNames.bind(styles);
+// Expose AMD required globals on the window
+window.define = window.SystemJS.amdDefine;
```
JakeLaCombe: Should we check the value of define before we override it?

Also, could we use an expose loader for this instead potentially? https://github.com/webpack-contrib/expose-loader

packages/terra-embedded-component/src/EmbeddedComponent.jsx
```diff
@@ -1,31 +1,144 @@
...
+  * */
+  basePath: PropTypes.string.isRequired,
+
+  /**
+   * A placeholder to display while the component is laoding.
+   * */
+  placeholder: PropTypes.element,
+
+  /**
+   * A component to display while the loading of css or module is timed out.
```
JakeLaCombe: I think this makes it a bit better

` A component to display when the css or module times out`

packages/terra-embedded-component/src/assets.js
```diff
@@ -0,0 +1,16 @@
+/**
+* Loads a stylsheet from the given url.
```
JakeLaCombe: *stylesheet

mjhenkes:
@mhemesath mind if we close this for the time being? Site has changed since this was posted and we're leaning away from pushing in skeleton packages before the package is ready to release.

Issue #28, opened 2018-01-17
Create Application Header Name component


dkasper-was-taken:
# Issue Description
Create ApplicationName/Logo component to be placed within the logo section of the application-header-layout.

## Issue Type
- [x] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

PR #42, opened 2018-01-24
Application Name Components

yuderekyu:
### Summary
Create two application-name components. These house a logo and application title and/or subtitle: one for `header-layout`, the other for `menu-layout`. Addresses #28.

### Deployment Links
https://terra-framework-deployed-pr-42.herokuapp.com/static/#/components/application-name
https://terra-framework-deployed-pr-42.herokuapp.com/static/#/tests/application-name

@cerner/terra


packages/terra-application-name/package.json
```diff
@@ -41,7 +42,7 @@
     "lint": "npm run lint:js && npm run lint:scss",
     "lint:js": "eslint --ext .js,.jsx . --ignore-path ../../.eslintignore",
     "lint:scss": "stylelint src/**/*.scss",
-    "props-table": "props-table ./src/ApplicationName.jsx --out-dir ./docs/props-table",
+    "props-table": "props-table ./src/ApplicationHeaderName.jsx --out-dir ./docs/props-table",
```
emilyrohrbough: Add `./src/ApplicationMenuName.jsx`

packages/terra-application-name/src/ApplicationHeaderName.jsx
```diff
@@ -0,0 +1,78 @@
...
+  /**
+   * The accessory element to be displayed next to the title.
+   * */
+  accessory: PropTypes.element,
+  /**
+   * The AppDelegate instance provided by the containing component. If present, its properties will propagate to the children components.
+   * */
+  app: AppDelegate.propType,
+  /**
+   * The breakpoint size that the window is currently at. Takes one of: `tiny`, `small`, `medium`, `large`, `huge`.
```
emilyrohrbough: Should we list the expected strings or the Utils Keys here?

emilyrohrbough:
It looks like the ApplicationHeaderName and ApplicationMenuName components are 1-to-1 in implementation. Could we condense these to an ApplicationName component that uses the same props and adds the prop `variant` with the types `header' and `menu`?

emilyrohrbough:
Actually, the styling is 1-to-1 as well. Is there a need to have two distinguishing components?

yuderekyu:
@emilyrohrbough they will most likely have different styles - splitting them allows for different themeable variables. At the moment, the functionality is the same too, but they may also diverge in future iterations when we have more info from UX.

emilyrohrbough:
@yuderekyu If the expectation that these will diverge in functionality, should they be their own packages then?

packages/terra-application-name/src/ApplicationHeaderName.jsx
StephenEsser: The util keys feel a little cumbersome for some situations, but they do help abstract the internal implementation. 

I think it's easier to read and write (both documentation and code syntax):
```
<ApplicationHeaderName size="tiny" />
```
vs
```
<ApplicationHeaderName size={ApplicationHeaderName.opts.SizeTypes.TINY} />
```
There's advantages and disadvantages to both.

emilyrohrbough: +1 to sizes

yuderekyu: Per offline discussion with @dkasper-was-taken, we'll let the higher order header handle the breakpoint screen sizes. The size prop has been removed, along with the util keys.

emilyrohrbough:
New package structure goes as follows:
```
component
    /docs
    /examples
        /test-examples  # what use to be in the nightwatch dir
        Index.site-page.js
    /src
    /tests
```

packages/terra-application-name/src/ApplicationHeaderName.jsx
```diff
@@ -0,0 +1,48 @@
...
+  }) => {
+  const ApplicationHeaderNameClassNames = cx([
+    'application-header-name',
+    customProps.className,
+  ]);
+
+  return (
+    <div {...customProps} className={ApplicationHeaderNameClassNames}>
+      {!!accessory && <div className={cx('accessory')}>{accessory}</div>}
+      {!!title && <div className={cx('title')}><strong>{title}</strong></div>}
```
bjankord: If you are just using the `strong` element because it makes the text bold, it is likely not the element you want to use. This can be done via CSS in the title class. The `<strong>` element is for content that is of "strong importance," including things of great seriousness or urgency (such as warnings). Source: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong

packages/terra-application-name/src/ApplicationMenuName.jsx
```diff
@@ -0,0 +1,48 @@
...
+  }) => {
+  const ApplicationMenuNameClassNames = cx([
+    'application-menu-name',
+    customProps.className,
+  ]);
+
+  return (
+    <div {...customProps} className={ApplicationMenuNameClassNames}>
+      {!!accessory && <div className={cx('accessory')}>{accessory}</div>}
+      {!!title && <div className={cx('title')}><strong>{title}</strong></div>}
```
bjankord: Same comment about strong element as in application-header-name

packages/terra-application-name/tests/nightwatch/application-header-name-spec.js
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable no-unused-expressions */
```
bjankord: We should write these as WDIO tests instead of nightwatch

packages/terra-application-name/tests/nightwatch/application-menu-name-spec.js
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable no-unused-expressions */
```
bjankord: We should write these as WDIO tests instead of nightwatch

README.md
```diff
@@ -42,13 +42,14 @@ Components in beta stage may include breaking changes, new features, and bug fix
...
-| [terra-modal-manager](https://github.com/cerner/terra-framework/tree/master/packages/terra-modal-manager) | [![NPM version](http://img.shields.io/npm/v/terra-modal-manager.svg)](https://www.npmjs.org/package/terra-modal-manager) | ![Stable](https://img.shields.io/badge/status-Stable-brightgreen.svg) |
-| [terra-navigation-layout](https://github.com/cerner/terra-framework/tree/master/packages/terra-navigation-layout) | [![NPM version](http://img.shields.io/npm/v/terra-navigation-layout.svg)](https://www.npmjs.org/package/terra-navigation-layout) | ![Stable](https://img.shields.io/badge/status-Stable-brightgreen.svg) |
-| [terra-theme-provider](https://github.com/cerner/terra-framework/tree/master/packages/terra-theme-provider) | [![NPM version](http://img.shields.io/npm/v/terra-theme-provider.svg)](https://www.npmjs.org/package/terra-theme-provider) | ![Stable](https://img.shields.io/badge/status-Stable-brightgreen.svg) |
+| [terra-app-delegate](https://github.com/cerner/terra-framework/tree/master/packages/terra-app-delegate) | [![NPM version](http://img.shields.io/npm/v/terra-app-delegate.svg)](https://www.npmjs.org/package/terra-app-delegate) | :white_check_mark: |
```
mjhenkes: You're reverting the status shields change here. We should be using `Stable` instead of `:white_check_mark:`

bjankord: Good catch, should be able to check this file out from master locally and add an entry for application-name.

yuderekyu: Oops. Updated

packages/terra-application-name/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "terra-application-name",
-  "main": "lib/ApplicationName.js",
+  "main": "lib/ApplicationHeaderName.js",
```
mjhenkes: This package contains the application header name and application menu name. I think your previous package name was correct. We should also keep this in sync with the package name in lerna to avoid confusion (in most cases).

packages/terra-application-name/package.json
```diff
@@ -41,9 +46,10 @@
     "lint": "npm run lint:js && npm run lint:scss",
     "lint:js": "eslint --ext .js,.jsx . --ignore-path ../../.eslintignore",
     "lint:scss": "stylelint src/**/*.scss",
-    "props-table": "props-table ./src/ApplicationName.jsx --out-dir ./docs/props-table",
-    "test": "npm run test:jest && npm run test:nightwatch",
+    "props-table": "props-table ./src/ApplicationHeaderName.jsx ./src/ApplicationMenuName.jsx --out-dir ./docs/props-table",
+    "test": "npm run test:jest && npm run test:wdio",
     "test:jest": "jest ./tests/jest/* --config ../../jestconfig.json",
-    "test:nightwatch": "nightwatch -c ../../nightwatch.conf.js"
+    "test:nightwatch": "nightwatch -c ../../nightwatch.conf.js",
```
mjhenkes: can we remove nightwatch?

wdio.conf.js
```diff
@@ -0,0 +1,43 @@
+/* eslint-disable import/no-extraneous-dependencies */
+const wdioConf = require('terra-toolkit/lib/wdio/conf');
+const WebpackDevService = require('terra-toolkit/lib/wdio/services/index').WebpackDevService;
+const localIP = require('ip');
+const path = require('path');
+const webpackConfig = require('./node_modules/terra-site/src/config/webpack.config.js');
```
emilyrohrbough: This can be updated to `require('terra-site/src/config/webpack.config.js');`

packages/terra-application-name/tests/wdio/application-menu-name-spec.js
```diff
@@ -0,0 +1,21 @@
+/* global browser, Terra */
+
+const viewports = Terra.viewports('tiny', 'medium');
+
+describe('ApplicationMenuName', () => {
+  describe('Displays a default application menu name', () => {
+    beforeEach(() => browser.url('/#/raw/tests/application-name/application-menu-name-default'));
+
+    Terra.should.matchScreenshot({ viewports, options: { selector: '#default' } });
```
emilyrohrbough: How come your default screenshots are so large/have so much whitespace?

packages/terra-application-name/examples/index-examples/ApplicationHeaderNameStandard.jsx
```diff
@@ -0,0 +1,61 @@
...
+import Image from 'terra-image';
+import ResponsiveElement from 'terra-responsive-element';
+import NavigationExample from 'terra-application-header-layout/examples/index-examples/NavigationExample';
+import ToggleExample from 'terra-application-header-layout/examples/index-examples/ToggleExample';
+import UtilitiesExample from 'terra-application-header-layout/examples/index-examples/UtilitiesExample';
+import ExtensionsExample from 'terra-application-header-layout/examples/index-examples/ExtensionsExample';
+import ApplicationHeaderName from '../../lib/ApplicationHeaderName';
+
+const ApplicationHeaderNameStandard = () => {
+  const color = { color: 'grey' };
```
JakeLaCombe: I think this is just used on one place. If that's so, I would leave out this variable.

JakeLaCombe: Do this throughout the rest of your examples

packages/terra-application-name/src/ApplicationHeaderName.jsx
```diff
@@ -0,0 +1,41 @@
...
+    ...customProps
+  }) => {
+  const ApplicationHeaderNameClassNames = cx([
+    'application-header-name',
+    customProps.className,
+  ]);
+
+  return (
+    <div {...customProps} className={ApplicationHeaderNameClassNames}>
+      {!!accessory && <div className={cx('accessory')}>{accessory}</div>}
```
JakeLaCombe: Do we need the double !! for this? In other code I've written, I've been able to get my desired functionality with just `{accessory && <div>{accessory}</div>}`

packages/terra-application-name/src/ApplicationHeaderName.scss
```diff
@@ -0,0 +1,24 @@
...
+    flex: var(--terra-application-header-name-flex, 0 0 auto);
+  }
+
+  .accessory {
+    padding-left: var(--terra-application-header-name-accessory-padding-left, 0.3571rem);
+  }
+
+  .title {
+    font-weight: var(--terra-application-header-name-title-font-weight, normal);
+    overflow: hidden;
```
JakeLaCombe: I think this value should be themed as well.

packages/terra-application-name/src/ApplicationMenuName.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+const propTypes = {
+  /**
+   * The accessory element to be displayed next to the title.
+   * */
+  accessory: PropTypes.element,
+  /**
+   * The title branding of the application name.
+   * */
+  title: PropTypes.string,
```
JakeLaCombe: On this issue, I proposed different ways we can account for overflowing text. (https://github.com/cerner/terra-core/issues/355).

While we don't need to implement these here, I think it will be easier to allow for developers to opt in and out of text truncation by making this a react node as opposed to just a node. In addition, React 16 allows to render a string. Given that, is it possible to change the proptype of the title argument?

packages/terra-application-name/examples/test-examples/ApplicationHeaderNameDefault.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import Image from 'terra-image';
+import ApplicationHeaderName from '../../lib/ApplicationHeaderName';
+
+const color = { color: 'black' };
```
emilyrohrbough: IMO It would be better to set the background color to be different than white, than overriding the default font color.

packages/terra-application-name/tests/jest/ApplicationHeaderName.test.jsx
```diff
@@ -0,0 +1,29 @@
...
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render with provided accessory', () => {
+    const wrapper = shallow(<ApplicationHeaderName accessory={<img alt="this is the accessory" />} />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render with provided title', () => {
+    const wrapper = shallow(<ApplicationHeaderName subtitle="subtitle" />);
```
emilyrohrbough: Looks like this test is out of date

packages/terra-application-name/tests/jest/ApplicationMenuName.test.jsx
```diff
@@ -0,0 +1,34 @@
...
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render with provided size', () => {
+    const wrapper = shallow(<ApplicationMenuName size="tiny" />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render with provided title', () => {
+    const wrapper = shallow(<ApplicationMenuName subtitle="subtitle" />);
```
emilyrohrbough: same with this one for the subtitle prop

packages/terra-application-name/src/ApplicationMenuName.scss
```diff
@@ -0,0 +1,24 @@
+/* stylelint-disable selector-class-pattern */
+:local {
+  .application-menu-name {
+    align-items: var(--terra-application-menu-name-align-items, center);
+    color: var(--terra-application-menu-name-color, #fff);
+    display: flex;
+    flex: var(--terra-application-menu-name-flex, 0 0 auto);
```
emilyrohrbough: Do we want flex to be themeable?

packages/terra-application-name/src/ApplicationMenuName.scss
```diff
@@ -0,0 +1,24 @@
+/* stylelint-disable selector-class-pattern */
+:local {
+  .application-menu-name {
+    align-items: var(--terra-application-menu-name-align-items, center);
```
emilyrohrbough: Looking at the screnshots, I am wondering if we need to add a width or min-width to get the centered behavior.

packages/terra-application-name/src/ApplicationMenuName.scss
```diff
@@ -2,7 +2,7 @@
 :local {
   .application-menu-name {
     align-items: var(--terra-application-menu-name-align-items, center);
-    color: var(--terra-application-menu-name-color, #fff);
+    color: var(--terra-application-menu-name-color, hsl(0, 0%, 0%));
```
emilyrohrbough: Should we use #000 instead of hsl(0, 0%, 0%) to be consistent with other component's styles?

packages/terra-application-name/tests/wdio/application-header-name-spec.js
```diff
@@ -6,7 +6,7 @@ describe('ApplicationHeaderName', () => {
   describe('Displays a default application header name', () => {
     beforeEach(() => browser.url('/#/raw/tests/application-name/application-header-name-default'));
 
-    Terra.should.matchScreenshot({ viewports, options: { selector: '#default' } });
+    Terra.should.matchScreenshot(viewports, { selector: '#default' });
```
emilyrohrbough: To correctly create the screenshots for each viewport, you write matchScreenshot like this:
`Terra.should.matchScreenshot({viewports, selector: '#default' });`

If you noticed, most of your VR screenshots were removed.

yuderekyu: Oops, updated

mmalaker:
IE10, 11, and Edge issue found. The icon in the app name component appears low res, but looks normal in all other browsers.

neilpfeiffer:
Tested: PR #42

**Reviewed:**
- Follows UX’s [universal design definition](https://zpl.io/an1vqoY) implementation with variations noted below.

**Approved Variations from design definition:**
- Adds the ability to include and accessory (Icon/Element) before the title
- Title can accept node which allows for simple text, dropdown app-switching feature, or providing Logo/Image as the application name 


\
\

**Pass** |
-----------|
UX Review|

neilpfeiffer:
Response to @mmalaker above - the icon used for for the demo in the Application Name/Title is actually a super huge (PNG)[https://raw.githubusercontent.com/cerner/terra-core/master/terra.png]

And IE is/does a really bad job of resampling images, so it is looking crusty. If it was an image PNG provided at size, or if it was an SVG it would look better. So nothing we need to worry about for actual usage for real application.

mmalaker:
Functional verification completed. The application name component displays and functions as expected.

PR #54, opened 2018-01-30
Creating Aggregator

tbiethman:
### Summary
Adding the terra-aggregator package and related files.

### Deployment URL
https://terra-framework-deployed-pr-54.herokuapp.com/

### Additional Details
This is a bit of a work in progress. Right now, the updated disclosure managers are included with the Aggregator examples in order to get some working examples out there for testing/review. However, I will work on getting additional PRs out for those now.


packages/terra-app-delegate/src/app-delegate/AppDelegate.js
```diff
@@ -5,7 +5,7 @@ import {
 } from '../disclosure-component-registry/DisclosureComponentRegistry';
 
 const supportedAttributes = [
-  'disclose', 'dismiss', 'closeDisclosure', 'goBack', 'maximize', 'minimize', 'requestFocus', 'releaseFocus',
+  'disclose', 'dismiss', 'closeDisclosure', 'goBack', 'maximize', 'minimize', 'requestFocus', 'releaseFocus', 'registerLock',
```
tbiethman: This change will also be removed and merged with a separate PR.

packages/terra-aggregator/examples/tmp/DisclosureManager.jsx
```diff
@@ -0,0 +1,316 @@
+import React from 'react';
```
tbiethman: This file will be moved into its own package shortly. Feel free to review if you want a preview.

packages/terra-aggregator/examples/tmp/NewModalManager.jsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
```
tbiethman: This file will be moved into the terra-modal-manager package and replace the existing ModalManager.

packages/terra-aggregator/examples/tmp/SlidePanelManager.jsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
```
tbiethman: This file will be moved into a terra-slide-panel-manager package.

packages/terra-aggregator/src/Aggregator.jsx
```diff
@@ -0,0 +1,217 @@
+import React from 'react';
```
tbiethman: This is the only non-example component in the package.

packages/terra-aggregator/package.json
```diff
@@ -0,0 +1,55 @@
...
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "prop-types": "^15.5.8",
+    "terra-app-delegate": "^1.10.0",
+    "terra-button": "^1.14.0",
+    "terra-list": "^1.19.0",
+    "terra-content-container": "^1.15.0",
+    "terra-clinical-header": "^1.7.0",
+    "terra-clinical-action-header": "^1.6.0",
+    "terra-form": "^1.16.0",
```
jeffmerten: Looks like some of these should be devDependencies

packages/terra-aggregator/src/Aggregator.jsx
```diff
@@ -0,0 +1,229 @@
...
+    return Promise.resolve()
+      .then(() =>
+        /**
+         * Focus is released on the currently focused item to ensure a clean start for the next component receiving focus.
+         * The releaseFocus's Promise is returned and inserted into the Promise chain to prevent disclosures from occurring
+         * if the focus release fails.
+         */
+        this.releaseFocus(focusedItemId)
+          .then(() => {
+            this.setFocusState(itemId, Object.freeze(itemState || {}));
```
jeffmerten: So you decided not to use redux for this?

packages/terra-aggregator/src/Aggregator.jsx
```diff
@@ -0,0 +1,229 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const propTypes = {
+  /**
+   * The components that will be managed by the Aggregator. Each component provided will be provided with an 'aggregatorDelegate' prop
+   * containing the Aggregator API. Keys must also be provided to allow the Aggregator to uniquely identify the component.
+   * Immutable objects are preferred.
+   */
+  items: PropTypes.arrayOf(PropTypes.shape({
```
jeffmerten: Why isn't items required?

packages/terra-aggregator/docs/README.md
```diff
@@ -0,0 +1,68 @@
+# Terra Aggregator
+
+The Aggregator provides focus-based mechanisms for managing actions across disparate components.
+
+I need to plump up this documentation a bit, but here's the gist:
```
emilyrohrbough: > I need to plump up 

😄

packages/terra-aggregator/src/Aggregator.jsx
tbiethman: You're right, no redux. This is mostly a side-effect of the Promise-based logic in play; it gets very hard to replay the state.

packages/terra-aggregator/src/Aggregator.jsx
tbiethman: I waffle a bit on that. The component will render appropriately without them, it just won't be very helpful with no items to manage.

packages/terra-aggregator/package.json
tbiethman: There a ton of weird dependencies now due to how the generated site works. As far as I know, devDependencies are not sufficient for the time being.

Some of these will go away as I create more packages and move the examples around.

jeffmerten: Should there be an issue for this? We don't want to bring in dependencies only needed in the generated site.

mjhenkes: We release the tests as runnable artifacts now. This won't affect your usage of the aggregator or pull more assets into your page, It just may make your node_modules folder larger.

packages/terra-aggregator/src/Aggregator.jsx
jeffmerten: +1 on not using redux like you were before to persist the focus state, I could have seen some consumers abusing this.

packages/terra-aggregator/src/Aggregator.jsx
```diff
@@ -0,0 +1,229 @@
...
+          requestFocus: state => this.requestFocus(item.key, state),
+          releaseFocus: childIsActive ? () => (this.releaseFocus(item.key)) : undefined,
+          itemState: childIsActive ? focusedItemState : undefined,
+        },
+      });
+    });
+  }
+
+  render() {
+    const renderedItems = this.renderItems();
```
jeffmerten: Is there any good way to avoid unnecessary renders when there is a selection change and it doesn't affect a particular child item?

packages/terra-aggregator/src/Aggregator.jsx
tbiethman: Yeah we're definitely going to have to come up with something. The Aggregator itself has few dirty data concerns. I moved a lot of that locking logic into the disclosure managers themselves; the Aggregator just hooks into the disclosure lifecycle and updates accordingly.  So we'll just have to figure out how to communicate disclosure locks to the larger overall application.

packages/terra-aggregator/src/Aggregator.jsx
tbiethman: The items themselves could implement 'shouldComponentUpdate' in order to block rerendering of their content. 

I'd like to avoid caching child instances in state, but that would be an option too if performance proves to be a concern. Until then, I say roll with it.

packages/terra-aggregator/src/Aggregator.jsx
jeffmerten: Makes me wonder if the management of focus and dirty data should be separate concerns....

packages/terra-aggregator/src/Aggregator.jsx
jeffmerten: fry

jeffmerten: I'd at least add that to the documentation.

packages/terra-aggregator/src/Aggregator.jsx
tbiethman: Let me rephrase: the Aggregator has zero dirty data concerns. The only place where it could run into dirty data issues is when it tries to close an open disclosure during a focus release event, and the DisclosureManager encapsulates all of that. A component having focus is not 'dirty' state from the Aggregator's perspective. A component can make itself 'dirty' after having received focus, but that is the component's concern, not the Aggregator's.

packages/terra-aggregator/src/Aggregator.jsx
```diff
@@ -0,0 +1,229 @@
...
+      // the focus is forcefully released to clean up any presented disclosures.
+      let focusItemIsPresent;
+      items.forEach((item) => {
+        if (item.key === focusedItemId) {
+          focusItemIsPresent = true;
+        }
+      });
+
+      if (!focusItemIsPresent) {
+        this.releaseFocus(undefined, true);
```
jeffmerten: Is this line necessary?

packages/terra-aggregator/src/Aggregator.jsx
```diff
@@ -0,0 +1,229 @@
...
+  /**
+   * A function that will override the Aggregators default render. The function will receive an Object parameter data necessary for the
+   * render process.
+   */
+  render: PropTypes.func,
+  /**
+   * A function that will be provided to Aggregator items that have received focus. The function must adhere to the standardized
+   * AppDelegate disclosure API.
+   */
+  disclose: PropTypes.func,
```
jeffmerten: Why isn't this required?

packages/terra-aggregator/src/Aggregator.jsx
```diff
@@ -0,0 +1,229 @@
...
+                  }
+                });
+
+                // We return the same API so as not to disrupt the chain.
+                return { afterDismiss, dismissDisclosure, ...other };
+              });
+          };
+        }
+
+        return focusRequestPayload;
```
jeffmerten: What's the use case for this?

packages/terra-aggregator/examples/components/DisclosureContent.jsx
```diff
@@ -0,0 +1,112 @@
...
+    this.checkLockState = this.checkLockState.bind(this);
+
+    this.state = {
+      text: undefined,
+    };
+  }
+
+  componentDidMount() {
+    if (this.props.app && this.props.app.registerLock) {
+      this.props.app.registerLock(this.checkLockState);
```
jeffmerten: I'm sorry I missed this, but how will any aggregator items (content) register locks? I see now how this works for any content disclosed, but nothing for anything in the main content:
https://github.com/cerner/terra-framework/blob/master/packages/terra-disclosure-manager/src/DisclosureManager.jsx#L257

packages/terra-aggregator/src/Aggregator.jsx
tbiethman: This is how components withing the Aggregator get access to the disclose function. They will not receive a traditional `app` prop (not from the Aggregator, at least), because we're trying to ensure that only the focused item has access to the progressive disclosure API.

This is what it'd look like from the item's standpoint.
```jsx
aggregatorDelegate.requestFocus({...})
  .then((focusRequestPayload) => {
    focusRequestPayload.disclose({...});
  });
```

packages/terra-aggregator/src/Aggregator.jsx
tbiethman: Yes. If the Aggregator item that holds focus is removed from the Aggregator on a subsequent render, we need to ensure the Aggregator's internal state is updated. That's all this is doing. It probably won't happen very often.

packages/terra-aggregator/src/Aggregator.jsx
tbiethman: Because the focus management works without it. Enforcement of the disclose prop presence could be managed by a higher component.

packages/terra-aggregator/src/Aggregator.jsx
jeffmerten: whoops read this wrong. 👍

lerna.json
```diff
@@ -7,7 +7,8 @@
         "DEPENDENCIES.md",
         "terra-application-name",
         "terra-application-utility",
-        "terra-embedded-component"
+        "terra-embedded-component",
+        "terra-aggregator"
```
bjankord: Is this something we want to keep in the lerna ignore config or is this ready to be released soon?

PR #57, opened 2018-02-01
Creating DisclosureManager

tbiethman:
### Summary
The DisclosureManager is a component that maintains the state and components necessary for a disclosure workflow. This component can be wrapped and implemented for more specific managers (ModalManager, SlidePanelManager, SomeOtherManager, etc.)

The README has been fleshed out quite a bit; I recommend reading it in its rendered format.


packages/terra-disclosure-manager/README.md
```diff
@@ -0,0 +1,24 @@
...
+- [Documentation](https://github.com/cerner/terra-framework/tree/master/packages/terra-disclosure-manager/docs)
+- [LICENSE](#license)
+
+## Getting Started
+
+- Install from [npmjs](https://www.npmjs.com): `npm install terra-disclosure-manager`
+
+## LICENSE
+
+Copyright 2017 Cerner Innovation, Inc.
```
JakeLaCombe: Should we update our generator to use 2018?

packages/terra-disclosure-manager/docs/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+The DisclosureManager is a stateful component used to manage disclosure presentation.
+
+## Getting Started
+
+- Install with [npmjs](https://www.npmjs.com):
+  - `npm install terra-disclosure-manager`
+
+## Usage
+
```
JakeLaCombe: Don't forget this section 😄

packages/terra-disclosure-manager/README.md
bjankord: We'll need to talk to James on the open source committee to get a clear direction on this. I would assume that we update this to 2018. Though as this gets into legal territory, we'll need guidance on this. Not sure if we update to `2018` or to `2017 - 2018`.

packages/terra-disclosure-manager/docs/README.md
```diff
@@ -0,0 +1,209 @@
...
+|`children`|An Object containing data relative to the children components provided to the DisclosureManager.|
+|`disclosure`|An Object containing data relative to the compoents in the disclosure stack.|
+|`dismissPresentedComponent`|A function that pops the currently disclosed component off the disclosure stack.|
+|`closeDisclosure`|A function that closes the disclosure and removes all components from the disclosure stack.|
+
+`children` Object API:
+
+|Key|Value|
+|---|---|
+|`components`|An Array of React components to render as face-up content. These components were updated with an AppDelegate prop with DisclosureManager integration.|
```
bjankord: Reading this, seems like `were` might be better worded as `are`

packages/terra-disclosure-manager/docs/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+`disclosure` Object API:
+
+|Key|Value|
+|---|---|
+|`components`|An Array of React components to render in a disclosure mechanism. These components were updated with an AppDelegate prop with DisclosureManager integration.|
```
bjankord: Is `components` part of the `disclosure` object API, or is this copy/paste from the `children` object API?

packages/terra-disclosure-manager/docs/README.md
```diff
@@ -0,0 +1,209 @@
...
+`content` Object API:
+
+|Key|Value|
+|---|---|
+|key|A String key uniquely identifying the component to the DisclosureManager. This key will be automatically added to the component when rendered.|
+|component|A React element that will be disclosed.|
+
+#### Disclosure Content
+
+The AppDelegate instances provided to the disclosure components are a little more complicated. In addition to a `disclose` function (with all the bells and whistles described in the above section), a number of other functions are exposed to manage various segments of the disclosure state. The included functions are:
```
bjankord: nit: I think you can trim this first paragraph in the disclosure content section down to: "In addition to a `disclose` function, a number of other functions are exposed to manage various segments of the disclosure state. The included functions are:"

packages/terra-disclosure-manager/docs/README.md
```diff
@@ -0,0 +1,209 @@
...
+      resolve();
+    }
+  });
+}
+```
+
+### Example
+
+```jsx
+// MyDisclosureManagerImpl.jsx
```
bjankord: If we want consumers to easily use this component, it is good to have a full working example which can be copy / pasted by consumers here. The modal disclosure example after the disclosure Object API docs is a good start to a working example.

packages/terra-disclosure-manager/tests/jest/DisclosureManager.test.jsx
```diff
@@ -0,0 +1,521 @@
...
+      expect(disclosureContentApp).toBeDefined();
+      expect(disclosureContentApp.disclose).toBeDefined();
+      expect(disclosureContentApp.dismiss).toBeDefined();
+      expect(disclosureContentApp.closeDisclosure).toBeDefined();
+      expect(disclosureContentApp.goBack).toBeUndefined();
+      expect(disclosureContentApp.maximize).toBeDefined();
+      expect(disclosureContentApp.minimize).toBeUndefined();
+      expect(disclosureContentApp.requestFocus).toBeDefined();
+      expect(disclosureContentApp.releaseFocus).toBeUndefined();
+      expect(disclosureContentApp.registerDismissCheck).toBeDefined(); // TODO: Flip this with AppDelegate release
```
bjankord: Should this todo be completed or has this been completed already?

packages/terra-disclosure-manager/docs/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+|Function|Description|
+|---|---|
+|`disclose(Object)`|Allows a component to disclose a component. This will open the disclosure if nothing was previously disclosed, or it will replace everything in the current disclosure stack. If disclosure was successful, it returns a Promise resolved with an Object containing the `dismissDisclosure` function and the `afterDismiss` deferred Promise. If disclosure was denied, it returns a rejected Promise.|
+
+`disclose` Argument API:
+
+|Key|Value|
+|---|---|
+|`preferredType`|The String describing the preferred disclosure type. This will be used to match the disclosure request to an appropriate DisclosureManager.|
```
mjhenkes: Lets add a sentence to the effect that the preferred type may not be the disclosure type used depending on availability and structure. ie if your component wants a slide panel disclosure but is embedded only in a modal manager, you get a modal.

packages/terra-disclosure-manager/docs/README.md
```diff
@@ -0,0 +1,209 @@
...
+|`size`|The String size desired for the disclosure. One of `tiny`, `small`, `medium`, `large`, `huge`, or `fullscreen`. The functional implementation of this size is determined by the rendering component.|
+|`content`| An Object containing items describing the component to be disclosed.|
+
+`content` Object API:
+
+|Key|Value|
+|---|---|
+|key|A String key uniquely identifying the component to the DisclosureManager. This key will be automatically added to the component when rendered.|
```
mjhenkes: Will the key be added as a prop?

packages/terra-disclosure-manager/docs/README.md
```diff
@@ -0,0 +1,209 @@
...
+|---|---|
+|`disclose(Object)`|Allows a component to disclose a component. This will open the disclosure if nothing was previously disclosed, or it will replace everything in the current disclosure stack. If disclosure was successful, it returns a Promise resolved with an Object containing the `dismissDisclosure` function and the `afterDismiss` deferred Promise. If disclosure was denied, it returns a rejected Promise.|
+
+`disclose` Argument API:
+
+|Key|Value|
+|---|---|
+|`preferredType`|The String describing the preferred disclosure type. This will be used to match the disclosure request to an appropriate DisclosureManager.|
+|`size`|The String size desired for the disclosure. One of `tiny`, `small`, `medium`, `large`, `huge`, or `fullscreen`. The functional implementation of this size is determined by the rendering component.|
+|`content`| An Object containing items describing the component to be disclosed.|
```
mjhenkes: Proposed re-wording?:
> An Object containing a key and a component describing the component to be disclosed. See the `content` api below.

packages/terra-disclosure-manager/docs/README.md
```diff
@@ -0,0 +1,209 @@
...
+|key|A String key uniquely identifying the component to the DisclosureManager. This key will be automatically added to the component when rendered.|
+|component|A React element that will be disclosed.|
+
+#### Disclosure Content
+
+The AppDelegate instances provided to the disclosure components are a little more complicated. In addition to a `disclose` function (with all the bells and whistles described in the above section), a number of other functions are exposed to manage various segments of the disclosure state. The included functions are:
+
+|Function|Description|
+|---|---|
+|`disclose(Object)`|Allows a component to disclose another component on top of itself. See above for argument API.|
```
mjhenkes: Proposed re-wording?:
>Allows a component to disclose another component. See above for argument API. 
The disclosure manager implementation will determine how the next object will be disclosed.

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -0,0 +1,351 @@
...
+
+    this.setState(newState);
+  }
+
+  closeDisclosure() {
+    this.setState({
+      disclosureIsOpen: false,
+      disclosureIsFocused: false,
+      disclosureIsMaximized: false,
+      disclosureSize: availableDisclosureSizes.SMALL,
```
mjhenkes: this seems awkward, should there be an undefined state?

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -0,0 +1,351 @@
...
+    const { disclosureComponentKeys, disclosureComponentData, disclosureIsMaximized, disclosureSize } = this.state;
+
+    return disclosureComponentKeys.map((componentKey, index) => {
+      const componentData = disclosureComponentData[componentKey];
+      const isFullscreen = disclosureSize === 'fullscreen';
+
+      const popContent = this.generatePopFunction(componentData.key);
+      const appDelegate = AppDelegate.create({
+        disclose: (data) => {
+          if (supportedDisclosureTypes.indexOf(data.preferredType) >= 0 || !app) {
```
mjhenkes: Lets dry up this line, it's used in `renderContentComponents` as well.

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -0,0 +1,351 @@
...
+    const { app, supportedDisclosureTypes } = this.props;
+    const { disclosureComponentKeys, disclosureComponentData, disclosureIsMaximized, disclosureSize } = this.state;
+
+    return disclosureComponentKeys.map((componentKey, index) => {
+      const componentData = disclosureComponentData[componentKey];
+      const isFullscreen = disclosureSize === 'fullscreen';
+
+      const popContent = this.generatePopFunction(componentData.key);
+      const appDelegate = AppDelegate.create({
+        disclose: (data) => {
```
mjhenkes: Maybe break off creating this function into a helper function to reduce the size of this method?

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -0,0 +1,351 @@
...
+          return app.disclose(data);
+        },
+        dismiss: index > 0 ? popContent : this.safelyCloseDisclosure,
+        closeDisclosure: this.safelyCloseDisclosure,
+        goBack: index > 0 ? popContent : undefined,
+        requestFocus: () => Promise.resolve().then(this.requestDisclosureFocus),
+        releaseFocus: () => Promise.resolve().then(this.releaseDisclosureFocus),
+        maximize: (!isFullscreen && !disclosureIsMaximized) ? () => (Promise.resolve().then(this.maximizeDisclosure)) : undefined,
+        minimize: (!isFullscreen && disclosureIsMaximized) ? () => (Promise.resolve().then(this.minimizeDisclosure)) : undefined,
+        registerDismissCheck: (checkFunc) => {
```
mjhenkes: Break this one off into it's own method too?

packages/terra-disclosure-manager/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2",
+    "terra-base": "^2.0.0"
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "prop-types": "^15.5.8",
+    "terra-base": "^2.0.0",
+    "terra-app-delegate": "^1.10.0"
```
jeffmerten: peer dependency?

packages/terra-disclosure-manager/docs/README.md
mjhenkes: I think the best example of the disclosure manager working would be the modal manager. We should point them there to look at that code. I'm not sure how many consumers will directly create another 'x type disclosure manager' outside of terra itself.

packages/terra-disclosure-manager/docs/README.md
tbiethman: I added subsections to divide up the Usage section; I think this reads a little more clearly when rendered.

packages/terra-disclosure-manager/docs/README.md
tbiethman: It is, these are the components that are supposed to be presented inside the disclosure mechanism. This components receive a different AppDelegate instance with more APIs (dismissal, maximization/minimization, focus stuff, dismiss checks, etc.). I'll update the description here to be more descriptive of the differences.

packages/terra-disclosure-manager/docs/README.md
tbiethman: Yep, I can clarify this.

packages/terra-disclosure-manager/docs/README.md
tbiethman: I'll remove my 1AM purple prose.

jeffmerten:
Do you have any examples of a ModalManager and SlidePanelManager both used in a single application?

packages/terra-disclosure-manager/docs/README.md
```diff
@@ -0,0 +1,209 @@
...
+### Rendering
+
+The DisclosureManager does not implement a traditional render function. A `render` prop should be provided to the DisclosureManager in order to render the various components managed by the DisclosureManager. The `render` function should accept an Object parameter containing the DisclosureManager's state.
+
+`render` Argument API:
+
+|Key|Value|
+|---|---|
+|`children`|An Object containing data relative to the children components provided to the DisclosureManager.|
+|`disclosure`|An Object containing data relative to the compoents in the disclosure stack.|
```
emilyrohrbough: typo, `compoents` to `components`

packages/terra-disclosure-manager/docs/README.md
```diff
@@ -0,0 +1,209 @@
...
+|`dismissPresentedComponent`|A function that pops the currently disclosed component off the disclosure stack.|
```
emilyrohrbough: A function that _props_?

packages/terra-disclosure-manager/docs/README.md
tbiethman: I agree that an fully working example included here could hurt more than help. Most consumers will not implement a DisclosureManager instance directly, and a working (yet un-opinionated) example included here would still be substandard in terms of actual usability. 

I can link out to the Modal-/SlidePanelManager for more reference, which will be more useful to consumers. If I do that, should this simple example be removed altogether? Or does seeing the props spelled out in their natural environment like this add just enough value to stick around?

packages/terra-disclosure-manager/docs/README.md
emilyrohrbough: From what I understand, this is saying The DisclosureManager takes `children` which must be of type object, then, children is passed as:
```
children ={{
    components: [
         <Content1 />,
         <Content2 />,
    ]
}}
``
Should children just take an array of components?

File: packages/terra-disclosure-manager/tests/jest/DisclosureManager.test.jsx (hunk 7 above)
Comment (tbiethman) [reply packages/terra-disclosure-manager/tests/jest/DisclosureManager.test.jsx]: It's been done now, I'll remove the comments.

File: packages/terra-disclosure-manager/docs/README.md [hunk 19]
```diff
@@ -0,0 +1,209 @@
...
+
+#### Children
+
+The AppDelegate provided to the child components contains a `disclose` function. This `disclose` function validates the disclosure type with which it is provided against the set of supported disclosure types given to the DisclosureManager as a prop. If the provided type is not supported, and if the DisclosureManager was given an AppDelegate prop to fall back to, the DisclosureManager will call the disclose function provided by its AppDelegate prop.
+
+If the type is supported, the DisclosureManager will check the currently disclosured content's state to ensure it can be replaced. If the disclosure is denied, then `disclose` returns a rejected Promise. If the disclosure is allowed, then a resolved Promise is returned. This Promise will be resolved with an Object containing functions and Promises that can be used to manipulate the disclosure, if necessary. Included are `dismissDisclosure`, a function that will dismiss the disclosed content, as well as `afterDismiss`, a deferred Promise that will be resolved when the disclosed content is dismissed by any means. Alternatively, if the additional logic isn't needed, the returned Promise can be completely ignored.
```
emilyrohrbough: currently _disclosed_ ~content's~ components state

emilyrohrbough: > This Promise will be resolved with an Object containing functions and Promises that can be used to manipulate the disclosure, if necessary.

I am a little unclear on this statement. Must this be provided or is this how the DisclosureManger is handling it.

emilyrohrbough: Included ~are~ is `dismissDisclosure`

packages/terra-disclosure-manager/docs/README.md
```diff
@@ -0,0 +1,209 @@
...
+    // If the dismissDisclosure function was cached, make sure it gets cleaned up.
+    this.myDismissFunction = false;
+  })
+}).catch((e) => {
+  // If this function is executed, the disclosure has been denied. You can use this to
+  // clean up any state that may be reliant on the disclosure ocurring.
+})
+```
+
+|Function|Description|
```
emilyrohrbough: Lets add a heading above this to make it clear what is being described

packages/terra-disclosure-manager/docs/README.md
```diff
@@ -0,0 +1,209 @@
...
+|component|A React element that will be disclosed.|
+
+#### Disclosure Content
+
+The AppDelegate instances provided to the disclosure components are a little more complicated. In addition to a `disclose` function (with all the bells and whistles described in the above section), a number of other functions are exposed to manage various segments of the disclosure state. The included functions are:
+
+|Function|Description|
+|---|---|
+|`disclose(Object)`|Allows a component to disclose another component on top of itself. See above for argument API.|
+|`dismiss()`|Allows a component to remove itself from the disclosure stack. If the component is the only element in the disclosure stack, the disclosure is closed.|
```
emilyrohrbough: some of these keys have `(-type-)` while others have `()` need to update to be consistent

packages/terra-disclosure-manager/docs/README.md
```diff
@@ -0,0 +1,209 @@
...
+|`requestFocus()`|Allows a component to request focus from the disclosure in the event that the disclosure mechanism in use utilizes a focus trap. This can be integrated with the Popup and similar focus-stealing controls.|
+|`requestFocus()`|Allows a component to release focus from itself and return it to the disclosure. This can be integrated with the Popup and similar focus-stealing controls.|
+|`registerDismissCheck(func)`|Allows a component to register a function with the DisclosureManager that will be called before the component is dismissed for any reason.|
+
+Each of these functions returns a Promise that can be used for chaining, if necessary.
+
+##### A Note on `registerDismissCheck`
+
+The function given to registerDismissCheck should return a resolved or rejected Promise. If the Promise is resolved, the component is guaranteed to be dismissed. If cleanup logic needs to execute before the component is dismissed, it is a good idea to execute before returning the resolved Promise. If a rejected Promise is returned, the component will not be dismissed. Components can render a prompt or confirmation window to give users control over the dismissal, if desired.
```
emilyrohrbough: Update to be `registerDismissCheck`

emilyrohrbough: > should return a resolved or rejected Promise

should or must?

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -0,0 +1,351 @@
...
+     * executed in stack order. Components will be dismissed in order up until a rejection occurs, at which point
+     * the blocking component will be presented.
+     */
+    return this.resolveDismissChecksInSequence(disclosureKeys).then(() => {
+      this.dismissChecks = {};
+      this.closeDisclosure();
+    });
+  }
+
+  generatePopFunction(key) {
```
emilyrohrbough: Given the complexity of this component, I think it could be valuable to add descriptions to any function that does more than set state

packages/terra-disclosure-manager/docs/README.md
emilyrohrbough: I do think referencing to Modal-/SlidePanelManager would be good, however, I think providing a simpler example may be beneficial as well. The examples on site use every possible prop options, with little distinguishing difference between the instances, which can be overwhelming.

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -0,0 +1,351 @@
...
+      disclosureIsMaximized: false,
+      disclosureSize: availableDisclosureSizes.SMALL,
+      disclosureComponentKeys: [],
+      disclosureComponentData: {},
+    });
+  }
+
+  requestDisclosureFocus() {
+    this.setState({
+      disclosureIsFocused: true,
```
jeffmerten: Anyway to avoid incurring a re-render with this update? Unless I'm mistaken, the lifecycle would be at the very least 
pushDisclosure and popDisclosure? 

Could how provide info on how many renders occur for the average disclosure?

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -0,0 +1,351 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import AppDelegate from 'terra-app-delegate';
+
+const availableDisclosureSizes = {
+  TINY: 'tiny',
+  SMALL: 'small',
+  MEDIUM: 'medium',
+  LARGE: 'large',
+  HUGE: 'huge',
```
jeffmerten: nit: Go with t-shirt sizes. Extra Large and extra small?

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -0,0 +1,351 @@
...
+    this.pushDisclosure = this.pushDisclosure.bind(this);
+    this.popDisclosure = this.popDisclosure.bind(this);
+    this.closeDisclosure = this.closeDisclosure.bind(this);
+    this.requestDisclosureFocus = this.requestDisclosureFocus.bind(this);
+    this.releaseDisclosureFocus = this.releaseDisclosureFocus.bind(this);
+    this.maximizeDisclosure = this.maximizeDisclosure.bind(this);
+    this.minimizeDisclosure = this.minimizeDisclosure.bind(this);
+
+    // These cached functions are stored outside of state to prevent unnecessary rerenders.
+    this.dismissChecks = {};
```
jeffmerten: Do you have any concern with these becoming unaligned with the state?

packages/terra-disclosure-manager/src/DisclosureManager.jsx
jeffmerten: Maybe have a DEFAULT one that is 'small'?

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -0,0 +1,351 @@
...
+    }));
+  }
+
+  renderDisclosureComponents() {
+    const { app, supportedDisclosureTypes } = this.props;
+    const { disclosureComponentKeys, disclosureComponentData, disclosureIsMaximized, disclosureIsFocused, disclosureSize } = this.state;
+
+    return disclosureComponentKeys.map((componentKey, index) => {
+      const componentData = disclosureComponentData[componentKey];
+      const isFullscreen = disclosureSize === 'fullscreen';
```
jeffmerten: nit: change to `availableDisclosureSizes.FULLSCREEN`

packages/terra-disclosure-manager/src/DisclosureManager.jsx
tbiethman: The registration of one of these functions shouldn't be a renderable event, which is why they're not stored in state. However, this object is cleaned up with the other state changing events to ensure state functions aren't hanging around.

packages/terra-disclosure-manager/docs/README.md
jeffmerten: > Most consumers will not implement a DisclosureManager instance directly

Are you saying this module should never be used by a module outside of terra-framework?

packages/terra-disclosure-manager/src/DisclosureManager.jsx
jeffmerten: ah, nevermind then.

packages/terra-disclosure-manager/docs/README.md
mjhenkes: Most likely yes, We'd expect consumers to use the modal manager or the slide panel manager. This is an abstraction that allows the guts of those components to be common. That said, consumers could conceivably use this to create their own special disclosure manager and we will still release this as it's own consumable package, as the slide panel and modal managers are.

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -0,0 +1,401 @@
...
+  }
+
+  disclosureTypeIsSupported(type) {
+    const { app, supportedDisclosureTypes } = this.props;
+
+    return supportedDisclosureTypes.indexOf(type) >= 0 || !app;
+  }
+
+  openDisclosure(data) {
+    this.setState({
```
jeffmerten: Certainly not something that needs to be addressed with this pr, but should a repeated disclosure event trigger a close (like tapping a row twice)?

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -0,0 +1,401 @@
...
+        .then(() => {
+          this.popDisclosure();
+        });
+    };
+  }
+
+  renderContentComponents() {
+    const { children, app } = this.props;
+
+    const appDelegate = {};
```
jeffmerten: Should this clone app?

tbiethman: Whoops, nice catch

jeffmerten: Same question in renderDisclosureComponents()

tbiethman: We could, but we're providing new functions for all AppDelegate APIs so nothing would actually be carried forward.

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -0,0 +1,416 @@
...
+  renderDisclosureComponents() {
+    const { app } = this.props;
+    const { disclosureComponentKeys, disclosureComponentData, disclosureIsMaximized, disclosureIsFocused, disclosureSize } = this.state;
+
+    return disclosureComponentKeys.map((componentKey, index) => {
+      const componentData = disclosureComponentData[componentKey];
+      const isFullscreen = disclosureSize === availableDisclosureSizes.FULLSCREEN;
+      const popContent = this.generatePopFunction(componentData.key);
+
+      const appDelegate = {};
```
jeffmerten: Consider renaming this or `app` so it's easier to make sense of the madness in this method.

packages/terra-disclosure-manager/src/DisclosureManager.jsx
tbiethman: There is only one state change for a given disclosure manager event, so there should only be one render at the DisclosureManager level. However, both the children and disclosure content are rendered, regardless of where the disclosure is actually happening.

I think we could be more performant by caching the built child and disclosure components in state; however, doing so would make a complicated implementation even more complicated. Let's optimize in the future if we identify bottlenecks. The important thing is that we have (passive) ways of doing so.

packages/terra-disclosure-manager/src/DisclosureManager.jsx
tbiethman: I think implementing such logic within the DisclosureManager itself would muddy up the APIs a bit. We should find a way to implement and share that behavior elsewhere. 

The only way we'd be able to do that here is by comparing the disclosure keys, which doesn't seem like a foolproof solution.

jeffmerten: Yeah, hopefully elsewhere isn't everywhere except the DisclosureManager (like in every component implementation). We can figure out that later if need be. Just a thought.

packages/terra-disclosure-manager/README.md
bjankord: Logged https://github.com/cerner/terra-framework/issues/65 to follow up on this issue.

PR #58, opened 2018-02-01
Adding dismiss check registration to AppDelegate

tbiethman:
### Summary
We're adding a new function to the AppDelegate that will allow components within disclosure managers to be notified of and prevent their dismissal.

I updated to the disclose API documentation as well to expose the new `component` key and mark parts of the API as deprecated. If we want to communicate the deprecation in a different way, I'm open to suggestions.

I also updated the documentation to remove references to Redux, because we're no longer using it for the disclosure managers, and it probably shouldn't have been mentioned in the first place.


packages/terra-app-delegate/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ Changelog
 Unreleased
 ----------
 
+### Added
+* Added registerDismissLock
+
+### Changed
+* Updated README to reflect new API
```
ryanthemanuel: Will this need to be a major version bump?

tbiethman: The change itself is passive. I'm leaving the old APIs in place, and the DisclosureManager will handle the old stuff appropriately. I'm content with marking things as deprecated and removing at them at the next-next version bump (after the React 16 bump). Or we can turn around immediately after React 16 is released and bump things again.

packages/terra-app-delegate/docs/README.md
```diff
@@ -78,15 +82,12 @@ Disclose API:
...
-|key|String|An identifying String that will be used as the React `key` for the disclosed component. Should be as specific as possible.|
-|name|String|The String value that will be used to retrieve the Component class from the AppDelegate component registry.|
-|props|Object|An Object containing the props for the disclosed component. These should be serializable.|
-
-The above structure allows us to disclose components dynamically without having to actually pass complex refs/functions through the API. This allows the parameters to be serializable and stored in Redux stores easily.
-
-The AppDelegate exposes two functions (`registerComponentForDisclosure` and `getComponentForDisclosure`) that allows components to be registered for and retrieved by an arbitrary key. Those functions should be used by components to build the disclosable components from the `disclose` arguments.
+|Key|Type|Required|Description|
+|---|---|---|---|
+|`key`|String|Yes|An identifying String that will be used as the React `key` for the disclosed component. Should be as specific as possible.|
```
emilyrohrbough: I'm assuming unique as well as well?

packages/terra-app-delegate/docs/README.md
```diff
@@ -145,3 +134,7 @@ ExampleRootComponent.propTypes = {
 }
 
 ```
+
+### Deprecated APIs
+
+The Disclose API for specifying which component to present was previously structured to support Redux-based disclosure managers. However, the disclosure managers were transitioned away from Redux to allow consumers to have greater control over the disclosure lifecycle. Therefore, the `name` and `props` keys in the disclose API are now deprecated, and the `component` key should be used to provide a component instance to present instead. This also means that the `registerComponentForDisclosure`/`getComponentForDisclosure` are deprecated as well, as there is no need for component retreival when an instance is given through the API. Those APIs will continue to function, but the usage of the `component` key is recommended.
```
emilyrohrbough: > Therefore, the `name` and `props` keys in the disclose API are now deprecated, and the `component` key should be used to provide a component instance to present instead.

This this be an enhancement we will need to make to terra-site?

tbiethman: If terra-site is using the deprecated APIs, things will still work for the time being. I'm not sure what our time frames would be for removing the deprecated APIs, but it will be at least after the React 16 bump.

packages/terra-app-delegate/docs/README.md
tbiethman: Yeah I can update this a bit.

packages/terra-app-delegate/docs/README.md
bjankord: Yeah let's try and keep the bump to react 16 as passive as possible. Post react 16 release, we can remove the deprecated APIs

jeffmerten: > Therefore, the `name` and `props` keys in the disclose API are now deprecated, and the `component` key should be used to provide a component instance to present instead.

The name and props approach works better than the component approach for ReactOnRails rendering...although I don't think that should be a dealbreaker as I'm fine with a slightly different API for the hybrid disclosures (we already need to know whether an application or container class should be disclosed).

+1 on deprecated registerComponentForDisclosure/getComponentForDisclosure though.

PR #70, opened 2018-02-09
Update layout hidden panel

### LINKED ISSUE #68: Enhance terra-layout to hide content in the menu when off screen.
# Issue Description
Set hidden styles on the layout's slide menu to prevent tab focus from component off of the screen.

## Issue Type
- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
When the value for isOpen is set to false, the content inside the layout menu should not be focusable, so hidden styles should be applied to the whole panel.

dkasper-was-taken:
### Summary
Update layout to hide it's content when off the page. This prevents the elements from being accessible by tabbing.

Fixes #68 

@cerner/terra


PaulSolDev:
Functional verification complete.  No issues identified.

PR #74, opened 2018-02-14
Correct blur on lower resolution monitors, and update screenshots

dkasper-was-taken:
### Summary
Correct blur on lower resolution monitors, and update screenshots to match styling without base link focus style.

This fixes the build issues on travis ci after the cache was cleaned.

@cerner/terra

### Deployment URL
https://terra-framework-deployed-pr-74.herokuapp.com/static/#/site/components/application-links/index


packages/terra-application-links/src/list/ApplicationList.scss
```diff
@@ -4,9 +4,7 @@
     color: var(--terra-application-list-color, #1c1f21);
     display: block;
     font-size: var(--terra-application-list-font-size, 1.071rem);
-    line-height: var(--terra-application-list-line-height, 3.214rem);
-    padding-left: var(--terra-application-list-padding-left 1.2143rem);
-    padding-right: var(--terra-application-list-padding-right, 0.7143rem);
+    padding: var(--terra-application-list-padding, 0.8225rem 0.7143rem 0.8225rem 1.2143rem);
```
bjankord: This will work fine in RTL since both left and right paddings are equal in this default theme. There may be a concern in other themes, if the left and right padding are not equal, not entirely sure how the PostCSS RTL plugin will handle this.

packages/terra-application-links/src/list/ApplicationList.scss
```diff
@@ -26,8 +24,8 @@
 
     &[aria-current='true'] {
       background-image: var(--terra-application-list-selected-background-image, linear-gradient(-270deg, #fff, #e8e9ea));
-      border-left: var(--terra-application-list-selected-border-left 0.5rem solid #007cc3);
-      padding-left: var(--terra-application-list-padding-left 0.7143rem);
+      border-left: var(--terra-application-list-selected-border-left, 0.5rem solid #007cc3);
+      padding: var(--terra-application-list-selected-padding, 0.8225rem 0.7143rem);
```
bjankord: This will work fine in RTL since both left and right paddings are equal in this default theme. There may be a concern in other themes, if the left and right padding are not equal, not entirely sure how the PostCSS RTL plugin will handle this.

packages/terra-application-links/src/list/ApplicationList.scss
dkasper-was-taken: Would you suggest just splitting the props into 4?

dkasper-was-taken: These two aren't equal left and right values.

bjankord: If it becomes an issue, we can split padding into into individual props. I'm not too worried about it.

dkasper-was-taken: Since it would be a non-passive theming change i was wondering if we should just do it now before release.

dkasper-was-taken: I can make another pr for it, that won't be as priority to get in.

PR #80, opened 2018-02-15
Remove hover style at medium breakpoint.

dkasper-was-taken:
### Summary
Remove hover style at medium breakpoint. Note that this is a UX request to remove the reveal-on-hover behavior from the left navigation menu. At larger breakpoints the nav menu simply displays, but at smaller breakpoints the nav menu is hidden with the option to toggle show/hide.

Deployment link:
https://terra-framework-deployed-pr-80.herokuapp.com/static/#/site/components/layout/index

@cerner/terra


packages/terra-layout/examples/test-examples/LayoutDefault.jsx
```diff
@@ -6,6 +6,7 @@ import TestContent from './test-content/TestContent';
 
 const LayoutDefault = () => (
   <Layout
+    style={{ height: '768px' }}
```
emilyrohrbough: you should be able to remove these heights now.

packages/terra-layout/src/_LayoutSlidePanel.jsx
```diff
@@ -148,29 +141,14 @@ class LayoutSlidePanel extends React.Component {
...
-      </div>
-    );
-
     return (
       <div
         {...customProps}
         className={slidePanelClassNames}
       >
-        {panel}
+        <div className={panelClasses} aria-hidden={this.isHidden ? 'true' : 'false'} ref={this.setPanelNode}>
```
emilyrohrbough: do we want to keep the quotes around true and false? Looks like you removed them above on [line 101](https://github.com/cerner/terra-framework/pull/80/files#diff-89a6c6532ba24f2bd1193b25680c573aR101)

bjankord: For html attributes that use the word true or false, e.g. aria-hidden="true", we should pass that value to the attribute as a string instead of a boolean. This is something react 16 seems to be more particular about. In react 15, we could pass true or false as a boolean type to the attribute and it would convert it to a string for us when passed to html attributes, or maybe the prop validation didn't really care. React 16 seems to throw prop errors now for invalid types if we pass boolean values to html attributes that use true or false.

packages/terra-layout/src/_LayoutSlidePanel.jsx
```diff
@@ -104,7 +98,7 @@ class LayoutSlidePanel extends React.Component {
     // As a result of this change, we are executing the code in the render block.
     if (this.props.isOpen && !this.lastIsOpen && this.panelNode) {
       // If the panel is opening remove the hidden attribute so the animation performs correctly.
-      this.contentNode.setAttribute('aria-hidden', 'false');
+      this.panelNode.setAttribute('aria-hidden', false);
```
emilyrohrbough: Based on @bjankord comment below, I think these should still be 'false'.

packages/terra-layout/src/_LayoutSlidePanel.jsx
```diff
@@ -88,13 +86,9 @@ class LayoutSlidePanel extends React.Component {
...
 
-  setContentNode(node) {
-    this.contentNode = node;
-  }
-
   handleTransitionEnd() {
-    if (!this.props.isOpen && this.contentNode) {
-      this.contentNode.setAttribute('aria-hidden', true);
+    if (!this.props.isOpen && this.panelNode) {
+      this.panelNode.setAttribute('aria-hidden', true);
```
emilyrohrbough: Based on @bjankord comment below, I think this should be 'true'.

packages/terra-layout/src/_LayoutSlidePanel.jsx
dkasper-was-taken: Updated.

mmalaker:
The following description is no longer correct:
"Small Desktop [medium] - The menu is disclosed using a Layout-provided hover mechanism. It can also be pinned open, at which point the menu will squish the content region instead of overlay it. Like the Large Desktop layout, the menu is presented under the header content."

Description should include medium with the rest of desktop, since it has the same behavior. Something like:
"Desktop [medium, large, huge] - The menu is pinned open and not dismissable. The content region is resized to account for the menu, and the menu is presented under the header content."

dkasper-was-taken:
@mmalaker Updated the documentation.

mmalaker:
Functional verification completed. The layout navigation menu example displays and functions correctly after removing the show-on-hover behavior at the medium breakpoint.

PR #86, opened 2018-02-19
Application links chevron

dkasper-was-taken:
### Summary
Add chevron style to the application-list version of links. Matches the implementation of terra-navigation-sub-menu.

@cerner/terra


packages/terra-application-links/src/list/ApplicationList.jsx
```diff
@@ -13,6 +14,7 @@ const propTypes = {
    * Navigational links that will generate list items that will update the path. These paths are matched with react-router to selection.
    */
   links: PropTypes.arrayOf(PropTypes.shape({
+    hasSubMenu: PropTypes.bool,
```
emilyrohrbough: In terra-site, the key [`hasSubNav`](https://github.com/cerner/terra-site/blob/master/src/config/navigation.config.js#L13) is used to determine if a ['more' indicator](https://github.com/cerner/terra-site/blob/master/src/app/common/menu/MenuList.jsx#L45) is needed and that the menus should drill to a sub menu. Should we align the terms or map them or change site's?

dkasper-was-taken:
#93 Makes this change superfluous. Closing.

Issue #77, opened 2018-02-15
Introduce terra-application-layout


tbiethman:
# Issue Description
We're going to reintroduce the terra-application package. Instead of the very simple wrapper that it previously exported, it will export a configurable Application component that utilizes the various 'terra-application-x' components and layouts that have been created in an opinionated fashion. This should be used as a starting point for application creation.

## Issue Type

- [X] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

bjankord:
I'm not really familiar with this intent or scope of this. My initial thought is this is probably outside of the scope of terra-ui. I'm thinking this is likely more a component named, orion-application which uses terra-ui components to compose the UI for the application.

mjhenkes:
The engineers working on this are getting close to delivering a pull request. I propose that they move forward with getting the pull request out and we can review intent, scope and naming in the war room on Friday.

bjankord:
Talked with @mjhenkes @ryanthemanuel @JakeLaCombe and reached consensus on moving forward with this as terra-application-layout.

mjhenkes:
merged with #116

PR #93, opened 2018-02-23
Update ApplicationLinks to use be buttons instead of NavLinks.

dkasper-was-taken:
### Summary
Convert tabs to use buttons instead of NavLinks and remove list style, as list functionality is rolled up into the side navigation menu.

@cerner/terra


packages/terra-application-links/src/tabs/_CollapsedTab.jsx
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * The path to push to the route.
+   */
+  path: PropTypes.string.isRequired,
+  /**
+   * The display text for the tab.
+   */
+  text: PropTypes.string.isRequired,
+  /**
+   * The display text for the tab.
```
tbiethman: Copy/paste comment

packages/terra-application-links/src/tabs/_Tab.jsx
```diff
@@ -20,33 +16,50 @@ const propTypes = {
    * The display text for the tab.
    */
   text: PropTypes.string.isRequired,
+  /**
+   * The display text for the tab.
```
tbiethman: Here too

packages/terra-application-links/src/tabs/_TabMenu.jsx
```diff
@@ -91,15 +91,16 @@ class TabMenu extends React.Component {
   }
 
   handleOnKeyDown(event) {
-    if (event.nativeEvent.keyCode === TabUtils.KEYCODES.ENTER && !this.state.isOpen) {
+    if ((event.nativeEvent.keyCode === TabUtils.KEYCODES.ENTER || event.nativeEvent.keyCode === TabUtils.KEYCODES.SPACE) && !this.state.isOpen) {
+      console.log('durp-2');
```
tbiethman: Probably don't need this in prod, though I would find it funny.

packages/terra-application-links/src/tabs/_CollapsedTab.jsx
```diff
@@ -0,0 +1,89 @@
...
+  path,
+  text,
+  ...customProps
+}) => (
+  <Route
+    render={({ location, history }) => {
+      const isActive = !!matchPath(location.pathname, { path });
+      const tabClassNames = cx([
+        'collapsed-tab',
+        { 'is-disabled': isActive },
```
emilyrohrbough: how does isActive map to is-disabled styling?

packages/terra-application-links/src/tabs/_CollapsedTab.jsx
```diff
@@ -0,0 +1,89 @@
...
+        customProps.className,
+      ]);
+      const tabAttr = { 'aria-current': isActive };
```
emilyrohrbough: did this need to be a string? or does a boolean value work here?

packages/terra-application-links/src/tabs/_CollapsedTab.jsx
```diff
@@ -0,0 +1,89 @@
...
+            onTabClick(event);
+          }
+        }
+      };
+
+      return (
+        /* eslint-disable jsx-a11y/no-static-element-interactions */
+        <div
+          {...customProps}
+          {...tabAttr}
```
emilyrohrbough: Could we just add `'aria-current': isActive` directly here?

packages/terra-application-links/src/tabs/_Tab.jsx
```diff
@@ -20,33 +16,50 @@ const propTypes = {
...
-  return (
-    <NavLink {...customProps} {...tabRole} className={tabClassNames} to={path} key={path} activeClassName={cx(['selected'])}>
-      <span className={cx(['tab-inner'])}>
-        {text}
-      </span>
-    </NavLink>
-  );
-};
+      return (
+        <button
```
emilyrohrbough: Do we want to use terra-button to enable consistent keyboard accessibility between browsers? or are all focus & active styling removed for these buttons?

packages/terra-application-links/src/ApplicationLinks.js
```diff
@@ -1,10 +1,25 @@
...
 
 const ApplicationLinks = {
   ApplicationTabs,
-  ApplicationList,
+  ApplicationLinksPropType,
 };
 
 export default ApplicationLinks;
-export { ApplicationTabs, ApplicationList };
+export { ApplicationTabs, ApplicationLinksPropType };
```
emilyrohrbough: Why are we exporting these proptypes?

emilyrohrbough:
Looks like the readme needs updated to remove the list example

packages/terra-application-links/src/ApplicationLinks.js
dkasper-was-taken: We are exporting them because header and application will both use them as part of their api.

packages/terra-application-links/src/tabs/_CollapsedTab.jsx
dkasper-was-taken: React doesn't like aria-current, so this is just a way around it.

packages/terra-application-links/src/tabs/_Tab.jsx
dkasper-was-taken: Terra button no longer accepts children or truncates text, so we need to managed it ourselves, also the the styles require a lot of overrides, so we just went with a basic button.

packages/terra-application-links/src/tabs/_CollapsedTab.jsx
dkasper-was-taken: I might be able to remove this, I'll look into it.

packages/terra-application-links/src/tabs/_CollapsedTab.jsx
dkasper-was-taken: It works, but i might update it, just to remain consistent.

packages/terra-application-links/src/tabs/_CollapsedTab.jsx
```diff
@@ -0,0 +1,88 @@
...
+import 'terra-base/lib/baseStyles';
+import { Route, matchPath } from 'react-router-dom';
+import TabUtils from './_TabUtils';
+import styles from './ApplicationTabs.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /**
+   * Indicates if the tab is be to desplayed in the tab menu.
```
mjhenkes: Typo: desplayed

emilyrohrbough:
Out of curiosity, will this change allow us to catch and 'correct' anchor link routing within a page? [This issue](https://github.com/cerner/terra-dev-site/issues/19) is logged in terra-dev-site and I wonder if we would run into similar issues with this component as well.

dkasper-was-taken:
Looks like the travis-ci build is glitched, might need a reset for this PR.

emilyrohrbough:
@JakeLaCombe I think what has been happening here (and in other PRs) is deleting the entire reference directory for a clean wdio run causes the timestamps to update on the images and check in as a different image.

PaulSolDev:
Functional verification complete.  Notes a visual inconsistency with selection style on IE10 but after discussing with @dkasper-was-taken this is already being addressed in #77 .  No further issues noted.

PR #87, opened 2018-02-19
Add webdriver.io tests for terra-theme-provider

rakshith91:
### Summary
Consuming wdio tests for the package and getting rid of the nightwatch tests.
Resolves #85  

Thanks for contributing to Terra. 
@cerner/terra

[CONTRIBUTORS.md]: ../blob/master/CONTRIBUTORS.md
[License]: ../blob/master/LICENSE

Issue #83, opened 2018-02-16
Add webdriver.io tests for terra-application-menu-layout


rakshith91:
# Issue Description
Consume wdio tests in terra-application-menu-layout and remove the existing nightwatch tests.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

PR #82, opened 2018-02-16
Add webdriver.io tests for terra-application-header-layout

rakshith91:
### Summary
Consuming wdio tests for the package and getting rid of the nightwatch tests.
Resolves #78 .

Thanks for contributing to Terra. 
@cerner/terra

[CONTRIBUTORS.md]: ../blob/master/CONTRIBUTORS.md
[License]: ../blob/master/LICENSE

PR #94, opened 2018-02-26
Add webdriver.io tests fro terra-navigation-layout and remove nightwatch dependency.

### LINKED ISSUE #88: Add webdriver.io tests for terra-navigation-layout and remove other nightwatch tests
# Issue Description

- Add webdriver.io tests to `terra-navigation-layout`
- Remove nightwatch tests from `terra-navigation-layout`, `terra-layout`, `terra-modal-manager`, `terra-application-utility` and `terra-embedded-component`.
- Remove nightwatch script from package.json and consume only jest and wdio tests during test.
 
## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

rakshith91:
### Summary

- Added wdio tests for `terra-navigation-layout`
- Removed nightwatch tests from packages that consumed wdio tests.
- Removed nightwatch dependency and scripts. 

### Additional Details
Resolves #88 
Blocked by PRs #87  #83  #82 

@cerner/terra


package.json
```diff
@@ -69,7 +69,7 @@
     "start": "webpack-dev-server --config node_modules/terra-site/src/config/webpack.config --progress",
     "start:build": "npm run generate-config && npm run start",
     "start:express": "node scripts/express/app.js",
-    "test": "jest && nightwatch && wdio",
```
emilyrohrbough: Awesome!

packages/terra-navigation-layout/tests/wdio/navigation-layout-spec.js
```diff
@@ -0,0 +1,41 @@
+/* global browser, Terra, before */
+
+const viewports = Terra.viewports('tiny', 'small', 'medium', 'large', 'huge');
+const selector = '#test-root';
```
emilyrohrbough: At this point, we should need to add a selector for these tests. The default should work fine.

rakshith91: The default selector does not work since React 16 upgrade. This error occurs: 
```
boundingRects.reduce is not a function
running chrome
TypeError: boundingRects.reduce is not a function
    at execute(<Function>) - selectorExecute.js:58:65
    at execute("return !!document.evaluate;") - ensureClientSideSelectorSupport.js:23:17
```

JakeLaCombe: You can update it in your wdio.conf.js file like so. (https://github.com/cerner/terra-core/blob/master/wdio.conf.js#L43). Will that work for you?

And then once it's on the conf file, I believe you can remove this selector.

emilyrohrbough: @JakeLaCombe we have a [default selector](https://github.com/cerner/terra-framework/blob/master/wdio.conf.js#L34) specified in the wdio.config. I have found there are instances when content is 'hidden' or animating off the page and it appears the element is larger than the document size which throws this error. Given we want the screenshot of the entire page, I feel this selector is fine in this instance.

PR #122, opened 2018-03-21
Remove ApplicationUtility internally managed selection state.

dkasper-was-taken:
### Summary
Remove the internally managed selection state and always dismiss the popup when an action takes place.

@cerner/terra


packages/terra-application-utility/CHANGELOG.md
```diff
@@ -4,3 +4,4 @@ ChangeLog
 Unreleased
 -----------------
 * Initial stable release.
+* Remove internal selection state logic.
```
JakeLaCombe: Is this released? If not, it feels like we don't really need to add this. Perhaps there is a way we can bypass the danger file?

emilyrohrbough: bypass danger by removing the `.` from ` * Initial stable release.` 😄

Issue #135, opened 2018-04-03
Update disclosure manager sizes


dkasper-was-taken:
# Issue Description
Current sizes for modal manager and slide manager are a limited subset. Consumers are requesting additional options for modal and slide panel sizes.

## Issue Type
- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

PaulSolDev:
JIRA created

PR #145, opened 2018-04-09
Add dimensions to the diclosure, modal, and slide-panel managers.

dkasper-was-taken:
### Summary
Added placeholder dimensions for the disclosure pattern. WDIO tests are waiting on the final sizes.

Resolves: #135 

@cerner/terra


dkasper-was-taken:
Adding 'do not merge' label until real sizes added.

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -97,10 +130,24 @@ class DisclosureManager extends React.Component {
   }
 
   openDisclosure(data) {
+    let dimensions = data.dimensions;
+    if (dimensions && !isValidDimensions(dimensions)) {
+      dimensions = defaultDimensions;
+    }
+
+    let size = data.size;
+    if (!size || (size && !isValidSize(size))) {
```
tbiethman: You could handle an undefined size in `isValidSize` and simplify this.

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -97,10 +130,24 @@ class DisclosureManager extends React.Component {
...
+      if (!dimensions) {
```
tbiethman: Add a comment about why this dimensions check is here.

packages/terra-modal-manager/src/ModalManager.jsx
```diff
@@ -20,36 +19,27 @@ const propTypes = {
...
     this.renderModal = this.renderModal.bind(this);
-
-    this.state = {
-      forceFullscreen: false,
-    };
-  }
-
-  componentDidMount() {
-    this.updateFullscreenState();
-    window.addEventListener('resize', this.updateFullscreenState);
```
tbiethman: wooot

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -11,7 +11,39 @@ const availableDisclosureSizes = {
   FULLSCREEN: 'fullscreen',
 };
 
-export { availableDisclosureSizes };
+const availableDisclosureHeights = {
```
JakeLaCombe: Minor:

Perhaps we should capitalize these constants as well?

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -97,10 +130,24 @@ class DisclosureManager extends React.Component {
   }
 
   openDisclosure(data) {
+    let dimensions = data.dimensions;
+    if (dimensions && !isValidDimensions(dimensions)) {
+      dimensions = defaultDimensions;
```
JakeLaCombe: Should we do a console warning  in development mode here if in valid dimensions are passed in, and that it's being defaulted?

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -11,7 +11,39 @@ const availableDisclosureSizes = {
...
-export { availableDisclosureSizes };
+const availableDisclosureHeights = {
+  240: 240,
+  420: 420,
+  600: 600,
+  690: 690,
+  780: 780,
+  870: 870,
+  960: 960,
+  1140: 1140,
```
nickpith: Can someone explain to why consumers of the ModalManager should have to specify a height at all? 

I can understand the width but why does the ModalManager care about height? As a consumer trying to use the ModalManager, I have to know my content size in order to know what height of a modal to specify. That is not always known. Why doesn't ModalManager just size the modal to the content being presented in it and have it max out so it doesn't go past the window's height?

nickpith: Your essentially pushing the hard work to all the consumers of the component. That's what I'm against. I can understand that it is hard but when you use a modal on other platforms like iOS or Android. You don't have to say what size you want. It just works because those platforms take care of the hard stuff for you.

What if a component's size is not easily deterministic on height and varies based on things like translations or content available like a list of things? I'm just supposed to pick one and if it is smaller than that size then it will have extra space.

dkasper-was-taken: Content presented with either a large data set or presented on a small form factor will be bound by the viewport size and result in the content overflowing. This presents a problem for header or footer aligned elements that we intend to remain fixed in their positions. 

If rendering of the header and footer was controlled by the Modal, we could be a little more flexible in our sizing. The presented component’s height could be used to determine whether maximum/minimum values have been met and updated sizes accordingly. However, it is likely that other “sticky” elements will want to be presented in the content region (toolbars, banners, etc.), resulting in the same problem now lower in the component hierarchy. 

On top of that, many consumers require very custom header/footer implementations that we couldn’t easily replicate with a genericized API. If we break up the Modal’s API to pass separate elements (i.e. buttons, actions, footers) that could be used to construct the final component, the component becomes harder to use and puts much greater responsibility on the disclosing component.

If we assume that header/footer rendering needs to be handled by the disclosed component, we’re left with three options for modal sizing:

1: Fixed sizes
Pros:
 - Users can provide their own headers/footers with no additional javascript (assuming the ContentContainer is used for header/footer positioning).
 - Consistent visual sizes for designers.
 - Disclosed components are responsive by default (or should be, in order to render appropriately for all sizes).
 - Modal size can remain consistent when navigating across a modal workflow (i.e. when a component inside a modal wants to present another component modally).
 - Components can be presented in different progressive disclosure mechanisms without any dynamic rendering.
Cons:
 - The component launching the modal must know how large the disclosed component will typically be.
 - White space or scrolling can occur if the size is not adequate for the disclosed content.

2: Dynamic heights
Pros:
 - Size fits to content automatically (no unnecessary whitespace or overflows).
 - No added responsibility for the component launching the modal since the size is automatically determined.
Cons:
 - Each modal content component needs to add resize listeners and custom scripts to appropriately position headers/footers.
 - Requires double renders to present (one to capture the standard size, another to bound the size if the standard size is beyond any present constraints).
 - Visual inconsistencies in modal size standards.
 - Disclosed components are not responsive by default. They could be written that way, but there’s no explicit requirement to enforce it.
 - Modal size can be inconsistent when navigating across a modal workflow.
 - Disclosed components needs to understand where they are being presented in order to size themselves appropriately.
- User implementing component needs to be aware of themeable AbstractModal inset values.

3: Resize javascript with child cloning
Pros:
 - Size fits to content.
 - Default Responsive as long as decorated props are handled properly.
Cons:
 - If content is dynamic, navigation across modal workflows will behave inconsistently.
 - Users need to correctly implement decorated props for every presented component (historically easily overlooked/abused with Popup).
 - Requires double renders to present (one to capture the standard size, another to bound the size if the standard size is beyond any present constraints).
 - Visual inconsistencies in modal size standards.
 - Disclosed components needs to understand where they are being presented in order to size themselves appropriately.
- Modal logic needs to be aware of themeable AbstractModal inset values.

nickpith: > On top of that, many consumers require very custom header/footer implementations that we couldn’t easily replicate with a genericized API.

It sounds like there may need to be some specialized components for these use cases rather than developing one generic API. I may not be aware of all the uses cases you elude to with respect to the "sticky" headers and footers. I thought a modal has a header and footer but I guess I assume Modal = Dialog (which has header and footer). 

What use cases are modal's used that are not using the [Dialog component](https://engineering.cerner.com/terra-ui/#/site/components/core/site/dialog/index)?

My main point is why are our modals and disclosures so complex to use when compared with things like https://elastic.github.io/eui/#/layout/modal, https://react-bootstrap.github.io/components/modal/ & http://www.material-ui.com/#/components/dialog?

From a complete outsiders perspective looking at these for the first time, all the disclosure stuff is very confusing and not very user friendly to developers who are using terra-framework. It feels like we have made it much more complex than what it really needs to be.

mjhenkes: Hey @nickpith, thanks for the feedback. As you can tell by the thesis above, we've spent a lot of time designing the modal and thinking through our use-cases and UX requirements. I know that the modal may seem overly complex, but providing an application grade modal that helps consumers avoid pitfalls is no small task. 

The modals you've linked to each have their own unique spin on the modal workflow and each have their own pitfalls, the modal scrolls off screen, the header is not appropriately customizable and their APIs are not simple. I know that our API is different, but we're pretty comfortable with where we've landed with the Modal and the level of complexity.

I'd appreciate if you'd try out the modal as we've designed it with providing responsive content. If it still seems like too much, you can fall back to providing your own styles on top of our abstract modal. 

http://engineering.cerner.com/terra-core/#/site/components/site/abstract-modal/index

nickpith: Can you point me to some examples of these use cases in applications or UX designs?

It would help me get my head around the history of why things are this way.

bjankord: Discussed this thread in our post-scrum today. We've identified needs for 4 distinct modals.
* abstract modal (powers underlying portal and positioning of all modals)
* prompt (A designed modal similar to native Javascript's window.alert or window.prompt methods)
* modal manager (handles complex workflows like wizard / multiple modal workflows, nested popups, etc)
* dialog modal (handles dynamic height modals, users wouldn't need to select a preset height)

Adding a topic to meeting on Friday to flush this out in more detail.

packages/terra-modal-manager/docs/README.md
```diff
@@ -15,15 +19,9 @@ The ModalManager utilizes the AppDelegate API to manage disclosure requests. The
...
-### Deprecations
-
-The ModalManager prior to version 1.20.0 utilized Redux to store its disclosure state. With the transition to a DiscosureManager-based implementation, Redux will no longer be used. As such, the provided actions and reducers are now deprecated. They are still included with the package for posterity, but they will be removed in a future major release.
-
-If your component currently implements a ModalManager, you can safely remove the ModalManager reducer from your reducer tree.
-
 ### withModalManager
 
-The main export from the `terra-modal-manager` package also exports a higher-order component generator called `withModalManager`. The component provided to the generator will be wrapped by a ModalManager and provided with an `app` prop.
+The main export from the `terra-modal-manager` package also exports a higher-order component generator called `withModalManager`. The component provided to the generator will be wrapped by a Modal and provided with an `app` prop.
```
mjhenkes: Change modal back to modal manager

packages/terra-modal-manager/tests/jest/ModalManager.test.jsx
```diff
@@ -5,7 +5,7 @@ const TestContainer = () => (
   <div>Hello World</div>
 );
 
-describe('ModalManger', () => {
+describe('ModalManager', () => {
```
mjhenkes: 😆

PR #147, opened 2018-04-09
[Hookshot] - Update attachment position none to be unbound

### LINKED ISSUE #146: [Hookshot] - Update attachment position none to be unbound
# Issue Description

Hookshot currently defaults the bounding region to the window. There are use cases where content should remain statically positioned to the specified attachment points even if the content is scrolled off screen.

## Issue Type

- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior

This enhancement should update the hookshot attachment behavior API to allow the behavior `none` to statically position content even if it is scrolled offscreen. A new behavior of `push` should be added to reposition content based on the bounded region.

## Current Behavior

Hookshot always binds to a region. Content cannot be position out of the visible viewport.

StephenEsser:
### Summary
This change updates the hookshot attachment behavior `none` to position content without a bounding region. This will allow content to scroll off screen and prevent content from being pushed to overlay the target.

An additional behavior of `push` has been added as a replacement for the behavior mentioned above. `push` will reposition content within a bounded region by pushing the content as it begins to move off screen.

### Additional Details
**This change requires a major version bump to terra-hookshot and terra-popup**

### Testing
The changes can be tested by setting the attachment behavior on the Hookshot example page to `push` and `none` and scrolling the screen to observe the repositioning behavior.

`none` - Should remain attached to the original target point even while scrolled off screen
`push` - Should reposition as it begins to scroll off screen. A pushing effect of the bound window should be observed.

Resolves #146 

@cerner/terra


packages/terra-application-layout/src/header/_ApplicationHeader.jsx
```diff
@@ -202,7 +202,7 @@ class ApplicationHeader extends React.Component {
     if (utilityComponent) {
       return (
         <Popup
-          attachmentBehavior="none"
+          attachmentBehavior="push"
```
StephenEsser: This change is worth pointing out. This should be passive for the ApplicationHeader due to being hardcoded.

PR #156, opened 2018-04-17
Consume changes from terra-toolkit and terra-dev-sever

mjhenkes:
### Summary
This PR pull's in the new terra-toolkit changes and terra-dev-site changes.
More deletions than additions. :)

@cerner/terra


package.json
```diff
@@ -44,11 +44,11 @@
     "clean:obsolete-snapshots": "npm test -- -u",
     "compile": "npm run compile:src",
     "compile:src": "lerna run compile",
-    "compile:prod": "NODE_ENV=production node --max_old_space_size=2500 node_modules/webpack/bin/webpack.js --config webpack.prod.config.js -p",
+    "compile:prod": "NODE_ENV=production node --max_old_space_size=3000 node_modules/webpack/bin/webpack.js --config node_modules/terra-dev-site/src/webpack/webpack.config.js -p",
```
mjhenkes: Anyone have a good idea for aliasing out this file to something more friendly without having to include a webpack config locally?
`node_modules/terra-dev-site/src/webpack/webpack.config.js`

emilyrohrbough: maybe provide a terra-dev-site script that will wrap webpack such that it takes args ?.
Thinking

```compile:prod: 'NODE_ENV=production node --max_old_space_size=3000 tds-prode'```

which maps to 

tds-prod.js which calls 

```shelljs(webpack.js --config ../../src/webpack/webpack.config.js -p)```

Comment (emilyrohrbough) [reply package.json]: Actually should we be using the servestatic script here by chance?

Comment (mjhenkes) [reply package.json]: I've thought about wrapping, but I want to avoid wrapping webpack because i don't want ot duplicate all the different webpack config options that exist (there are a ton).

Serve static could be used, but creating the artifacts is a by-product of running the server and there are cases where we just want the build output (deploying to gh-pages).

File: package.json [hunk 2]
```diff
@@ -38,17 +38,17 @@
     }
   },
   "scripts": {
-    "bootstrap": "lerna bootstrap",
+    "bootstrap": "lerna bootstrap --hoist",
```
bjankord: I'd hold of on using the --hoist flag until we can verify it works with the theme repos. Right now, I know it works with terra-ui, but I wasn't able to get it working the cerner-consumer-theme repo.

mjhenkes: Sounds good, but we should pause this PR until we can verify `--hoist` works. It significantly lowers the memory footprint of the dev site.

packages/terra-application-header-layout/package.json
```diff
@@ -46,6 +46,6 @@
     "lint:scss": "stylelint src/**/*.scss",
     "test": "npm run test:jest && npm run test:wdio",
     "test:jest": "jest ./tests/jest/* --config ../../jestconfig.json",
-    "test:wdio": "wdio ../../wdio.conf.js"
+    "test:wdio": "../../node_modules/.bin/wdio ../../node_modules/terra-dev-site/config/wdio/wdio.conf.js"
```
emilyrohrbough: We should make this change in the generator

mjhenkes: Agreed, but the downside is that it's different between core and framework because core uses it's own wdio config and framework uses the default.

emilyrohrbough: 😞 yeah...

PR #162, opened 2018-04-24
Create Terra Notification Dialog component 

### LINKED ISSUE #136: Create "notification dialog" component
# Issue Description
Create a prompt type component that can be presented over the top of all z-index. Built upon terra-modal, but outside of the modal manager stack. A common component to be used for confirmation/acceptance criteria style dialogs.

## Issue Type
- [x] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The dialog should present above the highest z-indexes of modal's 6000 and popup's 7000. A common visual pattern and size should be used for this component, and content should be limited, as we are placing a component outside of visual stacks. This needs design input as to what the limited content should be.

### LINKED ISSUE #190: [terra-infinite-list] InfiniteList.Item not re-rendering 'content' component
# Issue Description
Components assigned to the `content` prop of `InfiniteList.Item` do not re-render when their props change.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
When a prop is updated to a new value, then the component should re-render.

## Current Behavior
The custom component is not re-rendering.

## Steps to Reproduce
1. Create an `InfiniteList.Item` like so with a custom component:
```
<InfiniteList.Item
  key={myUniqueKey}
  content={MyCustomComponent myProp={123}}
/>
```
2. Update the prop of the custom component (e.g. `myProp={456}`).
3. Observe the infinite list does not re-render `MyCustomComponent`.

## Environment

* Component Version: 1.1.0
* Browser Name and Version: Chrome Version 66.0.3359.181
* Operating System and version (desktop or mobile): macOS 10.12.6

rakshith91:
### Summary
`terra-notification-dialog` is a notification dialog component built over the `terra-abstract-modal`. It has the highest z-index of 8001. It is a common component to be used for confirmation/acceptance criteria style dialogs. 

### Additional Details
Resolves #136 
Resolves #190 

### Sample Image
image

### Deployment Link
https://terra-framework-deploye-pr-162.herokuapp.com/#/site/components/notification-dialog/index

@cerner/terra


packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -0,0 +1,85 @@
...
+   */
+  message: PropTypes.string,
+  /**
+   * Array of buttons. Restricted to terra-button.
+   */
+  actions: PropTypes.arrayOf(elementType(Button)),
+  /**
+   * Toggle to show notification-dialog or not.
+   */
+  showNotificationDialog: PropTypes.bool.isRequired,
```
StephenEsser: I would gather a few opinions on this prop name. Many of our components use `isOpen` to address the open / closed state of a component.

packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -0,0 +1,85 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import AbstractModal from 'terra-abstract-modal';
+import Button from 'terra-button';
```
StephenEsser: This is a little tricky, the Notification Dialog itself doesn't actually use Buttons, it only needs the type check. I'm not sure if this component gains enough value from the type check to justify the hard Button dependency, potential package size increase, and multiple Button versions. Does anyone else have any thoughts on this?

packages/terra-notification-dialog/tests/jest/Prompt.test.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
```
StephenEsser: Many of these test files are named `Prompt`. I saw on the tech design and branch name this component was called `Prompt` but the code and package name look like they've been changed to `terra-notification-dialog`. I would update all occurrences of the conflicting names to match the desired package name.

packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -0,0 +1,85 @@
...
+      title,
+      message,
+      actions,
+      showNotificationDialog,
+      onRequestClose,
+      ...customProps
+    } = this.props;
+
+    return (
+      <div {...customProps} className={cx('notification-dialog')}>
```
StephenEsser: I almost think this outer div should be removed (if it can be) and the custom props spread onto the div with the class `notification-dialog-body`.

Right now this renders an empty div as a sibling to the triggering target, while all of the content is rendered into a portal somewhere else on the dom.

image

StephenEsser: The class names provided inside customProps would need to be merged into the className with notification-dialog. Here's an example: https://github.com/cerner/terra-core/blob/master/packages/terra-card/src/Card.jsx#L23

packages/terra-notification-dialog/src/NotificationDialog.jsx
rakshith91: One other option is to change the type checking to `componentWithName('Button')` .
This would allow any component with name Button and is not restricted to `terra-button`. Obviously, it's not foolproof though( anyone could rename any component to button ).

tbiethman: If we only want to accept Buttons here, would it make more sense to have the notification dialog render the buttons itself? Resulting in an API like:

```jsx
<NotificationDialog
  actions={[
    { text: 'OK', onClick: () => {} },
    { text: 'Cancel', onClick: () => {} }
  ],
  ...
/>
```

You could alternatively tie keys to each action and have a single selection function to filter them:
```jsx
<NotificationDialog
  actions={[
    { key: 'ok', text: 'OK' },
    { key: 'cancel', text: 'Cancel'}
  ],
  onSelect={(actionKey) => {
    if (actionKey === 'cancel') {
      // do something
    } else if (actionKey === 'ok') {
      // do something else
    }
  }}
/>
```
You may need more of the Button props exposed, but if we're really wanting to limit the actions to Buttons, it may be worth considering.

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,50 @@
...
+
+  .actions {
+    display: flex;
+    flex: 0 0 auto;
+    flex-direction: var(--terra-notification-dialog-actions-flex-direction-small-width, row);
+    justify-content: center;
+    padding: var(--terra-notification-dialog-actions-padding, 0.4rem);
+    width: 100%;
+
+    > button {
```
tbiethman: This is a brittle assumption (and more evidence that the NotificationDialog should be rendering the buttons itself).

packages/terra-notification-dialog/src/NotificationDialog.jsx
rakshith91: This sounds like an interesting solution. But till what extent do we want to restrict the keys?

yuderekyu:
screen shot 2018-04-30 at 10 25 00 am

When either the title or message is a continuous piece of text, the text overflows. Setting `word-wrap: break-word` should fix this edge case.

screen shot 2018-04-30 at 10 25 31 am

packages/terra-notification-dialog/README.md
```diff
@@ -0,0 +1,24 @@
+# Terra NotificationDialog
+
+[![NPM version](http://img.shields.io/npm/v/terra-notification-dialog.svg)](https://www.npmjs.org/package/terra-notification-dialog)
+[![Build Status](https://travis-ci.org/cerner/terra-framework.svg?branch=master)](https://travis-ci.org/cerner/terra-framework)
+
+`terra-notification-dialog` is a notification dialog component built over the `terra-abstract-modal`. It has the highest z-index of 8001. It is a common component to be used for confirmation/acceptance criteria style dialogs.
```
yuderekyu: *8000

packages/terra-notification-dialog/docs/README.md
```diff
@@ -0,0 +1,77 @@
+# Terra NotificationDialog
+
+`terra-notification-dialog` is a notification dialog component built over the `terra-abstract-modal`. It has the highest z-index of 8001. It is a common component to be used for confirmation/acceptance criteria style dialogs.
```
yuderekyu: *8000 here too

packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -0,0 +1,85 @@
...
+   */
+  title: PropTypes.string,
+  /**
+   * Message of the notification-dialog.
+   */
+  message: PropTypes.string,
+  /**
+   * Array of buttons. Restricted to terra-button.
+   */
+  actions: PropTypes.arrayOf(elementType(Button)),
```
yuderekyu: Curious, what's the reason for pulling in airbnb-prop-types to use`elementType`, versus using prop-type's `instanceOf`?

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,50 @@
+:local {
+  .notification-dialog {
+    align-items: center;
+    background: var(--notification-dialog-background, #fff);
+    border-radius: 0.6rem;
```
yuderekyu: Do we want `border-radius` themeable?

packages/terra-notification-dialog/tests/jest/NotificationDialog.test.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import CompleteNotificationDialog from '../../examples/test-examples/CompleteNotificationDialog.example';
+import CompleteNotificationDialogWithLongMessage from '../../examples/test-examples/CompleteNotificationDialogWithLongMessage.example';
+
+it('should shallow a complete notification-dialog', () => {
+  const modal = shallow(<CompleteNotificationDialog />);
+  expect(modal).toMatchSnapshot();
+});
+
+it('should mount an complete notification-dialog', () => {
```
yuderekyu: Nit: "a" instead of "an".

packages/terra-notification-dialog/tests/jest/NotificationDialog.test.jsx
```diff
@@ -0,0 +1,23 @@
...
+  const modal = mount(<CompleteNotificationDialog />);
+  expect(modal).toMatchSnapshot();
+});
+
+it('should shallow a complete notification-dialog with long message', () => {
+  const modal = shallow(<CompleteNotificationDialogWithLongMessage />);
+  expect(modal).toMatchSnapshot();
+});
+
+it('should mount an complete notification-dialog with long message', () => {
```
yuderekyu: Nit: "a" instead of "an".

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,50 @@
...
+    flex: 0 0 auto;
+    flex-direction: var(--terra-notification-dialog-actions-flex-direction-small-width, row);
+    justify-content: center;
+    padding: var(--terra-notification-dialog-actions-padding, 0.4rem);
+    width: 100%;
+
+    > button {
+      flex: var(--terra-notification-dialog-actions-button-flex, initial);
+      flex-shrink: 1;
+      margin: var(--terra-notification-dialog-actions-button-margin, 0.8rem 0.4rem 0 0.4rem);
```
yuderekyu: FYI, margin should be split up into separate properties.

packages/terra-notification-dialog/README.md
rakshith91: I think it's 8001.
https://github.com/cerner/terra-core/blob/master/packages/terra-abstract-modal/src/AbstractModal.scss

packages/terra-notification-dialog/src/NotificationDialog.jsx
rakshith91: Changed this to restrict using only two buttons.

packages/terra-notification-dialog/src/NotificationDialog.scss
rakshith91: made it themeable

packages/terra-notification-dialog/src/NotificationDialog.jsx
rakshith91: This is changed completely.

packages/terra-notification-dialog/README.md
yuderekyu: Ahhh I missed that

packages/terra-notification-dialog/src/NotificationDialog.scss
rakshith91: changed

packages/terra-notification-dialog/src/NotificationDialog.jsx
rakshith91: changed it to `isOpen` to maintain consistency.

packages/terra-notification-dialog/src/NotificationDialog.jsx
tbiethman: @neilpfeiffer Do we want to restrict the footer buttons this much? I see the value in controlling which button has emphasis, but it seems like a 3+ button footer is going to come up sooner or later.

packages/terra-notification-dialog/docs/README.md
```diff
@@ -0,0 +1,81 @@
...
+  handleCloseModal() {
+    this.setState({ showNotificationDialog: false });
+  }
+
+  render() {
+    return (
+      <div>
+        <NotificationDialog
+          variant={NotificationDialog.Opts.Variants.ALERT}
+          showNotificationDialog={this.state.showNotificationDialog}
```
tbiethman: I think this prop is out of date

packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -0,0 +1,173 @@
...
+const getIcon = (variant, customIcon = null) => {
+  switch (variant) {
+    case variants.ALERT:
+      return (<IconAlert />);
+    case variants.ERROR:
+      return (<IconError />);
+    case variants.WARNING:
+      return (<IconWarning />);
+    case variants.ADVISORY:
+      return null;
```
tbiethman: Do we need an explicit type for advisory if there's no icon for it and no separate logic for it?

packages/terra-notification-dialog/src/NotificationDialog.jsx
tbiethman: If we do only want two buttons, we could just expose two separate props to make that really clear:

```
<NotificationDialog
  primaryAction={{
    text: 'OK',
    onClick: () => {},
  }}
  secondaryAction={{
    text: 'Cancel',
    onClick: () => {},
  }}
>
```

neilpfeiffer: Thanks @tbiethman - @rakshith91 and I already discussed and I like taking the approach of more restrictive first (only 2 buttons) and only expand when we get requests (following Cory House's recommendation of being rigid to start https://youtu.be/MsTARVOl1H4?t=15m26s ).  

Current terra-alert (Notification Banners) restricts to only one custom action plus a dismiss, so it is safe to follow the same paradigm here to start as well as best practice, since the two are sister components with similar usage.

packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -0,0 +1,173 @@
...
+    );
+  }
+}
+
+NotificationDialog.propTypes = propTypes;
+NotificationDialog.defaultProps = defaultProps;
+NotificationDialog.Opts = {};
+NotificationDialog.Opts.Variants = variants;
+
+export default NotificationDialog;
```
tbiethman: This 'controlled' NotificationDialog allows for a lot of flexibility in the component's presentation. In addition to this, I think we should provide an 'uncontrolled' version that can be presented synchronously and managed outside of the calling component. This would be helpful to consumers who want to inject the notification dialog into Promise chains and control logic flow with the action selections.

Something like the confirmAlert function as seen here: https://github.com/GA-MO/react-confirm-alert/blob/master/src/index.js

I'm not sure if this needs to happen with the initial PR, but I do think it would be helpful.

packages/terra-notification-dialog/src/NotificationDialog.jsx
rakshith91: Removed it.

packages/terra-notification-dialog/src/NotificationDialog.jsx
rakshith91: Changed the API to pass `primaryAction` and `SecondaryAction` instead.

packages/terra-notification-dialog/src/NotificationDialog.jsx
rakshith91: https://github.com/cerner/terra-framework/issues/177
Logged an issue for now. I think this would be a good enhancement issue rather than pushing changes in the initial PR.

packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -0,0 +1,184 @@
...
+        <div className={cx('footer-body')}>{actionSection(primaryAction, secondaryAction)}</div>
+      </AbstractModal>
+    );
+  }
+}
+
+NotificationDialog.propTypes = propTypes;
+NotificationDialog.defaultProps = defaultProps;
+NotificationDialog.contextTypes = contextTypes;
+NotificationDialog.Opts = {};
```
bjankord: Instead of attaching Opts and Variants to NotificationDialog, we've landed exporting them as a separate export from the component file. This can be replaced with the following code:
```
export {  NotificationDialogVariants: Variants };
```

packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -0,0 +1,184 @@
...
+  /**
+   * The Action of the secondary button.
+   */
+  secondaryAction: PropTypes.shape({
+    text: PropTypes.string,
+    onClick: PropTypes.func,
+  }),
+  /**
+   * The variant of notification to be rendered.
+   * Use the Notification.Opts.Variants attribute of the Notification component for access to these variant strings.
```
bjankord: Let's update this callout the strings intended to be used with this. One of `alert`, `error`, `warning`, `info`, `success`, `custom` instead of using Notifications.Opts.Variants

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,99 @@
...
+    @media screen and (max-width: 768px) {
+      width: calc(100% - 2 * var(--notification-dialog-width, 10px));
+    }
+  }
+
+  .notification-dialog-body {
+    background-color: var(--notification-dialog-body-color, #fff);
+    display: inline-flex;
+    flex: 1 1 auto;
+    flex-direction: initial;
```
bjankord: Using `initial` as a value is not supported in IE 10 or IE 11. We should replace this with either `column` or `row`, whatever the intended design requires.

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  .notification-dialog-body-text {
+    background-color: var(--notification-dialog-body-color, #fff);
+    font-size: var(--notification-dialog-body-font-size, 1.25rem);
+  }
+
+  .header-body {
+    background-color: var(--notification-dialog-header-color, #f9f9f9);
+    border-color: var(--notification-dialog-header-border-color, #dedfe0);
```
bjankord: Group these into 1 border property. There is a bug with the sass-loader that mangles CSS custom properties in borders when it is split out into 3 individual properties.

e.g. `border: var(--notification-dialog-header-border, 0.0625rem solid #dedfe0);`

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,99 @@
...
+    border-style: var(--notification-dialog-header-border-style, solid);
+    border-width: var(--notification-dialog-header-border-width, 0.0625rem);
+    display: 0 0 auto;
```
bjankord: `0 0 auto` is not a valid display option. I believe this can be removed.

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,99 @@
...
+    font-family: var(--notification-dialog-header-font-family, Helvetica);
```
bjankord: If the header font family doesn't need to change from the font family set on the HTML element, I'd remove this. Check with @neilpfeiffer to be sure.

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,99 @@
...
+    font-size: var(--notification-dialog-header-font-size, 1.125rem);
+    padding: var(--notification-dialog-body-padding, 0.8rem);
+    text-align: left;
+  }
+
+  .footer-body {
+    background-color: var(--notification-dialog-footer-color, #f9f9f9);
+    border-color: var(--notification-dialog-footer-border-color, #dedfe0);
```
bjankord: Group these into 1 border property. There is a bug with the sass-loader that mangles CSS custom properties in borders when it is split out into 3 individual properties.

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,99 @@
...
+      margin-right: var(--terra-notification-dialog-icon-margin-right, 1.25rem);
+      width: 2rem;
+    }
+  }
+
+  .title {
+    color: var(--terra-notification-dialog-title-color, #000);
+    flex: 0 0 auto;
+    flex-wrap: wrap;
+    font-family: var(--notification-dialog-header-font-family, Helvetica);
```
bjankord: If the header font family doesn't need to change from the font family set on the HTML element, I'd remove this. Check with @neilpfeiffer to be sure.

dkasper-was-taken:
Looks like terra-abstract-modal was released so you should be able to update the z-index to 9000 now.

dkasper-was-taken:
If #178 has IE 10 & 11 issues, this one might too. We should create a test example with very long text content to validate this.

dkasper-was-taken:
This component may or may not need to handle the request and release focus functions. If it doesn't, the consumer needs to handle the behavior, if it does, then we need to update it's internal life cycle to handle mount, update, and willunmount functions minimum.

packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -0,0 +1,223 @@
...
+    if (context.intl === undefined) {
+      return new Error('Please add locale prop to Base component to load translations');
+    }
+  },
+};
+
+class NotificationDialog extends React.Component {
+
+  componentDidMount() {
+    if (this.props.isFocused && this.props.isOpen && this.props.requestFocus) {
```
dkasper-was-taken: For the time being I think we only need the isFocused on the dialog modal version, since the notification dialog is the highest level component, its focus should always take focus.

dkasper-was-taken:
You should look into upgrading your examples on the doc page to use the new doc template.

dkasper-was-taken:
Either in the documentation or in the test examples, we should provide an example of the notification dialog being presented from the terra-modal-manager.

rakshith91:
> You should look into upgrading your examples on the doc page to use the new doc template.

Changed. Also, added an example with modal manager.

packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -0,0 +1,216 @@
...
+class NotificationDialog extends React.Component {
+
+  componentDidMount() {
+    if (this.props.isOpen && this.props.requestFocus) {
+      this.props.requestFocus();
+    }
+  }
+
+  componentDidUpdate() {
+    if (this.props.isOpen && this.props.requestFocus) {
```
dkasper-was-taken: I had overlooked the close but mounted state. This should probably look something closer to:  
```
  componentDidUpdate() {
    if (this.props.isOpen) {
      if (this.props.requestFocus) {
        this.props.requestFocus();
      }
    } else if (this.props.releaseFocus) {
      this.props.releaseFocus();
    }
  }
```

tbiethman:
Regarding the example using the modal manager, can we update it so that the modal content is what presents the NotificationDialog? I think the idea is to see both the ModalManager modal and the NotificationDialog modal open at the same time.

Something similar to the DateTimePicker's ModalManager example: https://github.com/cerner/terra-core/blob/master/packages/terra-site/src/examples/date-time-picker/DateTimePickerModalContainer.jsx

mmalaker:
Nit, but the "OK" button must be all upper case, per the [Cerner Style Guide](https://wiki.ucern.com/display/CSG/Button)

packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -0,0 +1,216 @@
...
+    dismissButton = <Button text={secondaryAction.text} onClick={secondaryAction.onClick} />;
+  }
+
+  return <div className={cx('actions')}>{actionButton}{dismissButton}</div>;
+};
+
+const getIcon = (variant, customIcon = null) => {
+  switch (variant) {
+    case variants.ALERT:
+      return (<IconAlert />);
```
neilpfeiffer: Need to change icons from being imported from terra-icon to being included via CSS to be properly re-themeable. Status-View is similar with both specified icons for the variants and also allows for custom. 

Reference for how it is done in the JSX:
https://github.com/cerner/terra-core/blob/master/packages/terra-status-view/src/StatusView.jsx#L159-L174

And in the SCSS:
https://github.com/cerner/terra-core/blob/master/packages/terra-status-view/src/StatusView.scss#L18-L44

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,106 @@
...
+    width: 100%;
+    word-wrap: break-word;
+  }
+
+  .actions {
+    display: flex;
+    flex: 0 0 auto;
+    flex-direction: var(--terra-notification-dialog-actions-flex-direction-small-width, row);
+    justify-content: flex-end;
+    padding: var(--terra-notification-dialog-actions-padding, 0.4rem);
```
neilpfeiffer: Let's update this to  `padding: var(--terra-notification-dialog-actions-padding, 0.28571rem 0.35714rem);`
and move it to the `.footer-body` so it better matches with `.header-body` also having padding.

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,106 @@
...
+    background-color: var(--notification-dialog-header-color, #f9f9f9);
+    border: var(--notification-dialog-header-border, 0.0625rem solid #dedfe0);
+    font-size: var(--notification-dialog-header-font-size, 1.125rem);
+    padding: var(--notification-dialog-body-padding, 0.8rem);
+    text-align: left;
+  }
+
+  .footer-body {
+    background-color: var(--notification-dialog-footer-color, #f9f9f9);
+    border: var(--notification-dialog-footer-border, 0.0625rem solid #dedfe0);
```
neilpfeiffer: Border should only be `border-top` in the footer.

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,106 @@
...
+  .notification-dialog-container {
+    display: flex;
+    flex-direction: column;
+    max-height: inherit; // `inherit` value needed for IE10 to scroll modal-body when overflowed
+    width: 100%; // Needed for IE to wrap content correctly in flexbox
+  }
+
+  .header-body {
+    background-color: var(--notification-dialog-header-color, #f9f9f9);
+    border: var(--notification-dialog-header-border, 0.0625rem solid #dedfe0);
```
neilpfeiffer: Border should only be `border-bottom` in the header.

mmalaker:
Just discovered a new problem that is isolated to iOS. Safari and all other browsers are fine. When the dialog displays on any iOS device (regardless of viewport size), the text displays in a narrow window between the header and footer of the dialog. The space only allows for about two lines of text to display without scrolling.

rakshith91:
> Just discovered a new problem that is isolated to iOS

What version of iOS are you using? It seems to be working fine on iOS 11.3.1

mmalaker:
Functional verification completed. The notification dialog examples display and function as expected with the exception of the issue present in older iOS versions.

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,126 @@
+:local {
+  .notification-dialog {
+    background: var(--notification-dialog-background, #fff);
+    border-radius: var(--notification-dialog-border-radius, 0.3125rem);
```
neilpfeiffer: Let's update to match same value as modal-manager:
`border-radius: var(--notification-dialog-border-radius, 5px);`

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,126 @@
...
+
+  .notification-dialog-container {
+    display: flex;
+    flex-direction: column;
+    max-height: inherit; // `inherit` value needed for IE10 to scroll modal-body when overflowed
+    width: 100%; // Needed for IE to wrap content correctly in flexbox
+  }
+
+  .header-body {
+    background-color: var(--notification-dialog-header-color, #f9f9f9);
```
neilpfeiffer: this variable should be `--notification-dialog-header-background-color`

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,126 @@
...
+    border-bottom: var(--notification-dialog-header-border, 0.0625rem solid #dedfe0);
```
neilpfeiffer: this variable should be `--notification-dialog-header-border-bottom`

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,126 @@
...
+    font-size: var(--notification-dialog-header-font-size, 1.125rem);
+    padding: var(--notification-dialog-body-padding, 0.8rem);
+    text-align: left;
+  }
+
+  .footer-body {
+    background-color: var(--notification-dialog-footer-color, #f9f9f9);
```
neilpfeiffer: this variable should be `--notification-dialog-footer-background-color`

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,126 @@
...
+    border-top: var(--notification-dialog-footer-border, 0.0625rem solid #dedfe0);
```
neilpfeiffer: this variable should be `-notification-dialog-footer-border-bottom`

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,126 @@
...
+    padding-top: var(--terra-notification-dialog-message-padding-top, 1rem);
+    text-align: var(--terra-notification-dialog-message-text-align, left);
+    width: 100%;
+    word-wrap: break-word;
+  }
+
+  .actions {
+    display: flex;
+    flex: 0 0 auto;
+    flex-direction: var(--terra-notification-dialog-actions-flex-direction-small-width, row);
```
neilpfeiffer: update to `--terra-notification-dialog-actions-flex-direction`

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,126 @@
+:local {
+  .notification-dialog {
+    background: var(--notification-dialog-background, #fff);
+    border-radius: var(--notification-dialog-border-radius, 5px);
+    display: flex;
+    flex-direction: column;
+    position: fixed;
+
+    @media screen and (max-width: 768px) {
+      width: calc(100% - 2 * var(--notification-dialog-width, 10px));
```
neilpfeiffer: lets update variable to `--notification-dialog-screen-small-inset-width`

packages/terra-notification-dialog/tests/jest/intl-context-setup.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable react/jsx-boolean-value, import/no-extraneous-dependencies */
+import { IntlProvider, intlShape } from 'react-intl';
```
dkasper-was-taken: terra-base exports these two, you can remove this dependency.

packages/terra-notification-dialog/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "react-dom": "^16.2.0",
+    "terra-base": "^3.1.0"
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "prop-types": "^15.5.8",
+    "react-intl": "^2.4.0",
+    "terra-abstract-modal": "^1.2.0",
+    "terra-app-delegate": "^2.1.0",
+    "terra-base": "^3.1.0",
```
dkasper-was-taken: Change terra-base to peer only dependency.

packages/terra-notification-dialog/.npmignore
```diff
@@ -0,0 +1,4 @@
+*.log
+node_modules
+target
+reports
```
emilyrohrbough: add `tests/**/__snapshots__/`

packages/terra-notification-dialog/package.json
```diff
@@ -0,0 +1,52 @@
...
+  },
+  "scripts": {
+    "compile": "npm run compile:clean && npm run compile:build",
+    "compile:clean": "rimraf lib",
+    "compile:build": "babel src --out-dir lib --copy-files",
+    "lint": "npm run lint:js && npm run lint:scss",
+    "lint:js": "eslint --ext .js,.jsx . --ignore-path ../../.eslintignore",
+    "lint:scss": "stylelint src/**/*.scss",
+    "test": "npm run test:jest && npm run test:wdio",
+    "test:jest": "jest ./tests/jest/* --config ../../jestconfig.js",
```
emilyrohrbough: Lets update this script to be jest --config ../../jestconfig.js. This was a windows compat update made in #198

packages/terra-notification-dialog/tests/wdio/notification-dialog-spec.js
```diff
@@ -0,0 +1,98 @@
+/* global browser, Terra, before */
+const selector = '#root';
+
+Terra.viewports('tiny', 'medium', 'large').forEach((viewport) => {
+  before(() => {
+    browser.setViewportSize(viewport);
+  });
+
+  describe('Complete notification-dialog title, message and actions', () => {
```
emilyrohrbough: We should add a screenshot for each variant

packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -0,0 +1,215 @@
...
+  INFO: 'info',
+  SUCCESS: 'success',
+  CUSTOM: 'custom',
+};
+
+const propTypes = {
+  /**
+   * Header of the notification-dialog.
+   */
+  header: PropTypes.string,
```
emilyrohrbough: Should this be an element?

packages/terra-notification-dialog/translations/es-US.json
```diff
@@ -0,0 +1,7 @@
+{
+  "Terra.notification.dialog.info": "Information.",
```
emilyrohrbough: Is the `.` after Information a typo?

packages/terra-notification-dialog/tests/jest/NotificationDialog.test.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import intlContexts from './intl-context-setup';
+import CompleteNotificationDialog from '../../src/terra-dev-site/test/notification-dialog/CompleteNotificationDialog.test';
+import CompleteNotificationDialogWithLongMessage from '../../src/terra-dev-site/test/notification-dialog/CompleteNotificationDialogWithLongMessage.test';
+import NoVariantNotificationDialog from '../../src/terra-dev-site/test/notification-dialog/NoVariantNotificationDialog.test';
+
+it('should shallow a complete notification-dialog', () => {
```
emilyrohrbough: It seems there are a lot of prop combinations we haven't tests to ensure the notification dialog renders as expected if a prop is not provided.

packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -0,0 +1,215 @@
...
+   */
+  title: PropTypes.string,
+  /**
+   * Message of the notification-dialog.
+   */
+  message: PropTypes.string,
+  /**
+   * The Action of the primary button.
+   */
+  primaryAction: PropTypes.shape({
```
emilyrohrbough: Should the primary action be required? or what is the behavior if neither a primary or secondary is provided? Is it close on esc?

packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -0,0 +1,215 @@
...
+  variant: null,
+};
+
+const actionSection = (primaryAction, secondaryAction) => {
+  let actionButton = null;
+  let dismissButton = null;
+  if (!primaryAction && !secondaryAction) {
+    return null;
+  }
+  if (primaryAction) {
```
emilyrohrbough: If a primaryAction is not provided but a secondaryAction is, would we want render the secondary action with the Emphasis style?

rakshith91: I think making primaryAction as required would make more sense. Thoughts?

packages/terra-notification-dialog/src/NotificationDialog.jsx
rakshith91: No. header should be limited to string.

packages/terra-notification-dialog/translations/es-US.json
rakshith91: Looks like it. Changed it and propagated the change to other translation files.

packages/terra-notification-dialog/README.md
yuderekyu: *9001 xD

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,126 @@
...
+    flex-direction: column;
+    position: fixed;
+
+    @media screen and (max-width: 768px) {
+      width: calc(100% - 2 * var(--notification-dialog-small-horizontal-inset, 10px));
+    }
+  }
+
+  .notification-dialog-body {
+    background-color: var(--notification-dialog-body-color, #fff);
```
neilpfeiffer: Update variable from `--notification-dialog-body-color` to `--notification-dialog-body-background-color`

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,126 @@
...
+    flex: 1 1 auto;
+    flex-direction: row;
+    height: 100%;
+    overflow: auto;
+    padding: var(--notification-dialog-body-padding, 1.25rem);
+    width: 100%;
+  }
+
+  .notification-dialog-body-text {
+    background-color: var(--notification-dialog-body-color, #fff);
```
neilpfeiffer: remove this background-color from `.notification-dialog-body-text` since already applied in the `.notification-dialog-body` which is the correct place.

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,126 @@
...
+    flex-direction: column;
+    max-height: inherit; // `inherit` value needed for IE10 to scroll modal-body when overflowed
+    width: 100%; // Needed for IE to wrap content correctly in flexbox
+  }
+
+  .header-body {
+    background-color: var(--notification-dialog-header-background-color, #f9f9f9);
+    border-bottom: var(--notification-dialog-header-border-bottom, 0.0625rem solid #dedfe0);
+    font-size: var(--notification-dialog-header-font-size, 1.125rem);
+    padding: var(--notification-dialog-body-padding, 0.8rem);
```
neilpfeiffer: need to update this one `--notification-dialog-body-padding` (also used on line 21) to `--notification-dialog-header-padding` for here.

packages/terra-notification-dialog/src/NotificationDialog.scss
```diff
@@ -0,0 +1,126 @@
...
+    flex-direction: row;
+    height: 100%;
+    overflow: auto;
+    padding: var(--notification-dialog-body-padding, 1.25rem);
+    width: 100%;
+  }
+
+  .notification-dialog-body-text {
+    background-color: var(--notification-dialog-body-color, #fff);
+    font-size: var(--notification-dialog-body-font-size, 1.25rem);
```
neilpfeiffer: Also looks like the font-size here in unneeded since it is being set again in children both for `.title` (line 92) and `.message` (line 103) and could be removed.

packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -0,0 +1,215 @@
...
+
+    return (
+      <AbstractModal
+        ariaLabel="NotificationDialog"
+        classNameModal={notificationDialogClassNames}
+        isOpen={this.props.isOpen}
+        onRequestClose={this.props.onRequestClose}
+        closeOnEsc={false}
+        closeOnOutsideClick={false}
+        zIndex="9000"
```
neilpfeiffer: add prop definition for abstract-modal `role: 'dialog',`

neilpfeiffer:
Tested: PR #162
Resolves: Issue #136 and #190

**Reviewed:**
- Follows pattern considerations documented in Cerner Design Standards: [Notification Dialogs](https://wiki.ucern.com/x/8PTzXw#Dialogs-NotificationDialogs)
- Follows UX defined design implementation for Terra-Notification-Dialog
- Follows all defined theme designs: terra-theme

\
\
Verified by @scottwilmarth and  @neilpfeiffer, Ready for Release.


**Pass** |
-----------|
UX Review|

PR #163, opened 2018-04-24
Update theming on navigation-side-menu

bjankord:
### Summary
* Added theme styles to support legacy look
* Replaced use of action-header component with terra-core's action-header

## Example with mock styles applied
example1

**Note:** A mock stylesheet does not exist in this repo so there's no easy way to view this theme applied. This will likely need to be tested for themability in this repo and tested for theming in the terra-core repo with the terra-legacy-theme applied.


packages/terra-navigation-side-menu/src/MenuItem.scss
```diff
@@ -1,4 +1,11 @@
 :local {
+  .list-item {
```
bjankord: These are needed so we can apply borders to the side on one element and bottom on another element without creating a diagonal border effect (This occurs when both top and/or bottom borders are set with side borders). Only the bottom border is needed for the legacy theme, though I've included all border directions to match consistency with the other styles in this file.

packages/terra-navigation-side-menu/src/NavigationSideMenu.scss
```diff
@@ -4,4 +4,10 @@
     padding-bottom: var(--terra-navigation-side-menu-list-padding-bottom, 0.125rem);
     padding-top: var(--terra-navigation-side-menu-list-padding-top, 0.125rem);
   }
+
+  .side-menu-action-header {
```
bjankord: These allow us to apply a micro theme to the action header for just this specific location

PR #176, opened 2018-05-09
Update to use terra-doc-template

emilyrohrbough:
### Summary
Update site pages to use terra-doc-template.


bjankord:
With this change, I'd like to delete most of the usage sections from the README file. The ability to view the example source code should mean we don't need the usage examples anymore / we no longer need to maintain them.

However, I think there are cases where examples in the README are helpful, for instance, the examples in the app delegate package. 

I'd use this guidance, if the toggleable source code matches or nearly matches the usage example in the README file, delete the usage example from the README.

emilyrohrbough:
I can go through and look for duplicate examples, however are there any off the top that you know you want gone?

packages/terra-popup/examples/Index.site-page.jsx
```diff
@@ -19,83 +14,59 @@ import PopupNoHeader from './site-examples/PopupNoHeader';
...
-  </div>
+const DocPage = () => (
+  <DocTemplate
+    packageName={name}
+    readme={ReadMe}
+    srcPath={`https://github.com/cerner/terra-framework/tree/master/packages/${name}`}
+    examples={[
+      {
+        title: 'Popup Default Props',
+        example: <PopupStandard />,
```
StephenEsser: Should these have a source? Or should the doc template conditionally render the example?

image

emilyrohrbough: Both. I wasn't intending to add src for each popup example- updating now but the template should also be conditional IMO

Issue #179, opened 2018-05-11
InfiniteList component sending more than one next page request on scrolling down


srikantvadrevu:
# Issue Description
The callback method provided to the onRequestItems prop of InfiniteList component is being called more than once while scrolling down. This is causing more items to be loaded on in the list than required. 
## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
The list should send just one next page request on scrolling.

## Current Behavior
Currently, the next page request is being fired before the previous request results are loaded in list. This is causing two next page requests to be fired everytime we scroll down.

## Steps to Reproduce

Create a list using Infinitelist component that makes a service call on scrolling down.
Observe that two next page requests are fired on scrolling.

## Environment

* Component Version: 1.1.0
* Browser Name and Version: Chrome
* Operating System and version (desktop or mobile): Windows 10 and Ubuntu

PR #189, opened 2018-05-22
Fix bugs around updating with new child content.

dkasper-was-taken:
### Summary
Address bugs during update.

Resolves: #179
Resolves: #190

@cerner/terra


packages/terra-infinite-list/tests/wdio/infinite-list-spec.js
```diff
@@ -53,4 +53,17 @@ describe('InfiniteList', () => {
 
     Terra.should.matchScreenshot({ selector: '#test-infinite-list' });
   });
+
+  describe('Displays an infinite list same count and different content', () => {
```
emilyrohrbough: Should we have a before and after screenshot to verify the content was actually swapped?

PR #193, opened 2018-06-03
Introduce Hookshot targetCoordinates

dkasper-was-taken:
### Summary
Update hookshot to allow for a target set of screen coordinates to anchor the content.

@cerner/terra


packages/terra-hookshot/src/_HookshotUtils.js
```diff
@@ -183,6 +198,11 @@ const getBoundingRect = (boundingElement) => {
   return rect;
 };
 
+/**
+ * This function returns the attachment object for the target when using coordinate based positioning.
+ */
+const getCoordinateAttachment = () => ({ vertical: 'middle', horizontal: 'center' });
```
tbiethman: Any reason not to rename this coordinateAttachment and make it an object vs. a function that returns a static object?

dkasper-was-taken: good call

emilyrohrbough:
Is there an issue/ask linked to this update?

packages/terra-hookshot/src/_HookshotUtils.js
```diff
@@ -12,6 +12,21 @@ const MIRROR_TB = {
   bottom: 'top',
 };
 
+/**
+ * This function returns a rect 2px x 2px with the origin based at (1,1), allow for evenly centered attachment points.
+ * As a result of the 2x2 size, bounds checking errors as a result of rounding are eliminated.
+ *
+ * @ param {object} coords - THe x and y client coordinates.
+ */
+const getRectFromCoords = coords => ({
```
emilyrohrbough: Is there a check somewhere to verify that these coordinates are valid with the defined viewport?

dkasper-was-taken: Technically as long as they are evaluated by PropTypes as numbers the coordinates are valid. While the use-case for out of range values is more limited, I don't see number reason to prevent it if someone wants an element slightly off screen.

dkasper-was-taken:
@emilyrohrbough There isn't an immediate ask for this functionality outside of teams causally requesting right-click functionality from popup. Likely it would be a new component aside from popup, but the underlying implementation would be based off this work. It's not high priority, I just coded the resolution in my free time.

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -73,12 +73,20 @@ const propTypes = {
    */
   onPosition: PropTypes.func,
   /**
-   * Required element that the content will hookshot to.
+   * Client coordinates to serve as the anchor point for the hookshot'd content.
    */
-  targetRef: PropTypes.func.isRequired,
+  targetCoordinates: PropTypes.shape({
+    x: PropTypes.number,
```
emilyrohrbough: Although it wouldn't make sense to provide just an `x` or just an `y` coordinate, should we make each key required?

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -73,12 +73,20 @@ const propTypes = {
...
+    y: PropTypes.number,
+  }),
+  /**
+   * Element to serve as the anchor point for the hookshot'd content. (If targetCoordinates are provided, this is ignored.)
+   */
+  targetRef: PropTypes.func,
   /**
    * Object containing the vertical and horizontal attachment values for the target.
    * Valid values: { horizontal: ['start', 'center', 'end'], vertical: ['top', 'middle', 'bottom'] }.
+   * If targetCoordinates are provided { horizontal: 'center', vertical: 'middle' } will be applied.
```
emilyrohrbough: provided by default or is this the only option when using target coordinates?

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -186,17 +216,17 @@ class Hookshot extends React.Component {
   }
 
   enableListeners() {
-    const target = this.props.targetRef();
-    if (!target) {
+    const childElement = this.getTargetRef() || this.getBoundingRef();
```
emilyrohrbough: Why would would we use the bounding ref as the childElement fallback if target ref is undefined? Would we not use the target coordinates? I see below we grab the `scrollParents` it I would have though we'd still want to worry about the targeted scroll area

packages/terra-hookshot/src/_HookshotUtils.js
```diff
@@ -12,6 +12,21 @@ const MIRROR_TB = {
   bottom: 'top',
 };
 
+/**
+ * This function returns a rect 2px x 2px with the origin based at (1,1), allow for evenly centered attachment points.
+ * As a result of the 2x2 size, bounds checking errors as a result of rounding are eliminated.
```
emilyrohrbough: `bounds checking errors as a result of rounding are eliminated.` I think you meant `bounds checking errors are eliminated as a result of rounding.`

packages/terra-hookshot/src/Hookshot.jsx
dkasper-was-taken: Good call

packages/terra-hookshot/src/Hookshot.jsx
dkasper-was-taken: Since it'll be positioning with a single coordinate, we don't want to worry about non-standard positioning on a single point. Under the hood it uses a 2x2px grid to position it with origin at 1x1, in order to avoid rounding errors.

packages/terra-hookshot/src/Hookshot.jsx
dkasper-was-taken: It's an odd behavior, but one that I figured should work. If the user wants an element positioned according to screen coordinates, but also doesn't want it to leave a specific bounding element within the DOM flow, we still want to trap it.

Normally would be use the targetRef to find the elements to find scroll listeners from, but as we don't have an element to find parentNodes, we use the backup on the boundindRef as it could scroll as well. Technically I could see some wild scenario where the boundingRef and the targetRef aren't in the same flow and would need to merge the scroll listeners, but I figured we'll cross that bridge when we come to it.

packages/terra-hookshot/src/_HookshotUtils.js
```diff
@@ -12,6 +12,21 @@ const MIRROR_TB = {
   bottom: 'top',
 };
 
+/**
+ * This function returns a rect 2px x 2px with the origin based at (1,1), which allows for evenly centered attachment points.
```
bjankord: Nice

PR #202, opened 2018-06-13
[brand-footer] Add section headers to nav bar

Blackop778:
### Summary
Allows section headers to be put into the navigation bar of a `terra-brand-footer` by using a new `sections` prop. The `sections` prop overrides any provided `links` prop values.

### [Deployment](https://terra-framework-deploye-pr-202.herokuapp.com/#/components/terra-brand-footer/brand-footer)

@cerner/terra


packages/terra-brand-footer/src/BrandFooter.jsx
```diff
@@ -54,21 +54,27 @@ const BrandFooter = ({ links, contentLeft, contentRight, contentBottom, ...custo
...
-            {links.map(link => (
-              <li className={cx('list-item')} key={link.text + link.href}>
-                {link.target !== undefined ? (
-                  <a className={cx('link')} href={link.href} target={link.target} >{link.text}</a>
-                ) : (
-                  <a className={cx('link')} href={link.href} >{link.text}</a>
-                )}
-              </li>
-            ))}
+            {links.map((link) => {
```
emilyrohrbough: There is a lot of conditional logic happening here. I am wondering if it would be more readable if this was broken out to create the list items.

bjankord:
What are peoples thoughts on adding a custom content prop for the top bar of the brand footer. The approach of generating section headers bypassing an object to the `links` prop without an `href` key feels odd to me.

neilpfeiffer:
Seems like creating a linkgroup/optgroup type internal that takes a static subheader as prop (similar setup to terra-form-select, terra-list, etc.) might be the better solve.

Blackop778:
One option would be instead of taking the links as a prop directly, a `sections` prop could be added to hold an array of objects with a `title` or `header` property with the text to display and a `links` property which would work as it did before this PR.

bjankord:
I like the `sections` prop idea, it would allow us to keep `links` for passivity and we can add to the object if needed in the future.

packages/terra-brand-footer/src/BrandFooter.scss
```diff
@@ -7,19 +7,18 @@
     background-color: var(--terra-brand-footer-nav-background-color, #153d70);
     color: var(--terra-brand-footer-nav-color, #8baaca);
 
+    display: flex;
```
emilyrohrbough: Can you point to the screenshots / test examples that display this behavior?

Blackop778: Could you clarify what you mean?

emilyrohrbough: In your deployment link, which examples can I look at to see the flex style being applied? In the screenshots, it appears the navigational links are displaying horizontal and I though we were introducing an option to show a vertical list display with these changes.

Blackop778: Ah. https://terra-framework-deploye-pr-202.herokuapp.com/#/tests/terra-brand-footer/brand-footer/vertical-nav-brand-footer demonstrates the vertical layout, but I'll add an example to the documentation page.

packages/terra-brand-footer/src/BrandFooter.jsx
```diff
@@ -40,35 +94,81 @@ const propTypes = {
...
+                    return (
+                      <li className={cx('list-item')} key={link.text + link.href}>
+                        <a className={cx('link')} href={link.href} target={link.target} >
+                          {link.text}
+                        </a>
+                      </li>
+                    );
+                  }
+
+                  return (
```
tomleewu: Should this return something different when `link.target` is `undefined`?

packages/terra-brand-footer/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ Unreleased
 1.5.0 - (July 3, 2018)
 ------------------
 ### Added
+* Section headers can be put in the nav bar
```
emilyrohrbough: These need to be moved to the unreleased section

packages/terra-brand-footer/src/BrandFooter.jsx
```diff
@@ -9,21 +9,75 @@ const cx = classNames.bind(styles);
...
+  *         target: string,
+  *       ],
+  *     }
+  *   ],
+  * }
+  * ```
+  */
+  sections: PropTypes.shape({
+   /**
+    * If true link groups will be laid out from top to bottom then left to right
```
emilyrohrbough: I am wondering if `isVertical` is more aligned with our boolean prop naming convention.

Additionally, 
>If true link groups will be laid out from top to bottom then left to right

When will it stop laying out from top to bottom and flip to displaying left to right?

emilyrohrbough: Looking through this a bit more, I am wondering if it would make more sense to flatten the sections prop to be two props `hasVeriticalLinks` and `linkGroups`. Looking through the logic below it seems we are looping through `sections.linkGroups` or `processedSections.linkGroups` but you don't seem to be doing anything directly with `sections`.

packages/terra-brand-footer/src/BrandFooter.jsx
```diff
@@ -40,35 +94,81 @@ const propTypes = {
...
-      {linkslength > 0 ? (
-        <nav className={cx('nav')}>
-          <ul className={cx('menu')}>
-            {links.map(link => (
-              <li className={cx('list-item')} key={link.text + link.href}>
-                {link.target !== undefined ? (
-                  <a className={cx('link')} href={link.href} target={link.target} >{link.text}</a>
-                ) : (
-                  <a className={cx('link')} href={link.href} >{link.text}</a>
+      {processedSections.linkGroups.length > 0 && (
```
emilyrohrbough: Nit: Can we break this log out of the render?

packages/terra-brand-footer/src/BrandFooter.scss
```diff
@@ -1,53 +1,118 @@
...
+    }
+  
+    .list-item:first-child::before {
+      content: '';
+      padding: 0;
+    }
+
+    /* stylelint-disable selector-max-compound-selectors */
+    // A '/' isn't needed to separate an item from the header it follows
+    .list-header + .list-item::before {
```
emilyrohrbough: Do we need to apply these list item before styles to vertical displays as well?

packages/terra-brand-footer/src/BrandFooter.jsx
```diff
@@ -40,35 +94,81 @@ const propTypes = {
...
+        <nav className={cx(['nav', !processedSections.displayVertically ? 'nav-horizontal' : 'nav-vertical'])}>
+            {processedSections.linkGroups.map(linkGroup => (
+              <ul className={cx('menu')} key={linkGroup.id}>
+                { // When displaying vertically if one column has a header all columns are aligned as if they have a header
+                  ((containsASectionHeader && processedSections.displayVertically) || linkGroup.headerText) && (
+                    <li className={cx('list-header')} key={linkGroup.headerText}>
+                      { // Insert a zero width space to act as a placeholder section header that doesn't display but takes vertical space
```
emilyrohrbough: In regards to this comment:
>// Insert a zero width space to act as a placeholder section header that doesn't display but takes vertical space
why do we want to do this?

packages/terra-brand-footer/tests/wdio/brand-footer-spec.js
```diff
@@ -18,24 +18,77 @@ describe('BrandFooter', () => {
...
+      '--terra-brand-footer-link-color': 'green',
+      '--terra-brand-footer-footer-content-background-color': 'red',
+      '--terra-brand-footer-footer-content-color': 'black',
     });
   });
+
+  describe('Vertical Mixed Headers', () => {
+    beforeEach(() => browser.url('/#/raw/tests/terra-brand-footer/brand-footer/vertical-mixed-headers-brand-footer'));
+
+    Terra.should.matchScreenshot({ viewports });
```
emilyrohrbough: Can we add `Terra.should.beAccessible()` to these new tests?

packages/terra-brand-footer/tests/wdio/brand-footer-spec.js
```diff
@@ -18,24 +18,77 @@ describe('BrandFooter', () => {
...
 
-  describe('Active', () => {
-    beforeEach(() => {
-      browser.url('/#/raw/tests/terra-brand-footer/brand-footer/active-brand-footer');
+  describe('Section Headers', () => {
+    beforeEach(() => browser.url('/#/raw/tests/terra-brand-footer/brand-footer/section-header-brand-footer'));
+
+    Terra.should.beAccessible({ viewports, rules });
+    Terra.should.matchScreenshot({ viewports });
+    Terra.should.themeEachCustomProperty({
```
emilyrohrbough: Do we need to add themeable tests for Section Headers or are these vars covered in the previous tests?

packages/terra-brand-footer/tests/wdio/brand-footer-spec.js
```diff
@@ -18,24 +18,77 @@ describe('BrandFooter', () => {
...
+    Terra.should.beAccessible({ viewports, rules });
     Terra.should.matchScreenshot({ viewports });
     Terra.should.themeEachCustomProperty({
-      '--terra-brand-footer-link-hover-color': 'brown',
+      '--terra-brand-footer-footer-content-background-color': 'red',
+      '--terra-brand-footer-footer-content-color': 'black',
     });
   });
 
-  describe('Active', () => {
```
emilyrohrbough: How come you removed the hover and active tests?

packages/terra-brand-footer/tests/jest/BrandFooter.test.jsx
```diff
@@ -59,10 +122,14 @@ describe('BrandFooter', () => {
   it('should render with provided links and content left', () => {
```
emilyrohrbough: IMO we should keep these tests that are verify the links prop works and write new tests for the sections prop. Although you have marked `links` as deprecated, we still want to test and ensure both props render as expected until `links` is removed.

packages/terra-brand-footer/tests/jest/BrandFooter.test.jsx
```diff
@@ -12,6 +12,20 @@ describe('BrandFooter', () => {
```
emilyrohrbough: Can we add a jest test that adds both `links` and `sections` so we can verify the `sections` prop is honored as described?

packages/terra-brand-footer/src/BrandFooter.scss
Blackop778: No because vertical links are each given their own line to clearly separate them from other links. Horizontal links must insert a `/` between each other to show their separation. That is what this style controls.

packages/terra-brand-footer/src/BrandFooter.jsx
Blackop778: In the design specification link sections without a section header were shown to keep their first element aligned vertically with other link sections that do have a section header. The zero width space takes up as much vertical room as regular text would.

packages/terra-brand-footer/tests/wdio/brand-footer-spec.js
Blackop778: The tests weren't actually testing the active or hover styles, only applying a color variable that happens to be the same as it would be if the link was activated or hovered over. I don't think that merits testing.

I do now remember that there is actually a WDIO method to hover over an element, so that test will be added back.

Blackop778: The moveToObject method would not activate the hover styling in my testing, so the hover test will be left off again.

packages/terra-brand-footer/tests/jest/BrandFooter.test.jsx
Blackop778: Done in 0939f42333450ab5a565782924d03fc9904b8564

packages/terra-brand-footer/tests/jest/BrandFooter.test.jsx
Blackop778: Done in 0939f42333450ab5a565782924d03fc9904b8564

packages/terra-brand-footer/tests/wdio/brand-footer-spec.js
emilyrohrbough: Did you wrap `moveToObject` in an `it` block?

Blackop778: I did

packages/terra-brand-footer/src/BrandFooter.jsx
```diff
@@ -40,35 +86,82 @@ const propTypes = {
...
+              ((containsASectionHeader && isVertical) || linkGroup.headerText) && (
+                <li className={cx('list-header')} key={linkGroup.headerText}>
+                  { // Insert a zero width space to act as a placeholder section header that doesn't display but takes vertical space
+                    linkGroup.headerText || '\u200B'
+                  }
+                </li>
+              )
+            }
+            {linkGroup.links && linkGroup.links.map((link) => {
+              if (link.target !== undefined) {
```
dkasper-was-taken: Can probably reduce this code to some version of this:
```
      const spreadTarget = link.target !== undefined ? { target: link.target} : {};
      return (
        <li className={cx('list-item')} key={link.text + link.href}>
          <a className={cx('link')} href={link.href} {...spreadTarget}>
            {link.text}
          </a>
        </li>
      );
```

packages/terra-brand-footer/src/BrandFooter.scss
```diff
@@ -1,53 +1,116 @@
...
-    padding-right: 0;
-  }
+  .nav-vertical {
+    align-content: stretch;
+    flex-direction: row;
+
+    .menu {
+      flex: 1 1 0%;
+      min-width: 240px;
+      padding-right: 1.4285714286rem;
```
yuderekyu: Do we want to open up the padding vars to be themeable?

packages/terra-brand-footer/src/BrandFooter.scss
```diff
@@ -1,53 +1,116 @@
...
+  .menu {
+    list-style: none;
+    margin: 0;
+    padding-left: 0;
+    padding-right: 0;
   }
 
+  .list-header {
+    color: var(--terra-brand-footer-nav-header-color, #4bb8ef);
+    font-size: 0.7857rem;
```
yuderekyu: I'd assume someone would want the ability to control font css vars too

neilpfeiffer:
Tested: PR #202
 

**Reviewed:**
- N/A: Follows pattern considerations documented in Cerner Design Standards
- Follows UX defined design implementation for Terra-Brand-Footer
- Follows all defined theme designs: terra-theme

\
**Resolve with New Issue:**
- Need to add/transfer original colors for brand footer to [Terra-Legacy-Theme](https://github.com/cerner/terra-core/tree/master/packages/terra-legacy-theme/src/scss) to support legacy Blue Steel apps. 
\
\


**Pass** |
-----------|
UX Review|

mmalaker:
Functional verification completed. The brand footer examples display and function correctly with the new, section headers.

PR #221, opened 2018-07-06
Use Terra's Config Modules

emilyrohrbough:
### Summary
This PR is a WIP
- Use browerslist-config-terra, eslint-config-terra, and stylelint-config-terra modules.
- Fix lint errors.


package.json
```diff
@@ -8,7 +8,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/cerner/terra-framework.git"
+    "url": "githttps://github.com/cerner/terra-framework.git"
```
bjankord: why the removal of the `+` character here?

package.json
```diff
@@ -21,6 +21,15 @@
   "bugs": {
     "url": "https://github.com/cerner/terra-framework/issues"
   },
+  "browserslist": [
+    "extends browserslist-config-terra"
+  ],
+  "eslintConfig": {
+    "extends": "terra"
```
tomleewu: Should this be
```
"eslintConfig": {
  "extends": "eslint-config-terra"
},
```

emilyrohrbough: Good question! Either definition is valid, but eslint allows extends to obmit the `estlint-config` prefix. Here is the link to the docs: https://eslint.org/docs/user-guide/configuring#using-a-shareable-configuration-package.

bjankord: I feel like it should, I'm surprised using just `terra` works, though thats also what we use in terra-core. @emilyrohrbough what your thoughts on this?

emilyrohrbough: Eslint doesn't require the prefix to be included to successfully load the configuration. Eslint searched for the module `eslint-config-terra` when given `terra`. I am unopinionated to how we define it in the configuration.

bjankord: Let's keep it as is, TIL something new about ESLint configs 👍

packages/terra-application-layout/src/menu/RoutingMenu.jsx
```diff
@@ -58,22 +58,6 @@ class RoutingMenu extends React.Component {
     }));
   }
 
-  /**
```
emilyrohrbough: Change due to `react/sort-comp` ordering rules

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -95,30 +95,6 @@ const defaultProps = {
 };
 
 class PopupContent extends React.Component {
-  static getDimensionStyle(value, maxValue, isAutomatic) {
```
emilyrohrbough: Change due to `react/sort-comp` ordering rules

PR #236, opened 2018-07-25
Bump responsive-element version

bjankord:
### Summary
Missed bumping these to new major version in last update. This corrects the version so using the latest release of terra-ui components will all use terra-responsive-element v3.

### DIFF
--- a/packages/terra-aggregator/CHANGELOG.md
+++ b/packages/terra-aggregator/CHANGELOG.md
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Updated to use terra-responsive-element v3
 
 3.8.0 - (July 25, 2018)
 ------------------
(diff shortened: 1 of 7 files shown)


StephenEsser:
There is an example that spreads a custom style onto the responsive element, since there is no wrapping div within the responsive element any more we should double check this example still renders correctly: https://github.com/cerner/terra-framework/blob/master/packages/terra-application-header-layout/src/terra-dev-site/doc/example/HeaderWireframe.jsx#L50

bjankord:
@StephenEsser Update the example: https://github.com/cerner/terra-framework/pull/236/commits/cb10ba8816f16fb711f82f27a7a70908f7126ada

Blackop778:
Could we get this merged and released? `terra-layout` is trying to read `breakpoints.module.scss` from `terra-responsive-element` except in the 2.0.1 version that `terra-layout` is pulling in the file is still named `breakpoints.scss` so anything that depends on `terra-layout` cannot compile.

PR #246, opened 2018-08-15
Eslint 5 update

bjankord:
### Summary
* Update codebase to align with ESLint 5 compat Airbnb JavaScript style guide.
* Update ESLint to v5

There is a slight visual diff in the modal manager / slide panel manager example screenshots. The diff is caused from the example content and not the component itself. Before we had used [this in example content](https://github.com/cerner/terra-framework/blob/master/packages/terra-disclosure-manager/src/terra-dev-site/test/common/TestExample.jsx#L128).
```
<p>id: {identifier}</p>
```
After the lint auto fix, this has been updated to the following:
```
<p>
  id:
  {' '}
  {identifier}
</p>
```
I tried using the following but still got a `0.02` diff compared to original screenshots.
```
<p>{`id: ${identifier}`}</p>
```


package.json
```diff
@@ -86,8 +86,8 @@
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
     "enzyme-to-json": "^3.2.2",
-    "eslint": "^4.19.1",
-    "eslint-config-terra": "^1.0.0",
+    "eslint": "^5.0.0",
+    "eslint-config-terra": "git://github.com/cerner/eslint-config-terra.git#eslint-5-compat",
```
bjankord: will switch this to tag release once we cut / before this is merged

packages/terra-hookshot/src/terra-dev-site/test/common/HookshotTestTemplate.jsx
```diff
@@ -35,19 +35,19 @@ const defaultProps = {
...
+  } if (attachment === 'top end') {
     return { vertical: 'top', horizontal: 'end' };
-  } else if (attachment === 'top center') {
+  } if (attachment === 'top center') {
     return { vertical: 'top', horizontal: 'center' };
-  } else if (attachment === 'bottom start') {
+  } if (attachment === 'bottom start') {
     return { vertical: 'bottom', horizontal: 'start' };
-  } else if (attachment === 'bottom end') {
+  } if (attachment === 'bottom end') {
```
emilyrohrbough: The don't like else if?

dangerfile.js
```diff
@@ -1,4 +1,4 @@
-// eslint-disable-next-line import/no-extraneous-dependencies, no-unused-vars
+// eslint-disable-next-line import/no-extraneous-dependencies, no-unused-vars, object-curly-newline
```
emilyrohrbough: Should we just do the obj new line instead of disabling? I imagine it wants:
```
import {
   danger,
   warn,
   fail,
   message
} from 'danger';
```

packages/terra-aggregator/CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ Unreleased
 3.10.0 - (August 14, 2018)
 ------------------
 ### Changed
-* Formatting changes per eslint v4 update
+* Formatting changes per eslint v5 update
```
emilyrohrbough: should this be a new changelog?

bjankord: ah yes, I'll get a fix pushed up.

PR #252, opened 2018-09-04
Add main container to layout

### LINKED ISSUE #250: Modify Terra-Layout for the content container to be a main node with role=main instead of generic div.
# Feature Request

## Description

Screen Readers have a jump to content section that is toggled either by a keyboard shortcut, or a jump to content button that is toggled when a user first tabs into the navigation of the site, like how Airbnb does it.

screen shot 2018-08-28 at 10 31 55 am

screen shot 2018-08-28 at 10 31 49 am

I think adding the link would take a considerable amount of work and redesign of the navigation and layout, but for now, it would be an easy win if we make the content section of the node a main node with role="main", or at a minimum attaching role="main" to the content area.

JakeLaCombe:
### Summary
Resolves #250 

JAWS has the ability to switch to the main content area if one exists. It is defined by either the main node, or the main role. This pull request adds that feature.


emilyrohrbough:
I have a feeling this may affect our test screenshots by adding the height of 100%

JakeLaCombe:
Looks like they passed ok inside of framework. Also, the parent container was already at a height of 100%, and this node is essentially just an extra wrapper.

bjankord:
Since the `main` element is being added outside of the element we capture the screenshot from and it is not impacting the sizing of the element we capture the screenshot from, this should continue to work the same as it did before.

I added a `div` with a string of text inside of the div with  the `data-terra-dev-site-content` attribute which [is what we match on in terra-dev-site site](https://github.com/cerner/terra-dev-site/blob/master/config/wdio/wdio.conf.js#L10) to show that the first-child inside of `data-terra-dev-site-content` is still sized according to its content / own styles.

screen shot 2018-09-05 at 12 03 57 pm

PR #269, opened 2018-09-17
Removed duplicate terra-base dependencies

AlexBrizi:
### Summary
Removed any instances where terra-base was listed as both a peerDependency and either a dependency or devDependency (kept it as a peerDependency).

Closes [#200](https://github.com/cerner/terra-framework/issues/200)

### DIFF
--- a/packages/terra-aggregator/CHANGELOG.md
+++ b/packages/terra-aggregator/CHANGELOG.md
@@ -4,6 +4,11 @@ Changelog
 Unreleased
 ----------
 
+3.13.0 (September 17, 2018)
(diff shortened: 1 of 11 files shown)


packages/terra-aggregator/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ Changelog
 Unreleased
 ----------
 
+3.13.0 (September 17, 2018)
```
emilyrohrbough: Lets remove this, we update the release version & date right before release

PR #288, opened 2018-09-28
Script added to automate readme updates

amichaelparker:
### Summary
Similar to [this PR](https://github.com/cerner/terra-core/pull/1885) in terra-core, this adds a script to automatically update the package status portion of the readme to reflect the state of the packages folder.

Run 'npm run docs' to test.


markdown.config.js
```diff
@@ -0,0 +1,47 @@
+const path = require('path');
```
emilyrohrbough: We should place this under `scripts/generate-markdown/markdown.config.js`. We've followed this pattern for repos with release scripts

amichaelparker: I moved the script and altered it slightly to work from the new location. Works on Mac and Windows. Let me know if that is what you had in mind!

PR #289, opened 2018-10-01
Rearrange slide panel to change tab order

amichaelparker:
### Summary
Addresses issue [#1272](https://github.com/cerner/terra-core/issues/1272) from [terra-core](https://github.com/cerner/terra-core). Divs are rearranged so that tabbing will focus main panel first, slide panel second.

Closes cerner/terra-core#1272


bjankord:
@tbiethman Do you know if there is any concern with switching DOM order of panel and main content?

packages/terra-slide-panel/tests/jest/SlidePanel.test.jsx
```diff
@@ -1,112 +0,0 @@
-import React from 'react';
```
emilyrohrbough: Do we want to remove these jest tests?

packages/terra-slide-panel-manager/tests/jest/SlidePanelManager.test.jsx
```diff
@@ -1,68 +0,0 @@
-import React from 'react';
```
emilyrohrbough: Do we want to remove these jest tests?

tbiethman:
The DOM position of the panel should be dependent upon the `panelPosition` value. I would assume that if the panel is in the `start` position, the existing DOM order would be preferred.

packages/terra-slide-panel/tests/jest/SlidePanel.test.jsx
bjankord: We should keep the jest tests @amichaelparker

packages/terra-slide-panel-manager/tests/jest/SlidePanelManager.test.jsx
bjankord: We should keep the jest tests @amichaelparker

mmalaker:
Functional verification completed. The slide panel examples now have the correct tab order.

PR #293, opened 2018-10-01
Use injectIntl prop from terra-base

### LINKED ISSUE #203: Use injectIntl prop from terra-base
# Issue Description
terra-application-utility and terra-application-layout (_ApplicationHeader) use the `injectIntl` prop from `react-intl`. We should use the exposed `injectIntl` prop from terra-base instead (https://github.com/cerner/terra-core/pull/1379)

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

corymcdonald:
### Summary
Resolves #203

@cerner/terra

### DIFF
--- a/packages/terra-application-layout/package.json
+++ b/packages/terra-application-layout/package.json
@@ -22,18 +22,19 @@
   },
   "homepage": "https://github.com/cerner/terra-framework#readme",
   "devDependencies": {
-    "terra-app-delegate": "^2.7.0"
+    "terra-app-delegate": "^2.7.0",
+    "react-intl": "^2.4.0",
+    "terra-base": "^3.3.0"
   },
   "peerDependencies": {
     "react": "^16.4.2",
     "react-dom": "^16.4.2",
     "terra-app-delegate": "^2.0.0",
-    "terra-base": "^3.0.0"
+    "terra-base": "^3.3.0"
   },
   "dependencies": {
     "classnames": "^2.2.5",
     "prop-types": "^15.5.8",
-    "react-intl": "^2.4.0",
     "react-router-dom": "^4.2.2",
     "terra-action-header": "^1.14.0",
     "terra-application-header-layout": "^2.16.0",
--- a/packages/terra-application-layout/src/header/_ApplicationHeader.jsx
+++ b/packages/terra-application-layout/src/header/_ApplicationHeader.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
-import { injectIntl, intlShape } from 'react-intl';
+import { injectIntl, intlShape } from 'terra-base';
 import AppDelegate from 'terra-app-delegate';
 import ApplicationHeaderLayout from 'terra-application-header-layout';
 import { ApplicationHeaderUtility } from 'terra-application-utility';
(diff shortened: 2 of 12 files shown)


packages/terra-application-utility/package.json
```diff
@@ -21,16 +21,19 @@
     "url": "https://github.com/cerner/terra-framework/issues"
   },
   "homepage": "https://github.com/cerner/terra-framework#readme",
+  "devDependencies": {
```
emilyrohrbough: We can drop these devDependencies. Terra-base should already be provided by terra-base and then we want terra-base to manager the react-intl version.

packages/terra-application-layout/package.json
```diff
@@ -22,18 +22,19 @@
   },
   "homepage": "https://github.com/cerner/terra-framework#readme",
   "devDependencies": {
-    "terra-app-delegate": "^2.7.0"
+    "terra-app-delegate": "^2.7.0",
+    "react-intl": "^2.4.0",
```
emilyrohrbough: We can drop the `react-intl` and `terra-base` devDependencies. Terra-base should already be provided by terra-base and then we want terra-base to manager the react-intl version.

packages/terra-application-utility/package.json
corymcdonald: Unfortunately since terra-base is a peer dependency npm does not install those and therefore causes errors in the tests

emilyrohrbough: Ah you are right about terra-base. Whoops! I was thinking since terra-dev-site installs this as well it would be handled.

yuderekyu: i'm curious if tests pass when we replace [`IntlProvider`](https://github.com/cerner/terra-framework/blob/master/packages/terra-application-utility/tests/jest/UtilityButton.test.jsx#L2) with a terra-base component - this way we can remove `react-intl` as a dev dep

corymcdonald: I don't think we export that - https://github.com/cerner/terra-core/blob/master/packages/terra-base/src/Base.jsx

yuderekyu: booooooo

yuderekyu: wait i'm confused, aren't we exporting it here: https://github.com/cerner/terra-core/blob/master/packages/terra-base/src/Base.jsx#L108

corymcdonald: Nope that's the base component which features the render, then the next line is where we `export injectIntl`

packages/terra-application-layout/package.json
corymcdonald: Sure can! https://github.com/cerner/terra-core/issues/1944

emilyrohrbough:
@corymcdonald 

Looks like terra-base has been released with `intlProvider` being exported for testing purposes.

packages/terra-application-layout/src/header/_ApplicationHeader.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
-import { injectIntl, intlShape } from 'react-intl';
+import { injectIntl, intlShape } from 'terra-base';
```
bjankord: We've run into some issues with pulling in react-intl modules this way from terra-base. More info on this issue here: https://github.com/cerner/terra-core/issues/1982
For now, let's leave this as pulling from react-intl.

PR #296, opened 2018-10-02
Fix #295 (terra-base devDependency clean up)

### LINKED ISSUE #295: terra-base devDependency clean up
# Feature Request

## Description
We recently removed terra-base from being listed as a dependency and listed it only as a peerDependency. We missed an instance where its [listed as a devDependency in terra-disclosure-manager](https://github.com/cerner/terra-framework/blob/master/packages/terra-disclosure-manager/package.json#L26). We should remove this devDependency from terra-disclosure-manager and make terra-base a devDependency in the terra-framework root package.json file similar to what we do for react, react-dom, etc.

tatchi:
Fix #295

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
     "shelljs": "^0.8.2",
     "stylelint": "^9.2.0",
     "stylelint-config-terra": "^1.0.0",
-    "terra-base": "3.0.0",
+    "terra-base": "^3.0.0",
     "terra-dev-site": "^2.0.0",
     "terra-toolkit": "^4.5.0",
     "xfc": "^1.2.1"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -106,6 +106,7 @@
     "shelljs": "^0.8.2",
     "stylelint": "^9.2.0",
     "stylelint-config-terra": "^1.0.0",
+    "terra-base": "3.0.0",
```
ryanthemanuel: Why is this locked to 3.0.0?

bjankord: Ah, we should add a caret in front. `^3.0.0`

tatchi: That was already like that in the disclosure-manager package. I fixed it :)

emilyrohrbough:
@tatchi You should add yourself to the contributors list 😄 https://github.com/cerner/terra-framework/blob/master/CONTRIBUTORS.md

tatchi:
> @tatchi You should add yourself to the contributors list 😄 https://github.com/cerner/terra-framework/blob/master/CONTRIBUTORS.md

Done :tada: :tada:

PR #299, opened 2018-10-04
[terra-application-utility] - Fix divider focus

### LINKED ISSUE #298: [terra-application-utility] - Divider allows click focus
# Bug Report
The divider has [`tabIndex` set to -1](https://github.com/cerner/terra-framework/blob/master/packages/terra-application-utility/src/utility/_UtilityMenuDivider.jsx#L24), focus skips the divider when tabbing. But, selecting the divider shows a focus style.

divider-focus mov

yuderekyu:
### Summary
Closes #298. 

Compare [pr build](https://terra-framework-deploye-pr-299.herokuapp.com/#/components/terra-application-utility/application-utility/application-utility) with [master build](https://engineering.cerner.com/terra-framework/#/components/terra-application-utility/application-utility/application-utility).

### DIFF
--- a/packages/terra-application-utility/src/utility/_UtilityMenuDivider.module.scss
+++ b/packages/terra-application-utility/src/utility/_UtilityMenuDivider.module.scss
@@ -5,6 +5,7 @@
     border-right-width: var(--terra-application-utility-menu-divider-border-bottom-right, 0);
     border-top: var(--terra-application-utility-menu-divider-border-bottom, 1px solid #dedfe0);
     height: var(--terra-application-utility-menu-divider-height, 0.2143rem);
+    outline: none;
     width: 100%;
 
     &::after {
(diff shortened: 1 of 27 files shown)


packages/terra-application-utility/src/utility/_UtilityMenuDivider.module.scss
```diff
@@ -5,6 +5,7 @@
     border-right-width: var(--terra-application-utility-menu-divider-border-bottom-right, 0);
     border-top: var(--terra-application-utility-menu-divider-border-bottom, 1px solid #dedfe0);
     height: var(--terra-application-utility-menu-divider-height, 0.2143rem);
+    outline: none;
```
StephenEsser: If the diver should never receive focus removing the tabIndex attribute would also prevent tab and click focus entirely, assuming the tabIndex attribute isn't playing another role.

> Because tabindex="-1" makes elements non-navigable via the keyboard, make sure you don’t include it on any form, object, or link elements a user could click on with a mouse or receive focus via touch.

https://www.wufoo.com/html5/tabindex-attribute/

yuderekyu: ohh i didn't know that. that's a better solution - updated

Issue #177, opened 2018-05-09
Add an 'uncontrolled' version of notification-dialog


rakshith91:
# Issue Description
This 'controlled' NotificationDialog allows for a lot of flexibility in the component's presentation. In addition to this, I think we should provide an 'uncontrolled' version that can be presented synchronously and managed outside of the calling component. This would be helpful to consumers who want to inject the notification dialog into Promise chains and control logic flow with the action selections.

Something like the confirmAlert function as seen here: https://github.com/GA-MO/react-confirm-alert/blob/master/src/index.js#L121

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

PR #300, opened 2018-10-04
[DO_NOT_MERGE] Adding ConfirmNotificationDialog

tbiethman:
### Summary
Issue: #177

I thought it may be easier to talk about this with the code mostly written and examples to play with. Feel free to take a look at the deployment to see how the examples are set up. I'll be annotating the PR with questions that I still have or would like suggestions on.

Two big changes here:
1. Creation of `_transientUtils.jsx`
  * This file exports `mount`/`unmount` functions that will generically mount/unmount a React tree separate from the main React tree. An id is required to identify the mount node, along with the component instance to be mounted and locale/customMessages values to be given to the Base instance that wraps the component automatically.
  * A HOC-generator called `injectTransientComponent` is also exported so that relevant data can provided to the presented component for self-management, if desired. This is currently utilizes the Context for transmission.
  * This is meant to be a agnostic set of utilities that will allow any component to be transiently mounted.

2. Creation of `ConfirmNotificationDialog.jsx`
  * This component is a version of the NotificationDialog meant to be presented with the utilities provided by `transientUtils.jsx`.
  * It exports its own mount/unmount functions for ease of use by consumers.
  * A subset of the NotificationDialog props are supported (isOpen/header should no longer be necessary).

Any alternatives to this type of implementation would create complicated or inefficient logic within components that want to integrate NotificationDialog presentation with Promise chains and other synchronous-ish flows.

I just want to get some opinions on this and pick some brains before I harden it too much. I'll be happy to get naming suggestions 😄


packages/terra-notification-dialog/src/_transientUtils.jsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  const mountElement = document.createElement('div');
+  mountElement.id = containerId;
+  mountElement.setAttribute('data-terra-transient-component', true);
+  document.body.appendChild(mountElement);
+
+  return new Promise((resolve) => {
+    render((
+      <Base locale={locale} customMessages={customMessages}>
```
tbiethman: I figured it'd be convenient to wrap these components with Base automatically. It did require locale/customMessages to be exposed in the API, though. 

Would it make more sense for people to have the ConfirmNotificationDialog implement Base instead? locale/customMessages would become props of the ConfirmNotificationDialog instead.

packages/terra-notification-dialog/src/_transientUtils.jsx
```diff
@@ -0,0 +1,94 @@
...
+        <TransientComponentContext.Provider
```
tbiethman: The provider is currently only exposing the containerId to child components so that those child components could unmount themselves, if necessary. 

I had toyed with the notion of providing a custom 'remove' function instead that would capture the id in an enclosure, but I wasn't sure if that was overcomplicating things.

Of course, we could clone the object direction on the component as a prop. It'd be straightforward for the ConfirmNotificationDialog's implementation, but anything more complex may benefit from the context implementation.

packages/terra-disclosure-manager/src/terra-dev-site/doc/example/DisclosureComponent.jsx
```diff
@@ -39,15 +41,37 @@ class DisclosureComponent extends React.Component {
...
+              resolve();
+            },
+          },
+          secondaryAction: {
+            text: 'Cancel',
+            onClick: () => {
+              reject();
+            },
+          },
+          releaseFocus: () => {
```
tbiethman: The focus stuff is a little squirrely. The ConfirmNotificationDialog needs to request/release focus from the Modal, however it does not get a new version of the AppDelegate after the initial presentation. So we have to write custom functions here bound to the mounting component (which does get the updated version). Ideally this logic could be obfuscated in some way, but no good ideas yet.

packages/terra-notification-dialog/src/_transientUtils.jsx
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * If an element with the chosen id is already present, the new presentation is aborted.
+   */
+  if (document.getElementById(containerId)) {
+    return Promise.reject();
+  }
+
+  const mountElement = document.createElement('div');
+  mountElement.id = containerId;
+  mountElement.setAttribute('data-terra-transient-component', true);
```
tbiethman: I added an attribute here in case we ever needed to forcefully find and remove all transient components (in the case of session logouts or something). There's a lot more thought that needs to go into that.

packages/terra-disclosure-manager/src/terra-dev-site/doc/example/DisclosureComponent.jsx
```diff
@@ -1,12 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
+import { injectIntl, intlShape } from 'terra-base';
```
bjankord: We need to import `injectIntl, intlShape` this from `react-intl`. Pulling from terra-base ends up executing more of terra-base than just the `injectIntl` and `intlShape`. 
See https://github.com/cerner/terra-core/issues/1982 for more info.

packages/terra-notification-dialog/src/terra-dev-site/doc/example/ConfirmNotificationDialogExample.jsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import { injectIntl, intlShape } from 'terra-base';
```
bjankord: We need to import `injectIntl, intlShape` this from `react-intl`. Pulling from terra-base ends up executing more of terra-base than just the `injectIntl` and `intlShape`. 
See https://github.com/cerner/terra-core/issues/1982 for more info.

tbiethman:
Closing this because it's outdated as all hell. I'll come back to it.

PR #309, opened 2018-10-10
Improve popup focus on iOS with VoiceOver

### LINKED ISSUE #285: When popup is opened, focus is not shifted to popup
# Bug Report

## Description
When using VoiceOver, when the popup is opened, focus is not shifted to opened popup. This makes it difficult for users which are using screen readers to navigate through the popup quickly after opening it.

## Steps to Reproduce

1. Navigate to https://engineering.cerner.com/terra-framework/#/components/terra-popup/popup/popup
2. Scroll down page to example
3. Open popup

## Expected Behavior
When opening popup, we should shift focus to the popup.

## Possible Solution
Look into using a ref on the popup to shift focus to popup container when its initially opened. This will likely resolve the other issues where we are using popup and the focus is not shifted when opened.

## Environment

* Component Name and Version: terra-popup

bjankord:
### Summary
Resolves #285 

Shifts focus to the first focusable item in hookshot when mounted. This allows focus to shift in popups. The helps make popups usable on iOS when using VoiceOver and improves reading experience in JAWs and VoiceOver on Mac.

### DIFF
--- a/packages/terra-popup/src/Popup.jsx
+++ b/packages/terra-popup/src/Popup.jsx
@@ -76,7 +76,7 @@ const propTypes = {
    */
   isArrowDisplayed: PropTypes.bool,
   /**
-   * Should the popup content have tab focus. Set this to `true` if your content doesn't contain any focusable elements.
+   * Set this to `true` if your content has focusable elements and you want them to receive focus instead of focusing on the default popup frame when the popup is opened.
    */
   isContentFocusDisabled: PropTypes.bool,
   /**
--- a/packages/terra-popup/src/Popup.jsx
+++ b/packages/terra-popup/src/Popup.jsx
@@ -76,7 +76,7 @@ const propTypes = {
    */
   isArrowDisplayed: PropTypes.bool,
   /**
-   * Should the popup content have tab focus. Set this is your content doesn't contain any focusable elements.
+   * Should the popup content have tab focus. Set this to `true` if your content doesn't contain any focusable elements.
    */
   isContentFocusDisabled: PropTypes.bool,
   /**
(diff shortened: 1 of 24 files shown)


dkasper-was-taken:
This logic should probably be moved to specifically popup, rather than hookshot. This change will likely break the select control.
What do you think @StephenEsser

packages/terra-popup/src/Popup.jsx
```diff
@@ -76,7 +76,7 @@ const propTypes = {
    */
   isArrowDisplayed: PropTypes.bool,
   /**
-   * Should the popup content have tab focus. Set this is your content doesn't contain any focusable elements.
+   * Should the popup content have tab focus. Set this to `true` if your content doesn't contain any focusable elements.
```
dkasper-was-taken: I also think I originally documented this incorrectly. Setting the value to true should be done when you want to disable the popups default tab focus. Meaning you "do" have focusable elements and want to take focus away from the default popup frame.

dkasper-was-taken: Using that prop may actually fix the screen reader issue by itself.

StephenEsser:
There is a good chance this change would break the Select control which maintains synthetic focus within the hookshot content using aria-owns and arian-activedescedent. If possible I think this logic should be inside of popup or hidden behind an opt-in API within hookshot

packages/terra-popup/src/Popup.jsx
bjankord: Did a quick test with VO on iOS and setting the prop to true allows the virutal focus indicator to shift correctly. I'll work on the wording to make this more clear what the prop does and update our popup examples that have focusable content within them to use this prop.

bjankord:
Added tech-only as this is only adding a prop to examples we already have.

PR #318, opened 2018-10-30
Updating DisclosureManager to use a context-based API

### LINKED ISSUE #315: Refactor DisclosureManager with Context-based API
# Feature Request

The DisclosureManager and related components should provide a Context-based API to allow child components to interact with the DisclosureManager without using prop-injection.

## Description

Today, the DisclosureManager and related components do a lot of Component cloning to inject instances of the AppDelegate object into child components. The AppDelegate instances expose the various functions used to control DisclosureManager state. 

I propose that the DisclosureManager implement and expose a DisclosureManagerContext to provide those APIs to its child components.  I also propose that the terra-app-delegate package be completely deprecated, with its contents being moved into the terra-disclosure-manager package (and renamed to DisclosureManagerDelegate).

tbiethman:
### Summary
Switching the DisclosureManager to expose its APIs through context vs. prop injection. This should allow us to remove a lot of the prop-chaining logic we're doing today across various components while making the relationships between the DisclosureManager and components that use it more explicit. 

I locked down the dependencies in other packages so the non-passive changes don't start breaking things; I'll update those dependencies in a subsequent PR. I wanted to have the DisclosureManager changes reviewed in isolation first.

Fixes #315.

### Deployment Links
https://terra-framework-deploye-pr-318.herokuapp.com/#/components/terra-slide-panel-manager/slide-panel-manager/slide-panel-manager

https://terra-framework-deploye-pr-318.herokuapp.com/#/components/terra-modal-manager/modal-manager/example


packages/terra-disclosure-manager/CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
```
mjhenkes: Call this out as a breaking change
https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/CHANGELOG.md#breaking-changes---upgrade-guide
Also, we should add an upgrade guide.

```suggestion
### Breaking Change
```

packages/terra-disclosure-manager/docs/README.md
```diff
@@ -71,15 +67,17 @@ Example (using the Modal and SlideGroup):
 
 ### Interacting with the Disclosure Manager
 
+The DisclosureManager wraps its contents in a context provider that exposes an instance of a DisclosureManagerDelegate, an object containing DisclosureManager APIs,  to components based upon their presented location. Components can use the higher order component generator `withDisclosureManager()` to wrap themselves automatically in a context consumer and receive a prop named `disclosureManager` containing the DisclosureManagerDelegate instance. The various APIs are detailed below.
```
yuderekyu: xtra space

packages/terra-aggregator/package.json
```diff
@@ -37,7 +37,7 @@
     "terra-arrange": "^2.19.0",
     "terra-button": "^2.0.1",
     "terra-content-container": "^2.4.0",
-    "terra-disclosure-manager": "^2.14.0",
+    "terra-disclosure-manager": "2.13.0",
     "terra-doc-template": "^1.1.0",
     "terra-form": "^2.0.1",
```
yuderekyu: how are we using `terra-form`? should probably remove it anyway since it's deprecated

tbiethman: It's being used in the examples. I will update this in a subsequent PR; terra-aggregator will have more changes once this PR is merged.

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -369,74 +373,68 @@ class DisclosureManager extends React.Component {
...
-      }
-
-      const ComponentClass = AppDelegate.getComponentForDisclosure(componentData.name);
-      if (!ComponentClass) {
-        return undefined;
-      }
-
-      return <ComponentClass key={componentData.key} {...componentData.props} app={AppDelegate.create(disclosureApp)} />;
+      return (
+        <React.Fragment key={componentData.key}>
```
tbiethman: I can actually remove this Fragment and put the key directly on the Provider.

packages/terra-aggregator/package.json
JakeLaCombe: Be gentile! That's still my baby, even though it's getting removed!!!

mmalaker:
Issue found during functional verification. When using keyboard navigation, the modal manager examples change tab focus as you would expect, from the container to the modal disclosure window when triggered. However, the slide panel manager examples keep focus within the parent container until the last element has focus, then move focus to the disclosure window. Modal and slide panel focus behavior should be the same.

I checked https://engineering.cerner.com/terra-ui/#/components/terra-slide-panel-manager/slide-panel-manager/slide-panel-manager and it looks like it is an issue with our current examples, not one introduced in this PR.

tbiethman:
@mmalaker We've logged #321 to correct that issue with the focus with the SlidePanelManager.

mmalaker:
Functional verification completed. The modal and slide panel manager examples display and function as expected. Focus issue to be addressed in #321

PR #325, opened 2018-11-05
[terra-layout] - Regenerated jest snapshots

StephenEsser:
### Summary
The following pull request introduced a default z-index that caused the terra-layout snapshot tests to fail. 

https://github.com/cerner/terra-core/pull/1955

### Additional Details
Regenerated the jest snapshots to account for the z-index

Thanks for contributing to Terra. 
@cerner/terra

[CONTRIBUTORS.md]: ../blob/master/CONTRIBUTORS.md
[License]: ../blob/master/LICENSE

PR #333, opened 2018-11-12
[terra-layout] - Regenerate jest and wdio snapshots

StephenEsser:
### Summary
The following pull requests introduced some visually non-passive changes that require updating the wdio screenshots and jest snapshots for terra-layout.

* Added a root selector prop that caused the terra-layout snapshot tests to fail.
  * https://github.com/cerner/terra-core/pull/1983 
* Adjusted the styling of the action footer resulting in a screenshot mismatch
  * https://github.com/cerner/terra-core/pull/1985

### Additional Details
Regenerated the wdio and jest snapshots to account for the default root selector and style changes of the terra-action-footer.

@cerner/terra


packages/terra-layout/tests/jest/__snapshots__/LayoutSlidePanel.test.jsx.snap
```diff
@@ -21,6 +21,7 @@ exports[`LayoutSlidePanel should render when small 1`] = `
       isRelativeToContainer={true}
```
StephenEsser: Coincidentally this package already has a change log update titled "Regenerate jest snapshots" from a similar issue resolved on this pull request: https://github.com/cerner/terra-framework/pull/325

Because the change log update would be identical I didn't add one for this fix.

StephenEsser:
There were some breaking changing to the action-footer within this component as well, I'll regenerate the screenshots on the next commit.

PR #324, opened 2018-11-05
[terra-dialog-modal] - Restructure theme tests

StephenEsser:
### Summary
* Moved theme tests into a separate file and theme directory.
* Removed custom colors in test that were causing theme libraries to fail accessibility
* Removed stale reference screenshots

### Additional Details
Part of #323 

Thanks for contributing to Terra. 
@cerner/terra

[CONTRIBUTORS.md]: ../blob/master/CONTRIBUTORS.md
[License]: ../blob/master/LICENSE

Issue #323, opened 2018-11-05
Decouple Functional and Custom Property WebDriver Tests


StephenEsser:
## Issue
WebDriver tests contain both functional test steps and custom property theme tests within the same spec file. When tests are shared across libraries, specifically theming libraries, it is not possible to skip tests that are applying a mock theme. This introduces a risk of generating unnecessary and inaccurate screenshots.

Tests that contain both functional steps and custom property alteration steps generate a screenshot that includes a mock theme and inaccurately represents the component during that step.

## Proposal
Decouple functional and custom property WebDriver tests by separating them into different files and moving the custom property tests into a unique directory. This enables explicit control over which set of tests consumers want to run and makes each spec file's intention more clear.

Functional Tests: 
```
/wdio/test-name.spec
```

WDIO Tests:
```
        ↓ Directory
/wdio/theme/test-name.spec
```
## Component Checklist
- [ ] terra-aggregator
- [ ] terra-app-delegate
- [ ] terra-application-header-layout
- [ ] terra-application-layout
- [ ] terra-application-links
- [ ] terra-application-menu-layout
- [ ] terra-application-name
- [ ] terra-application-utility
- [ ] terra-brand-footer
- [x] [terra-dialog-modal](https://github.com/cerner/terra-framework/pull/324)
- [ ] terra-disclosure-manager
- [ ] terra-form-validation
- [ ] terra-hookshot
- [ ] terra-infinite-list
- [ ] terra-layout
- [ ] terra-modal-manager
- [ ] terra-navigation-layout
- [ ] terra-navigation-side-menu
- [ ] terra-notification-dialog
- [ ] terra-popup
- [ ] terra-slide-panel
- [ ] terra-slide-panel-manager
- [ ] terra-theme-provider

PR #334, opened 2018-11-12
Webdriver Custom Property Test Updates

amichaelparker:
### Summary
Ref: [#305](https://github.com/cerner/terra-framework/issues/305). Webdriver tests updated to use themeCombinationOfCustomProperties where possible.

Closes [#305](https://github.com/cerner/terra-framework/issues/305)

### Additional Details
Application utility required having the property combinations divided up for coherence. Other components may require the same after review.


packages/terra-application-layout/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
```
StephenEsser: There is an [open issue](https://github.com/cerner/terra-framework/issues/323) for restructuring these theme tests into a different file. An example of what those changes look like can be found here: https://github.com/cerner/terra-framework/pull/324

Do we want to migrate these theme tests into a new file with this pull request or as a separate issue?

bjankord: If you want to tackle https://github.com/cerner/terra-framework/issues/323 in this PR, that sounds good to me. We can scope it to just the packages you've already modified in this PR or do the whole repo. Either way sounds good to me.

+1 on changes so far.

StephenEsser: Breaking the tests out is no longer required for the theme repositories to skip these tests. However, there is some discussions around whether or not we want to continue breaking the tests out to segregate functional tests from theme tests. 

cc: @mjhenkes @bjankord

bjankord: I propose we check in what's in the PR. It looks like only 2 of the theme tests had functional steps that would be duplicated with the primary wdio tests. The rest looked to be just defining the theme variables statically without any functionality being invoked to verifiy component is in a set state before capturing a screenshot.

With that being said, I wouldn't worry about doing this work in the other repos.

bjankord: @amichaelparker  Let's create a branch off of this to keep the changes to split the theming tests into new files.
And then back on this branch, lets revert those change and just keep the code that updates to use themeCombinationOfCustomProperties where possible.

PR #337, opened 2018-11-15
Replace use of componentWillMount in popup

bjankord:
### Summary
Part of the work to resolve this issue: https://github.com/cerner/terra-core/issues/1884


packages/terra-popup/src/terra-dev-site/test/popup/OverlayPopup.test.jsx
```diff
@@ -11,13 +11,10 @@ class DefaultPopup extends React.Component {
     this.state = { open: false };
   }
 
-  componentWillMount() {
```
dkasper-was-taken: This was originally using componentWillMount to ensure the overflow was turned off before the popup was positioned, this was mainly an issue with body scrolling.  We'll likely need to move this to render or something.  Similar to what the other overlay uses. Or perhaps switch to the terra-overlay.

packages/terra-popup/src/_PopupOverlay.jsx
```diff
@@ -18,7 +18,7 @@ const defaultProps = {
 };
 
 class PopupOverlay extends React.Component {
-  componentWillMount() {
+  componentDidMount() {
```
dkasper-was-taken: This was originally using componentWillMount to ensure the overflow was turned off before the popup was positioned, this was mainly an issue with body scrolling. We'll likely need to move this to render or something. Similar to what the other overlay uses. Or perhaps switch to the terra-overlay.

bjankord: We could look into moving it to the render, though when I tested opening a popup and tried scolling the body in [the page from the PR](https://terra-framework-deploye-pr-337.herokuapp.com/#/components/terra-popup/popup/popup) compared to the [terra-ui page](https://engineering.cerner.com/terra-ui/#/components/terra-popup/popup/popup), I didn't notice a difference other than maybe a 0.5 second delay after clicking on the button for the html element to have it's overflow set to hidden. Functionally, it seems to work the same as before.

JakeLaCombe: I don't think a half second delay is the worst. It's not amazon thankfully! https://www.fastcompany.com/1825005/how-one-second-could-cost-amazon-16-billion-sales

dkasper-was-taken: The issue originally presented itself on IE and on browsers with a mouse plugged in on mac with the body scrolling. When the scrollbar was removed it, caused the popup to be misaligned. If this is no longer an issue I'm good with that.

Though looking at the code, I think the shift to use hookshot may have fixed it. The mounting order is different than was used with tether.

bjankord: I'll test in those scenarios before merging and will update the code if needed.

PR #349, opened 2018-11-27
Notification Dialog example update

### LINKED ISSUE #329: Notification Dialog example update
# Bug Report

## Description
Latest update to notification-dialog made [here](https://github.com/cerner/terra-framework/pull/308/files#diff-c22091f816a343666f7c6e4d7aa42eebL116) changed the example to no longer show the really long message as original intended. 

https://github.com/cerner/terra-framework/blob/05f1e4870a1a3d2f9cab81d1b836248077ccb33f/packages/terra-notification-dialog/src/terra-dev-site/doc/example/NotificationDialogWithLongText.jsx#L116

## Possible Solution
Example could probably be better updated to not use an array, but an ES6 multi-line string instead, then won't need to reference the array

noahbenham:
### Summary
Resolves #329. Improves documentation examples.

### Additional Details
N/A

@cerner/terra

### DIFF
--- a/packages/terra-notification-dialog/src/terra-dev-site/doc/example/CompleteNotificationDialogWithLongMessage.jsx
+++ b/packages/terra-notification-dialog/src/terra-dev-site/doc/example/CompleteNotificationDialogWithLongMessage.jsx
@@ -33,8 +33,8 @@ class CompleteNotificationDialogWithLongMessage extends React.Component {
           variant={NotificationDialogVariants.SUCCESS}
           isOpen={this.state.isOpen}
           onRequestClose={this.handleCloseModal}
-          title="This is the title. This is the title. This is the title. This is the title. This is the title. This is the title. This is the title. This is the title."
-          message="This is the message. This is the message. This is the message. This is the message. This is the message. This is the message. This is the message. This is the message. This is the message. This is the message."
+          title="Lorem ipsum dolor sit amet, consectetur adipiscing elit. In ornare odio quis turpis viverra, volutpat laoreet magna porta."
+          message="Quisque egestas ullamcorper velit vitae volutpat. Quisque vestibulum nulla nunc, eget pharetra massa semper ac. In sit amet felis tincidunt, laoreet tortor nec, tempus ipsum."
           primaryAction={{
             text: 'OK',
             onClick: clickOK,
--- a/packages/terra-notification-dialog/src/terra-dev-site/doc/example/NotificationDialogWithLongText.jsx
+++ b/packages/terra-notification-dialog/src/terra-dev-site/doc/example/NotificationDialogWithLongText.jsx
@@ -27,7 +27,7 @@ class NotificationDialogWithLongText extends React.Component {
   }
 
   render() {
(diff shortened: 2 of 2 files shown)


packages/terra-notification-dialog/src/terra-dev-site/doc/example/NotificationDialogWithLongText.jsx
```diff
@@ -27,85 +27,85 @@ class NotificationDialogWithLongText extends React.Component {
...
-      'Lorem ipsum dolor sit amet, consectetur adipiscing elit. ',
-      'Maecenas molestie in lorem vel facilisis. Quisque ac enim nec lectus malesuada faucibus.',
-      'Integer congue feugiat ultricies.',
-      ' Nunc non mauris sed tellus cursus vestibulum nec quis ipsum.',
-      'Vivamus ornare magna justo, et volutpat tortor congue ut. Nulla facilisi.',
-      ' Cras in venenatis turpis. Nullam id odio justo. Etiam vehicula lectus vel purus consectetur cursus id sit amet diam.',
-      'Donec facilisis dui non orci hendrerit pharetra. Suspendisse blandit dictum turpis, in consectetur ipsum hendrerit eget.',
-      'Nam vehicula, arcu vitae egestas porttitor,',
-      'turpis nisi pulvinar neque, ut lacinia urna purus sit amet elit.'];
+    const message = [`Lorem ipsum dolor sit amet, consectetur adipiscing elit.
```
StephenEsser: Does this need to be an array at all after updating to use back ticks?

packages/terra-notification-dialog/src/terra-dev-site/doc/example/CompleteNotificationDialogWithLongMessage.jsx
```diff
@@ -33,8 +33,8 @@ class CompleteNotificationDialogWithLongMessage extends React.Component {
           variant={NotificationDialogVariants.SUCCESS}
           isOpen={this.state.isOpen}
           onRequestClose={this.handleCloseModal}
-          title="This is the titleThis is the titleThis is the titleThis is the titleThis is the titleThis is the titleThis is the titleThis is the title"
-          message="This is the messageThis is the messageThis is the messageThis is the messageThis is the messageThis is the messageThis is the messageThis is the messageThis is the messageThis is the message"
+          title="This is the title. This is the title. This is the title. This is the title. This is the title. This is the title. This is the title. This is the title."
+          message="This is the message. This is the message. This is the message. This is the message. This is the message. This is the message. This is the message. This is the message. This is the message. This is the message."
```
StephenEsser: It might be worth looking into a more realistic title / message example we could substitute for this phrase while we're updating it. I think it would help our examples demo a lot better. If we can't find a long enough phrase I think even the lorem ipsum text would be an improvement.

packages/terra-notification-dialog/src/terra-dev-site/doc/example/NotificationDialogWithLongText.jsx
noahbenham: Whoops, nope. Thanks.

PR #357, opened 2018-12-04
Update popup preventDefault behavior

bjankord:
### Summary
Remove preventDefault prop on PopupContent. This was preventing events inside the popup content, where we just want to prevent events outside the popup content when it is open. This behavior should be handled by default with react-onclickoutside. 

* Need to cross-browser test / cross-device test this change.

### DIFF
--- a/packages/terra-popup/src/_PopupContent.jsx
+++ b/packages/terra-popup/src/_PopupContent.jsx
@@ -229,7 +229,6 @@ class PopupContent extends React.Component {
           onEsc={onRequestClose}
           onOutsideClick={this.props.onRequestClose}
           onResize={this.handleOnResize}
-          preventDefault
           refCallback={refCallback}
           role="dialog"
         >
--- a/packages/terra-popup/CHANGELOG.md
+++ b/packages/terra-popup/CHANGELOG.md
@@ -4,7 +4,12 @@ Changelog
 Unreleased
 ----------
 
-4.19.1 - (December 3, 2018)
+4.19.1 - (December 4, 2018)
+------------------
+### Removed
+* Removed preventDefault prop on PopupContent component. This is handled by react-onClickOutside.
+
+4.19.0 - (December 3, 2018)
 ------------------
 ### Changed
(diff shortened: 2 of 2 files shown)


bjankord:
It seems removing the `preventDefault` prop reintroduces [the bug logged here](https://github.com/cerner/terra-core/issues/1602). Clicking on the popup button results in the following error when the prop is removed.

```sh
index.js:186 [Intervention] Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/features/5093566007214080
```

packages/terra-hookshot/src/HookshotContent.jsx
```diff
@@ -216,3 +216,4 @@ HookshotContent.propTypes = propTypes;
 const onClickOutsideContent = onClickOutside(HookshotContent);
 
 export default onClickOutsideContent;
+export { HookshotContent };
```
bjankord: Export version of HookshotContent without onClickOutside HOC. Using popup overlay to handle onClickOutisde but keeping this HOC for passivity.

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -4,7 +4,7 @@ import classNames from 'classnames/bind';
 import Button from 'terra-button';
 import ContentContainer from 'terra-content-container';
 import FocusTrap from 'focus-trap-react';
-import Hookshot from 'terra-hookshot';
+import HookshotContent from 'terra-hookshot/lib/HookshotContent';
```
bjankord: Use Hookshot without onClickOutside HOC.

packages/terra-popup/src/_PopupOverlay.jsx
```diff
@@ -18,6 +18,11 @@ const defaultProps = {
 };
 
 class PopupOverlay extends React.Component {
+  constructor() {
+    super();
+    this.handleOnClick = this.handleOnClick.bind(this);
```
bjankord: Handle "onClickOutside" via popup overlay.

Issue #351, opened 2018-11-28
Main element is undefined in IE and needs shimmed


bjankord:
# Bug Report

## Description
IE throws a console error when `<main />` element is used. Need to shim it for IE support.

## Steps to Reproduce

1. Navigate to a page using the application-layout
2. Open the console in IE
3. See error message about main element

## Expected Behavior
No console errors

## Possible Solution
Shim main element in IE.

Add this line in componentDidMount [here](https://github.com/cerner/terra-framework/blob/75e009f148abc8712eaf61ca93cdc51c62ad38f8/packages/terra-layout/src/_LayoutSlidePanel.jsx#L69).

```
document.createElement('main');
```

We should also add `display: block;` [here](https://github.com/cerner/terra-framework/blob/75e009f148abc8712eaf61ca93cdc51c62ad38f8/packages/terra-layout/src/LayoutSlidePanel.module.scss#L72).

## Environment

* Component Name and Version: terra-layout
* Browser Name and Version: IE

## @ Mentions
@JakeLaCombe

PR #360, opened 2018-12-05
Shim for HTML Main Tag in IE Added to Layoutslidepanel

amichaelparker:
### Summary
Adds `document.createElement('main')` to resolve error in IE regarding the use of the <main> tag. 

Closes [#351](https://github.com/cerner/terra-framework/issues/351)


packages/terra-layout/src/_LayoutSlidePanel.jsx
```diff
@@ -68,6 +68,7 @@ class LayoutSlidePanel extends React.Component {
   }
 
   componentDidMount() {
+    document.createElement('main'); // shim for <main> tag not being recognized in IE
```
StephenEsser: How does this work? Is it just magic?

bjankord: https://weblog.west-wind.com/posts/2015/Jan/12/main-HTML5-Tag-not-working-in-Internet-Explorer-91011

Reading through that link, I think we may be able to drop the JS part and just keep the CSS setting `main { display: block}` I think the element creation may only be needed for IE8 and below is not included in our browser support list.

amichaelparker: I'll check it out.

amichaelparker: image

This is what I get in IE11 if I remove the JS line.

bjankord: ah, good testing. Ok, let's leave the script.

PR #362, opened 2018-12-06
Making ApplicationLinks test more reliable

tbiethman:
### Summary
The '--terra-application-tabs-spacer-width' variable was the offending style that was causing some timing issues with the application tabs test. I removed it for the time being, and I'd argue that it probably should be a static value going forward. But this should make the builds more stable.


packages/terra-application-links/tests/wdio/application-tabs-spec.js
```diff
@@ -25,7 +25,7 @@ describe('ApplicationTabs', () => {
         '--terra-application-tabs-color': '#000',
         '--terra-application-tabs-font-size': '1rem',
         '--terra-application-tabs-line-height': '1',
-        '--terra-application-tabs-padding': '1rem 2rem',
```
emilyrohrbough: We should probably remove the `--terra-application-tabs-padding` and `--terra-application-tabs-spacer-width` reference screenshots

yuderekyu: ah, this is using `themeCombinationOfCustomProperties`, so I'd think we only need to regenerate this single screenshot

PR #367, opened 2018-12-06
Remove use of componentWillReceiveProps from layout

bjankord:
### Summary

### DIFF
--- a/packages/terra-layout/src/Layout.jsx
+++ b/packages/terra-layout/src/Layout.jsx
@@ -44,6 +44,14 @@ class Layout extends React.Component {
     });
   }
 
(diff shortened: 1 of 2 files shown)


packages/terra-layout/src/Layout.jsx
```diff
@@ -44,6 +44,14 @@ class Layout extends React.Component {
     });
   }
 
+  static getDerivedStateFromProps(nextProps, prevState) {
+    if (nextProps !== prevState.prevProps) {
```
tbiethman: Is this conditional ever false?

bjankord: Yes, it will evaluate to true when changing pages on the doc site but on initial render it evaluates to false

PR #370, opened 2018-12-10
Remove use of componentWillReceiveProps from hookshot

bjankord:
### Summary


packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -128,6 +128,10 @@ class Hookshot extends React.Component {
     this.pendingTimeout = null;
   }
 
+  static getDerivedStateFromProps(nextProps) {
+    return { isEnabled: nextProps.isEnabled && nextProps.isOpen };
```
bjankord: Kinda feel like this should be wrapped in a conditional, check if these values are different then their prior state, but we weren't conditionally checking them before. Thoughts?

dkasper-was-taken: I believe we can just replace the two instances where we are using the state in componentDidMount and componentDidUpdate:
```
if (this.state.isEnabled) {
```
for:
```
if (this.props.isEnabled && this.props.isOpen) {
```

bjankord: Updated in this commit: https://github.com/cerner/terra-framework/pull/370/commits/5657cd58d5d022c71323767bf03ab8af9591301c

PR #371, opened 2018-12-10
Remove use of componentWillReceiveProps from application utility

bjankord:
### Summary


packages/terra-application-utility/src/utility/_UtilityMenu.jsx
```diff
@@ -79,21 +79,21 @@ class UtilityMenu extends React.Component {
...
+    if (nextProps.initialSelectedKey !== prevState.prevPropsInitialSelectedKey) {
+      return { currentKey: nextProps.initialSelectedKey };
     }
 
-    if (nextProps.menuItems !== this.props.menuItems) {
-      this.setState({
-        map: processMenuItems(nextProps.menuItems),
-      });
+    if (nextProps.menuItems !== prevState.prevPropsMenuItems) {
+      return { map: processMenuItems(nextProps.menuItems) };
```
tbiethman: I know this is how it was coded before, but do we really want two different returns here? It seems like we'd want to regenerate the map even if the initialSelectedKey changed.

PR #372, opened 2018-12-10
Remove use of componentWillReceiveProps from navigation side menu

bjankord:
### Summary


packages/terra-navigation-side-menu/src/NavigationSideMenu.jsx
```diff
@@ -104,12 +104,19 @@ class NavigationSideMenu extends React.Component {
...
+      prevPropsMenuItem: props.menuItems,
+    };
   }
 
-  componentWillReceiveProps(nextProps) {
-    const { items, parents } = processMenuItems(nextProps.menuItems);
-    this.setState({ items, parents });
+  static getDerivedStateFromProps(nextProps, prevState) {
+    if (nextProps.menuItems !== prevState.prevPropsMenuItem) {
+      const { items, parents } = processMenuItems(nextProps.menuItems);
```
dkasper-was-taken: I can't remember why I initially used two lines for this. Did esLInt need it? Can we change it to just return?
 ```
return processMenuItems(nextProps.menuItems);
```

bjankord: Updated in this commit: https://github.com/cerner/terra-framework/pull/372/commits/10caf55880034d18b3c8f3d7a409df60195432ad

Issue #359, opened 2018-12-04
Utility Menu Read Only Item


foxannefoxanne:
# Feature Request

## Description
We have a use case where we show a correlation id in the Utility Member. This ID is used to help correlate IDs to requests for better investigation. This would typically be something that a user would highlight and copy paste. However, there does not seem to be an option to have a read-only utility item, so any attempt to highlight/copy-paste this item will count as a button click and closes the menu. A read only option would greatly assist this workflow. 

## Additional Context / Screenshots
<img width="240" alt="screen shot 2018-12-04 at 1 50 32 pm" src="https://user-images.githubusercontent.com/7482294/49468885-af85bf80-f7cb-11e8-8f9b-a2b337fd1dea.png">

## @ Mentions
Consulted with @bjankord and he recommended logging a Read Only option terra bug. He also recommended opening a modal to allow users to copy, which I will follow-up on to see if it is a viable option on our side.

bjankord:
Looks like here is where we would want to start adding changes: https://github.com/cerner/terra-framework/blob/master/packages/terra-application-utility/src/utility/_UtilityMenuItem.jsx

yuderekyu:
### Potential Non Modal Solution

- Add a new prop, `isReadOnly`, to the [itemShape](https://github.com/cerner/terra-framework/blob/master/packages/terra-application-utility/src/Utils.js#L24). 
- Add corresponding prop `isReadOnly` to [`_UtilityMenuItem.jsx`](https://github.com/cerner/terra-framework/blob/master/packages/terra-application-utility/src/utility/_UtilityMenuItem.jsx).
- Inside `_UtilityMenu.jsx`'s menu item generation, conditionally check `isReadOnly` to not attach the `onChange` function to that menu item. Do this [here](https://github.com/cerner/terra-framework/blob/master/packages/terra-application-utility/src/utility/_UtilityMenu.jsx#L132).  
- Add an `onChange` existence check [here](https://github.com/cerner/terra-framework/blob/master/packages/terra-application-utility/src/utility/_UtilityMenuItem.jsx#L104-L106). 
- Use `readOnly` prop to create read-only styles for the menu item.

We'll have to handle edge cases. One example - if someone defines a menu item containing both `isReadOnly` and `isSelectable` -  I'd vote `isReadOnly` takes precedence.

Issue #272, opened 2018-09-17
Terra Utility Menu needs to have the option to switch it's role.


JakeLaCombe:
# Bug Report

## Description
The Terra Utility Menu is not always designed to be a navigation component. In our dev site for example, it's used as a menu. To ensure screen readers do not always read this as a navigation menu, we need to allow for the developer to change the value of role. We can default it to "navigation" for passivity purposes.

https://github.com/cerner/terra-framework/blob/01daffb58489853b7663bd406ba0dbd53df96e46/packages/terra-application-utility/src/utility/_UtilityMenu.jsx#L364

yuderekyu:
### Tech Design
Add new prop `menuRole` to ApplicationMenuUtility and ApplicationHeaderUtility components. This allows a developer to change the value of role.

| Prop     | Type | Default | Description  |
|----------|------|---------|--------------|
| `menuRole` | string | 'navigation' | Set the menu's role for accessibility. |

bjankord:
I think for this, we'll want to move the role attribute onto the `UL` element.
<img width="956" alt="screen shot 2019-02-11 at 2 35 32 pm" src="https://user-images.githubusercontent.com/633148/52592144-404fcb80-2e0b-11e9-8f85-c738196233f8.png">

The latest version of axe warns that the UL element inside of terra-utility, "must only directly contain `<li>`, `<script>` or `<template>` elements". 

I believe this issue is from the `role="button"` attributes on the `li` element. If we update the `ul` role to navigation or menu, it resolve the axe warning and seems to be more inline with the intention of the `<li role="button" />` markup.

supreethmr:
I would like to know can we have role at both utilitymenu and ul tag so that it will help user in terms accessibility as screen readers will read the role even when we click on outer div of ul tag which is content container in case of utility menu. Since keeping the role only at ul tag will make screen readers to read role only when we click on menu items and not on click of menu itself. Please share your thoughts on this. @JakeLaCombe

JakeLaCombe:
That's an interesting axe issue. I'm looking at this issue, and it looks like role is actually attached to the container of the navigation, which is the outer div. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role

It almost feels like we should attach role="navigation" to the div, and role="menu" to the ul. Maybe that approach will resolve the issue?

bjankord:
[...]

PR #373, opened 2018-12-10
Add Read-Only Property to UtilityMenuItems

amichaelparker:
### Summary
Adds a read-only property to the menu item shape, allowing menu items to be clicked on and copy/pasted without closing the menu on click.

Closes [#359](https://github.com/cerner/terra-framework/issues/359)


packages/terra-application-utility/src/utility/_UtilityMenuItem.jsx
```diff
@@ -127,6 +132,7 @@ class UtilityMenuItem extends React.Component {
     const bodyItemClassNames = cx([
       { 'header-utility-body-item': variant === Utils.VARIANTS.HEADER },
       { 'menu-utility-body-item': variant === Utils.VARIANTS.MENU },
+      { 'read-only': isReadOnly },
```
emilyrohrbough: If this is readOnly, would we need to change the role type of the list item and remove the click / keydown handlers?

amichaelparker: Thinking something like this to address this:
```
const roles = isReadOnly ? 'note' : 'button';
/* eslint-disable jsx-a11y/no-static-element-interactions, jsx-a11y/no-noninteractive-element-to-interactive-role */
const renderBodyItem = (fill, wrapOnKeyDown, handleSelection) => (
  <li
    {...customProps}
    tabIndex={!isReadOnly ? '0' : undefined}
    key={itemKey}
    onClick={!isReadOnly ? event => handleSelection(event, itemKey) : undefined}
    onKeyDown={!isReadOnly ? wrapOnKeyDown(itemKey, onKeyDown) : undefined}
    role={roles}
    className={bodyItemClassNames}
    aria-label={title}
    ref={this.setItemNode}
  >
    <Arrange
      fitStart={leftInset ? <IconCheckmark className={checkmarkClassNames} /> : null}
      fill={fill}
      fillAttributes={{ className: cx('menu-item-fill') }}
      fitEnd={rightInset ? <IconChevronRight className={chevronClassNames} /> : null}
      align="center"
      className={arrangeClassNames}
    />
  </li>);
```
Lint pops up an error about the tabindex on a non-interactive element, but if I do what is in the code above, that creates an accessibility issue (screen readers won't be able to tab to it to read it). Otherwise, I can just disable the rule for that line.

packages/terra-application-utility/src/utility/_UtilityMenuItem.module.scss
```diff
@@ -145,4 +145,17 @@
   .menu-item-fill {
     width: 100%;
   }
+
+  .read-only {
+    cursor: text;
+
+    &:hover {
+      background-color: var(--terra-application-header-utility-menu-body-item-hover-background-color);
```
yuderekyu: we should sync with UX, but I think a read only item will never have hover and focus states. We'd probably set `background-color: transparent` and `box-shadow: none`, instead of opening them up for themes

amichaelparker: I thought so too, but wasn't sure if I should leave the option there just in case. Will take it out!

neilpfeiffer: I need to review PR further, but generally anything that is "read-only" or non-interactive elements on the page never should have have any visual state changes (pointer/color,background-color/shadow etc.) for hover, focus, active, etc. (for all themes)

packages/terra-application-utility/src/utility/_UtilityMenuItem.jsx
```diff
@@ -149,15 +155,17 @@ class UtilityMenuItem extends React.Component {
...
       <li
         {...customProps}
         tabIndex="0"
         key={itemKey}
-        onClick={event => handleSelection(event, itemKey)}
-        onKeyDown={wrapOnKeyDown(itemKey, onKeyDown)}
-        role="button"
+        onClick={!isReadOnly ? event => handleSelection(event, itemKey) : undefined}
+        onKeyDown={!isReadOnly ? wrapOnKeyDown(itemKey, onKeyDown) : undefined}
+        role={roles}
```
yuderekyu: Nitpick - since we're doing inline conditionals, it would be fine do the same for role
```jsx
role={isReadOnly ? 'note' : 'button'}
```

I don't have much accessibility knowledge, but we may rework how these roles are set. Looks like we we may open a first class prop to configure this: https://github.com/cerner/terra-framework/issues/272. That is out of scope though, just something to ponder

packages/terra-application-utility/src/utility/_UtilityMenuItem.jsx
bjankord: What you have looks good to me. role="note" seems fine to me as well, hadn't heard of it before so I had to look it up.

packages/terra-application-utility/src/utility/_UtilityMenuItem.jsx
bjankord: Inlining this conditional should be fine. Let's tackle opening the role prop as a first-class prop in a separate PR.

packages/terra-application-utility/src/utility/_UtilityMenuItem.module.scss
bjankord: All the focus styles have been removed, the only remaining question I have is should we still allow people to focus to it via keyboard tab? I'm thinking we drop that functionality from read-only items in the menu as it should be non-interactive, e.g. no clicking on it to trigger something.

amichaelparker: I was originally thinking altering the tabindex would also hide the information from a screen reader, but I realized you can't tab to it anyway, so I threw the conditional back in there.

yuderekyu:
Let's document `isReadOnly` in the readme. https://github.com/cerner/terra-framework/blob/master/packages/terra-application-utility/docs/README.md

amichaelparker:
Added a (probably rough draft) explanation of the isReadOnly prop.

emilyrohrbough:
Looks good. Kinda random, but I am noticing the themed screenshots have grown in height? Any clue why?

amichaelparker:
> Looks good. Kinda random, but I am noticing the themed screenshots have grown in height? Any clue why?

This was some odd behavior I came across while working on this - I've only tested this with the examples, but essentially if you add items to the scroll list, and the number of items makes the scroll list longer than the containing menu height, it actually increases the height of the menu as well.

In the comparison screenshots, the read-only menu item has been added to the new pics, so it has pushed the total height of the menu down as well. If anyone has any thoughts on addressing that, I can slip in a fix here, or it can be another PR.

mmalaker:
Functional verification completed. The utility menu example displays and functions as expected with the read-only item included.

PR #376, opened 2018-12-10
Remove use of componentWillReceiveProps from popup

bjankord:
### Summary


packages/terra-popup/src/Popup.jsx
```diff
@@ -130,10 +130,10 @@ class Popup extends React.Component {
     this.contentWidth = PopupWidths[props.contentWidth];
   }
 
-  componentWillReceiveProps(newProps) {
-    this.isContentSized = newProps.contentHeight !== 'auto' && newProps.contentWidth !== 'auto';
-    this.contentHeight = PopupHeights[newProps.contentHeight];
-    this.contentWidth = PopupWidths[newProps.contentWidth];
+  componentDidUpdate() {
```
dkasper-was-taken: We'll want to switch this to use shouldComponentUpdate, otherwise we'll render without updating the variables.

bjankord: @dkasper-was-taken I tried using `shouldComponentUpdate` and some variations of calling the code from `componentWillReceiveProps` in the render method. I ended up getting a working solution by moving the code into `getDerivedStateFromProps`. I'm open to other ways to solve this.

dkasper-was-taken: Hrmm.. I'll take another look. I know in using with getDerivedStateFromProps it'll trigger every render, I am mainly worried about the scenario where the popup is constrained and/or using auto sizing, then triggers an additional render, but maybe it's not a problem.

bjankord: @dkasper-was-taken I'm planning to hold off on merging this until you get back so you have time to review when you get back, no need to worry about it now. Enjoy your break!

dkasper-was-taken: @bjankord Cool Thanks!

bjankord: Updated to use `shouldComponentUpdate`: https://github.com/cerner/terra-framework/pull/376/commits/4fb95055597e6b73df8fcd4c38741bd954c8bdf1

PR #378, opened 2018-12-10
Remove use of componentWillReceiveProps from infinite list

bjankord:
### Summary


packages/terra-infinite-list/src/InfiniteList.jsx
```diff
@@ -111,20 +111,18 @@ class InfiniteList extends React.Component {
     this.handleRenderCompletion();
   }
 
-  componentWillReceiveProps(newProps) {
-    const newChildCount = React.Children.count(newProps.children);
+  componentDidUpdate() {
```
dkasper-was-taken: Switching to componentDidUpdate here will have unintended consquences. We'll want to use shouldComponentUpdate in this case:
```
  shouldComponentUpdate(nextProps) {
    const newChildCount = React.Children.count(nextProps.children);
    if (newChildCount > this.childCount) {
      this.lastChildIndex = this.childCount;
      this.loadingIndex += 1;
      this.updateItemCache(nextProps);
    } else if (newChildCount < this.childCount) {
      this.initializeItemCache(nextProps);
    } else {
      this.childrenArray = React.Children.toArray(nextProps.children);
    }

    return true;
  }
```

bjankord: Thanks @dkasper-was-taken! Updated in this commit: https://github.com/cerner/terra-framework/pull/378/commits/054e3283e231fe3bffde04fa0218eed79456940a

Issue #265, opened 2018-09-13
Correct MenuItem management in NavigationSideMenu


tbiethman:
# Bug Report

## Description
There are 2 separate issues to be addressed:
1. The shapes required for the menuItems prop accept an `id` value that is supposed to be added to the MenuItem markup as an actual `id` attribute. However, that is not occurring.
2. The data-menu-item attribute added to MenuItems currently has a value based on the MenuItem's text. However, this value may be an internationalized string and is non-deterministic, making it difficult to use for selector queries. We need to use the MenuItem's `key` value instead of `text` for that attribute value.

## Additional Context / Screenshots

* Location of `id` in shape: https://github.com/cerner/terra-framework/blob/79b795cdbcab1211b5dc09f653d3af48d3106e41/packages/terra-navigation-side-menu/src/NavigationSideMenu.jsx#L40
* Location of data-menu-item: https://github.com/cerner/terra-framework/blob/79b795cdbcab1211b5dc09f653d3af48d3106e41/packages/terra-navigation-side-menu/src/NavigationSideMenu.jsx#L170

## Expected Behavior
* `id` attribute is appropriately added to the MenuItem markup
* `data-menu-item` has `key` based value.

## Environment

* Component Name and Version: All
* Browser Name and Version: All
* Node/npm Version: [e.g. Node 8/npm 5] All
* Webpack Version: All 
* Operating System and version (desktop or mobile): All

ShettyAkarsh:
The issue 1 mentioned above is working correctly. we are setting the id attribute value of Menuitem in BuildListItem function of NavigationSideMenu.jsx.

tbiethman:
This was fixed by #407

PR #407, opened 2018-12-14
Set Data-menu-item attribute value of Menuitem to Unique key value.  #265

ShettyAkarsh:
### Summary
Resolves the second issue mentioned in #265. The fix is to set the Data-menu-item attribute value of div containing each menu item of Navigation menu to unique value. 

### Additional Details
As the first issue mentioned in #265 is working correctly. it is not taken into consideration during this fix.


packages/terra-navigation-side-menu/src/MenuItem.module.scss
```diff
@@ -15,7 +15,7 @@
     color: var(--terra-navigation-side-menu-item-color, #1c1f21);
     cursor: pointer;
     display: flex;
-    font-size: var(--terra-navigation-side-menu-item-size, 1.071rem);
+    font-size: var(--terra-navigation-side-menu-item-font-size, 1.071rem);
     line-height: var(--terra-navigation-side-menu-item-line-height, 1.42857);
```
emilyrohrbough: @yuderekyu Is this non-passive?

yuderekyu: yeah, i'd say revert this change until we major version bump this with the rest of the refactored theme variables

ShettyAkarsh: Reverted this changes.

PR #410, opened 2018-12-19
Remove _PopupContent Unused Styling, Add Text Label for A11y

amichaelparker:
### Summary
Removes unused 'close' styling class from _PopupContent close button. 'text' prop added to same, for conversion by Button component into aria-label.

Closes [#173](https://github.com/cerner/terra-framework/issues/173)


packages/terra-popup/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ Unreleased
 ### Changed
 * Prevent click events from propagating from popup overlay
 
+### Fixed
+* Removed 'close' class from _PopupContent button
```
ryanthemanuel: I believe you might want to wrap _PopupContent in `` (i.e. `PopupContent`).  Otherwise, I think the markdown will treat this as italicized

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -191,14 +200,16 @@ class PopupContent extends React.Component {
       ...customProps
     } = this.props;
 
+    const { intl } = this.context;
```
dkasper-was-taken: Move the intl and close declarations inside the block:
```
if (isFullScreen && !isHeaderDisabled) {
```

amichaelparker: Done!

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -94,10 +94,19 @@ const defaultProps = {
   isWidthAutomatic: false,
 };
 
+const contextTypes = {
+  /* eslint-disable consistent-return */
```
emilyrohrbough: why do we need to add this here? I don't think we've had to in other places

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -94,10 +94,19 @@ const defaultProps = {
   isWidthAutomatic: false,
 };
 
+const contextTypes = {
```
yuderekyu: We now recommend using `<FormattedMessage/>` and `injectIntl` over context. This PR should help: https://github.com/cerner/terra-ui/pull/132

amichaelparker: Added here (probably): [103fc45](https://github.com/cerner/terra-framework/pull/410/commits/103fc4576d7b40a2090f51de7e32e198446e7530)

packages/terra-popup/CHANGELOG.md
```diff
@@ -3,16 +3,19 @@ Changelog
...
+* intl prop added to `_PopupContent` for i18n message formatting
 
 4.21.0 - (January 2, 2019)
 ------------------
 ### Changed
 * Prevent click events from propagating from popup overlay
 
-### Fixed
-* Removed 'close' class from `_PopupContent` button
-* Added text prop to `_PopupContent` button for conversion to aria-label in Button component
```
amichaelparker: This is from this PR, I must have resolved a merge conflict incorrectly, so I moved this back to unreleased.

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -50,6 +51,10 @@ const propTypes = {
    * The maximum width value in px, to be applied to the content container. Used with responsive behavior for actual width.
    */
   contentWidthMax: PropTypes.number,
+  /**
```
bjankord: You can add `@private` here and this will hide the intl prop from being displayed in the docs. Since this prop is added via the `injectIntl` HOC, consumers don't need to be aware/interface with it. Here's [the related PR](https://github.com/cerner/terra-core/pull/2135) that adds support for `@private`

emilyrohrbough: @bjankord do we need to log an issue to add this @private key to existing private props?

bjankord: Yeah that would be good to go through and update.

PR #423, opened 2019-01-04
[application-layout] Use Avatar 2.0 In Examples

yuderekyu:
### Summary
Use Avatar version 2 in documentation and examples.

This will also fix clinical-theme failures. When application-layout screenshots were generated, version 2 was somehow pulled in. Now on a clean install, version 1 is correctly pulled in. 

We have documentation that uses Avatar, so let's opt to upgrade the documentation and examples to use version 2 instead of regenerating clinical-theme screenshots to use version 1.


yuderekyu:
@bjankord sure, that wouldn't require another PR in `terra-framework` since we already pull in v1. Still not sure how we pulled in v2 on the clinical theme side o.O

Issue #379, opened 2018-12-10
Ability to theme text-decoration of navigation items in application-layout


nickpith:
# Feature Request

## Description
There is a need to be able to theme the text-decoration CSS property of the navigation item labels in the header portion of the layout. The specific request is to be able to make every nav item appear in uppercase.

mhemesath:
Minor comment, but the css property to change to uppercase would be `text-transform` I believe.

nickpith:
My bad. I don't know why I put text-decoration.

alex-bezek:
@nickpith does the text-transform have to be a theme property? They can't just pass in props with capitalized text (configured in the backend for the client that wants it this way)?

nickpith:
It doesn't have to be in my mind. I mentioned that as a possible workaround.

alex-bezek:
oh, i assumed side nav since I think the client that utilizes this style with our existing setup only has side nav

alex-bezek:
Confirmed with our team that the expectation is to allow the side nav menu and the application-tabs in the top header to text-transform. There is no need for the text shown by the logo, the user's name, or the buttons in the profile pop up to handle text transform

alex-bezek:
@bjankord I have started already https://github.com/cerner/terra-framework/compare/support-text-transform-nav?expand=1 wrapping up some tests

PR #435, opened 2019-01-15
Support text transform nav

alex-bezek:
### Summary
Some of our clients choose to theme their components heavily to match the design of the content they embed. One specific request is the ability to make their navigation text be all caps. This exists in their current navigation implementation, and in order to switch over to fully utilize the terra-application-layout, we need this functionality.

Our initial reaction to this request was that since the nav text is configurable, the client could just provide all caps text to store in that service, which would then be passed to the terra nav and display as all caps.

However, this was shot down as "it should be a theme option". The goal is to let the text be theme-able, and not to try to achieve this look by having to programmatically or manually go update all of a clients nav items.

Open issue here https://github.com/cerner/terra-framework/issues/379

@cerner/terra


alex-bezek:
I saw this class being added https://github.com/cerner/terra-framework/blob/master/packages/terra-application-links/src/tabs/_Tab.jsx#L149 but didn't see it in the modules.scss file. I noticed while reading through the dom structure and saw it was just "tab-inner" with no module obfuscation. Is this used for something?

alex-bezek:
oh come on @StephenEsser it's a perfect 5/7 so far. no bots had to yell at me

bjankord:
I'd like to get @neilpfeiffer to take a look so he is aware of the change.

neilpfeiffer:
Reviewed: PR #435
Resolves: Issue #379

**Reviewed:**
- Adds ability for Client-based themes to UPPERCASE their top navigation labels

**Related:**
- @BenBoersma this change will need to be incorporated into your ocs-theme-variables branch
- @tbiethman @dkasper-was-taken this change will need to be incorporated into your framework refactor branch

\
\

**Pass** |
-----------|
UX Review|

PR #437, opened 2019-01-17
Updates for FF and IE Driver Compatibility

emilyrohrbough:
### Summary
An update for IE and FF driver compatibility.


packages/terra-time-input/tests/wdio/time-input-twelve-hour-mobile-spec.js
```diff
@@ -8,11 +8,7 @@ const ignoredA11y = {
 describe('Time Input Twelve Hour Mobile', () => {
   before(() => {
     browser.setViewportSize(Terra.viewports('tiny')[0]);
-    browser.execute(() => {
-      if (!window.ontouchstart) {
-        window.ontouchstart = 'true';
-      }
-    });
+    browser.execute('if (!window.ontouchstart) { window.ontouchstart = "true"; }');
```
JakeLaCombe: For every webdriver io test we have, does it do the test as a new page refreshed in the browser, or does it use the same page and just render different markup? Making sure we don't need to reset this when the test is complete.

emilyrohrbough: @JakeLaCombe Our current wdio setup refreshes the page after each test. This is something we need to remove to speed up testing. See https://github.com/cerner/terra-dev-site/issues/136

emilyrohrbough: That being said, I can add the removal now so we don't have to find this when we remove it 😄

JakeLaCombe: sounds good!

PR #442, opened 2019-01-22
Add Example for Validation with Datepicker

### LINKED ISSUE #441: Document integration of terra-date-picker with react-final-form
# Feature Request

## Description
It's unclear to teams how terra-date-picker should be integrated with react-final-form. Since the component is uncontrolled, it doesn't work like the other traditional form inputs handle.

JakeLaCombe:
### Summary
Resolves #441 

People have asked for an example of integrating react-final-form with terra-date-picker. This PR addresses that.

One thing to note: Since DatePicker is uncontrolled, it can not be combined with parsing logic.


JakeLaCombe:
Input masking. React final form calls is parsing

bjankord:
@JakeLaCombe Ah, like adding `/` to the date as the types into the input.

PR #446, opened 2019-01-25
Add terra-layout-slide-panel-transition custom property

bjankord:
### Summary
Made transition property themable so if teams want the layout slide panel to transition in, they can theme it to do so. By default, there will be no transition.


bjankord:
Closing in favor of picking this work up post-release of the CSS custom property update. We use the `transitionEnd` event to manage some internal state of the layout slide panel. If we disable the transition, we lose the ability to listen for the `transitionEnd` event and thus, lose the ability to infer the state of the panel. Will need to enhance the layout component to work without `transitionEnd` event if we make the transition themeable.

PR #479, opened 2019-01-30
clean up dependencies

emilyrohrbough:
### Summary
- bump terra-dev-site
- bump stylelint-config-terra
- Specify all dependencies as MVB to remove peer dependency warnings


package.json
```diff
@@ -108,10 +101,9 @@
     "react-test-renderer": "^16.4.2",
     "shx": "^0.3.2",
     "stylelint": "^9.2.0",
-    "stylelint-config-terra": "^1.0.0",
+    "stylelint-config-terra": "^2.0.0",
     "terra-base": "^4.0.0",
-    "terra-dev-site": "^2.0.0",
-    "terra-toolkit": "^4.16.0",
-    "xfc": "^1.2.1"
```
bjankord: I believe we need to keep xfc for the terra-embedded-content-consumer package:
https://github.com/cerner/terra-framework/blob/master/packages/terra-embedded-content-consumer/package.json#L27

emilyrohrbough: Ah, we defined xfc in terra-dev-site, so I didn't realize we needed to defining it here. Makes sense. 👌 will re-add

Issue #490, opened 2019-01-02
[terra-abstract-modal] Disable 'escapeDeactivates' option for AbstractModal's focus trap


tbiethman:
# Bug Report

## Description
The focus-trap-react package we use will disable the focus trap by default when the escape key is pressed. This is normally fine, because in the vast majority of cases, the AbstractModal also closes when the escape key is pressed. However, for implementations like the NotificationDialog in terra-framework that do not close when escape is pressed, the focus trap will be disabled and focus will leave the NotificationDialog. 

## Steps to Reproduce

1. Go here: https://engineering.cerner.com/terra-framework/#/components/terra-notification-dialog/notification-dialog/notification-dialog
2. Click 'Disclose' button in 'Notification Dialog On Modal Manager' example.
3. Click 'Trigger NotificationDialog' button within modal
4. Press Escape
5. Press Tab multiple times and see focus being set outside the NotificationDialog

## Expected Behavior
Focus should be trapped within the modal as long as it's open.

## Possible Solution
We need to disable the `escapeDeactivates` option [1] for the focus trap within the AbstractModal's implementation [2].

[1] https://github.com/davidtheclark/focus-trap#focustrap--createfocustrapelement-createoptions
[2] https://github.com/cerner/terra-core/blob/master/packages/terra-abstract-modal/src/_ModalContent.jsx#L128

PR #493, opened 2019-01-31
Add 'escapeDeactivates' Option to 'abstract-modal' Focus Trap

amichaelparker:
### Summary
Updated focus trap option 'escapeDeactivates' to default to 'false' to prevent the escape key from dropping focus.

Closes [#490](https://github.com/cerner/terra-framework/issues/490)


packages/terra-abstract-modal/src/AbstractModal.jsx
```diff
@@ -31,6 +31,10 @@ const propTypes = {
    * If set to true, the modal will close when a mouseclick is triggered outside the modal.
    */
   closeOnOutsideClick: PropTypes.bool,
+  /**
+   * If set to true, the Escape key will trigger deactivation of the focus trap.
+   */
+  escapeDeactivates: PropTypes.bool,
```
tbiethman: I don't think we need a prop to enable this. There isn't a scenario where we'd want the modal to be open and the focus trap to be disabled, right?

bjankord: Here where a few scenarios I thought where we need a modal open but the focus to not be within the modal focus trap.

* Notification Dialog open above a modal
* Popup / Select / Datepicker open inside a modal (shift focus to these which are portalled outside of the modal DOM/focus trap)

I haven't thought through what impact the ecapseDeactive functionality would have on those scenarios much. I think we have test cases for these we could test and see if having it always be default works. If so, I'm fine with removing the prop.

JakeLaCombe: It's mostly a personal opinion, but I really like to use the is syntax when dealing with boolean variables. For this instance, I think isDeactivatedByEscape or isDeactivatedByEsc would be great candidate names for that variable.

emilyrohrbough: @JakeLaCombe  I actually think we defined the `is` / `has` boolean prop naming as a terra-convention for boolean props. We have it documented here: https://engineering.cerner.com/terra-ui/#/contributing/terra-ui/conventions

amichaelparker: @tbiethman @bjankord Where did we land on having the prop at all? Yay/nay?

tbiethman: I see two solutions:
1. Don't expose a prop and never allow the FocusTrap to deactivate itself.
2. Set the escapeDeactivates option based on our existing closeOnEsc prop, set the onDeactivate option with a function that will call the onRequestClose prop, and remove our own keydown handlers for escape handling. 

I think 2 would allow us to more easily transition to newer versions of the focus-trap-react that support auto-pausing nested focus traps: https://github.com/davidtheclark/focus-trap#one-at-a-time. However, we'd definitely have to investigate the cascading effects of such a change.

amichaelparker: I talked with @bjankord, and opted for option 1 for the time being. He suggested coming back around to option 2 after investigating the effects of that change.

PR #499, opened 2019-02-07
Wrapped collapsible menu items in <li> tags

AlexBrizi:
### Summary
Wrapped the CollapsibleMenuItems in li tags. Also ensured that doing so would not add any extra spacing (see before & after screenshot).

before and after change

Resolves [#396](https://github.com/cerner/terra-framework/issues/396)

### DIFF
--- a/packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItemGroup.jsx
+++ b/packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItemGroup.jsx
@@ -70,22 +70,22 @@ class CollapsibleMenuViewItemGroup extends React.Component {
     const { isCollapsibleMenuItem } = this.context;
 
     if (isCollapsibleMenuItem && selectedKeys.length) {
-          return (
-            <li>
-              <Menu.ItemGroup {...customProps} onChange={this.handleOnChange}>
-                {children}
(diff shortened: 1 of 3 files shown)


packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItemGroup.jsx
```diff
@@ -70,18 +70,22 @@ class CollapsibleMenuViewItemGroup extends React.Component {
...
+            <li>
+              <Menu.ItemGroup {...customProps} onChange={this.handleOnChange}>
+                {children}
+              </Menu.ItemGroup>
+            </li>
+          );
+        } if (isCollapsibleMenuItem) {
+          return (
+            <li>
+              <List {...customProps} onChange={this.handleOnChange}>
```
emilyrohrbough: Hmmm doesn't this throw an accessibility error or screen reader error by having a list nested in a list item??

bjankord: I believe these get nested in a UL. Before this change we had the following structure:

```
ul
  ul 
    li
```

This updates wraps the nested `ul` in a `li`.

```diff
ul
+  li
    ul 
      li
```

PR #500, opened 2019-02-07
Fix lint errors related to eslint-react-plugin update

bjankord:
Fixes all but 1 lint error. I'm curious how to handle the remaining lint error.

### DIFF
--- a/packages/terra-application-utility/src/utility/_UtilityMenu.jsx
+++ b/packages/terra-application-utility/src/utility/_UtilityMenu.jsx
@@ -73,7 +73,6 @@ class UtilityMenu extends React.Component {
     this.handleOnChange = this.handleOnChange.bind(this);
     this.handleOnKeyDown = this.handleOnKeyDown.bind(this);
     this.pop = this.pop.bind(this);
-    this.push = this.push.bind(this);
     this.state = {
       map: processMenuItems(props.menuItems),
       currentKey: props.initialSelectedKey,
@@ -189,9 +188,8 @@ class UtilityMenu extends React.Component {
 
   /**
    * Function to trigger when an item is selected.
-   * 1. Has children: navigate to the next page
-   * 2. Toggles: trigger onChange without closing the menu.
-   * 3. Endpoint: close menu and trigger onChange.
+   * * Has children: navigate to the next page
+   * * Endpoint: close menu and trigger onChange.
    * @param {*} event
    * @param {*} key
    */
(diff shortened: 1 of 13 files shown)


packages/terra-application-utility/src/utility/_UtilityMenu.jsx
```diff
@@ -199,6 +199,7 @@ class UtilityMenu extends React.Component {
     const { childKeys } = this.getItem(key);
     const item = this.getItem(key);
     if (childKeys && childKeys.length > 0) {
+      // TODO: Figure out intention of previousKey variable
```
bjankord: The linter throws an error saying `previousKey` is unused state. Looking in this file, it doesn't seem to be referenced elsewhere. @yuderekyu do you have any thoughts on what this was intended for? There are refernces for `previousKeyStack`.

packages/terra-application-utility/src/utility/_UtilityMenu.jsx
```diff
@@ -78,7 +78,7 @@ class UtilityMenu extends React.Component {
       map: processMenuItems(props.menuItems),
       currentKey: props.initialSelectedKey,
       focusIndex: -1,
-      previousKeyStack: [],
+      previousKeyStack: [], // eslint-disable-line react/no-unused-state
```
bjankord: This lint error seems erroneous, looking through this file we use `previousKeyStack` in a few places. I'm wondering if this is a case that the eslint-react-plugin doesn't account for correctly and is throwing a warning for it.

tbiethman: It's exclusively being referenced in setState function arguments; I could understand it being hidden from from the linter.

packages/terra-application-utility/src/utility/_UtilityMenu.jsx
bjankord: Refactored this with the help of @tbiethman and @yuderekyu: https://github.com/cerner/terra-framework/pull/500/commits/dbd7a710b5ace0e20de2824186b3e885aa2647f2

packages/terra-application-utility/src/utility/_UtilityMenu.jsx
bjankord: It seems there may be an update to eslint-plugin-react that will view state used in this way as used instead of unused: https://github.com/yannickcr/eslint-plugin-react/issues/1964

PR #502, opened 2019-02-07
[terra-disclosure-manager] Removing unused dependencies

tbiethman:
### Summary
I'm removing the dependencies that are no longer being used by the package. They were once used by the disclosure manager's examples, but those examples were removed some time ago.

This should result in fewer major version bumps to the package going forward.

### DIFF
--- a/packages/terra-disclosure-manager/package.json
+++ b/packages/terra-disclosure-manager/package.json
@@ -23,15 +23,10 @@
   "homepage": "https://github.com/cerner/terra-framework#readme",
   "peerDependencies": {
     "react": "^16.4.2",
-    "react-dom": "^16.4.2",
-    "terra-base": "^4.0.0"
+    "react-dom": "^16.4.2"
   },
   "dependencies": {
-    "classnames": "^2.2.5",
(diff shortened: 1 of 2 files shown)


tbiethman:
A successful build is dependent on #500

packages/terra-disclosure-manager/package.json
```diff
@@ -23,15 +23,10 @@
   "homepage": "https://github.com/cerner/terra-framework#readme",
   "peerDependencies": {
     "react": "^16.4.2",
-    "react-dom": "^16.4.2",
-    "terra-base": "^4.0.0"
```
tbiethman: I removed the peer dependency on terra-base because the disclosure manager doesn't actually import anything from it.

emilyrohrbough: terra-doc-template requires a peer of terra-base. Should we remove it from this level?

tbiethman: If terra-doc-template requires a peer dependency of terra-base, it'll still complain at install if one isn't provided. Removing it here will just result in one less package complaining about it missing, right?

PR #503, opened 2019-02-08
 Fixes issue #2074 DatetimePicker throws NPE if manually entered valid datetime is outside of allotted datetime range

ShettyAkarsh:
### Summary
Resolves the NPE error when user tries to enter time along with valid date outside of the allotted datetime range.

The fix is to validate whether the date entered is within the datetime range after the date value is validated.

Resolves #2074

### DIFF
--- a/packages/terra-date-time-picker/src/DateTimeUtils.js
+++ b/packages/terra-date-time-picker/src/DateTimeUtils.js
@@ -1,4 +1,5 @@
 import moment from 'moment-timezone';
+import DateUtil from 'terra-date-picker/lib/DateUtil';
 
 class DateTimeUtils {
   static createSafeDate(date) {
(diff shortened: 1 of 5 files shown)


packages/terra-date-time-picker/tests/jest/DateTimePicker.test.jsx
```diff
@@ -116,3 +118,19 @@ it('should render a disabled date time picker', () => {
     </IntlProvider>));
   expect(datePicker).toMatchSnapshot();
 });
+
+it('Should not throw any errors while date value is outside of the Min, Max range and new time value is being entered ', () => {
+  const allMessages = Object.assign(messages, dateInputMessages, timeInputMessages);
+  const datePicker = mount((
```
JakeLaCombe: Can you put more new lines around some of these blocks. It feels very cluttered here.

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -109,6 +110,10 @@ class DateTimeUtils {
   static getStandardTZDisplay() {
     return moment('2017-01-01').tz(moment.tz.guess()).format('z');
   }
+
+  static convertDateTimeStringToMomentObject(date, time, dateformat) {
```
JakeLaCombe: Good move!

packages/terra-date-time-picker/tests/jest/DateTimePicker.test.jsx
ShettyAkarsh: Addressed

PR #504, opened 2019-02-08
Add Required Prop to Terra Date Picker

### LINKED ISSUE #470: Datepicker add `required` prop
# Issue Description
Add a `required` prop to datepicker to match the forms api.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Datepicker component includes required prop.

## Current Behavior
Datepicker does not include required prop.

gabeparra01:
### Summary
Added required prop to terra-date-picker so that it matches the required prop in terra-form components.

### Additional Details
Resolves #470 

@cerner/terra


packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -342,12 +352,16 @@ class DatePicker extends React.Component {
...
-        <ResponsiveElement
-          responsiveTo="window"
-          defaultElement={portalPicker}
-          small={popupPicker}
-        />
+      <div className={cx(['container-div'])}>
+        {required && <span className={cx(['required'])}>*</span>}
+        <div className={cx(['date-picker'])}>
+          <ResponsiveElement
+            className={cx(['date-picker'])}
```
JakeLaCombe: It looks like date-picker is added as a class twice now.

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -342,12 +352,16 @@ class DatePicker extends React.Component {
     );
 
     return (
-      <div className={styles['date-picker']}>
-        <ResponsiveElement
-          responsiveTo="window"
-          defaultElement={portalPicker}
-          small={popupPicker}
-        />
+      <div className={cx(['container-div'])}>
```
JakeLaCombe: instead of container-div, I would like a more specific name, like date-picker-container. I would be ok with that one.

JakeLaCombe: date-picker-wrapper is ok as well, as another possible idea.

packages/terra-date-picker/tests/wdio/date-picker-spec.js
```diff
@@ -330,4 +330,22 @@ describe('Date Picker', () => {
     Terra.should.matchScreenshot();
     Terra.should.beAccessible({ rules: ignoredA11y });
   });
+
+  describe('Required Date Picker', () => {
+    before(() => browser.url('/#/raw/tests/terra-date-picker/date-picker/date-picker-required'));
+
+    it('creates a hidden input and empty value attr. when no date is entered', () => {
```
JakeLaCombe: Don't worry about checking for attributes inside the wdio tests. Those are best left to the jests tests.

JakeLaCombe:
Nice work! Address those, and you are good!

gabeparra01:
Thanks! :)

packages/terra-date-picker/src/DatePicker.jsx
gabeparra01: Ooops. fixed it here: https://github.com/cerner/terra-framework/pull/504/commits/8035961724c3db36d70929ae027084bd24b5a293

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -342,12 +352,16 @@ class DatePicker extends React.Component {
...
+        {required && <span className={cx(['required'])}>*</span>}
```
benbcai: Have you checked with UX that adding an asterisk to the front of the date input is how it should look like for required? This is just the date input instead of a date field as in https://engineering.cerner.com/terra-core/#/tests/terra-form-field/form-field/field-label-required

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -239,6 +248,7 @@ class DatePicker extends React.Component {
       onClickOutside,
       onSelect,
       requestFocus,
+      required,
```
benbcai: Should add `aria-required`. See [form-input](https://github.com/cerner/terra-core/blob/master/packages/terra-form-input/src/Input.jsx#L109-L111 ).

packages/terra-date-picker/src/DatePicker.jsx
gabeparra01: date-picker-wrapper is a good one: https://github.com/cerner/terra-framework/pull/504/commits/2203fba9c165fddf755cb6251760768116a5b678

packages/terra-date-picker/tests/wdio/date-picker-spec.js
gabeparra01: Oh I see. I removed the attribute checks: https://github.com/cerner/terra-framework/pull/504/commits/49decfbc4d5c6a3500eff04684e94fbe0c23eb30

packages/terra-date-picker/src/DatePicker.jsx
gabeparra01: Since the terra-date-picker doesn't have a default label, anyone that uses it can add a label and then wrap both of those elements int a terra-form-field to get the required asterisk next to the label. My understanding of #470 is that consumers wanted to be able to add a required asterisk by just passing in a prop

packages/terra-date-picker/src/DatePicker.jsx
gabeparra01: Thanks for catching that: https://github.com/cerner/terra-framework/pull/504/commits/85f234d80bd58813cc7d4a3f3cdd119875811db6

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -250,6 +260,12 @@ class DatePicker extends React.Component {
     delete customProps.onCalendarButtonClick;
     delete customProps.onInputFocus;
 
+    const attributes = Object.assign({}, customProps);
+
+    if (required) {
+      attributes['aria-required'] = 'true';
```
emilyrohrbough: would it be helpful to always have the aria-required attribute and we toggle true or false on the required prop value? @JakeLaCombe

packages/terra-date-picker/src/DatePicker.jsx
emilyrohrbough: Should we be directly add this to the date-pick? Or should we create a new component called date-picker-field which uses terra-form-field to wrap the date picker? That would give the date-picker access styling, etc that form-field provides.

For example, what about the error text that may be associated with the required field?

JakeLaCombe: I would be fine honestly with leaving it off for now, and waiting for when we add a field component to this for the rendering of the *. The key here would be making sure the inputs are getting their correct attributes.

packages/terra-date-picker/src/DatePicker.jsx
JakeLaCombe: I think how it is should be what we need. the aria should only be set if the input is required.

packages/terra-date-picker/src/DatePicker.jsx
benbcai: I'm also fine with leaving out the rendering of the * and just set the required attribute...just like what Input does today. I just think rendering the * directly in front of the date input is inconsistent with all other form fields where the * is rendered in front of the label.

bjankord:
@neilpfeiffer do you UX guidance on the placement of this asterisk.

packages/terra-date-picker/src/DatePicker.jsx
bjankord: Talked with our UX designer @neilpfeiffer. The preference from UX is to leaving the asterisk out of the terra-datepicker component. This will be added in a a terra-datepicker-field component we create instead. 

For this PR, I agree with the others, let's leave the required prop and functionality of adding the aria-required and just drop the rendering of the asterisk.

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -261,7 +273,7 @@ class DatePicker extends React.Component {
 
     const portalPicker = (
       <ReactDatePicker
-        {...customProps}
+        {...attributes}
```
benbcai: I'm not sure that this is going to work as expected because these attributes are spread onto ReactDatePicker and I don't think it takes custom props. The first class `required` prop should be added to the inputAttributes prop in DateInput. Adding `aria-required` and setting the `required` prop to Input would need to be done in [DateInput](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/src/DateInput.jsx#L161-L170).

packages/terra-date-picker/src/DatePicker.jsx
benbcai: Looks like you don't need to set `aria-required` here after all since [DateInput](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/src/DateInput.jsx#L161-L170) uses Input and Input will add [aria-required](https://github.com/cerner/terra-core/blob/master/packages/terra-form-input/src/Input.jsx#L109-L111) automatically if `required` is set. Having said that, you should pass `required` prop down to the `required` prop in ReactDatePicker.

packages/terra-date-picker/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ Changelog
 
 Unreleased
 ----------
+### Added
+* Added 'required' prop to match terra-form components' required prop
+git
```
StephenEsser: Suspicious

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -250,6 +256,12 @@ class DatePicker extends React.Component {
     delete customProps.onCalendarButtonClick;
     delete customProps.onInputFocus;
 
+    const attributes = Object.assign({}, customProps);
```
StephenEsser: `customProps` is already be a local object. You should be able to just conditionally add the aria required attribute directly to customProps. ( If we want to keep the aria attribute pending the other discussion )

```
if (required) {
  customProps['aria-required'] = 'true';
}
```

packages/terra-date-picker/package.json
```diff
@@ -41,6 +41,9 @@
     "terra-modal-manager": "^5.6.0",
     "terra-responsive-element": "^4.0.0"
   },
+  "devDependencies": {
+    "terra-enzyme-intl": "^3.0.0"
```
gabeparra01: The Jest tests could no longer resolve the path to terra-enzyme-intl so I added it as a dev dependency

packages/terra-date-picker/src/DatePicker.jsx
gabeparra01: @benbcai Is this what you had in mind? https://github.com/cerner/terra-framework/pull/504/commits/14324e7a538b9f6496afbbf408ae5c2f12e73378

packages/terra-date-picker/CHANGELOG.md
gabeparra01: 😖https://github.com/cerner/terra-framework/pull/504/commits/ea6129825b194a995a2325d06c46f8cdc84fd4c6

packages/terra-date-picker/src/DatePicker.jsx
benbcai: @gabeparra01 Yes. If you inspect the date input when you passed in the `required` prop, it should have the required attribute and `aria-label=true`.

packages/terra-date-picker/package.json
benbcai: You shouldn't need to add terra-enzyme-intl as a dev dependency here. It is already added [here](https://github.com/cerner/terra-framework/blob/master/package.json#L106). Did you tried to do a clean:install before adding it here to see if it works?

gabeparra01: I tried installing in the datepicker package but not the entire project. I will try that now

gabeparra01: Yep, that fixed it: https://github.com/cerner/terra-framework/pull/504/commits/6a98f84f536470ae7926bf7fb6e772eda34f5e78

benbcai:
You can delete the `Required_Date_Picker[date_picker_required].png` wdio reference screenshot since it is no longer needed.

gabeparra01:
@benbcai I removed the old wdio test here. I also added the required prop to the popup picker and updated the Jest tests. https://github.com/cerner/terra-framework/pull/504/commits/dd0b5f4b71531b98f556c1059f2b5ca0e3e4b547

packages/terra-date-picker/src/DatePicker.jsx
gabeparra01: Your edit comment is correct, assignment was removed in this commit: 14324e7

PR #515, opened 2019-02-13
fix popup content opts export

### LINKED ISSUE #513: PopupContent Opts Object Isn't Visible Because The PopupContent Component is Wrapped in `injectIntl()`
# Bug Report

## Description

Terra PopupContent Opts object isn't accessible since the component is [wrapped in injectIntl](https://github.com/cerner/terra-framework/blob/master/packages/terra-popup/src/_PopupContent.jsx#L253-L257) so any mention of `PopupContent.Opts.*` will error out:
* [example 1](https://github.com/cerner/terra-framework/blob/master/packages/terra-popup/src/Popup.jsx#L141)
* [example 2](https://github.com/cerner/terra-framework/blob/master/packages/terra-popup/src/Popup.jsx#L149)
* [example 3](https://github.com/cerner/terra-framework/blob/master/packages/terra-popup/src/Popup.jsx#L153)

## Steps to Reproduce

1. enzyme mount (with intl) a component that consumes `terra-popup`, e.g. `terra-menu`
2. watch as jest complains about accessing `PopupContent.Opts`

Failed jest test:
image

Compiled Popup line Jest test complains about:
image

## Expected Behavior

Accessing any `PopupContent.Ops.*` attributes doesn't error out.

## Possible Solution

Attach the `Opts` object to the exported component (the one wrapped in the `injectIntl` HOC):

```jsx
// terra-framework/packages/terra-popup/src/_PopupContent.jsx

class PopupContent extends React.Component {
  ...
}

PopupContent.propTypes = propTypes;
PopupContent.defaultProps = defaultProps;

const PopupContentWithIntl = injectIntl(PopupContent);
PopupContentWithIntl.Opts = {
  cornerSize: CORNER_SIZE,
};

export default PopupContentWithIntl;
```

And/Or:

```jsx
export const PopupContentOpts = {
  cornerSize: CORNER_SIZE,
};
```

## Environment

* Component Name and Version: Menu v4.5.0, Popup v5.1.0
* Node/npm Version: lts/carbon (8.15.0)

dev-cprice:
### Summary
* fixes #513


packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -250,8 +250,11 @@ class PopupContent extends React.Component {
...
-PopupContent.Opts = {
+
+const PopupContentWithIntl = injectIntl(PopupContent);
+
+PopupContentWithIntl.Opts = {
   cornerSize: CORNER_SIZE,
 };
 
-export default injectIntl(PopupContent);
+export default PopupContentWithIntl;
```
dkasper-was-taken: This might be better solved exporting the constant separately.
```
export default injectIntl(PopupContent);
export {
  CORNER_SIZE as cornerSize,
};
```

dev-cprice: I'm okay with either approach. I mentioned something similar to what you describe in #513

image

The current method just involves less moving parts and is passive.

bjankord: Let's stick with the passive solution.

dkasper-was-taken: It would be passive in either case. Corner size is only exposed internally for the use of Popup.jsx itself here:
https://github.com/cerner/terra-framework/blob/d0360e660cb4ce8e99c73a2ba366971825015689/packages/terra-popup/src/Popup.jsx#L141

I lean toward exposing the value by itself, potentially updating arrow to match the same pattern.

bjankord: Sounds good to me, let's move forward with that @dev-cprice

packages/terra-popup/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+* Add Opts object onto Intl wrapped component so it's available when accessed by `PopupContent.Opts`
```
bjankord: Let's get the changelog updated to match the change.

dev-cprice:
So idk why I didn't actually look into how the `intl` object was being used, but this can very easily be enhanced so that `_PopupContent` ***doesn't*** need to be wrapped in `injectIntl`.

We just need to change the "pseudo-private" (as in only used in the `_PopupContent` component) method [addPopupHeader](https://github.com/cerner/terra-framework/blob/master/packages/terra-popup/src/_PopupContent.jsx#L103-L108) to:

```jsx
static addPopupHeader(children, onRequestClose) {
  const icon = <span className={cx('close-icon')} />;
  const header = (
    <div className={cx('header')}>
      <FormattedMessage id="Terra.popup.header.close">
        {text => <Button variant="utility" isIconOnly icon={icon} onClick={onRequestClose} text={text} />}
      </FormattedMessage>
    </div>
  );
  return <ContentContainer header={header} fill>{children}</ContentContainer>;
}
```

I can leave the named export of the `cornerSize` as is, or revert it. With the above change, we don't need to wrap the `_PopupContent` in `injectIntl`, thus we can access any exports attached to the default export directly (like `Opts`, as from before).

bjankord:
@dev-cprice Sounds good to me, let's leave the named export of the `cornerSize` and update to use the render prop version of FormattedMessage.

package.json
```diff
@@ -83,6 +83,7 @@
     "danger": "^3.7.14",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
+    "enzyme-intl": "^3.0.0",
```
dev-cprice: this will be replaced by terra-toolkit's enzyme helpers when they're available

packages/terra-popup/tests/jest/PopupContent.test.jsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import { shallowWithIntl, mountWithIntl } from 'enzyme-intl'; // eslint-disable-line import/no-extraneous-dependencies
```
dev-cprice: this will be replaced by terra-toolkit's enzyme helpers when they're available

package.json
dev-cprice: https://github.com/cerner/terra-toolkit/pull/255

packages/terra-popup/tests/jest/PopupContent.test.jsx
dev-cprice: https://github.com/cerner/terra-toolkit/pull/255

package.json
dev-cprice: https://github.com/cerner/terra-toolkit/pull/255#issuecomment-466157733

packages/terra-popup/tests/jest/PopupContent.test.jsx
dev-cprice: https://github.com/cerner/terra-toolkit/pull/255#issuecomment-466157733

dev-cprice:
@emilyrohrbough that's intentional because the owner/releaser of [enzyme-react-intl](https://www.npmjs.com/package/enzyme-react-intl) still has [not released](https://github.com/joetidee/enzyme-react-intl/issues/22) the changes which allow us to write tests that do not depend on translations to npm registry.

image

This is a fix that is blocking consumers from using `terra-menu`, which uses `terra-popup`.

The assumption when recommending `enzyme-react-intl` was that the owner knew how releases to the npm registry worked and that just because you tag something as a release on github does not mean that it's automatically released to artifactory. 

It still has yet to be released, see https://github.com/joetidee/enzyme-react-intl/issues/22#issuecomment-452802234

dev-cprice:
@bjankord @emilyrohrbough https://github.com/cerner/terra-framework/pull/515/commits/d2325ac39704a4fbed80511080780b123e1a223c uses the intl-context-setup and context types shape instead of `enzyme-intl`

PR #540, opened 2019-03-07
Replaced local KeyCode objects with keycode-js npm package

AlexBrizi:
### Summary
We have been using case-by-case keycode objects for use of referencing certain keys such as:

`const KEYCODES = {
  TAB: 9,
  ENTER: 13,
  SPACE: 32,
  UP_ARROW: 38,
  DOWN_ARROW: 40,
  LEFT_ARROW: 37,
  RIGHT_ARROW: 39,
};
`

Now we will import the [keycode-js](https://www.npmjs.com/package/keycode-js) npm package to reduce duplication and redundancy between packages.

Resolves [#2235](https://github.com/cerner/terra-core/issues/2235)


packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -239,7 +236,7 @@ class DateTimePicker extends React.Component {
     if (validDate && validTime) {
       const updatedDateTime = DateTimeUtils.updateTime(previousDateTime, time);
 
-      if (event.keyCode === keyCodes.ARROWDOWN
+      if (event.keyCode === KeyCode.KEY_DOWN
```
emilyrohrbough: is this the correct key down code?

package.json
```diff
@@ -105,5 +105,8 @@
     "terra-dev-site": "^3.0.0",
     "terra-toolkit": "^4.16.0",
     "xfc": "^1.2.1"
+  },
+  "dependencies": {
+    "keycode-js": "^1.0.4"
```
emilyrohrbough: This shouldn't be included at the root level

AlexBrizi: You are correct :) Forgot to remove it the 2nd time. Ran into an issue where npm installing keycode-js was deleting the terra-list package. Made no sense, but forgot to remove it the 2nd time around.

Issue #507, opened 2019-02-11
Update heading order in examples / docs


bjankord:
# Bug Report

## Description
In a few of our examples and docs, heading order is out of sync.

## Steps to Reproduce

1. Navigate to https://engineering.cerner.com/terra-ui/#/components/terra-dialog-modal/dialog-modal/dialog-modal
2. Run [HTML CodeSniffer bookmarklet](https://squizlabs.github.io/HTML_CodeSniffer/) and view Section 508 errors.
3. Navigate to https://engineering.cerner.com/terra-ui/#/components/terra-disclosure-manager/disclosure-manager/upgrade-guide
4. Run [HTML CodeSniffer bookmarklet](https://squizlabs.github.io/HTML_CodeSniffer/) and view Section 508 errors.
5. Navigate to https://engineering.cerner.com/terra-ui/#/components/terra-popup/popup/popup
6. Run [HTML CodeSniffer bookmarklet](https://squizlabs.github.io/HTML_CodeSniffer/) and view Section 508 errors.
## Expected Behavior
Heading order matches Section 508 guidelines

## Possible Solution
Update heading order in docs and examples

## Environment

* Component Name and Version: terra-dialog-modal, terra-disclosure-manager, terra-popup
* Browser Name and Version: All

PR #541, opened 2019-03-08
Doc Site Updates to Correct Section 508 Errors and Warnings

amichaelparker:
### Summary
Updates several component doc site examples to correct Section 508 errors and warnings.

Closes [#507](https://github.com/cerner/terra-framework/issues/507)


packages/terra-brand-footer/src/terra-dev-site/doc/example/DefaultBrandFooter.jsx
```diff
@@ -32,6 +33,7 @@ export default () => (
           textDecoration: 'none',
         }}
       >
+        <VisuallyHiddenText text="Terra UI" />
```
emilyrohrbough: What are Section 508 errors and warnings?

Do we want to include another dependency here? Is this a change we should make to the brand-footer instead of the examples?

amichaelparker: Essentially [Section 508](https://www.access-board.gov/guidelines-and-standards/communications-and-it/about-the-section-508-standards/guide-to-the-section-508-standards/web-based-intranet-and-internet-information-and-applications-1194-22) is another form of accessibility guidelines.

The dependency on VisuallyHiddenText is something @bjankord suggested as a solution to the TerraUI link in the footer not actually having any content in it; it’s just svg art, so a screen reader will think that link is empty.

The only other option I could think of was to just wipe out the SVG art, and go with standard, styled text.

As far as I know, the only place that art exists is in those examples, so I threw one option in here and put it up for some input. I’m obviously fine either way, but it looked a little complex for me to just delete outright so, for the moment, I opted for the compromise option.

emilyrohrbough: hmmm what are people's thoughts of dropping the 'prettiness' of the example? I feel like its visually-hidden-text is light weight enough it isn't a big deal but I think it would be nice to keep as many dependencies as possible out of our packages when possible

bjankord: That sounds good to me. Lets replace the Terra UI SVG and drop the visually-hidden-text component and just pass the following to contentLeft prop:

```
contentLeft={(
      <a
        href="http://terra-ui.com/"
        style={{
          backgroundColor: 'transparent',
          color: '#0065a3',
          textDecoration: 'none',
        }}
      >
        Terra UI
      </a>
)}
```

amichaelparker: Updated: [76bf079](https://github.com/cerner/terra-framework/pull/541/commits/76bf0796cea810cd57b6d92e44f3fd6f185fb36e). Removing the svg also got rid of the logo; I removed the portion that was just text, and replaced it with styled text. It corrects the Section 508 warning, and leaves the examples looking essentially the same.

PR #543, opened 2019-03-08
Terra Notification Dialog- Copied Translation from nl to nl-BE 

### LINKED ISSUE #537: Missing translation
# Bug Report

## Description

The new translation script is live and is flagging the following as missing:

nl-BE translation missing for Terra.notification.dialog.success, nl translation string will be used instead.

gabeparra01:
### Summary
The success translation is missing for nl-BE.json and can be added using the success string in nl.json. Closes #537


packages/terra-notification-dialog/translations/nl-BE.json
```diff
@@ -1,7 +1,7 @@
 {
-  "Terra.notification.dialog.advisory": "Adviserend",
```
gabeparra01: This string was added when the component was created: https://github.com/cerner/terra-framework/pull/162
I couldn't find any information about why it was added. It is not being used by the component so I removed it

PR #555, opened 2019-03-21
[terra-date-time-picker] Tweaking workflow logic for onChange when given ambiguous date-time.

BenBoersma:
### Summary
Changed workflow logic when internal functions call prop `onChange` function. Addresses https://github.com/cerner/terra-framework/issues/521

Now if given an ambiguous date-time when the last date-time was not, we wait till after the TimeClarification dialog resolves to call `onChange`. Also added defensive code to prevent unnecassary `onChange` calls when resolving the TimeClarification dialog that was launched via the Offset button and date-time was not changed.

@cerner/terra


packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -335,11 +344,20 @@ class DateTimePicker extends React.Component {
...
     if (newDateTime.isDST()) {
       newDateTime.add(1, 'hour');
-      this.handleChange(event, newDateTime);
+      this.setState({
+        dateTime: newDateTime,
+      });
+      if (this.props.onChange) {
+        this.props.onChange(event, newDateTime && newDateTime.isValid() ? newDateTime.format() : '');
+      }
+    } else if (this.props.onChange && !this.wasOffsetButtonClicked) {
```
benbcai: It might be helpful to add a comment to explain the workflow and why we're making the check on wasOffsetButtonClicked. I know I will forget several months later.

BenBoersma: Added in https://github.com/cerner/terra-framework/pull/555/commits/105f1cae0c24d88d4c2096cf5bd7c3fd28a755ec.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -262,7 +263,8 @@ class DateTimePicker extends React.Component {
       dateTime: newDateTime,
     });
 
-    if (this.props.onChange) {
+    // We want to not fire onChange in the event that the new time is ambiguous and the old time was not. This way the user can use TimeClarification before we fire the onChange function.
```
yuderekyu: nitpick - want to make this more concise?
```suggestion
    // If the new time is ambiguous and the old time is not, do not fire onChange. This allows a user to use TimeClarification before onChange is fired.
```

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -324,8 +326,19 @@ class DateTimePicker extends React.Component {
...
       newDateTime.subtract(1, 'hour');
-      this.handleChange(event, newDateTime);
+      this.setState({
+        dateTime: newDateTime,
+      });
+      if (this.props.onChange) {
+        this.props.onChange(event, newDateTime && newDateTime.isValid() ? newDateTime.format() : '');
+      }
+    } else if (this.props.onChange && !this.wasOffsetButtonClicked) {
+      // Since we now don't fire onChange in our handleChange function if the new value is ambiguous and the old value wasn't,
```
yuderekyu: same here
```suggestion
      // This fires onChange if the TimeClarification dialog was launched without using the OffsetButton.
      // If the user clicks the OffsetButton, onChange should have already been fired and does not need to be fired 
      // again (unless they change the DateTime).
```

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -334,11 +347,23 @@ class DateTimePicker extends React.Component {
...
       newDateTime.add(1, 'hour');
-      this.handleChange(event, newDateTime);
+      this.setState({
+        dateTime: newDateTime,
+      });
+      if (this.props.onChange) {
+        this.props.onChange(event, newDateTime && newDateTime.isValid() ? newDateTime.format() : '');
+      }
+    } else if (this.props.onChange && !this.wasOffsetButtonClicked) {
+      // Since we now don't fire onChange in our handleChange function if the new value is ambiguous and the old value wasn't,
```
yuderekyu: same suggestion as above

packages/terra-date-time-picker/src/DateTimePicker.jsx
BenBoersma: Fixed https://github.com/cerner/terra-framework/pull/555/commits/32443749bb26dca02597dd2efc4dc95cea1efcba

packages/terra-date-time-picker/src/DateTimePicker.jsx
BenBoersma: Fixed https://github.com/cerner/terra-framework/pull/555/commits/32443749bb26dca02597dd2efc4dc95cea1efcba

packages/terra-date-time-picker/src/DateTimePicker.jsx
BenBoersma: Fixed https://github.com/cerner/terra-framework/pull/555/commits/32443749bb26dca02597dd2efc4dc95cea1efcba

mmalaker:
Functional verification completed. The date time examples display and function as expected with ambiguous date-times. Found a minor, existing issue in IE that causes the modal to be invoked prematurely. That will be addressed in #522

Issue #563, opened 2019-03-22
Slide panel panelSize medium size


Chris-Boyle:
# Feature Request
I would really like to have a medium size for the slide panel component.  Currently there is a large and a small.  Neither of these look good for our content.
## Description
Create a medium panelSize in the Slide Panel component.
There are two options here:
1.  Make a medium size that is scalar .4
2. Make a medium size that is fixed at about 500 px.
## Additional Context / Screenshots
Changes in:
1. https://github.com/cerner/terra-framework/blob/master/packages/terra-slide-panel/src/SlidePanel.module.scss#L65
2. https://github.com/cerner/terra-framework/blob/master/packages/terra-slide-panel/src/SlidePanel.module.scss#L114
3. https://github.com/cerner/terra-framework/blob/master/packages/terra-slide-panel/src/SlidePanel.jsx#L32
Small
![image](https://user-images.githubusercontent.com/20329437/54853390-c9390d00-4cbd-11e9-9465-acd9e9d3df6c.png)
Large
![image](https://user-images.githubusercontent.com/20329437/54853426-e077fa80-4cbd-11e9-87ae-4e0ab01134d5.png)
We would like something similar to this screen shot, or 500 px
![image](https://user-images.githubusercontent.com/20329437/54853500-0a312180-4cbe-11e9-8aae-41fa2566729f.png)

Chris-Boyle:
I am working on a PR for this.

neilpfeiffer:
Thanks @Chris-Boyle - curious to know more about your content and workflow, and if you have been working with a designer to guide your layouts. UX has currently kept the slide-panel reserved to only two sizes, large and small, with defined template layouts and most likely will not want to add another size. Responsive details are also a consideration of how best to achieve a layout you are desiring, it might be there is another pattern or layout UX can provide you as good direction.

FWIW - hard to tell is this is exactly what you need, but from your screenshots, if you are looking for a split-view with a 40% left panel, 60% right panel (slide-panel set to squish) - the dimension can actually be achieved by switching your main-content and panel-content.

Chris-Boyle:
@neilpfeiffer we need to have the the left side of our slide panel always in view as the screen is resized.  This would work if the panelContent was set to the start position but the only two options for it being on the left side of the view is large or small.  These two sizes are either not big enough for the panelContent or obnoxiously large.  We worked with Scott on our design some but he couldn't be on board anymore due to time constraints.
Small
[...]

PR #565, opened 2019-03-23
Created a medium slide panel size that is scalar .4

Chris-Boyle:
### Summary
Created a medium size for the Slide Panel that is calculated by scalar .4.  This is needed so that the panelContent can be 40% of the view.  We can not achieve this by setting the panelContent to large and flipping the mainContent and panelContent because once the screen is resized the mainContent(40%) is overlayed by the panelContent or squished behind it.  We want the 40% side to always be visible when resizing.

Resolves 
https://github.com/cerner/terra-framework/issues/563
### Additional Details
The large size  uses scalar .6 so it felt natural to make the medium use .4.

image


neilpfeiffer:
Added comment on initial issue here before reviewing PR further: https://github.com/cerner/terra-framework/issues/563#issuecomment-476243564

JakeLaCombe:
This pull request is no longer applicable, as UX has left feedback that this size is not desirable.

PR #573, opened 2019-03-27
Fork react datepicker

### LINKED ISSUE #572: Fork React Datepicker
# Feature Request

## Description
We have specific needs for the datepicker that need to be addressed on a Cerner customization level in regards to styling, accessibility, and technical capabilities. To do this, we need to fork the react datepicker and use it inside our terra-date-picker.

JakeLaCombe:
### Summary

Resolves #572 

This forks react-datepicker and puts it inside our repo.

To ensure nothing breaks, we ignore the scss lint as it violates some of our rules such as selector count and class name syntax.


packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -161,6 +161,7 @@ class DateTimePicker extends React.Component {
   }
 
   handleOnDateBlur(event) {
+    debugger;
```
bjankord: We should remove the `debugger;` statement here

.eslintrc
```diff
@@ -0,0 +1,8 @@
+{
```
emilyrohrbough: This invalidates the [terra-eslint-config setup in our package.json](https://github.com/cerner/terra-framework/pull/573/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R27). Lets move this to the data-picker package and leverage this configuration for just that package.

More info can be found here: https://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy

.babelrc
```diff
@@ -1,4 +1,9 @@
 {
   "presets": ["env", "react"],
-  "plugins": ["transform-object-assign", "transform-object-rest-spread", "transform-async-to-generator"],
+  "plugins": [
+    "transform-object-assign",
+    "transform-object-rest-spread",
+    "transform-async-to-generator",
+    "transform-class-properties"
```
emilyrohrbough: Should we be scoping these changes to the date-picker project as well?

https://babeljs.io/docs/en/config-files#monorepos

packages/terra-date-picker/src/react-datepicker/stylesheets/datepicker.scss
```diff
@@ -0,0 +1,501 @@
+@import "variables.scss";
+@import "mixins.scss";
+
+.react-datepicker-wrapper {
+  display: inline-block;
+}
+
+.react-datepicker {
+  font-family: $datepicker__font-family;
```
emilyrohrbough: Do we need to log an issue to migrate this to using css modules?

packages/terra-date-picker/src/react-datepicker/day.jsx
```diff
@@ -0,0 +1,175 @@
...
+  now,
+  isSameDay,
+  isDayDisabled,
+  isDayInRange,
+  getDayOfWeekCode
+} from './date_utils'
+
+export default class Day extends React.Component {
+  static propTypes = {
+    day: PropTypes.object.isRequired,
```
emilyrohrbough: Is there existing documentation we can pull forward to use as inline comments for these props?

emilyrohrbough:
I know we are forking this and intended for it to be 1-to-1 but should we attempt to follow terra's file naming for using capitals for React component files and using `_` for private modules we do not intend for consumers to use directly?

packages/terra-date-picker/src/react-datepicker/day.jsx
JakeLaCombe: react-datepicker has a ton of them. https://github.com/Hacker0x01/react-datepicker/tree/v0.64.0/docs

I don't want to make them public in our repo to consumers, but I would have no problem adding a react-datepicker-docs directory for our use and reference.

JakeLaCombe:
Right now, we want to keep as close to their source code as possible so we can easily pull in any defects that they fix. As we progress and modify this datepicker, we can definitely look into restructuring and redoing our files to fall better in line with our standards.

.eslintrc
JakeLaCombe: Gotcha. I think what would benefit here is a .gitignore file for this package.

bjankord:
@emilyrohrbough I'd like to move the code over as-is into our repo. Then, in separate PRs, we can work on updating the code to align more with terra conventions.

packages/terra-date-picker/package.json
```diff
@@ -30,6 +30,8 @@
     "popper.js": "^1.14.3",
     "prop-types": "^15.5.8",
     "react-datepicker": "^0.64.0",
```
emilyrohrbough: Can we drop the react-datapicker dep now?

packages/terra-date-picker/package.json
```diff
@@ -29,7 +29,8 @@
     "moment": "^2.21.0",
     "popper.js": "^1.14.3",
     "prop-types": "^15.5.8",
-    "react-datepicker": "^0.64.0",
+    "react-onclickoutside": "^6.6.3",
+    "react-popper": "^0.7.4",
```
benbcai: Are we able to depend on 1.x.x of react-popper to at least match [react-datepicker's dependency version](https://github.com/Hacker0x01/react-datepicker/blob/v1.8.0/package.json#L96)? And are we able to get rid of the dependency on popper.js on line 30?

packages/terra-date-picker/src/react-datepicker/index.jsx
```diff
@@ -0,0 +1,560 @@
...
+    onFocus: PropTypes.func,
+    onKeyDown: PropTypes.func,
+    onMonthChange: PropTypes.func,
+    openToDate: PropTypes.object,
+    peekNextMonth: PropTypes.bool,
+    placeholderText: PropTypes.string,
+    popperContainer: PropTypes.func,
+    popperClassName: PropTypes.string, // <PopperComponent/> props
+    popperModifiers: PropTypes.object, // <PopperComponent/> props
+    popperPlacement: PropTypes.oneOf(popperPlacementPositions), // <PopperComponent/> props
```
benbcai: Which version of react-datepicker was this forked from? I was expecting there to be the `preventOpenOnFocus` prop as in [v1.8.0](https://github.com/Hacker0x01/react-datepicker/blob/v1.8.0/src/index.jsx#L127) that we could use in the future.

bjankord: Should be forked from v1.8.0, if it is not based on the v1.8.0 tag we should update before this is merged.

.eslintignore
```diff
@@ -3,3 +3,4 @@ node_modules
 coverage
 build
 **/aggregated-translations/*.*
+**/react-datepicker/*.*
```
emilyrohrbough: Can we make this a direct path ?
```
packages/terra-datapicker/src/react-data-picker/**
```

.stylelintignore
```diff
@@ -0,0 +1 @@
+**/src/react-datepicker/**
```
emilyrohrbough: Can we make this a direct path ?
```
packages/terra-datapicker/src/react-data-picker/**
```

emilyrohrbough: Also, this this is specific to thie date-picker pacakge, can we move this .stylelintingore to the package directory level? This would clean up this script as well
```diff
- "lint:scss": "stylelint src/**/*.scss --ignore-path ../../.stylelintignore",
+ "lint:scss": "stylelint src/**/*.scss --ignore-path .stylelintignore",
```

package.json
```diff
@@ -78,6 +78,7 @@
     "babel-plugin-transform-async-to-generator": "^6.24.1",
     "babel-plugin-transform-object-assign": "^6.22.0",
     "babel-plugin-transform-object-rest-spread": "^6.23.0",
+    "babel-plugin-transform-class-properties": "^6.24.1",
```
emilyrohrbough: Since this is a dev-dep of the data picker package, do we need it here?

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -1,7 +1,10 @@
+
 import React from 'react';
 import PropTypes from 'prop-types';
-import ReactDatePicker from 'react-datepicker';
 import ResponsiveElement from 'terra-responsive-element';
+
+/* eslint-disable-next-line  */
```
emilyrohrbough: nit: can we specify the exact rule we are disabling?

packages/terra-date-picker/src/react-datepicker/day.jsx
emilyrohrbough: I guess I was thinking the Inline docs so when code changes are made later one, we knew the purpose of each prop. I don't think this should be exposed to consumers but it would make code maintenance easier (IMO) later on. If we don't want to do it now, can we log an issue to do this later on?

packages/terra-date-time-picker/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Unreleased
 ------------------
 ### Changed
 * Update react/react-dom peer-dependency to latest stable minor release
+* Forked React Datepicker
```
emilyrohrbough: Can we add more context to this change log. We should include the added/removed date picker dependencies in the changelog.

package.json
JakeLaCombe: with out this here, the Travis CI breaks.

packages/terra-date-picker/src/DatePicker.jsx
JakeLaCombe: react-datepicker uses syntax that isn't currently supported by our linter, so it's throwing parse errors when it tries to run the linter on this file. As a result, it's always throwing lint errors, even when I specify the rules that need to be ignored.

packages/terra-date-picker/src/react-datepicker/index.jsx
bjankord: You can disregard my comment of using 1.8.0 as the basis for the fork. Talked with @benbcai  and he mentioned he had explored updating terra-date-picker to use react-datepicker@1.8.0 and the UI crashed when interacting with the datepicker when using that version. He is going to post some more details. 

Based on this, I'd recommend we fork the [current version of react-datepicker that we're using today](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/package.json#L32). From talking with @JakeLaCombe, it sounds like that is the version used in this PR as the basis for the fork. Once this is merged, we can start doing work to cherry pick additional code changes from react-datepicker that we want to incorporate.

packages/terra-date-time-picker/CHANGELOG.md
JakeLaCombe: https://github.com/cerner/terra-framework/pull/573/commits/1a0686975d2d7af2a3614d0d3492e447ae6e89e6

.stylelintignore
JakeLaCombe: That we can't do. We have two ways of doing the lint. We do it at the individual package level, and also at the terra-framework level. Even if we have this file here, we would still need to have the stylelineignore file at the root, so I think to reduce duplication, it's better to have it up top.

.eslintignore
JakeLaCombe: https://github.com/cerner/terra-framework/pull/573/commits/07793bd62b5b0afc00c050a9c5f83cba38745cbf

packages/terra-date-picker/src/react-datepicker/index.jsx
JakeLaCombe: I forked it based off of our current version in the repo of 0.64.0.

packages/terra-date-picker/package.json
JakeLaCombe: I had troubles with the blur event when I tried to uplift to react-popper 1.x.x. To alleviate those concerns, I used the version that react-datepicker was using.

packages/terra-date-picker/package.json
JakeLaCombe: https://github.com/cerner/terra-framework/pull/573/commits/c9128e3b0ebe8368a5a63e61c6ba99ff7a79484d

.babelrc
JakeLaCombe: https://github.com/cerner/terra-framework/pull/573/commits/b65c948b8bb448c5411e1e8ac76b65a387da50f5

packages/terra-date-picker/src/react-datepicker/stylesheets/datepicker.scss
JakeLaCombe: https://github.com/cerner/terra-framework/issues/575

packages/terra-date-picker/src/react-datepicker/day.jsx
JakeLaCombe: I looked through the docs, and they don't have any descriptions on them theirselves. https://github.com/Hacker0x01/react-datepicker/tree/master/docs.

I'd love to add more, but the docs don't give enough of a description to do so.

packages/terra-date-picker/src/react-datepicker/index.jsx
benbcai: If we were to fork V1.8.0, we would need to cherry pick the following PR:
- PR [#1548](https://github.com/Hacker0x01/react-datepicker/pull/1548) to fix issue  [#1547](https://github.com/Hacker0x01/react-datepicker/issues/1547). This was fixed in V2.1.0.

If we were to fork V0.64.0, we would need to cherry pick the following PRs:
- PR [#1181](https://github.com/Hacker0x01/react-datepicker/pull/1181) which introduces the `preventOpenOnFocus` prop that we would like to use.
- PR [#1374](https://github.com/Hacker0x01/react-datepicker/pull/1374) to remove unsafe react lifecycles.

V1.8.0 is obviously more updated than V0.64.0. But since we don't know what other broken changes (other than [#1547](https://github.com/Hacker0x01/react-datepicker/issues/1547))  we would run into if we were to use 1.8.0, I'm fine forking V0.64.0 and then subsequently pull in these additional pull requests that we need.

packages/terra-date-picker/src/react-datepicker/day.jsx
bjankord: Let's get an issue logged to add inline doc to the prop-types in a follow up PR.

packages/terra-date-time-picker/src/DateTimePicker.jsx
bjankord: https://github.com/cerner/terra-framework/pull/573/commits/e5022b9b87707ab87e848ee59e1b73eca9442c18

packages/terra-date-picker/src/react-datepicker/day.jsx
JakeLaCombe: https://github.com/cerner/terra-framework/issues/579

packages/terra-date-time-picker/CHANGELOG.md
benbcai: I don't know if the changelog is the best place but it would be helpful to call out the version of react-datepicker that we forked so that we can use it as the base for comparison or to lookup which fixes we have or don't have.

JakeLaCombe: https://github.com/cerner/terra-framework/pull/573/commits/671facc0b342a9226a59e6b6b302369411194eeb

packages/terra-date-picker/package.json
JakeLaCombe: Popper is removed though. https://github.com/cerner/terra-framework/pull/573/commits/671facc0b342a9226a59e6b6b302369411194eeb

PR #583, opened 2019-04-01
[terra-disclosure-manager] Introduced safety checks for nested disclosure dismissals

tbiethman:
### Summary
I recently found a consumer of the disclosure manager that was calling the dismiss implementation of a disclosed component that was deep within the disclosure stack. Calling that dismiss function would dismiss the top-most component instead (while bypassing it's dismissal prevention mechanisms 😱). While this is occurring due to a consumer implementation error, it has a really bad result.

I updated the dismissal logic to no-op when the component attempting dismissal is not on top of the stack. I also made sure that every dismissal goes through this updated branch of logic to ensure close and goBack are also handled in the same way.


packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -444,7 +461,7 @@ class DisclosureManager extends React.Component {
     }
 
     return render({
-      dismissPresentedComponent: (disclosureComponentKeys.length > 1) ? this.generatePopFunction(disclosureComponentKeys[disclosureComponentKeys.length - 1]) : this.safelyCloseDisclosure,
+      dismissPresentedComponent: (disclosureComponentKeys.length > 0) ? this.generatePopFunction(disclosureComponentKeys[disclosureComponentKeys.length - 1]) : () => {},
```
tbiethman: This is a little weird, but I'm forcing this function to go through the pop logic rather than deferring to `safelyCloseDisclosure` when we're popping the last item. I'm providing a no-op function in the alternate flow for passivity, but I'm fairly certain this API isn't being used by anyone so 🤷‍♂️

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -177,7 +177,7 @@ class DisclosureManager extends React.Component {
      * Allows a component to remove itself from the disclosure stack. If the component is the only element in the disclosure stack,
      * the disclosure is closed.
      */
-    delegate.dismiss = componentIndex > 0 ? popContent : this.safelyCloseDisclosure;
+    delegate.dismiss = popContent;
```
emilyrohrbough: Does this change need to be reflected in the goBack assignment as well?

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -407,9 +407,19 @@ class DisclosureManager extends React.Component {
    */
   generatePopFunction(key) {
     return () => {
-      let promiseRoot = Promise.resolve();
+      const { disclosureComponentKeys } = this.state;
+
+      if (disclosureComponentKeys[disclosureComponentKeys.length - 1] !== key) {
```
emilyrohrbough: With this change, can we remove the `disclosureComponentKeys.length` check throughout this file?

packages/terra-disclosure-manager/src/DisclosureManager.jsx
emilyrohrbough: This is a little odd. We can always assume we can use 
```
this.generatePopFunction(disclosureComponentKeys[disclosureComponentKeys.length - 1]
```
here with the checks for the key values to match?

Either way we should add an inline comment to describe why this is being done

packages/terra-disclosure-manager/src/DisclosureManager.jsx
tbiethman: goBack is already using popContent (or nothing when we don't want to provide it)

packages/terra-disclosure-manager/src/DisclosureManager.jsx
tbiethman: If we use generatePopFunction all time, we'll end up calling it with `undefined` when there are no disclosure keys present. It might be worthwhile to just throw an `undefined` check along with the key equality and use it all the time.

packages/terra-disclosure-manager/src/DisclosureManager.jsx
tbiethman: I might be missing something, but I don't think so

emilyrohrbough: I'm not sure what I was thinking here... 🤷‍♀️

packages/terra-disclosure-manager/src/DisclosureManager.jsx
jmsv6d: It seems cleaner to me to just use generatePopFunction all the time with the `undefined` check.

emilyrohrbough: I agree with @jmsv6d on just moving this check to the `generatePopFunction`. Then we could (not sure if needed) update the comment for `generatePopFunction`

tbiethman: Updatd: b66b17e

packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -407,9 +407,19 @@ class DisclosureManager extends React.Component {
    */
   generatePopFunction(key) {
     return () => {
-      let promiseRoot = Promise.resolve();
+      const { disclosureComponentKeys } = this.state;
+
+      if (!key || disclosureComponentKeys[disclosureComponentKeys.length - 1] !== key) {
+        /**
+         * If the top component key in the disclosure stack does not match
```
bjankord: +1 on adding comments here

PR #584, opened 2019-04-01
Added this.panelNode.Focus() to set focus for Slide panel on Open

### LINKED ISSUE #321: [terra-slide-panel-manager] The SlidePanelManager should set focus to the panel when opened
# Feature Request

## Description
To better align ModalManager and SlidePanelManager behavior, we will set focus to the SlidePanel's panel container when the SlidePanelManager is triggered open.

Changes will also be required in the SlidePanel to expose a ref to the panel container.

See issue: https://github.com/cerner/terra-core/issues/1973

supreethmr:
### Summary
Setting focus to slide panel's panel content on Open and on Close returning focus back to slide panel's main content.

Resolves #321


packages/terra-slide-panel/src/SlidePanel.jsx
```diff
@@ -134,10 +141,10 @@ class SlidePanel extends React.Component {
         data-slide-panel-panel-position={panelPosition}
         data-slide-panel-panel-size={panelSize}
       >
-        <div className={cx('main')}>
+        <div className={cx('main')} tabIndex="-1" ref={this.setMainNode}>
```
ShettyAkarsh: We can set the MainNode value here itself inline. instead of having a separate event handler.

nramamurth:
+1 but why is there a discrepancy here? 

**Reference:** https://github.com/cerner/terra-framework/blob/master/packages/terra-aggregator/tests/wdio/__snapshots__/reference/en/chrome_large/aggregator-spec/With_Disclosure_-_Close_from_focus_release%5Bdefault%5D.png

**Latest:** https://github.com/cerner/terra-framework/blob/b8687bf1e03a48b5ed33e44fc04062af2129e7a3/packages/terra-aggregator/tests/wdio/__snapshots__/reference/en/chrome_large/aggregator-spec/With_Disclosure_-_Close_from_focus_release%5Bdefault%5D.png

packages/terra-slide-panel/src/SlidePanel.jsx
supreethmr: Updated [reference for main node](https://github.com/cerner/terra-framework/pull/584/commits/eebe761af198b2489aed3529bea7870d57599092#diff-810d8b37f7659eab5a75af63e03e20e8R139)

supreethmr:
> +1 but why is there a discrepancy here?
> 
> **Reference:** https://github.com/cerner/terra-framework/blob/master/packages/terra-aggregator/tests/wdio/__snapshots__/reference/en/chrome_large/aggregator-spec/With_Disclosure_-_Close_from_focus_release%5Bdefault%5D.png
> 
> **Latest:** https://github.com/cerner/terra-framework/blob/b8687bf1e03a48b5ed33e44fc04062af2129e7a3/packages/terra-aggregator/tests/wdio/__snapshots__/reference/en/chrome_large/aggregator-spec/With_Disclosure_-_Close_from_focus_release%5Bdefault%5D.png

the Difference is due to setting focus back to main node on close of side panel. since here we are setting focus back to main node each time on close of side panel the focus is not on section1. 

I assume that having focus on section1 even after slide panel close might give impression to user that slide panel is still open. So I feel this is expected behavior of slide panel with this fix.

packages/terra-slide-panel/src/SlidePanel.jsx
tbiethman: You could also switch over to the createRef API that was introduced a few versions back: https://reactjs.org/docs/refs-and-the-dom.html#creating-refs

I'm finding myself using that pretty much exclusively at this point.

supreethmr: updated to [CreateRef](https://github.com/cerner/terra-framework/pull/584/commits/507edc3acdddf9a8b03db274d1f1e9028f95387d)

PR #585, opened 2019-04-01
[terra-date-time-picker] Expose onSelect

BenBoersma:
### Summary
Adding onSelect prop that is called when a date is picked via date picker. Addresses https://github.com/cerner/terra-framework/issues/523.

@cerner/terra


packages/terra-date-time-picker/src/terra-dev-site/test/date-time-picker/DateTimePickerOnSelect.test.jsx
```diff
@@ -0,0 +1,42 @@
...
+  handleDateSelect(event, onselectdate) {
+    this.setState({ onselectdate });
+  }
+
+  render() {
+    return (
+      <div>
+        <h3>
+OnChange Date:
+          <span id="selected-date">{this.state.onchangedate}</span>
```
benbcai: nit: Perhaps, choose an id that is different from the one below. (i.e. "changed-date"). But it doesn't affect anything here.

BenBoersma: Fixed in https://github.com/cerner/terra-framework/pull/585/commits/5e89405a15d090636a60395b922c116a068c2f58

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -212,6 +212,27 @@ describe('DateTimePicker', () => {
     Terra.should.matchScreenshot('1', { viewports });
   });
 
+  describe('OnSelect', () => {
+    before(() => {
+      browser.url('/#/raw/tests/terra-date-time-picker/date-time-picker/date-time-picker-on-select');
+      browser.click('[class*="button"]');
+    });
+
+    Terra.should.matchScreenshot('0', { viewports });
```
emilyrohrbough: The "tiny" OnSelect[0].png screenshot does not match the large screenshot. Should they?

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -63,6 +63,11 @@ const propTypes = {
    * The first parameter is the event. The second parameter is the changed input value.
    */
   onChangeRaw: PropTypes.func,
+  /**
+   * A callback function to execute when a selection is made in the date picker.
+   * The first parameter is the event. The second parameter is the selected input value.
```
emilyrohrbough: nit: should this be update to indicate the returned input value is formatted or would consumers be expecting the formatted date?

Below:
```
if (this.props.onSelect) {
   this.props.onSelect(event, updatedDateTime.format());
}
```

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
BenBoersma: A similar thing happens between some of the other screenshots so I assumed that this was expected behavior. Probably gonna poke around to double check though.
Excluded_Dates_are_Disabled[default].png for instance
Large https://github.com/cerner/terra-framework/blob/master/packages/terra-date-time-picker/tests/wdio/__snapshots__/reference/en/chrome_large/date-time-picker-spec/Excluded_Dates_are_Disabled%5Bdefault%5D.png
Tiny https://github.com/cerner/terra-framework/blob/master/packages/terra-date-time-picker/tests/wdio/__snapshots__/reference/en/chrome_tiny/date-time-picker-spec/Excluded_Dates_are_Disabled%5Bdefault%5D.png

emilyrohrbough: Hmm actually, I think this test should be using beforeEach() instead of before

BenBoersma: Yep that was it according to @benbcai, I can just remove the {viewports} argument and we should be okay.

packages/terra-date-time-picker/src/DateTimePicker.jsx
benbcai: I agree to update the comment that the returned value is in the ISO format. The props takes an ISO format so I think it makes sense that onSelect also returned the value in ISO format similar to the onSelect event in DatePicker and the onChange event.)

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
BenBoersma: Fixed here https://github.com/cerner/terra-framework/pull/585/commits/bf7770d8eb04e4701794086433d6b1f6f1609afe
Specifically I reworked the tests. Since the purpose of the tests was to show that we start with no "onSelect" or "onChange" values and how we set these values, I have changed the tests to not show the calendar at all in `tiny` view so that it matches `large` view.

packages/terra-date-time-picker/src/DateTimePicker.jsx
BenBoersma: Updated here https://github.com/cerner/terra-framework/pull/585/commits/bf7770d8eb04e4701794086433d6b1f6f1609afe

mmalaker:
Functional verification completed. The date time picker examples display and function as expected using the OnSelect prop.

Issue #498, opened 2019-01-09
[terra-date-picker] expose onFocus method


SharathAlur:
# Feature Request
Expose onFocus method

## Description
onFocus method is required to focus on the control programatically.

Since onFocus is not there, we can not use the auto-focus on error functionality in react-final-form.

benbcai:
@SharathAlur We delete the onFocus prop [here](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/src/DateInput.jsx#L144-L146) because we do not want the calendar picker to automatically display when the input receives focus. You can pass in a handler to the `onInputFocus` prop in DatePicker to see if that works for you. We should make `onInputFocus` a first class prop so that consumers are aware of it. I'll leave this issue open to make this change.

JakeLaCombe:
What goal are you looking to achieve with this issue? Perhaps there is something else we can do to resolve it?

PR #588, opened 2019-04-03
[terra-date-time-picker] Consolidate the onBlur and onFocus events

### LINKED ISSUE #522: [terra-date-time-picker] Create/expose consolidated 'onBlur' for date-time-picker
# Feature Request

## Description
A way for consumers to know when a user shifts focus away from any of the date-time-picker child components.

We have a use case that requires updating a form as values are entered individually.  For most components, this means listening for either onChange or onBlur to signify that the user is done setting a new value.

For the more complex date time component neither of these fit our use case.  Both of these approaches fire when setting the date, the hour, and the minute separately.

Additionally, we'd like to not force a user to always have to input an hour or a minute, so we don't want to 'delay' the call by waiting for all values to be entered.

Ideally, a consolidated 'onBlur'-type event that only fired when leaving a child component and not focusing on a different child component would allow us to infer that the user considered their value complete and then firing off the update call provided the value was valid.

## @ Mentions
@benbcai

benbcai:
### Summary
Consolidate the onBlur and onFocus events for DatePicker, DateTimePicker and TimeInput.

### Additional Details
- Fixes #522; #498 
- DatePicker, DateTimePicker and TimeInput used to trigger onBlur every time when losing focus from any of its child elements (date input, calendar button, hour input, minute input, meridiem select, meridiem button group, tz offset button). It now will trigger onBlur only when losing focus from the container. That is, if focus is moved from the date input to the hour input then to the minute input, onBlur will not fire.
- Pulled in the `preventOpenOnFocus` [change](https://github.com/Hacker0x01/react-datepicker/pull/1181/files) from react-datepicker so we can remove the custom `onInputFocus` prop. We previously had to delete the onFocus prop to prevent the picker from showing whenever the date input gains focus.
- Add onFocus as a first-class prop. onFocus is also consolidated similar to onBlur.
- When a valid date is entered/selected, the focus will automatically be moved to the hour input.

@cerner/terra


packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -53,6 +53,11 @@ const propTypes = {
    * Name of the date input. The name should be unique.
    */
   name: PropTypes.string.isRequired,
+  /**
+   * A callback function triggered when the entire date time picker component loses focus.
+   * This event does not get triggered when the focus is moved from the date input to the time input because the focus is still within the main date time picker component.
+   */
+  onBlur: PropTypes.func,
```
jmsv6d: Should we allow consumers to pass in individual onDateBlur and onTimeBlur callbacks as well?

benbcai: Do you know of a use case where consumers would need the individual onDateBlur and onTimeBlur? I'm not against providing these more granular onBlurs. Perhaps, add them when the need arises.

bjankord: My preference would be to add as a use-case arises.

packages/terra-date-picker/docs/DatePickerEvents.md
```diff
@@ -0,0 +1,27 @@
+# Terra DatePicker - Events
```
yuderekyu: nit: space
```suggestion
# Terra Date Picker - Events
```

packages/terra-date-picker/docs/DatePickerEvents.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Events
```
yuderekyu: nit: to be concise, either remove this header or remove "Events" on line 1

packages/terra-date-picker/docs/DatePickerEvents.md
```diff
@@ -0,0 +1,27 @@
...
+
+### onFocus
+* The `onFocus` callback prop is triggered when the component initially gains focus. If the focus is moved from the date input to the calendar button, the `onFocus` event will not be triggered since the focus is still within the date picker component. Like the `onBlur` event, the `onFocus` event does not get triggered when the focus is moved within child node of the Date Picker component to another child node of the same Date Picker component.
+
+### onSelect
+* The `onSelect` callback prop is triggered when a date is selected from the date picker. Once a date is selected, the `onChange` event will be triggered first followed by the `onSelect` event.
+
+## Event Trigger Order
+When these actions are taken, the events are triggered in the following order.
+* Manual date entry - onChangeRaw followed by onChange (if the date entry is a valid date).
```
yuderekyu: nit: want to use a numbered list?
### Manual date entry
1. `onChangeRaw`
2. `onChange`

packages/terra-date-picker/src/terra-dev-site/doc/guides/DatePickerEvents.jsx
```diff
@@ -0,0 +1,94 @@
...
+  handleOnClickOutside() {
+    this.clickOutsideCount = this.clickOutsideCount + 1;
+    this.setState({ clickOutsideTriggerCount: this.clickOutsideCount });
+  }
+
+  render() {
+    return (
+      <div>
+        <p>
+          onBlur Trigger Count:
```
yuderekyu: nit: for readability, let's add a space after each p string
```suggestion
          onBlur Trigger Count: 
```

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -63,6 +68,10 @@ const propTypes = {
    * The first parameter is the event. The second parameter is the changed input value.
    */
   onChangeRaw: PropTypes.func,
+  /**
+   * A callback function triggered when the date input, hour input, or minute input receives focus. function(event)
```
yuderekyu: should we remove function(event)
```suggestion
   * A callback function triggered when the date input, hour input, or minute input receives focus.
```

packages/terra-date-picker/docs/DatePickerEvents.md
benbcai: Updated: https://github.com/cerner/terra-framework/pull/588/commits/19718a8d69dc2de79da2896ad1657dd178cd6b8d

packages/terra-date-picker/docs/DatePickerEvents.md
benbcai: Updated: https://github.com/cerner/terra-framework/pull/588/commits/608599424d554dff2a4908132605c72032d51661

packages/terra-date-picker/docs/DatePickerEvents.md
benbcai: Updated: https://github.com/cerner/terra-framework/pull/588/commits/608599424d554dff2a4908132605c72032d51661

packages/terra-date-picker/src/terra-dev-site/doc/guides/DatePickerEvents.jsx
benbcai: Updated: https://github.com/cerner/terra-framework/pull/588/commits/608599424d554dff2a4908132605c72032d51661

packages/terra-date-time-picker/src/DateTimePicker.jsx
benbcai: Updated: https://github.com/cerner/terra-framework/pull/588/commits/608599424d554dff2a4908132605c72032d51661

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -192,16 +220,21 @@ class DatePicker extends React.Component {
...
 
     if (!this.isDefaultDateAcceptable) {
       this.handleChange(null, event);
       this.isDefaultDateAcceptable = true;
     }
   }
 
+  handleFocus(event) {
+    // Handle focus only if focus is gained from outside of the entire date picker component.
+    if (this.props.onFocus && !this.datePickerContainer.current.contains(event.relatedTarget)) {
```
tbiethman: The comment above mentions IE10/11 issues with `relatedTarget`. Should we be doing a the same presence here as above?

benbcai: Ah, yes. I need to do something similar for IE as above but just not the same as above. I can't use activeElement since it references the current active element. I need to get the _previous_ active element. I haven't found a native way to do so I may have to keep track of it internal.

benbcai: Added change to make it work in IE. https://github.com/cerner/terra-framework/pull/588/commits/dc4b6b92c993c0c6ea2aeec03e1844783473ba99 and https://github.com/cerner/terra-framework/pull/588/commits/4253be45c6ac86ee205bff37876d6ffda377e5fd

packages/terra-date-time-picker/docs/DateTimePickerEvents.md
```diff
@@ -0,0 +1,31 @@
+# Terra Date Time Picker
```
bjankord: Nice work on this doc, this will be really helpful for our consumers.

benbcai:
Functional Verification steps:

1. Go to the example for the bottom of https://terra-framework-deploye-pr-588.herokuapp.com/#/components/terra-date-time-picker/date-time-picker/events.
2. There are counters that tracks whenever the date-time-picker gains focus or loses focus.
3. When one of the date-time-picker elements (date input, calendar button, hour input, minute input) _initially_ gains focus, the onFocus counter will increment by 1. Moving focus from any of these elements would not trigger focus or blur. The onBlur counter will increment by 1 only when the focus is moved anywhere outside of the entire date-time-picker component.
4. Note: The other counters and event triggers on the example can be ignored as they are not related to the onBlur and onFocus changes in this PR. They are just additional information to display when events occur. But it doesn't hurt to also ensure that the information is correct when these events take place.

mmalaker:
Functional verification completed. The date time example displays and functions as expected with the consolidated onBlur and onFocus events.

PR #590, opened 2019-04-03
[terra-date-picker] Change filterDate to return an ISO string

### LINKED ISSUE #589: [terra-date-picker] filterDate return a moment object
# Bug Report

## Description
The data type of the parameter in the `filterDate` prop callback is a moment object. It should be a date represented in the ISO format so that it is consistent with the parameter in other callbacks (`onChange`;`onChangeRaw`; `onSelect`)

See [example code](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/src/terra-dev-site/doc/example/DatePickerFilterDates.jsx#L52-L55).

benbcai:
### Summary
Change the data type of the parameter in the `filterDate` callback from a moment object to a date string in the ISO format so it is consistent with other callbacks.

### Additional Details
Fixes #589

@cerner/terra


benbcai:
Sounds good. I'll add that as a topic.

Issue #536, opened 2019-03-04
Limit the characters that can be typed in the date/time picker


ryanthemanuel:
# Feature Request

## Description

Treat the date picker like the time picker and only allow certain values to be entered to prevent nonsense characters from being entered (like alpha characters).

bjankord:
We should test and see if this is available today with masking features of react-final-form.

PR #595, opened 2019-04-05
[terra-date-picker] Character limit on Date input

BenBoersma:
### Summary
Limit the characters that can be put in the Date input field to only `.`, `/`, and numeric.
Addresses https://github.com/cerner/terra-framework/issues/536 

@cerner/terra


packages/terra-date-picker/src/DateUtil.js
```diff
@@ -83,6 +83,17 @@ class DateUtil {
 
     return undefined;
   }
+
+  /**
+   * Determines if a provided date input value is valid.
+   * Valid inputs are either empty strings or contain only numeric and the `/` character.
```
benbcai: Should also call out the `.`

packages/terra-date-picker/tests/wdio/date-picker-spec.js
```diff
@@ -328,4 +328,18 @@ describe('Date Picker', () => {
...
+
+  describe('Key Limitations', () => {
+    before(() => browser.url('/#/raw/tests/terra-date-picker/date-picker/date-picker-key-limits'));
+
+    it('sets the date', () => {
+      browser.setValue('input[name="terra-date-date-input"]', '0');
+      browser.keys('a1.b2/;3');
+      // Ensures the mouse pointer doesn't appear in the screenshot
+      browser.click('h3');
+    });
```
benbcai: Since the DatePickerKeyLimits.test.jsx doesn't require to set any specific props, you could get rid of it and just use the [DatePickerDefault](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/src/terra-dev-site/test/date-picker/DatePickerDefault.test.jsx) test. You can do [this](https://github.com/cerner/terra-framework/blob/master/packages/terra-time-input/tests/wdio/time-input-spec.js#L114-L116) to make the cursor transparent and won't need the `h3`.

packages/terra-date-picker/src/DateUtil.js
BenBoersma: Fixed https://github.com/cerner/terra-framework/pull/595/commits/9af826c35a73bb5fad6c3023e10fbe23063035a5

packages/terra-date-picker/tests/wdio/date-picker-spec.js
BenBoersma: Fixed https://github.com/cerner/terra-framework/pull/595/commits/9af826c35a73bb5fad6c3023e10fbe23063035a5

packages/terra-date-picker/src/terra-dev-site/test/date-picker/DatePickerKeyLimits.test.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
```
jmsv6d: I don't think that this test page can be removed as it's not being used anymore.

BenBoersma: Removed https://github.com/cerner/terra-framework/pull/595/commits/edfd4d2c34f522b1a8a4742eec7e6fc6bfbd4f00

packages/terra-date-picker/tests/wdio/date-picker-spec.js
```diff
@@ -328,4 +328,22 @@ describe('Date Picker', () => {
...
+      });
+    });
+
+    it('sets the date', () => {
+      browser.setValue('input[name="terra-date-date-input"]', '0');
+      browser.keys('a1.b2/;3');
+    });
+
+    Terra.should.matchScreenshot();
+    Terra.should.beAccessible({ rules: ignoredA11y });
```
yuderekyu: want to use `Terra.should.validateElement` instead of these two functions?

BenBoersma: should be fixed here https://github.com/cerner/terra-framework/pull/595/commits/abba523319b419b80af7a36f29c869ca441d6e8f, but I noticed most of the rest tests in this test use the older way of doing tests.

yuderekyu: Yeah, eventually we should refactor tests to use that function. I'm not opposed to you doing that for this file though :D

mmalaker:
Functional verification completed. The date and date time picker examples now limit the date input to slash, period, and numeric characters.

PR #596, opened 2019-04-05
[terra-application-layout] Updated RoutingMenu to correctly synchronize internal state with props s provided after initial mount

tbiethman:
### Summary
We went through our components a while back and replaced the `componentWillReceiveProps` implementations with `getDerivedStateFromProps` in anticipation of the former's coming deprecation. When updated the RoutingMenu, we didn't remember to update the in-state copies of the prop values with the new prop values. This caused the logic to sometimes not execute as expected.

I updated the ApplicationLayout example to validate the bug was both reproducible without and corrected with this change.


ss037751:
Do you have any idea when the new version will be available in npm?

ss037751:
I appreciate the progress on this fix; any idea when the new version will be available in npm?

mjhenkes:
@ss037751 Terra released weekly. This fix would be available approximately one week after it has been merged.

ss037751:
@mjhenkes (or anyone else) - Is there a specific holdup to the Merge, and can you predict a date this fix will be available?

Issue #539, opened 2019-03-06
[terra-hookshot] Remove circular dependency.


dkasper-was-taken:
# Bug Report
Remove circular dependency on terra-form-select within terra-hookshot caused by selects in the documentation.

https://github.com/cerner/terra-framework/blob/master/packages/terra-hookshot/package.json#L40

## @ Mentions
@mjhenkes @tbiethman

ryanthemanuel:
We should use native selects instead.

ryanthemanuel:
@ShettyAkarsh Spoke about this in our KC war room and we decided that we will use native selects instead of terra-form-selects and not worry about any styling/browser differences.

bjankord:
Resolved in https://github.com/cerner/terra-framework/pull/600

PR #598, opened 2019-04-07
Fixes circular dependency issue in terra-hookshot due to dependency on terra-form-select . 

ShettyAkarsh:
### Summary
Resolves issue #539. Removed direct dependency on terra-form-select for terra-hookshot by updating the terra-dev-site documentation examples using native html select along with terra-hookhot.

### Additional Details

Created new class under terra-devsite doc refferred from all the exisiting example doc. which internally uses native html select and terra-hookshot component to build the dropdown by accepting the  following props  

labelName : Text applied to the Select label.
selectName:  Name applied to the html select.
value : Default value to be binded to the select.
selectOptions :  Options to be binded to html select.
hookshotName: Name applied to hookshot.
idAppedName: Text to be appended when setting the id attribute for either select or hookshot components.

Created new style compoenent under terra-devsite docs to match the dropdown appearance of terra-form-select and have similar appearance for html select across all the browsers. this styles is refferred from the above class .

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -23,7 +23,7 @@ Cerner Corporation
 - Shetty Akarsh [@@ShettyAkarsh]
 - Alex Bezek [@alex-bezek]
 - Supreeth MR [@supreethmr]
-- Shetty Akarsh [@ShettyAkarsh]
+- Akarsh Shetty [@ShettyAkarsh]
 - Gabe Parra [@gabeparra01]
 - Naveen Kumar Ramamurthy [@nramamurth]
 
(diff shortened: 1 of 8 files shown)


dkasper-was-taken:
Hookshot examples do not behave according to select value chosen.

dkasper-was-taken:
Is there a reason we are using hookshot rather than a native select control in the examples?

nramamurth:
Why is it that the `HTMLSelect` component renders a `Hookshot`? 

https://github.com/cerner/terra-framework/blob/580b781fe56310acedce799ba86de750416ec233/packages/terra-hookshot/src/terra-dev-site/doc/example/HTMLSelectExample.jsx#L168-L180

ShettyAkarsh:
> Is there a reason we are using hookshot rather than a native select control in the examples?

@dkasper-was-taken  there was a confusion here. thought we would need to replace the dropdown with hookshot of native select . as terra-form-select internally uses hookshot to display the values in the dropdown. myself will go ahead and close this PR and log the new pull request with only native selects being used.

PR #600, opened 2019-04-09
 Fixes circular dependency issue in terra-hookshot due to dependency on terra-form-select

ShettyAkarsh:
### Summary
Resolves issue #539. Removed direct dependency on terra-form-select for terra-hookshot by updating the terra-dev-site documentation examples using native html select.


packages/terra-hookshot/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Updated the terra-dev-site documentation examples to use native select and Removed Dependency on terra-form-select.
```
emilyrohrbough: Lets add a comment stating we removed the `terra-form-select` dependency

PR #601, opened 2019-04-09
Replace react-popper usage in terra-date-picker with terra-hookshot

### LINKED ISSUE #597: Update date picker to use hookshot for positioning
# Feature Request

## Description
Now that we've forked the date picker we were using, we should look into updating the positioning logic to use terra-hookshot instead of react-popper.

bjankord:
### Summary
Resolves #597


packages/terra-date-picker/src/DatePositionUtil.js
```diff
@@ -0,0 +1,59 @@
+/**
+ * Calculates the bottom position of the element.
```
bjankord: This is based on [logic we use for positioning arrows in terra-popup](https://github.com/cerner/terra-framework/blob/master/packages/terra-popup/src/_PopupUtils.js#L187).

packages/terra-date-picker/src/react-datepicker/index.jsx
```diff
@@ -542,19 +550,31 @@ export default class DatePicker extends React.Component {
...
-          <div className="react-datepicker__input-container">
-            {this.renderDateInput()}
-            {this.renderClearButton()}
-          </div>
-        }
-        popperContainer={this.props.popperContainer}
-        popperComponent={calendar}
-        popperPlacement={this.props.popperPlacement}/>
+      <React.Fragment>
+        <div ref={(container) => { this.datePickerContainer = container; }} className="react-datepicker__input-container">
```
tbiethman: Not a huge deal, but it'd be nice to start using `createRef` for our ref needs: https://reactjs.org/docs/refs-and-the-dom.html#creating-refs

mmalaker:
Functional verification completed. The date picker examples display and function as expected using Hookshot for positioning.

packages/terra-date-picker/src/react-datepicker/index.jsx
bjankord: Good call, updated in this commit: https://github.com/cerner/terra-framework/pull/601/commits/dea77f089b4c8d6dd9e45945272ad2e9d3101423

PR #605, opened 2019-04-10
Creating the NavigationPrompt

tbiethman:
### Summary
This PR introduces the terra-navigation-prompt package and the NavigationPrompt and NavigationPromptCheckpoint components.

### Additional Details

A React component has complete control over its own rendering logic. It can render child components and unmount those child components as it sees fit. It follows that any individual component has no direct control over its own mounting and unmounting; it is rendered (or not) at the discretion of its parent. A component must rely on its lifecycle events to determine how it is being implemented.

However, a component's children can be stateful and responsible for important user input or interactions. If the user can navigate away from that child component, it may be necessary for that child component to notify the user that their workflow is incomplete or that some transient state has not yet been persisted. If the component has no control over its own unmounting, how can it communicate this state to the overall application before the component is already being unmounted?

The most straightforward solution is to lift the state up. If the state is lifted into a parent component, it can be more directly managed by the component responsible for navigation, or it could be persisted indefinitely. **This is the strategy that should be implemented by most components.** However, for very large applications or highly composable applications built out of many stateful components, lifting the state may not be feasible.

Another solution is to expose callbacks from the stateful components. Whenever the stateful component's state changes, the callbacks are executed to notify the consumer of the current state. This works well at a small scale, but this becomes hard to maintain as the component tree increases in depth. One would have to chain these callbacks many levels deep to ensure every component is kept abreast of the current state, and every component, even those without these state concerns, would have to adhere to a consistent prop interface for those callbacks.

The NavigationPrompt and the NavigationPromptCheckpoint were created smooth over the rough edges of the aforementioned strategies. Components that have sensitive state can render a NavigationPrompt to make their state known to the application. Components that navigate between components can use the NavigationPromptCheckpoint to check for the presence of NavigationPrompts and inform the user in a consistent way. State remains localized to the component that needs it, communication of state changes occurs automatically, and components that don't share these concerns can safely ignore all of it.

### Deployment
https://terra-framework-deploye-pr-605.herokuapp.com/#/components/terra-navigation-prompt/navigation-prompt/about


packages/terra-navigation-prompt/docs/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+Another solution is to expose callbacks from the stateful components. Whenever the stateful component's state changes, the callbacks are executed to notify the consumer of the current state. This works well at a small scale, but this becomes hard to maintain as the component tree increases in depth. One would have to chain these callbacks many levels deep to ensure every component is kept abreast of the current state, and every component, even those without these state concerns, would have to adhere to a consistent prop interface for those callbacks.
+
+The NavigationPrompt and the NavigationPromptCheckpoint were created smooth over the rough edges of the aforementioned strategies. Components that have sensitive state can render a NavigationPrompt to make their state known to the application. Components that navigate between components can use the NavigationPromptCheckpoint to check for the presence of NavigationPrompts and inform the user in a consistent way. State remains localized to the component that needs it, communication of state changes occurs automatically, and components that don't share these concerns can safely ignore all of it.
```
jmsv6d: Grammar nit: were created _to_ smooth over

packages/terra-navigation-prompt/docs/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+#### `onPromptChange`
+
+The `onPromptChange` function prop is used to communicate NavigationPrompt registrations to the implementer of the NavigationPromptCheckpoint. The function is called with an array of objects containing each registered NavigationPrompt's `description` and `metaData` properties.
+
+> Note: The `onPromptChange` prop will be called after each NavigationPrompt registration and removal. It may be called more than once with the same array of NavigationPrompt data. Calls to `setState` within onPromptChange should be performed only when necessary to minimize component updates.
+
+#### `resolvePrompts`
+
+The `resolvePrompts` function be accessed from the ref to a NavigationPromptCheckpoint. Calling `resolvePrompts` results in a Promise being returned and a NotificationDialog being presented to the user with options to either confirm or cancel their action. If the user confirms the action, the dialog will close, and the returned Promise will be resolved. If the user cancels the action, the dialog will close, and the returned Promise will be rejected. If no NavigationPrompts are detected, no dialog is presented, and a resolved Promise is returned.
```
mjhenkes: > The `resolvePrompts` function be accessed from the ref to a NavigationPromptCheckpoint.

packages/terra-navigation-prompt/src/terra-dev-site/doc/navigation-prompt/examples/NavigationPromptExample.jsx
```diff
@@ -0,0 +1,195 @@
...
+    </div>
+  );
+};
+
+Form.propTypes = {
+  title: PropTypes.string,
+};
+
+/**
+ * The FormSwitcher toggles between to versions a Form to demonstrate the functionality of nested NavigationPromptCheckpoints.
```
mjhenkes: > between to versions a Form
between two version of a Form

packages/terra-navigation-prompt/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "terra-navigation-prompt",
+  "main": "lib/index.js",
+  "version": "0.0.0",
+  "description": "TODO",
```
mjhenkes: TODO

packages/terra-navigation-prompt/src/NavigationPrompt.jsx
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * An object containing any other pertinent information related to the NavigationPrompt.
+   */
+  metaData: PropTypes.object,
+  /**
+   * @private
+   * An object containing prompt registration APIs provided through the PromptRegistrationContext.
+   */
+  promptRegistration: PropTypes.shape({
+    isDefaultContextValue: PropTypes.bool,
```
mjhenkes: Add some documentation about what this means maybe?

packages/terra-navigation-prompt/docs/README.md
```diff
@@ -0,0 +1,138 @@
+# Terra Navigation Prompt
+
+The NavigationPrompt and NavigationPromptCheckpoint components allow for the detection and management of components with unsaved state.
+
+## Getting Started
+
+- Install with [npmjs](https://www.npmjs.com):
+  - `npm install terra-navigation-prompt`
```
emilyrohrbough: Out of curiosity, should an application only pull in one version of navigation prompt, like react?

 If so, I think it would be helpful to expand on how this should be pulled into a project as a dev-dependency and peer dependency to later be pulled into an different project which is the application.

packages/terra-navigation-prompt/docs/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+However, a component's children can be stateful and responsible for important user input or interactions. If the user can navigate away from that child component, it may be necessary for that child component to notify the user that their workflow is incomplete or that some transient state has not yet been persisted. If the component has no control over its own unmounting, how can it communicate this state to the overall application before the component is already being unmounted?
+
+The most straightforward solution is to lift the state up. If the state is lifted into a parent component, it can be more directly managed by the component responsible for navigation, or it could be persisted indefinitely. **This is the strategy that should be implemented by most components.** However, for very large applications or highly composable applications built out of many stateful components, lifting the state may not be feasible.
```
emilyrohrbough: The most straightforward solution is to lift the state up. If the **child component's** state is lifted into a parent component, it can **either** be ~more~ directly managed by the **parent** component **that is** responsible for navigation, or it **can** ~could~ be persisted indefinitely.

packages/terra-navigation-prompt/docs/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+Another solution is to expose callbacks from the stateful components. Whenever the stateful component's state changes, the callbacks are executed to notify the consumer of the current state. This works well at a small scale, but this becomes hard to maintain as the component tree increases in depth. One would have to chain these callbacks many levels deep to ensure every component is kept abreast of the current state, and every component, even those without these state concerns, would have to adhere to a consistent prop interface for those callbacks.
```
emilyrohrbough: Another solution is to expose callbacks from the stateful **children** components. Whenever the stateful component's state changes, ~the~ **these** callbacks ~are~ **would** executed to notify the ~consumer~ **parent component** of the ~current~ **updated** state.

packages/terra-navigation-prompt/docs/README.md
emilyrohrbough: - The `NavigationPrompt` and the `NavigationPromptCheckpoint`

- _rough edges of the aforementioned strategies._ should this be strategy?

- Components that have ~sensitive~ ** transient** state 

> Components that have sensitive state can render a NavigationPrompt to make their state known to the application.

its known to the application if it has implemented the `NavigationPromptCheckpoint`  or because the application has access to the context?

packages/terra-navigation-prompt/docs/README.md
```diff
@@ -0,0 +1,138 @@
...
+When a component renders a NavigationPrompt, it is telling its parent components two things: that the component is in the middle of something important, and that users may care about the component going away.
+
+The NavigationPrompt can be rendered with two props: a `description` string that gives context to the NavigationPrompt, and a `metaData` object that can be used to provide additional data that may be relevant to the NavigationPrompt.
+
+The NavigationPrompt can be rendered like any other component, but prop consistency should be maintained between renders when possible. Creating a new object for the `metaData` prop within a render function will cause the NavigationPrompt to deregister and reregister itself during every update:
+
+```jsx
+render() {
+  return (
+    <NavigationPrompt description="This prompt will register and notify every time the component renders" metaData={{ important: 'info' }} />
```
emilyrohrbough: Might be easier to read like:
```
    <NavigationPrompt
        description="This prompt will register and notify every time the component renders"
         metaData={{ important: 'info' }}
      />
```

packages/terra-navigation-prompt/docs/README.md
```diff
@@ -0,0 +1,138 @@
...
+  );
+}
+```
+
+To avoid this, the `metaData` value can be stored inside a component's state or an internal property and updated only when necessary:
+
+```jsx
```
emilyrohrbough: It'd be nice to update this to use diff syntax to quick see the changes you suggested

packages/terra-navigation-prompt/docs/README.md
```diff
@@ -0,0 +1,138 @@
...
+render() {
+  return (
+    <NavigationPrompt description="This prompt will register and notify every time the component renders" metaData={{ important: 'info' }} />
+  );
+}
+```
+
+To avoid this, the `metaData` value can be stored inside a component's state or an internal property and updated only when necessary:
```
emilyrohrbough: To avoid ~this,~ **a deregister and reregister for an update,**

packages/terra-navigation-prompt/docs/README.md
emilyrohrbough:
```suggestion
```diff
+ constructor() {
+   this.state = {
+     metaDataValue: {
+       important: 'info',
+     }
+   }
+ }

render() {
  return (
    <NavigationPrompt
        description="My props are consistent between renders, so this prompt will not re-register or notify" 
-       metaData={{ important: 'info' }}
+       metaData={this.state.metaDataValue}
     />
  );
}

```

packages/terra-navigation-prompt/docs/README.md
```diff
@@ -0,0 +1,138 @@
...
+render() {
+  return (
+    <NavigationPrompt description="My props are consistent between renders, so this prompt will not re-register or notify" metaData={this.state.metaDataValue} />
+  );
+}
+```
+
+### `NavigationPromptCheckpoint`
+
+The NavigationPromptCheckpoint can be used to detect the presence of NavigationPrompts rendered within it.
```
emilyrohrbough: playing dumb, this should be the parent component that should implement this? Or top level navigation or application root?

packages/terra-navigation-prompt/docs/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+The `resolvePrompts` function be accessed from the ref to a NavigationPromptCheckpoint. Calling `resolvePrompts` results in a Promise being returned and a NotificationDialog being presented to the user with options to either confirm or cancel their action. If the user confirms the action, the dialog will close, and the returned Promise will be resolved. If the user cancels the action, the dialog will close, and the returned Promise will be rejected. If no NavigationPrompts are detected, no dialog is presented, and a resolved Promise is returned.
+
+`resolvePrompts` accepts a single Object argument that should contain the text strings used to render the NotificationDialog. If these strings need to present information based on the currently registered NavigationPrompts, the `onPromptChange` prop can be used to receive that data and build those strings as needed.
+
+The supported keys in the resolvePropts argument include:
+
+|Key|Type|Description|
+|---|---|---|
+|`title`|String|The title of the NotificationDialog|
```
emilyrohrbough: Do we provide default strings?

packages/terra-navigation-prompt/docs/README.md
```diff
@@ -0,0 +1,138 @@
...
+        ref={this.checkpointRef}
+      >
+        <Content key={contentKey} />
+      </NavigationPromptCheckpoint>
+    </div>
+  );
+}
+```
+
+### `PromptRegistrationContext`
```
emilyrohrbough: This is the first mention of PromptRegistrationContext. As a consumer, should I be using this or is this just to explain how this works behind the scenes? I'm not seeing reference in the guides

packages/terra-navigation-prompt/docs/implementation-guide/IMPLEMENTATION_GUIDE_PART_0.md
```diff
@@ -0,0 +1,81 @@
+# Terra Navigation Prompt - Implementation Guide
+
+Assume we have built a simple component, the FormSwitcher, that toggles between showing two different stateful components, Form 1 and Form 2. Users of the FormSwitcher have recently complained that they can switch between forms before submitting their in progress form, losing their data in the process. We want to implement the NavigationPrompt and NavigationPromptCheckpoint to give our users more control over the FormSwitcher's navigation.
+
+> Note: This implementation guide features examples and use cases for the various APIs provided by the NavigationPrompt and NavigationPromptCheckpoint.
+> Please review your individual requirements to see which APIs are necessary for your specific needs.
```
emilyrohrbough: > ‼️ Please review your individual requirements to see which APIs are necessary for your specific needs.

packages/terra-navigation-prompt/docs/implementation-guide/IMPLEMENTATION_GUIDE_PART_0.md
```diff
@@ -0,0 +1,81 @@
...
+
+  return (
+    <div>
+      <h2>Form Switcher</h2>
+      <p>The NavigationPrompt is not implemented, so no prompting occurs.</p>
+      <button
+        type="button"
+        disabled={activeForm === 'Form 1'}
+        onClick={onSwitchForm.bind(null, 'Form 1')}
+
```
emilyrohrbough: nit: extra line

packages/terra-navigation-prompt/docs/implementation-guide/IMPLEMENTATION_GUIDE_PART_0.md
```diff
@@ -0,0 +1,81 @@
...
+    setActiveForm(formKey);
+  }
+
+  return (
+    <div>
+      <h2>Form Switcher</h2>
+      <p>The NavigationPrompt is not implemented, so no prompting occurs.</p>
+      <button
+        type="button"
+        disabled={activeForm === 'Form 1'}
```
emilyrohrbough:
```suggestion
        disabled={activeForm !== 'Form 1'}
```

packages/terra-navigation-prompt/docs/implementation-guide/IMPLEMENTATION_GUIDE_PART_0.md
```diff
@@ -0,0 +1,81 @@
...
+        onClick={onSwitchForm.bind(null, 'Form 1')}
+
+      >
+        Switch to Form 1
+      </button>
+      <button
+        type="button"
+        disabled={activeForm === 'Form 2'}
```
emilyrohrbough:
```suggestion
        disabled={activeForm !== 'Form 2'}
```

packages/terra-navigation-prompt/docs/implementation-guide/IMPLEMENTATION_GUIDE_PART_0.md
```diff
@@ -0,0 +1,81 @@
...
+      </button>
+      <Form title={activeForm} key={activeForm} />
+    </div>
+  );
+};
+
+export default FormSwitcher;
+```
+
+The FormSwitcher renders two buttons that allow users to toggle between showing Form 1 and Form 2. Form 1 and Form 2 are unique instances of the Form component.
```
emilyrohrbough: I think it might be nice to move this description above the code snippet

packages/terra-navigation-prompt/docs/implementation-guide/IMPLEMENTATION_GUIDE_PART_1.md
```diff
@@ -0,0 +1,95 @@
+# Terra Navigation Prompt - Implementation Guide
+
+## Part 1 - Implementing the NavigationPrompt and NavigationPromptCheckpoint
+
+To better notify the user of the Form's transient state, we update the Form to render a NavigationPrompt.
+
+```diff
+- import React, { useState } from 'react';
++ import React, { useState, useRef } from 'react';
```
emilyrohrbough: Do we use `useRef` here?

packages/terra-navigation-prompt/docs/implementation-guide/IMPLEMENTATION_GUIDE_PART_1.md
```diff
@@ -0,0 +1,95 @@
...
++      </NavigationPromptCheckpoint>
+    </div>
+  );
+};
+```
+
+The FormSwitcher gets a ref to the NavigationPromptCheckpoint; this example uses the [useRef hook](https://reactjs.org/docs/hooks-reference.html#useref), but any of the supported ref retrieval methods would be sufficient. The FormSwitcher then uses this ref to call the `resolvePrompts` function when it wants to switch to a new Form.
+
+The `resolvePrompts` function is an instance function of the NavigationPromptCheckpoint. When `resolvePrompts` is called, the NavigationPromptCheckpoint will render a NotificationDialog and return a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises). The Promise will be resolved immediately if no NavigationPrompts have been rendered below the checkpoint, so we don't need to check for the presence of any NavigationPrompts before we call it. The Promise will also resolve if a user presses the NotificationDialog's secondary action button. The Promise will reject if a user presses the NotificationDialog's primary action button. In our example, we only update the FormSwitcher's state when the Promise is resolved.
```
emilyrohrbough: The Promise will also resolve if a user presses the NotificationDialog's ~secondary~ ** accept** action button. The Promise will reject if a user presses the NotificationDialog's ~primary~ **reject** action button.

packages/terra-navigation-prompt/docs/implementation-guide/IMPLEMENTATION_GUIDE_PART_2.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Part 3 - Implementing a Custom Prompt Solution
+
+At this point, we realize that we do not want to **ever** allow users to navigate away from a Form with unsaved state. We can do that by using the NavigationPromptCheckpoint's `onPromptChange` prop.
+
+```diff
+- import React, { useState, useRef } from 'react';
++ import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import NavigationPrompt, { NavigationPromptCheckpoint } from '../../../../index';
```
emilyrohrbough:
```suggestion
import NavigationPrompt, { NavigationPromptCheckpoint } from 'terra-navigation-prompt';
```

packages/terra-navigation-prompt/src/NavigationPrompt.jsx
emilyrohrbough: Should we export this shape from `PromptRegistrationContext` ?

packages/terra-navigation-prompt/src/NavigationPromptCheckpoint.jsx
```diff
@@ -0,0 +1,148 @@
...
+   *   this.myLocalPromptRegistry = arrayOfPrompts;
+   * }`
+   */
+  onPromptChange: PropTypes.func,
+  /**
+   * @private
+   * An object containing prompt registration APIs provided through the PromptRegistrationContext.
+   */
+  promptRegistration: PropTypes.shape({
+    registerPrompt: PropTypes.func.isRequired,
```
emilyrohrbough: could `isDefaultContextValue: PropTypes.bool,` be passed?

packages/terra-navigation-prompt/src/NavigationPromptCheckpoint.jsx
```diff
@@ -0,0 +1,148 @@
...
+       * The consumer is notified on unmount with an empty set of prompt data to clean up any previously mounted prompts.
+       */
+      onPromptChange([]);
+    }
+  }
+
+  registerPrompt(promptId, promptDescription, metaData) {
+    const { onPromptChange, promptRegistration } = this.props;
+
+    if (!promptId) {
```
emilyrohrbough: If we are using `uuidv4` to generate a unique ID for consumers will this check ever be hit?
If it is hit, will this be a hard gottchya to track down?

packages/terra-navigation-prompt/src/NavigationPromptCheckpoint.jsx
```diff
@@ -0,0 +1,148 @@
...
+      return;
+    }
+
+    this.registeredPrompts[promptId] = [promptDescription, metaData];
```
emilyrohrbough: Out of curiosity, why store this as an array instead of a hash? I notice above you accessed these like
```
description: prompts[id][0], metaData: prompts[id][1]
```

packages/terra-navigation-prompt/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "scripts": {
+    "compile": "npm run compile:clean && npm run compile:build",
+    "compile:clean": "rm -rf lib",
+    "compile:build": "babel src --out-dir lib --copy-files",
+    "lint": "npm run lint:js && npm run lint:scss",
+    "lint:js": "eslint --ext .js,.jsx . --ignore-path ../../.eslintignore",
+    "lint:scss": "stylelint src/**/*.scss",
+    "test": "npm run test:jest && npm run test:wdio",
+    "test:jest": "jest --config ../../jestconfig.js",
+    "test:wdio": "../../node_modules/.bin/wdio ../../node_modules/terra-dev-site/config/wdio/wdio.conf.js"
```
emilyrohrbough: It looks like we have no wdio tests. I think we can remove this

packages/terra-navigation-prompt/docs/README.md
tbiethman: There are no default strings. We could include them if we can land on a reasonable set of terms, but from what I can tell, the messaging will likely need to be specific to each use case.

Consumer's could wrap the NavigationPromptCheckpoint to provide their own default strings if that's desirable to them.

packages/terra-navigation-prompt/docs/README.md
tbiethman: I waffled on including this at all, but I figured I should document this for anyone wanting to interface with the context themselves. I can further document my expectations around this.

packages/terra-navigation-prompt/src/NavigationPromptCheckpoint.jsx
tbiethman: It should always be provided by the NavigationPrompt, but I can't guarantee that from other custom implementations should they exist. We could do a dev-only warn here in that case.

packages/terra-navigation-prompt/src/_CheckpointNotificationDialog.jsx
```diff
@@ -0,0 +1,54 @@
...
+        message={dialogProps.message}
+        primaryAction={{
+          text: dialogProps.rejectButtonText,
+          onClick: this.hideDialog.bind(null, dialogProps.onReject),
+        }}
+        secondaryAction={{
+          text: dialogProps.acceptButtonText,
+          onClick: this.hideDialog.bind(null, dialogProps.onAccept),
+        }}
+        variant="warning"
```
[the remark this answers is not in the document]
tbiethman: Maybe. I don't want to expose it prematurely, but we could expose it at a later date or outright change it to something else.

packages/terra-navigation-prompt/src/NavigationPromptCheckpoint.jsx
tbiethman: I know that I'm always dealing with these two values, so I figured the looking them up by index would be (ever so slightly) more efficient. Its probably pretty arbitrary with a key space this small.

tbiethman: Given that I'm converting ultimately converting to an Object, I may as well make an Object from the get go.

packages/terra-navigation-prompt/src/NavigationPromptCheckpoint.jsx
tbiethman: It will be passed with the default context value. This will happen for the top-most checkpoint in the hierarchy. I didn't include it here because the NavigationPromptCheckpoint doesn't actually use that value and thus doesn't really have an expectation that it is there, even optionally. But I think I'll export the shape from the context and reuse it here and in the NavigationPrompt, rendering this all moot.

packages/terra-navigation-prompt/docs/implementation-guide/IMPLEMENTATION_GUIDE_PART_2.md
tbiethman: Argh missed one

packages/terra-navigation-prompt/docs/README.md
tbiethman: I waffled some more. I'm just going to remove it. I don't want to open the door to that level of customization out of the gate.

packages/terra-navigation-prompt/docs/README.md
tbiethman: Any component that navigates between stateful child components should render NavigationPromptCheckpoints around those children. Each NavigationPrompt rendered by a child will be registered to the NavigationPromptCheckpoints above it in the component tree. The implementer of each NavigationPromptCheckpoint can use the registered data to prompt the user before navigating or prevent navigation altogether.

packages/terra-navigation-prompt/docs/README.md
tbiethman: > its known to the application if it has implemented the NavigationPromptCheckpoint or because the application has access to the context?

 The former. I'll rephrase.

packages/terra-navigation-prompt/src/NavigationPrompt.jsx
tbiethman: Yes and yes

packages/terra-navigation-prompt/docs/README.md
tbiethman: I too think that would help.

packages/terra-navigation-prompt/docs/implementation-guide/IMPLEMENTATION_GUIDE_PART_2.md
```diff
@@ -0,0 +1,88 @@
...
++       }}
+      >
+        <Form title={activeForm} key={activeForm} />
+      </NavigationPromptCheckpoint>
+    </div>
+  );
+};
+```
+
+We remove the `ref` from the NavigationPromptCheckpoint and instead implement `onPromptChange` to store the currently rendered NavigationPrompt's in the FormSwitcher's state. We disable the 'Switch to' buttons altogether if a NavigationPrompt is detected and render our own custom message within the FormSwitcher to notify our users. We could also use this state to control other navigation-capable components, like client-side routers.
```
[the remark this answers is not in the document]
tbiethman: That was referring to something that no longer exists. I'll remove.

packages/terra-navigation-prompt/src/terra-dev-site/doc/navigation-prompt/examples/NavigationPromptExample.jsx
tbiethman: 68704dd

packages/terra-navigation-prompt/docs/README.md
tbiethman: 68704dd

packages/terra-navigation-prompt/docs/README.md
tbiethman: 68704dd

packages/terra-navigation-prompt/docs/README.md
tbiethman: 68704dd

packages/terra-navigation-prompt/docs/implementation-guide/IMPLEMENTATION_GUIDE_PART_0.md
tbiethman: 68704dd

packages/terra-navigation-prompt/docs/implementation-guide/IMPLEMENTATION_GUIDE_PART_1.md
tbiethman: 68704dd

packages/terra-navigation-prompt/src/NavigationPromptCheckpoint.jsx
tbiethman: 68704dd

packages/terra-navigation-prompt/src/NavigationPromptCheckpoint.jsx
tbiethman: 68704dd

packages/terra-navigation-prompt/docs/README.md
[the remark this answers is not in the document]
tbiethman: 68704dd

packages/terra-navigation-prompt/docs/README.md
```diff
@@ -0,0 +1,138 @@
...
+The PromptRegistrationContext defines the communication contract between NavigationPrompts and NavigationPromptCheckpoints. A PromptRegistrationContext.Provider must provide an Object value with the following attributes:
+
+|Key|Type|Description|
+|---|---|---|
+|`registerPrompt(promptId, promptDescription, metaData)`|Function|A function which should be called when a NavigationPrompt is mounted.|
+|`deregisterPrompt(promptId)`|Function|A function which should be called when a NavigationPrompt is unmounted.|
+
+## Extending the NavigationPrompt
+
+The various APIs provided and consumed by the NavigationPrompt and NavigationPromptCheckpoint are purposefully broad and unrestrictive. If more defined or restricted APIs are desirable for a application platform, these components can and should be wrapped by components that provide the necessary interfaces.
```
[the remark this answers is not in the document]
tbiethman: d'oh 68704dd

packages/terra-navigation-prompt/docs/implementation-guide/IMPLEMENTATION_GUIDE_PART_2.md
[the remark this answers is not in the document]
tbiethman: 68704dd

packages/terra-navigation-prompt/src/NavigationPromptCheckpoint.jsx
```diff
@@ -0,0 +1,148 @@
...
+  /**
+   * `resolvePrompts` returns a Promise that will be resolved or rejected based upon the presence of child prompts and
+   * the the actions taken by a user from the checkpoint's NotificationDialog.
+   *
+   * This function is part of the NavigationPromptCheckpoint's public API. Changes to this function name or overall functionality
+   * will impact the package's version.
+   */
+  resolvePrompts(dialogOptions) {
+    if (!Object.keys(this.registeredPrompts).length) {
+      /**
```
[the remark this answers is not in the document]
tbiethman: 68704dd

packages/terra-navigation-prompt/src/NavigationPromptCheckpoint.jsx
```diff
@@ -0,0 +1,148 @@
...
+  children: PropTypes.node,
+  /**
+   * A function that will be executed when NavigationPrompts are registered to or deregistered from the NavigationPromptCheckpoint instance.
+   * This can be used to track registered prompts outside of a NavigationPromptCheckpoint and handle prompt resolution directly, if necessary.
+   * The function will be provided with an array of data objects representing the registered NavigationPrompts as the sole argument. An empty
+   * Array will be provided when no prompts are registered.
+   *
+   * Function Example:
+   *
+   * `(arrayOfPrompts) => {
```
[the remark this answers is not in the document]
tbiethman: 68704dd

packages/terra-navigation-prompt/docs/implementation-guide/IMPLEMENTATION_GUIDE_PART_0.md
```diff
@@ -0,0 +1,81 @@
+# Terra Navigation Prompt - Implementation Guide
```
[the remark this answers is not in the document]
tbiethman: c45f197

packages/terra-navigation-prompt/docs/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+#### `resolvePrompts`
+
+The `resolvePrompts` function be accessed from the ref to a NavigationPromptCheckpoint. Calling `resolvePrompts` results in a Promise being returned and a NotificationDialog being presented to the user with options to either confirm or cancel their action. If the user confirms the action, the dialog will close, and the returned Promise will be resolved. If the user cancels the action, the dialog will close, and the returned Promise will be rejected. If no NavigationPrompts are detected, no dialog is presented, and a resolved Promise is returned.
+
+`resolvePrompts` accepts a single Object argument that should contain the text strings used to render the NotificationDialog. If these strings need to present information based on the currently registered NavigationPrompts, the `onPromptChange` prop can be used to receive that data and build those strings as needed.
+
+The supported keys in the resolvePropts argument include:
```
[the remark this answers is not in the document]
tbiethman: 68704dd

packages/terra-navigation-prompt/package.json
tbiethman: 68704dd

packages/terra-navigation-prompt/docs/implementation-guide/IMPLEMENTATION_GUIDE_PART_2.md
tbiethman: 68704dd

tbiethman:
With 3fa87db I added support for functions in the `resolvePrompts` API. This will allow consumers to create custom prompt strings, based on the set of actively registered prompts, without having to implement `onPromptChange` and storing those registered prompts locally. I ran into this with the NavigationPrompt implementation and found this makes my life a lot easier there.

mmalaker:
Functional review in progress. The prompt works as expected in all supported desktop browsers; however, I see some unusual behavior on mobile. On iOS, the prompt does not flex properly. It seems to be zooming in when you enter data into the form fields, then when the prompt is initially triggered, it is displayed beyond the bounds of the viewport. Once manually resized, it functions properly again.

On an Android device, the page does not render at all.

bjankord:
@tbiethman The sizing issues may be related to sizing issues identified in https://github.com/cerner/terra-framework/issues/621

tbiethman:
@mmalaker Regarding the zoom, that is an iOS feature for inputs. The examples here uses a bog standard <input> element. One way to prevent the zoom from happening is by setting a font-size on the input larger than 16px (the terra-form-input does this and thus does not exhibit this same zoom behavior). The NotificationDialog itself is rendering appropriately, as the viewport itself is still zoomed-in after presenting it. Zooming out with your fingers will reset the view and show the whole dialog. If we're worried about it, I could update the examples to set an appropriate font-size and prevent the zoom from happening. 

Regarding the Android device not rendering the page at all, that's news to me. I'll have to look into that one.

mmalaker:
@tbiethman thanks for the info. That all makes sense. I updated my Android device and now the page displays and the prompt works as expected. Interestingly, with Android I'm not seeing any zoom behavior. 

To make a better show on iOS in our example, could we do what you mentioned above and adjust the font size? Otherwise, everything looks good and is working correctly.

mmalaker:
Functional verification completed. The nav prompt now displays and functions as expected on all supported browsers and devices.

mjhenkes:
This build is failing because the docker-compose.yml file in terra-toolkit updated to use helium chrome version.
https://github.com/cerner/terra-toolkit/blob/master/config/docker/ci-docker-compose.yml#L10

Travis-ci pulls the docker-compose.yml file from master here.
https://github.com/cerner/terra-framework/blob/master/.travis.yml#L29

A cheap hack to get this PR done would be to change the travis-ci to point to the sha prior to our updating to helium

It cares not for your versions

PR #607, opened 2019-04-10
[terra-date-picker]  Fixed crash when selectedDate is not ISO compliant

### LINKED ISSUE #528: Failed propType warning for terra-date-picker using selectedDate prop
# Bug Report

## Description
terra-date-picker throws a console warning when rendering with the selectedDate prop filled out as a string.  That string is then given to the underlying react-datepicker component that expects it to be a date.

API for terra-date-picker showing that the selectedDate prop is a string:
https://engineering.cerner.com/terra-ui/#/components/terra-date-picker/date-picker/date-picker

API for the react-datepicker component that expect it to be a date:
https://github.com/Hacker0x01/react-datepicker/blob/master/src/index.jsx#L134

## Steps to Reproduce
1.  Render the terra-date-picker and give it a selectedDate prop on initial render

screen shot 2019-02-18 at 4 41 04 pm 1

## Expected Behavior
No warning to be shown in the console

## Possible Solution
Convert the string to a date object before giving it to react-datepicker

benbcai:
### Summary
Fixed crash when selectedDate is not ISO compliant

### Additional Details 
- Fixes #528
- If the selectedDate prop is set to an empty string, the selectedDate prop would be ignored and the proptype mismatch warning will no longer occur.
- If the selectedDate prop is set to a non ISO compliant string (e.g.; "abc"; "!@#$"), the selectedDate prop would be ignored and will not crash in render.

@cerner/terra


packages/terra-date-picker/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Fixed propType warning if selectedDate prop is an empty string.
```
StephenEsser: This might need rephrased. If it's a fix it should move under a fixed section. If we're considering this an enhancement we might just find an alternative to the word fixed.

StephenEsser: Allow empty string to represent no selected date

Updated the selectedDate to accept an empty string when no date is selected

packages/terra-date-picker/src/DateUtil.js
```diff
@@ -2,34 +2,48 @@
...
+  /**
+   * Creates a moment object using the provided date string.
+   * @param {string|undefined} date - The date to convert. Expect to be in ISO format.
+   * @return {object|undefined} - The moment object. Undefined if unable to convert.
+   */
   static createSafeDate(date) {
-    if (date) {
-      const momentDate = moment(date);
-      return momentDate.isValid() ? momentDate : date;
+    if (!date || (date && date.length === 0)) {
```
StephenEsser: Empty string is falsey, is there a case where date can be truthy and of length zero?

packages/terra-date-picker/src/DateUtil.js
```diff
@@ -2,34 +2,48 @@
...
-    return momentDates.length > 0 ? momentDates : dates;
+    return validMomentDates.length > 0 ? validMomentDates : undefined;
   }
 
-  // Checks if a given date is out of the range between the start and end dates.
+  /**
+   * Determines if a date if in the range of two given dates.
+   * @param {object} sourceDate - The moment date to check if it is within range.
+   * @param {object} minDate - The moment date that represents the minimum date of the range.
+   * @param {object} maxDate - The moment date that represents the maximum date of the range.
```
StephenEsser: A few of these params don't match the method signature below

packages/terra-date-picker/CHANGELOG.md
benbcai: Changed to use `Fix` and updated the description since it used to crash if the selectedProp is not ISO compliant and unable to convert to a moment object.
https://github.com/cerner/terra-framework/pull/607/commits/fa9fadf949ff404eb5b8fb2b5a5ca40e414af04b

packages/terra-date-picker/src/DateUtil.js
benbcai: You are correct. The check for length is not needed.
https://github.com/cerner/terra-framework/pull/607/commits/fa9fadf949ff404eb5b8fb2b5a5ca40e414af04b

packages/terra-date-picker/src/DateUtil.js
benbcai: In [PR #578](https://github.com/cerner/terra-framework/pull/578/files#diff-bed1c9aab9f3944cbb524e8668862592), I renamed the name of these parameters to minDate and maxDate. I'm using the updated names now so that I don't have to go back and update again.

jmsv6d: Since you're not making any changes to this function in this PR, I think it makes more sense to put the beefed up doc in #578

benbcai: Reverted the documentation and will add it using PR #578 
https://github.com/cerner/terra-framework/pull/607/commits/28984dc7cbf6d82a2063bc3e0b8c017f0d6e98f2

packages/terra-date-picker/src/DateUtil.js
```diff
@@ -2,31 +2,40 @@
...
+   * null, empty string, or alpha characters and undefined would be returned.
+   * @param {string|undefined} date - The date to convert. Expect to be in ISO format.
+   * @return {object|undefined} - The moment object. Undefined if unable to convert.
+   */
   static createSafeDate(date) {
-    if (date) {
-      const momentDate = moment(date);
-      return momentDate.isValid() ? momentDate : date;
+    if (!date) {
+      return undefined;
```
yuderekyu: nit - an empty return statement automatically returns undefined
```suggestion
      return;
```

benbcai: Having an empty return statement here would introduce a [consistent-return](https://eslint.org/docs/rules/consistent-return) linter error since it also needs to return a moment object at the end of this function.

Issue #609, opened 2019-04-15
Default browser focusing style adds blue line wrapping slide panel


FloraChenCerner:
# Bug Report

## Description

#584 set the detailed panel to be focused when slide panel is open. With this change the browser's default focusing style adds a blue line on the left of detail panel, and when dismissing the detail view, master view is wrapped with blue outline. This behavior fails most of our components wdio tests, and we don't want the blue line from the UX perspective. 

One of the failing build jobs: https://jenkins.cerner.com/ion/job/ion-chart/job/ion-conditions-js/job/master/159/

## Steps to Reproduce

1. Implement a slide panel with any behavior. 
2. Open the slide panel and see the blue outline around detail panel
3. Close the slide panel and see the blue outline around master panel

## Additional Context / Screenshots

![1128283 _Renders_Detail_on_Condition_Selection default 1300x768](https://user-images.githubusercontent.com/43655107/56145845-b0ddaa80-5f6a-11e9-9066-96ef962cb71f.png)

## Expected Behavior

No blue outline around master or detail panel when opening/closing

## Possible Solution

Probably to make the style of focusing to be clear

## Environment

* Component Name and Version: terra-slide-panel 3.3.0
* Browser Name and Version: Chrome Version 73.0.3683.10
* Operating System and version (desktop or mobile): Mac OS

## @ Mentions

@dkasper-was-taken @tbiethman

tbiethman:
We'd just need to set `outline` to `none` for the two SlidePanel content areas that are now focusable.

neilpfeiffer:
As long as the element that receives the focus when the panel gets the focus-trap is the parent container and not an actionable element (and the parent container could/should never have actionability on it, it would be the similar as the when the base body technically has the focus before user clicks tab for the first time) -- lets go ahead and add the outline:none, since there is a clear and obvious visual interaction of the panel opening which is the same effect as a visual indication of focus location.

Issue #606, opened 2019-04-10
terra-slide-panel Takes focus while panel is closed


Me999999999:
# Bug Report

## Description
After pulling in terra-slide-panel version 3.3.0 our functionality broke for an input-field inside the slide panel. While the slide-panel was closed the input-field would lose focus after each letter you entered. This was not the case while the slide-panel was open, in this case the text area functioned as expected.

## Steps to Reproduce
1. Create a SlidePanel with an InputField as part of its mainContent
2. Type into the InputField while the SlidePanel is closed
Focus will be removed from the InputField

## Additional Context / Screenshots
For our specific case our InputField is the child of an Item, which is the child of a List, which is the mainContent of the SlidePanel.

## Expected Behavior
The InputField should remain focused while typing in to it.

## Possible Solution
It seems the problem that handleTransitionEnd will always be called when the panel is closed; giving the mainNode focus.

## Environment
* Component Name and Version: terra-slide-panel 3.3.0
* Browser Name and Version: Chrome
* Operating System and version (desktop or mobile): Mac Desktop

tbiethman:
This would be resolved with #447. We need to remove the handleTransitionEnd/preparePanelForTransition logic here: https://github.com/cerner/terra-framework/blob/master/packages/terra-slide-panel/src/SlidePanel.jsx#L87-L108

mfried1:
I am experiencing a similar occurrence. I have a SlidePanel and a Select on a modal dialog (The Select is not contained within the SlidePanel.) When I click on the Select, the dropdown opens and immediately closes because the SlidePanel takes focus.

makr28:
I am seeing a similar issue. Whenever our code calls setState, the slidepanel gets the focus and our page scrolls down to it

Issue #447, opened 2019-01-25
Update slide panels to work without relying on transitionEnd event


bjankord:
# Feature Request

## Description
We currently use `transitionEnd` event to infer the state of the slide panels. With some styling changes we added recently, the transition isn't as performant as we'd like it to be. Need to find a way where we can handle the slide panel state when the panel transition is set to none.

## Additional Context / Screenshots
Originated on this PR: https://github.com/cerner/terra-framework/pull/446#issuecomment-457652996

## @ Mentions
@ryanthemanuel

PR #617, opened 2019-04-17
[terra-slide-panel] Turn focus outline off; replace transition logic

yuderekyu:
### Summary
Closes https://github.com/cerner/terra-framework/issues/609, https://github.com/cerner/terra-framework/issues/606, https://github.com/cerner/terra-framework/issues/447

#### Changes
* Set main + content panel outline to none
* Move focus transition logic to `componentDidUpdate`


packages/terra-slide-panel/src/SlidePanel.jsx
```diff
@@ -58,53 +58,24 @@ class SlidePanel extends React.Component {
     super(props);
     this.setPanelNode = this.setPanelNode.bind(this);
     this.mainNode = React.createRef();
-    this.handleTransitionEnd = this.handleTransitionEnd.bind(this);
-    this.preparePanelForTransition = this.preparePanelForTransition.bind(this);
 
     this.isHidden = !props.isOpen;
```
tbiethman: From what I'm seeing, I don't think we need to manage this internal property anymore?

tbiethman: without animations, isHidden is just !isOpen all the time.

yuderekyu: https://github.com/cerner/terra-framework/pull/617/commits/f8bc8503300c1f9a66e7a18aa33b80ed7cf64d3b

yuderekyu: wait a sec....do we even need to toggle `aria-hidden` inside componentUpdate if we use `isOpen` inside the render function?

tbiethman: I don't think so. Remove all the things.

PR #619, opened 2019-04-17
[terra-date-time-picker]  Keep time updates while date is invalid.

### LINKED ISSUE #611: [terra-date-time-picker] Time updates lost if occured while date was invalid
# Bug Report

## Description
If date portion is invalid, updates to time are lost upon update date to a valid value.

## Steps to Reproduce
1. Enter a valid date/time: D;T
2. Update date to make it invalid (e.g., 01/1/2019): D`:T
3. Update time to a valid value: D`:T+1
4. Update date to a valid value: D+1:T+1
Date time is actually updated to D+1;T.  T+1 is not retained.

No screenshots. **Occurs more subtly when resetting date to the original value (D), as in this case onChange does not fire.  So, the underlying selected date is D;T but D;T+1 is displayed face up**.

image
(ok one screenshot)

## Expected Behavior
I expect the valid updates to time to be retained upon inputting a valid date.

## Possible Solution
I'm not sure. I think the component sometimes looks at date and time together (is this valid, should I apply an update), and sometimes separately (what changed?), and the complications that arise from that creates room for these sorts of gaps. In this instance, that 'what changed' (time) was valid, but date and time together were not.

## Environment
* Component Name and Version: terra-date-time-picker
* Browser Name and Version: chrome

## @ Mentions
@benbcai

benbcai:
### Summary
If the date portion is invalid, changes made to the time are no longer lost upon updating the date to a valid value.

### Additional Details
- Fixes #611 
- When the date is invalid, any updates made to the time is now preserved upon changing the date back to a valid value.
- onChange used to be triggered when there is only a valid date with no time value. Having a valid date without a time should not be considered a valid date-time value. Hence, onChange should not be triggered in this situation. With this change, onChange is now triggered only when both a valid date and valid time are present or when both the date and time are cleared.

@cerner/terra


packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -48,10 +48,15 @@ class DateTimeUtils {
...
+
+    if (!date.isValid()) {
+      return momentDate;
+    }
+
     let newDate = momentDate ? momentDate.clone() : date;
 
     // If momentDate was null, a new moment date needs to be created and sync'd with the entered time.
-    if (momentDate === null && time && time.length === 5) {
+    if (time && time.length === 5) {
```
emilyrohrbough: Should we update the inline comment since we no longer check if `momentDate === null`

benbcai: Definitely. https://github.com/cerner/terra-framework/pull/619/commits/3512b63fa2bb09f6ee9e79b42000edc697dc7237

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -216,17 +216,24 @@ class DateTimePicker extends React.Component {
...
 
-      if (DateTimeUtils.isValidTime(this.timeValue)) {
+      if (isTimeValid) {
+        // Update the timeValue in case the updatedDateTime falls in the missing hour and needs to bump the hour up.
         this.timeValue = DateTimeUtils.formatISODateTime(updatedDateTime.format(), 'HH:mm');
       }
     }
 
-    this.handleChange(event, updatedDateTime);
+    if ((isDateValid && isTimeValid) || (this.dateValue === '' && this.timeValue === '')) {
```
bjankord: May be good to add a comment here clarifying why were adding this logic.

benbcai: Added comment: https://github.com/cerner/terra-framework/pull/619/commits/0fdc394ea8acf93005f45918763597ecb9d98c90

mmalaker:
Functional verification completed. The date time examples now only update when a valid date and time are entered.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -250,13 +258,18 @@ class DateTimePicker extends React.Component {
         updatedDateTime.subtract(1, 'hours');
       }
 
-      this.timeValue = DateTimeUtils.formatISODateTime(updatedDateTime.format(), 'HH:mm');
+      if (updatedDateTime) {
+        this.timeValue = DateTimeUtils.formatISODateTime(updatedDateTime.format(), 'HH:mm');
```
emilyrohrbough: above we are subtracting 1 hour from the `updatedDateTime` value. Should that logic occur in this check or why do we need that check here? It might help to add an inline comment to describe what this check is validating. I think that it's a non-empty date ?

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -62,10 +62,15 @@ class DateTimeUtils {
    */
   static syncDateTime(momentDate, iOSdate, time) {
     const date = moment(iOSdate);
-    let newDate = momentDate ? momentDate.clone() : date;
 
-    // If momentDate was null, a new moment date needs to be created and sync'd with the entered time.
-    if (momentDate === null && time && time.length === 5) {
+    if (!date.isValid()) {
+      return momentDate;
```
emilyrohrbough: Should we be checking that the momentDate is valid here like we are below? `momentDate.isValid()`

packages/terra-date-time-picker/src/DateTimeUtils.js
emilyrohrbough: I'm a little confused, but its likely my lack of code knowledge for this module. 

We are updating to return the moment date if the generated iSOdate is invalid but shouldn't we still attempt to update the time regardless of if the iOSdate or momentDate's value if we are updating the time when date is invalid?

packages/terra-date-time-picker/src/DateTimePicker.jsx
benbcai: I added comment to indicate that the face-up time input value would need to be updated to account for the missing hour scenario. The check is to make sure we have a valid object before calling the format function on it.

https://github.com/cerner/terra-framework/pull/619/commits/784e52a725189be07a1cc7da27ca567b630385ed

packages/terra-date-time-picker/src/DateTimeUtils.js
benbcai: This is to return the passed in moment object in the event that the iSOdate is invalid. I have made some updates to this function. https://github.com/cerner/terra-framework/pull/619/commits/784e52a725189be07a1cc7da27ca567b630385ed. See above comment for details.

packages/terra-date-time-picker/src/DateTimeUtils.js
benbcai: I agree that _generally_ if the iSOdate is invalid, it should proceed to update the (base) moment object using the time. In our use case, if the iSOdate is invalid (the date input is cleared), the moment object in our state doesn't really care about the time. I have updated it to account for the general use case to proceed to update the time even if the date is invalid https://github.com/cerner/terra-framework/pull/619/commits/784e52a725189be07a1cc7da27ca567b630385ed

Issue #569, opened 2019-03-26
Fix color contrast issues in terra-application-links example


bjankord:
# Bug Report

## Description
Currently, the terra-application-links examples use demo colors for background and foreground colors which do not meet WCAG 2 AA recommended contrast ratios. We should update these demo examples to meet the recommended contrast ratio.

## Steps to Reproduce

1. Navigate to https://engineering.cerner.com/terra-framework/#/tests/terra-application-links/application-links/application-tabs-center
2. Run axe-core chrome extension
3. View axe-core report

## Additional Context / Screenshots
We'll want to make the demo colors themable so in other themes where we change the text color, (consumer theme), the background demo color can be themed for our wdio screenshot tests. You can [view this PR](https://github.com/cerner/terra-framework/pull/550/files) as an example of similar work we've done for other components.

PR #620, opened 2019-04-18
[terra-application-links] Update Color Contrast in Examples

amichaelparker:
### Summary
This updates the test examples in application-links to use themable background color styling, and updates that background color / makes the examples more uniform to address color contrast issues.

Closes [#569](https://github.com/cerner/terra-framework/issues/569).


emilyrohrbough:
Out of curiosity, how many components have custom demo styles that we need to theme now? Would it be worth push a few demo colors to doc template ?

bjankord:
@emilyrohrbough Yeah I think that would be a good addition to terra-doc-template.
I could see adding a dark container and a light container.

Issue #623, opened 2019-04-19
[terra-disclosure-manager] Prevent progressive disclosure from occurring if no key is provided during disclosure request


tbiethman:
# Feature Request

## Description
The disclosure manager's `disclose` API expects a unique key to be provided alongside the component to be disclosed. This key is used for internal mappings and bounds checking within the stack of disclosed components.

We do not currently check for the presence of that key during the disclose request. If a key is not provided, the disclose request should be rejected, and the user should be warned in development mode.

tbiethman:
Thought about this some more. I think we should remove the key from the API altogether and generate it internally. This will guarantee presence and uniqueness while making consumption easier for our consumers.  If we do this, then we wouldn't necessarily need a Major Version Bump; the key provided by consumers would just go unused.

PR #624, opened 2019-04-19
[terra-disclosure-manager] Explicitly documenting the required arguments for the disclose behavior

tbiethman:
### Summary
As of #583, the DisclosureManager will prevent dismissals of components disclosed without a key. A unique key is expected to be provided with the disclose request, per the documentation, and this was not expected to be a breaking change.

However, some consumers of the disclosure manager were **not** providing a key with their disclose requests. As of #583, this results in modals that can be opened but not closed. While I consider this a bug in their implementation, it is a regression in their functionality and leaves their applications in an unfortunate state.

This PR updates the documentation to be more explicit and removes the undefined check within the dismiss functions that is causing the above issue. I've logged #623 to better handle this in the future.

### DIFF
--- a/packages/terra-disclosure-manager/docs/README.md
+++ b/packages/terra-disclosure-manager/docs/README.md
@@ -154,18 +154,18 @@ disclosureManager.disclose({
 
 `disclose` Argument API:
 
-|Key|Value|
-|---|---|
-|`preferredType`|The String describing the preferred disclosure type. This will be used to match the disclosure request to an appropriate DisclosureManager. Depending on the structuring of managers in a given component, the `preferredType` value may not be honored. If the provided `preferredType` is not supported by any present disclosure managers, the root disclosure manager will perform the disclosure using its own disclosure type, regardless of the `preferredType` value.
-|`size`|The String size desired for the disclosure. One of `tiny`, `small`, `medium`, `large`, `huge`, or `fullscreen`. The functional implementation of this size is determined by the rendering component.|
-|`content`|An Object containing a key and a component describing the component to be disclosed. See the `content` API below.|
+|Key|Is Required|Value|
+|---|---|---|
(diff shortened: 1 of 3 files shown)


packages/terra-disclosure-manager/docs/README.md
```diff
@@ -154,18 +154,18 @@ disclosureManager.disclose({
...
-|`size`|The String size desired for the disclosure. One of `tiny`, `small`, `medium`, `large`, `huge`, or `fullscreen`. The functional implementation of this size is determined by the rendering component.|
-|`content`|An Object containing a key and a component describing the component to be disclosed. See the `content` API below.|
+|Key|Is Required|Value|
+|---|---|---|
+|`preferredType`|optional|The String describing the preferred disclosure type. This will be used to match the disclosure request to an appropriate DisclosureManager. Depending on the structuring of managers in a given component, the `preferredType` value may not be honored. If the provided `preferredType` is not supported by any present disclosure managers, the root disclosure manager will perform the disclosure using its own disclosure type, regardless of the `preferredType` value.|
+|`size`|optional|The String size desired for the disclosure. One of `tiny`, `small`, `medium`, `large`, `huge`, or `fullscreen`. The functional implementation of this size is determined by the rendering component. If not provided, a default size will be used.|
```
nramamurth: I think `dimensions` should also be documented. 

https://github.com/cerner/terra-framework/blob/ac87285ac89dbc005e5f3483c0cc863585bd0191/packages/terra-disclosure-manager/src/DisclosureManager.jsx#L245

emilyrohrbough:
@tbiethman is this a MVB?

tbiethman:
No MVB is necessary because there's no change in API or behavior here. Anyone not providing a key will get the same experience they had before #583 was released. I'm updating the docs to be crystal clear that a key *should* be provided, because it will be absolutely required after #623 is complete.

packages/terra-disclosure-manager/docs/README.md
tbiethman: you got it 1b87cb7

PR #626, opened 2019-04-19
Moved packages we use solely in doc examples and tests to devDependencies

bjankord:
### Summary
Phase 1 of work for https://github.com/cerner/terra-ui/issues/165


packages/terra-aggregator/package.json
```diff
@@ -26,8 +26,9 @@
     "react-dom": "^16.8.5"
   },
   "dependencies": {
-    "classnames": "^2.2.5",
```
bjankord: classnames is not used in terra-aggregator so I've removed it altogether from the package.json file.

packages/terra-abstract-modal/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ Changelog
 Unreleased
 ----------
 
+Unreleased
```
emilyrohrbough: Whoops added Unreleased

packages/terra-slide-panel-manager/package.json
```diff
@@ -29,14 +29,16 @@
   "dependencies": {
     "classnames": "^2.2.5",
     "prop-types": "^15.5.8",
+    "terra-disclosure-manager": "^4.8.0",
```
emilyrohrbough: Should the disclosure manager be a peer dep?

packages/terra-abstract-modal/CHANGELOG.md
bjankord: De-duped in this commit: https://github.com/cerner/terra-framework/pull/626/commits/b2ba9614b3f3761e7459d2f7efe6d5a9cf3d1d87

packages/terra-slide-panel-manager/package.json
bjankord: Not sure, I'll defer to @tbiethman on this

tbiethman: It should be, but moving it to peer now won't bepassive. I'll get a PR out to make it a peer dependency (and the one in terra-modal-manager) that we can time with another major release.

tbiethman: Basically any component that provides Context-based APIs should be a peer dependency in non-application component packages. I want to warn on this at build-time with a webpack plugin or something; I might work on that over my sabbatical.

PR #634, opened 2019-04-25
Replace terra-disclosure-manager dependencies with peerDependencies

tbiethman:
### Summary
The terra-modal-manager and terra-slide-panel-manager packages depend on the terra-disclosure-manager. However, those packages specify dependencies, not peerDependencies, which makes aligning on a single terra-disclosure-manager version difficult for consumers.

With this change, consumers will be responsible for installing the appropriate version of the terra-disclosure-manager for these packages. If they don't, they will get warned by npm during installation (assuming the dependency isn't supplied transitively, we cannot do much about that).


emilyrohrbough:
We should update the docs for these components to describe how these peer dependencies should be added to projects vs applications.

tbiethman:
@bjankord We may as well since it's already getting a major bump. I'll get that changed here.

@emilyrohrbough The documentation would need to go on the terra-disclosure-manager (and anything else providing context-based APIs, like terra-breakpoints, terra-navigation-prompt, etc.). The basic gist of the messaging could be:

"This package uses React's Context for inter-component communication. A single instance of this package must be installed to ensure that communication occurs between all Context Providers and Consumers. To help ensure a single instance is installed, this package should be defined as a peer dependency in packages that are reusable libraries or otherwise consumed by other packages."

Does that make sense?

tbiethman:
With f92aeca, the terra-application-layout has been updated, and the terra-disclosure-manager docs have been updated to include the blurb above.

PR #635, opened 2019-04-25
[terra-form-validation] Update form validation docs, add FormValidationUtil, mark package as public.

BenBoersma:
### Summary
Added a new example that shows how to build a form that validates.
Added a numeric validation example that checks minimum value, maximum value, and the value for up to a certain level of precision.
Added FormValidationUtil class which has some basic validation checks such as minimum length and maximum length.

@cerner/terra


packages/terra-form-validation/package.json
```diff
@@ -3,7 +3,6 @@
   "main": "docs/README.md",
```
bjankord: Let's move [final-form](https://github.com/cerner/terra-framework/blob/e1ac6427b914cb570d40b4fc2d91279ef473c50d/packages/terra-form-validation/package.json#L29) and [react-final-form](https://github.com/cerner/terra-framework/blob/e1ac6427b914cb570d40b4fc2d91279ef473c50d/packages/terra-form-validation/package.json#L31) to be both a peerDependency as well as devDependency

BenBoersma: Added here https://github.com/cerner/terra-framework/pull/635/commits/8e55e43d8a7dcc9d77396ff2e45042877751fc94

packages/terra-form-validation/docs/MaxLengthInput.md
```diff
@@ -0,0 +1,234 @@
...
+    >
+      ...
+    </Field>
+    <Button text="Submit" type={Button.Opts.Types.SUBMIT} />
+    </form>
+  );
+}
+```
+
+And with that we have created an input that is required and has max length of 7 characters. This showed how to make the validate function by yourself, but if we wanted to we could shorten the code a bit more under the validateLength function through the use of `terra-form-validation`'s FormValidationUtil class like so.
```
yuderekyu: maybe something like
```suggestion
We have created an input that is required and has max length of 7 characters by creating our own validate function. For ease of use, Terra provides utils for common validation use cases. In this case, we can use the `validateLength` function.
```

packages/terra-form-validation/docs/NumericValidation.md
```diff
@@ -0,0 +1,169 @@
...
+          </div>
+          )
+        }
+      </Spacer>
+    );
+  }
+}
+```
+
+With this, we have a pretty simple number type input with a basic validation tied to the `<Form>` component that checks to see whether or not we have a value for the input.
```
yuderekyu:
```suggestion
We have a simple number type input with validation, tied to the `<Form>` component, that checks whether we have a value for the input.
```

packages/terra-form-validation/docs/NumericValidation.md
```diff
@@ -0,0 +1,169 @@
...
+
+We also have our `<Field>`'s validation function defined and ready to be added to, with it only having a basic check to see if there is a value.
+
+Because we have added to the `inputAttrs` prop  of `<InputField>` the value of `type: 'number'` this also means we have a bit of browser enforcement on the inputted value. Mainly that it has to be a number of some sort so it only accepts certain characters.
```
yuderekyu:
```suggestion
We have added browser enforcement on the inputted value by passing `type: 'number'` into `<Inputfield>`'s `inputAttrs`.
```

packages/terra-form-validation/docs/NumericValidation.md
```diff
@@ -0,0 +1,169 @@
...
+
+A thing of note is that even though we specified the input as a number type, it will still take some 'wrong' values and place them in the field but give them as `undefined`. Specifically the input will accept values such as `--1` or `2.0.32` but then when the validation functions run the value given to them will be `undefined`. With our code is currently setup that means we will just return a value of `Required` via the `<Form>`'s validation function.
```
yuderekyu:
```suggestion
Note that specifying `type: 'number'` still allows negatives and decimal points, such as `--1` or `2.0.32`. These will be placed in the field as `undefined`. With our code, we will just return a value of `Required`, via the `<Form>`'s validation function, if values are undefined.
```

packages/terra-form-validation/docs/NumericValidation.md
```diff
@@ -0,0 +1,169 @@
...
+
+The 2nd thing of note is that our validation function, `validateNumber`, will be given it's argument in the form of a string despite it being a number type input. This doesn't impact simple operators too much like `<` and `>` but it is information you should know for when your building custom validation functions.
```
yuderekyu:
```suggestion
Second thing - our validation function, `validateNumber`, will be given a *string*, despite the input being of `type: 'number'`. This does not impact simple operators but is useful to know when building custom validation functions.
```

packages/terra-form-validation/docs/NumericValidation.md
```diff
@@ -0,0 +1,169 @@
...
+
+With this in mind, we'll first start by giving a minimum and maximum to our value. For simplicity's sake lets go with a minimum of 10 and a maximum of 100.
```
yuderekyu:
```suggestion
Let's start by giving a minimum and maximum to our value. For simplicity, set a minimum to 10 and a maximum to 100.
```

packages/terra-form-validation/docs/NumericValidation.md
```diff
@@ -0,0 +1,169 @@
...
+  }
+  if (value > 100) {
+    return 'Value should not be over 100.';
+  }
+
+  return undefined;
+};
+```
+
+Now for the more complicated validation which is the level of precision we want in our input. For the example let us say that we want 3 decimals worth of precision and we don't want to accept more than that. There are multiple ways of validating this but since our value is a string, we'll use string functions to validate.
```
yuderekyu:
```suggestion
Now, let's check for precision. For this example, lets validate to the thousandth place (three decimal places). Since our value is a string, we'll use string functions to validate.
```

packages/terra-form-validation/docs/NumericValidation.md
```diff
@@ -0,0 +1,169 @@
...
+
+We will split the given string on its `.` character, and then check to make sure the 2nd string we get from doing this has a length of 3 or less.
```
yuderekyu:
```suggestion
Use `.` as a delimiter and check for a length less than or equal to 3.
```

packages/terra-form-validation/docs/NumericValidation.md
```diff
@@ -0,0 +1,169 @@
...
+  const valueSplit = value.split('.');
+  if (valueSplit.length === 2 && valueSplit[1].length > 3) {
+    return 'Value has more than 3 decimal points';
+  }
+
+  return undefined;
+};
+```
+
+This way of validating for precision has a few quirks with it, mainly that if we give it a value of `23.023000` it would fail the validation, because of the extra `0`s at the end despite this number effectively being the same as `23.023`.
```
yuderekyu: This validation has a quirk. If the input contains extra `0`s, like `23.0230000`, validation will fail.

packages/terra-form-validation/docs/NumericValidation.md
```diff
@@ -0,0 +1,169 @@
...
+
+For the sake of the example lets try a different way of validating. Let us, instead of using string function to validate, convert the value given to us to a `Number` instead. With it being a number we can use the `Number` class function `toFixed` which allows us to automatically take a value to a given precision.
```
yuderekyu:
```suggestion
Lets try a different way of validating. Convert the value a `Number` instead. This way we can use the `toFixed` function to automatically set precision, and use this as a comparison.
```

packages/terra-form-validation/docs/NumericValidation.md
```diff
@@ -0,0 +1,169 @@
...
+  const valueNum = Number(value);
+  if (valueNum.toFixed(3) != valueNum) {
+    return 'Value has more than 3 decimal points';
+  }
+
+  return undefined;
+};
+```
+
+With this we compare the value given us to same value to a given precision. Which means if we are handed `45.4321`, we will compare it to `45.432`, which is not equal and will return the failure message. The quirk with this validation is that if you give it `23.023000` it will succeed because it will be compared to `23.023` and be found equal thus you probably will need to include something to shorten the String value in this case.
```
yuderekyu: This validation has the opposite quirk. If the input contains extra 0s, like 23.0230000, validation will succeed.

packages/terra-form-validation/docs/MaxLengthInput.md
```diff
@@ -0,0 +1,234 @@
...
+
+As it currently stands this render would render nothing so we need to start filling out the `<Form>` component with some props to render what we need, what we want to happen on submit, initial values for our field(if we want), and our validate function.
+
+We are going to add the `render` prop to `<Form>` first, but before we do that we need to define what we want the `<Form>` to render so we're going to define a `renderForm` function. All this function will do is return a `<form>` tag containing a `<Field>` component, terra's `<InputField>` component, and a `<Button>` to be used to submit the form.
+```javascript
+render() {
+  return (
+    <Spacer marginBottom="small">
+      <Form
+        render={this.renderForm} // added line
```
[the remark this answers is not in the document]
BenBoersma: Fixed https://github.com/cerner/terra-framework/pull/635/commits/a0241f8008eacb2cb9bba04c762e6018716f0ff5

packages/terra-form-validation/docs/MaxLengthInput.md
```diff
@@ -0,0 +1,234 @@
...
+        )
+      }
+    </Spacer>
+  );
+}
+```
+
+Submit will now visually update with a JSON of our `submittedValues` state.
+
+Next is to set up an initial value for our field. This is fairly simple and just requires us to add a prop called `initialValues` to our `<Form>` component that has an object defining our values in properties with the property being the inputID of our `<InputField>`, which in this case is `example`, and the property's value being the value of the `<InputField>`.
```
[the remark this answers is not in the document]
BenBoersma: Attempted to fix https://github.com/cerner/terra-framework/pull/635/commits/a0241f8008eacb2cb9bba04c762e6018716f0ff5

packages/terra-form-validation/docs/NumericValidation.md
BenBoersma: Swapped https://github.com/cerner/terra-framework/pull/635/commits/a0241f8008eacb2cb9bba04c762e6018716f0ff5

packages/terra-form-validation/docs/NumericValidation.md
BenBoersma: Swapped https://github.com/cerner/terra-framework/pull/635/commits/a0241f8008eacb2cb9bba04c762e6018716f0ff5

BenBoersma:
Added references to the new functions inside the `Validation Functions` page and switched all the examples to use `diff` markdown

packages/terra-form-validation/docs/MaxLengthInput.md
```diff
@@ -0,0 +1,232 @@
...
+      </Spacer>
+    );
+  }
+}
+```
+
+Current we render nothing. Lets create a `<Form>` component that contains initial values, and functionality for submission and validation.
+
+Let's use the render props pattern to return our necessary components. Call this function `renderForm`.
+```javascript
```
yuderekyu: in order to highlight additions and removals, update this to diff. other code blocks need this too
```suggestion
```diff
```

BenBoersma: Fixed now https://github.com/cerner/terra-framework/pull/635/commits/a0efa6047e843fd48aa3b656d2583afd0c9d15c2

packages/terra-form-validation/docs/MaxLengthInput.md
```diff
@@ -0,0 +1,232 @@
+# Making an Input With a Maximum Length
+
+This is a step-by-step guide to making a form with a single input that has a maximum length as its validations.
```
ryanthemanuel: "step-by-step guide to make"

packages/terra-form-validation/src/FormValidationUtil.js
```diff
@@ -0,0 +1,80 @@
...
+/* eslint-disable no-restricted-globals */
+
+class FormValidationUtil {
+  /* Checks given value's length against maximum. Returns true if value's length is less than the maximum.
+   * @value is the value string, @maximum is a nonnegative whole number. */
+  static isUnderMaxLength(value, maximum) {
+    if (!(typeof value === 'string' || value instanceof String)
+    || !(typeof maximum === 'number' && isFinite(maximum))
+    || maximum < 0 || !Number.isInteger(maximum)) { return false; }
+    return value.length <= maximum;
```
benbcai: Should this be < instead of <= based on the function description?

packages/terra-form-validation/src/terra-dev-site/doc/example/MaxLengthInput.jsx
```diff
@@ -0,0 +1,90 @@
...
+import React from 'react';
+import { Form, Field } from 'react-final-form';
+import InputField from 'terra-form-input/lib/InputField';
+import Button from 'terra-button';
+import Spacer from 'terra-spacer';
+import FormValidationUtil from '../../../FormValidationUtil';
+
+const validateLength = (name) => {
+  if (!FormValidationUtil.isUnderMaxLength(name, 8)) {
+    return 'Name needs to be less than 8 characters long';
```
benbcai: Should this say "...less than 9 characters..."? The number (8) in the return string is inconsistent with the return string in the validateLength function in [packages/terra-form-validation/docs/MaxLengthInput.md](https://github.com/cerner/terra-framework/pull/635/files#diff-00d3067043a9948e4371f570ef81a357R227). One passes in 8 while the other passes in 7 but the number in the return string is still 8.

packages/terra-form-validation/src/FormValidationUtil.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /* Checks given value's length against minimum. Returns true if value's length is greater than the minimum.
+   * @value is the value string, @minimum is a nonnegative whole number. */
+  static isOverMinLength(value, minimum) {
+    if (!(typeof value === 'string' || value instanceof String)
+    || !(typeof minimum === 'number' && isFinite(minimum))
+    || minimum < 0 || !Number.isInteger(minimum)) { return false; }
+    return value.length >= minimum;
```
benbcai: Should this be > instead of >= based on the function description?

packages/terra-form-validation/src/FormValidationUtil.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /* Checks given value for extraneous whitespace. return true if there is no whitespace.
+   * @value is the value string */
+  static hasNoWhitespace(value) {
+    if (!(typeof value === 'string' || value instanceof String)) { return false; }
+    return value.trim() === value;
```
benbcai: If this function only intends to check for leading or trailing whitespaces, then it would work. If it should also check for white spaces in between characters, then a value of `white space` would return false. This function could call containsCharacters(value, ' ').

packages/terra-form-validation/src/FormValidationUtil.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /* Checks given value for any string in given array. return true if value does not contain any of them.
+   * @value is the value string, @chars is an array of strings  */
+  static containsCharacters(value, chars) {
+    if (!(typeof value === 'string' || value instanceof String)
+    || !(chars && typeof chars === 'object' && chars.constructor === Array)) { return false; }
+    for (let i = 0; i < chars.length; i += 1) {
+      if (value.includes(chars[i])) {
+        return false;
```
yuderekyu: shouldn't we flip the return statements?  this function is called containsCharacters, so i'd expect it to return true if the value contains one of the strings, not false

packages/terra-form-validation/docs/MaxLengthInput.md
```diff
@@ -0,0 +1,232 @@
...
+
+First we start with out imports
+```javascript
+import React from 'react';
+import { Form, Field } from 'react-final-form';
```
bjankord: Might be good to cover running `npm install final-form react-final-form` before the imports.

packages/terra-form-validation/docs/MaxLengthInput.md
```diff
@@ -0,0 +1,232 @@
...
+    return (
+      <Spacer marginBottom="small">
+        <Form/>
+      </Spacer>
+    );
+  }
+}
+```
+
+Current we render nothing. Lets create a `<Form>` component that contains initial values, and functionality for submission and validation.
```
bjankord: nit: `Currently, this does not render anything.`

packages/terra-form-validation/docs/MaxLengthInput.md
```diff
@@ -0,0 +1,232 @@
...
+const validateLength = (name) => {
+  if (name && name.length > 7) {
+    return 'Name needs to be less than 8 characters long';
+  }
+
+  return undefined;
+};
+```
+
+Finally, let's add pass the function as the `validate` prop to `<Field>` within our `renderForm` function..
```
bjankord: nit: Let's go with `Finally, let's add the function as the` and remove the double period at the end of the sentence.

packages/terra-form-validation/docs/MaxLengthInput.md
```diff
@@ -0,0 +1,232 @@
...
+
+```javascript
+submitForm(values) {
+  this.setState({
+    submittedValues: values,
+  });
+}
+```
+
+Bind `onSubmit`, pass `onSubmit` as a prop to `<Form>`, pass `handleSubmit` to our render prop function, and pass `handleSubmit` to the native form's `onSubmit` prop.
```
bjankord: It feels like `handleSubmit` needs to be defined. I didn't see any code for it in the step-by-step guide.

packages/terra-form-validation/docs/NumericValidation.md
```diff
@@ -0,0 +1,177 @@
...
+Note that specifying `type: 'number'` still allows negatives and decimal points, such as `--1` or `2.0.32`. These will be placed in the field as `undefined`. With our code, we will just return a value of `Required`, via the `<Form>`'s validation function, if values are undefined.
+
+Second thing - our validation function, `validateNumber`, will be given a *string*, despite the input being of `type: 'number'`. This does not impact simple operators but is useful to know when building custom validation functions.
+
+Let's start by giving a minimum and maximum to our value. For simplicity, set a minimum to 10 and a maximum to 100.
+
+```diff
+const validateNumber = (value) => {
+  if (!value) { return undefined; }
++ if (value < 10) {
```
bjankord: nit: I feel like we should be casting `value` to a number data type here: `if (Number(value) < 10)` even though this will work without it.

packages/terra-form-validation/docs/NumericValidation.md
```diff
@@ -0,0 +1,177 @@
...
++   return 'Value has more than 3 decimal points';
++ }
+
+  return undefined;
+};
+```
+
+This validation has a quirk. If the input contains extra `0`s, like `23.0230000`, validation will fail.
+
+Lets try a different way of validating. Convert the value a `Number` instead. This way we can use the `toFixed` function to automatically set precision, and use this as a comparison.
```
bjankord: 👍

packages/terra-form-validation/docs/NumericValidation.md
```diff
@@ -0,0 +1,177 @@
...
+- if (valueNum.toFixed(3) != valueNum) {
++ if (!FormValidationUtil.isPrecise(value, 3)) {
+    return 'Value has more than 3 decimal points';
+  }
+
+  return undefined;
+};
+```
+
+Note: `FormValidationUtil.precisionCheck` uses the first method check precision.
```
bjankord: Hmm, Im not sure what this means.

packages/terra-form-validation/docs/MaxLengthInput.md
BenBoersma: Fixed here https://github.com/cerner/terra-framework/pull/635/commits/415c46ab9a0b74893a686c38430b1d8decd708ad#diff-00d3067043a9948e4371f570ef81a357R3

packages/terra-form-validation/docs/MaxLengthInput.md
BenBoersma: Fixed here https://github.com/cerner/terra-framework/pull/635/commits/415c46ab9a0b74893a686c38430b1d8decd708ad#diff-00d3067043a9948e4371f570ef81a357R5

packages/terra-form-validation/docs/MaxLengthInput.md
BenBoersma: Fixed here https://github.com/cerner/terra-framework/pull/635/commits/415c46ab9a0b74893a686c38430b1d8decd708ad#diff-00d3067043a9948e4371f570ef81a357R35

packages/terra-form-validation/docs/MaxLengthInput.md
BenBoersma: Fixed here https://github.com/cerner/terra-framework/pull/635/commits/415c46ab9a0b74893a686c38430b1d8decd708ad#diff-00d3067043a9948e4371f570ef81a357R201

packages/terra-form-validation/docs/NumericValidation.md
BenBoersma: Fixed here https://github.com/cerner/terra-framework/pull/635/commits/415c46ab9a0b74893a686c38430b1d8decd708ad#diff-efb0b388a04429df36f1ce4eb79ef872R108
The other references to this code later in the example should also be fixed.

packages/terra-form-validation/docs/NumericValidation.md
BenBoersma: Cleared it up here https://github.com/cerner/terra-framework/pull/635/commits/415c46ab9a0b74893a686c38430b1d8decd708ad#diff-efb0b388a04429df36f1ce4eb79ef872R108

packages/terra-form-validation/src/FormValidationUtil.js
BenBoersma: I meant for it to be <= since its a `maximum` value which usually means the value can be the maximum amount. Fixed the description to show this https://github.com/cerner/terra-framework/pull/635/commits/415c46ab9a0b74893a686c38430b1d8decd708ad#diff-efb0b388a04429df36f1ce4eb79ef872R108

packages/terra-form-validation/src/FormValidationUtil.js
BenBoersma: Same deal with the maximum comment. Meant for the minimum to be an acceptable value, fixed description here https://github.com/cerner/terra-framework/pull/635/commits/415c46ab9a0b74893a686c38430b1d8decd708ad#diff-efb0b388a04429df36f1ce4eb79ef872R108

packages/terra-form-validation/src/FormValidationUtil.js
BenBoersma: Fixed it here https://github.com/cerner/terra-framework/pull/635/commits/415c46ab9a0b74893a686c38430b1d8decd708ad#diff-e0dd485bcf22c1ac58c3d5cfd0103965R27
As per offline discussion instead of using containsCharacters, using a Regex test to see if we have anything that counts as a white space character.

packages/terra-form-validation/src/terra-dev-site/doc/example/MaxLengthInput.jsx
BenBoersma: It should say `less than 9`, changed all the supporting text and said 8, and fixed the examples that used to say 7 to now say 8. https://github.com/cerner/terra-framework/pull/635/commits/415c46ab9a0b74893a686c38430b1d8decd708ad

packages/terra-form-validation/src/FormValidationUtil.js
BenBoersma: Yep point, fixed here https://github.com/cerner/terra-framework/pull/635/commits/415c46ab9a0b74893a686c38430b1d8decd708ad#diff-e0dd485bcf22c1ac58c3d5cfd0103965R37, and swapped the jest tests around.

Issue #456, opened 2019-01-29
Add a Date/Time Field


ryanthemanuel:
# Feature Request

## Description

The Terra date/time picker is sufficiently complex that there is a need to create a common encapsulation around it that supports various field type capabilities.  These would potentially include (but are not limited to):

- Functionality supported by terra-field
- Validation
- Masking

The goal here would be to provide default implementations for the common cases.

gerard-isdell:
Is there an equivalent for the combined control -  DateTimePicker?

leducquan:
Do we have a target date for this issue to be completed?

ryanthemanuel:
Splitting out field to be done with this: https://github.com/cerner/terra-framework/issues/1228

PR #608, opened 2019-04-15
Invert abstract modal focus trap to mitigate issues with focus management in portals

bjankord:
# Major version bump to various framework components
* Resolves https://github.com/cerner/terra-core/issues/2330
* Resolves part of https://github.com/cerner/terra-framework/issues/527 (date-picker rendered in overlay now renders in a react portal)

## These changes will trigger major version bumps in the following packages:
- terra-abstract-modal
- terra-application-layout (via popup dependency MVB)
- terra-application-links (via popup dependency MVB)
- terra-collapsible-menu-view (via menu dependency MVB)
- terra-date-picker
- terra-date-time-picker
- terra-dialog-modal
- terra-menu (via popup dependency MVB))
- terra-notification-dialog
- terra-popup
- terra-tabs (via menu dependency MVB)

## These changes will also trigger minor version bumps for the following:
- terra-aggregator
- terra-modal-manager
- terra-slide-panel-manager

### Abstract Modal, Modal Manager, Dialog Modal, Notification Dialog opens
* Focus allowed within modal but not the rest of the app.
#### Test URLS
* https://terra-framework-deploye-pr-608.herokuapp.com/#/tests/terra-abstract-modal/abstract-modal/abstract-modal-is-open
* https://terra-framework-deploye-pr-608.herokuapp.com/#/tests/terra-modal-manager/modal-manager/modal-manager-default
* https://terra-framework-deploye-pr-608.herokuapp.com/#/tests/terra-dialog-modal/dialog-modal/default-dialog-modal
* https://terra-framework-deploye-pr-608.herokuapp.com/#/tests/terra-notification-dialog/notification-dialog/complete-notification-dialog
### Select opened inside of Modal Manager or Dialog Modal
* Focus allowed to move within select
#### Test URLS
* https://terra-framework-deploye-pr-608.herokuapp.com/#/tests/terra-modal-manager/modal-manager/modal-manager-default
* https://terra-framework-deploye-pr-608.herokuapp.com/#/tests/terra-dialog-modal/dialog-modal/diaglog-modal-with-select
### Datepicker opened inside of Modal Manager or Dialog Modal
* Focus allowed to move within Date Picker
#### Test URLS
* https://terra-framework-deploye-pr-608.herokuapp.com/#/tests/terra-date-picker/date-picker/date-picker-inside-modal
* https://terra-framework-deploye-pr-608.herokuapp.com/#/tests/terra-dialog-modal/dialog-modal/dialog-modal-with-date-picker
[...]

PR #642, opened 2019-04-30
Add DatePickerField

jmsv6d:
### Summary
Addresses #456 Will follow with a separate PR for TimeInputField

### Deployment link
https://terra-framework-deploye-pr-642.herokuapp.com/#/components/terra-date-picker/date-picker/date-picker-field

@cerner/terra


packages/terra-date-picker/docs/DatePickerField.md
```diff
@@ -0,0 +1,59 @@
...
+    filterDate={isWeekday}
+    name="admission-date"
+    label="Admission Date"
+    required
+  />
+</Base>
+
+
+const isWeekday = (date) => {
+  const day = date.day();
```
benbcai: #590 changes the parameter in filterDate from a moment object to be an ISO string. This should be updated to something like [this](https://github.com/cerner/terra-framework/pull/590/files#diff-4d64f16fdc0d286ca4cfba4a4e10601b). If we wanted to, we could just remove the illustration of filterDate in this example code to minimize maintenance.

packages/terra-date-picker/docs/DatePickerField.md
```diff
@@ -0,0 +1,59 @@
...
+};
+```
+
+## Implementation Notes:
+DatePickerField is required to be composed inside the [Base][6] component with locale in order for it to load the correct date format and translation strings.
+
+The [Modal][4] component by default will trap focus. In order to interact with the date picker when rendered inside a modal, the date picker must request focus from the modal. The [App Delegate][5] in the [Modal Manager][3] component provides support for requesting and releasing focus from the modal. Therefore, the [Modal Manager][3] component must be used when rendering the date picker in a modal. When composing the date picker in the [Modal Manager][3], pass in the requestFocus and releaseFocus functions in the [App Delegate][5] to the requestFocus and releaseFocus props in the date picker. The [Modal Manager][3] will facilitate requesting and releasing focus when the picker opens and closes. Take a look at this example for composing a [modal manager][7] with a date picker as the [content][8].
+
+[1]: https://github.com/Hacker0x01/react-datepicker
+[2]: http://momentjs.com/docs/
```
benbcai: Links 1 and 2 can be removed since they are no longer being referenced here.

packages/terra-date-picker/docs/DatePickerField.md
emilyrohrbough: +1 to removing hard coded example

packages/terra-date-picker/src/DatePickerField.jsx
```diff
@@ -0,0 +1,248 @@
...
+   */
+  onSelect: PropTypes.func,
+  /**
+   * A callback function to let the containing component (e.g. modal) to regain focus.
+   */
+  releaseFocus: PropTypes.func,
+  /**
+   * A callback function to request focus from the containing component (e.g. modal).
+   */
+  requestFocus: PropTypes.func,
```
benbcai: The `releaseFocus` and `requestFocus` are being removed in [#608](https://github.com/cerner/terra-framework/pull/608/files#diff-06f07d3cca7012107347dc63be00c147). These two props will no longer be needed here.

packages/terra-date-picker/src/DatePickerField.jsx
```diff
@@ -0,0 +1,248 @@
...
+   */
+  hideRequired: PropTypes.bool,
+  /**
+   * Whether or not the field is invalid.
+   */
+  isInvalid: PropTypes.bool,
+  /**
+   * Whether or not the field is an inline field.
+   */
+  isInline: PropTypes.bool,
```
benbcai: nit: `isInline` should be placed before`isInvalid`.

packages/terra-date-picker/docs/DatePickerField.md
```diff
@@ -0,0 +1,21 @@
...
+[terra-date-picker](https://github.com/cerner/terra-framework/tree/master/packages/terra-date-picker) component that is wrapped inside a
+[terra-form-field](https://github.com/cerner/terra-core/tree/master/packages/terra-form-field) component.
+
+## Getting Started
+
+- Install with [npmjs](https://www.npmjs.com):
+  - `npm install terra-date-picker`
+  - `yarn add terra-date-picker`
+
+
```
emilyrohrbough: nit: extra line return

packages/terra-date-picker/src/DatePickerField.jsx
```diff
@@ -0,0 +1,234 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Field from 'terra-form-field';
+import DatePicker from './DatePicker';
+
```
emilyrohrbough: nit: extra line return

packages/terra-date-picker/src/DatePickerField.jsx
```diff
@@ -0,0 +1,234 @@
...
+  filterDate: undefined,
+  help: null,
+  hideRequired: false,
+  includeDates: undefined,
+  inputAttributes: undefined,
+  isInline: false,
+  isInvalid: false,
+  isLabelHidden: false,
+  labelAttrs: {},
+  maxDate: undefined,
```
emilyrohrbough: do we need to explicitly set default prop values of undefined?

packages/terra-date-picker/src/DatePickerField.jsx
```diff
@@ -0,0 +1,234 @@
...
+const defaultProps = {
+  disabled: false,
+  error: null,
+  errorIcon: Field.defaultProps.errorIcon,
+  excludeDates: undefined,
+  filterDate: undefined,
+  help: null,
+  hideRequired: false,
+  includeDates: undefined,
+  inputAttributes: undefined,
```
emilyrohrbough: should inputAttributes be `{}`?

packages/terra-date-picker/src/DatePickerField.jsx
```diff
@@ -0,0 +1,234 @@
...
+  /**
+   * An array of ISO 8601 string representation of the dates to enable in the picker. All Other dates will be disabled.
+   */
+  includeDates: PropTypes.arrayOf(PropTypes.string),
+  /**
+   * Custom input attributes to apply to the date input. Use the name prop to set the name for the input.
+   * Do not set the name in inputAttribute as it will be ignored.
+   */
+  // eslint-disable-next-line react/forbid-prop-types
+  inputAttributes: PropTypes.object,
```
emilyrohrbough: Should we use `inputAttrs` to match other form-field wrappers? 

https://github.com/cerner/terra-core/blob/master/packages/terra-form-input/src/InputField.jsx#L46

packages/terra-date-picker/src/terra-dev-site/test/date-picker/DisabledDatePickerField.test.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import DatePickerField from '../../../../lib/DatePickerField';
```
emilyrohrbough: I think these imports should just be: `'../../../DatePickerField';`

@dkasper-was-taken @mjhenkes

packages/terra-date-picker/tests/wdio/date-picker-spec.js
```diff
@@ -320,4 +320,31 @@ describe('Date Picker', () => {
...
+      Terra.should.validateElement({ axeRules: { rules: ignoredA11y } });
+    });
+
+    describe('Invalid DatePickerField', () => {
+      before(() => {
+        browser.url('/#/raw/tests/terra-date-picker/date-picker/date-picker-field');
+        browser.click('#validity-toggle');
+      });
+
+      Terra.should.validateElement({ axeRules: { rules: ignoredA11y } });
```
emilyrohrbough: With this new component, can we remove the disabled axe rule `label: { enabled: false },`?

packages/terra-date-picker/src/DatePickerField.jsx
jmsv6d: I used `inputAttributes` here because it's the name of the prop on the DatePicker

packages/terra-date-picker/src/DatePickerField.jsx
jmsv6d: I don't think that any of these default props really need to be set. They were just copied from field and date-picker default props, I was following the pattern of what was done for the other field components.

packages/terra-date-picker/tests/wdio/date-picker-spec.js
jmsv6d: I don't think that the DatePickerField component should be used in the DatePicker tests, but they could be updated to have aria-labels to be able to re-enable that rule.

jmsv6d: Actually, looking at the test failures when I enable that rule, there a11y failures are within the date picker calendar, for the year and month select components. We should probably log a separate issue for that.

emilyrohrbough: @jmsv6d ah makes sense

packages/terra-date-picker/src/DatePickerField.jsx
bjankord: We've used both [inputAttrs](https://github.com/cerner/terra-core/blob/master/packages/terra-form-input/src/InputField.jsx#L46) and [inputAttributes](https://github.com/cerner/terra-core/blob/master/packages/terra-search-field/src/SearchField.jsx#L79) on different components. I like `inputAttrs` but I think consumers aren't really sure what it is used for so I lean towards using `inputAttributes` going forward as it is more explicit what the prop is used for.

benbcai: Here is the original comment on deciding on the name for this prop and decided to go with `inputAttributes`: https://github.com/cerner/terra-core/pull/292#discussion_r114352918

packages/terra-date-picker/src/DatePickerField.jsx
```diff
@@ -0,0 +1,237 @@
...
+   */
+  minDate: PropTypes.string,
+  /**
+   * Name of the date input. The name should be unique.
+   */
+  name: PropTypes.string.isRequired,
+  /**
+   * A callback function to execute when a valid date is selected or entered.
+   * The first parameter is the event. The second parameter is the changed date value.
+   */
```
benbcai: Please include the [onBlur](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/src/DatePicker.jsx#L52) and [onFocus](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/src/DatePicker.jsx#L71) props that were recently merged.

jmsv6d: Done: https://github.com/cerner/terra-framework/pull/642/commits/9cd02301e80b98aa1af8dd879f017f0a741e6d09

mmalaker:
Functional verification completed. The date picker field example displays and functions as expected.

PR #645, opened 2019-05-02
[terra-notification-dialog] Added max-width for IE.

### LINKED ISSUE #621: Notification Dialog taking entire screen width in IE
# Bug Report

## Description
Notification Dialog taking entire screen width in IE for long texts

## Steps to Reproduce

1. Open [terra-notification-dialog](https://engineering.cerner.com/terra-ui/#/components/terra-notification-dialog/notification-dialog/notification-dialog) in IE 11
2. Check for 'Complete Notification Dialog With Long Message' example
3. Trigger notification dialog
4. Notice that it takes more width than normal (width varies with the number of characters in the text)

**IE:**
image
**Chrome:**
image

## Expected Behavior
It should display as it is displaying in chrome

## Environment

* Component Name and Version: terra-notification-dialog (2.10.0)
* Browser Name and Version: IE 11
* Operating System and version (desktop or mobile): Mac/Windows

pranav300:
### Summary
Resolves #621


pranav300:
In Chrome - 
Chrome
In IE - 
IE

packages/terra-notification-dialog/src/NotificationDialog.module.scss
```diff
@@ -8,6 +8,10 @@
     flex-direction: column;
     position: fixed;
 
+    //The query checks whether the browser is IE or not.
+    @media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
```
ShettyAkarsh: Do we need to check for max-width: 768px as well here ?

ShettyAkarsh: As per MDN doc it looks like this media feature is not suggested one https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-ms-high-contrast

pranav300: > Do we need to check for max-width: 768px as well here ?

I have added the max-width to be 768px for this query.

packages/terra-notification-dialog/src/NotificationDialog.module.scss
```diff
@@ -8,6 +8,10 @@
     flex-direction: column;
     position: fixed;
 
+    //The query checks whether the browser is IE or not.
+    @media screen and (max-width: 768px) and (-ms-high-contrast: active), (-ms-high-contrast: none) {
```
tbiethman: I think we should just go ahead and add a max-width, regardless of the browser. Based on the styles that are present, I'm not sure why Chrome/FF/Safari are rendering the way they are.

@neilpfeiffer 49% is probably pretty close, but do you have any recommendations for a max-width for this?

tbiethman: On second thought, I'm not sure why this is a max-width media query, as we want this to take effect when the viewport is larger than 768px.

I think we could do something like:
```scss
@import '~terra-breakpoints/lib/media-queries';

.notification-dialog {
  // This will ensure that by default, the notification dialog sizes to the viewport (minus the insets).
  width: calc(100% - 2 * var(--terra-notification-dialog-small-horizontal-inset, 10px));

  // This will ensure that above the medium breakpoint, a max-width is enforced
  @include terra-mq-medium-up {
    max-width: 49%; // or whatever Neil comes up with
  }
}
```

packages/terra-notification-dialog/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+* Updated max-width to be `49%` for screen size greater than 768px.
+* Updated wdio snapshot
```
nramamurth: This is not needed.

bjankord: You can remove the the `* Updated wdio snapshot`, it is not pertinent to the change from a consumers perspective.

packages/terra-notification-dialog/src/NotificationDialog.module.scss
neilpfeiffer: Investigating to see what would be an appropriate width to set.

packages/terra-notification-dialog/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+* Updated max-width to be `49%` for screen size greater than 768px.
+
+=======
```
emilyrohrbough: remove merge divider

pranav300:
Screenshots:
**IE10**
IE10

**IE11**
IE11

packages/terra-notification-dialog/src/NotificationDialog.module.scss
neilpfeiffer: @pranav300 Lets set max width to be a fixed `800px` to match one of our [disclosure widths](https://github.com/cerner/terra-framework/blob/1e9427e3312236e517ae8691fd64a6d9d109a0a4/packages/terra-disclosure-manager/src/DisclosureManager.jsx#L18) 

Which should be an appropriate max-width for the primary type of content included in the notification-dialog. If we find that is too small after hearing feedback, we can increase it to `960px`.

neilpfeiffer:
@pranav300 
Lets set max-width to be a fixed `800px` (instead of 49%) to match one of our [disclosure widths](https://github.com/cerner/terra-framework/blob/1e9427e3312236e517ae8691fd64a6d9d109a0a4/packages/terra-disclosure-manager/src/DisclosureManager.jsx#L18) 

Which should be an appropriate max-width for the primary type of content included in the notification-dialog. If we find that is too small after hearing feedback, we can increase it to `960px`.

packages/terra-notification-dialog/src/NotificationDialog.module.scss
```diff
@@ -124,4 +128,8 @@
       margin-top: var(--terra-notification-dialog-actions-button-margin-top, 0.4rem);
     }
   }
+
+  .text-wrapper {
+    flex-shrink: 1;
```
bjankord: Might be good to add a inline comment on why this still is needed so we remember.

pranav300: Added here https://github.com/cerner/terra-framework/pull/645/commits/a4dd349cdf9f28036fbc54b1a24a1d2419c09364.

neilpfeiffer:
+1 UX review - Ready to Merge

PR #658, opened 2019-05-07
Update react-transition-group to latest stable major release

bjankord:
### Summary
Update react-transition-group to latest stable major release


tbiethman:
I looked at the CHANGELOG: https://github.com/reactjs/react-transition-group/blob/v4.0.0/CHANGELOG.md

Doesn't seem to be anything worrisome there. The terra-slide-group package does not expose or export any functionality of the react-transition-group, so the switch to the new Context shouldn't impact things.

The deployment also seems to be behaving like I'd expect.

PR #663, opened 2019-05-13
[terra-slide-panels] Resolved the DOM order

### LINKED ISSUE #647: slidepanel's dom ordering  flexing based on prop
# Feature Request

## Description

requesting that the slidepanel's dom ordering is flexed when the start vs end panelPosition prop is updated. This would allow a developer to control the tab order of the slide panel component. Currently it is impossible to have the side panel's contents take tab precedence over the main  

## @ Mentions

@dkasper-was-taken  talked to me over teams about this

avinashg1994:
### Summary
Resolves #647


packages/terra-slide-panel/src/SlidePanel.jsx
```diff
@@ -92,6 +92,26 @@ class SlidePanel extends React.Component {
...
     ]);
 
+    const content = (panelPosition === 'start') ? (
+      <React.Fragment>
+        <div className={cx(['panel'])} tabIndex="-1" aria-hidden={!isOpen ? 'true' : 'false'} ref={this.setPanelNode}>
+          {panelContent}
+        </div>
+        <div className={cx('main')} tabIndex="-1" ref={this.mainNode}>
+          {mainContent}
+        </div>
```
ShettyAkarsh: we can declare these as two separate constants and call those one in both if and else. do avoid code duplication.

ShettyAkarsh:
Please update the contributors.md file

packages/terra-slide-panel/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+* Corrected slide panel's dom ordering
```
nramamurth: I think it is worth mentioning that the DOM ordering is being flexed based upon the `panelPosition` prop in the CHANGELOG.

packages/terra-slide-panel/src/SlidePanel.jsx
```diff
@@ -92,6 +92,26 @@ class SlidePanel extends React.Component {
       customProps.className,
     ]);
 
+    const content = (panelPosition === 'start') ? (
```
nramamurth: Is there an `enum`-like object for 'start' and 'end' that could be leveraged here?

nramamurth:
Has this change been unit tested in Jest to cover the scenario when the `panelPosition` is 'start' v/s 'end'?

avinashg1994:
> Has this change been unit tested in Jest to cover the scenario when the `panelPosition` is 'start' v/s 'end'?
Yes, there already exists an Jest which tests the DOM sequence for both the scenarios when 'panelPosition' is either 'start' or 'end'.

packages/terra-slide-panel/src/SlidePanel.jsx
avinashg1994: Made the appropriate changes of declaring the two separate constants for panel and main

avinashg1994:
> Please update the contributors.md file

Updated the CONTRIBUTORS.md file

packages/terra-slide-panel/CHANGELOG.md
avinashg1994: Corrected the CHANGELOG.md file to be more descriptive.

packages/terra-slide-panel/src/SlidePanel.jsx
avinashg1994: Added an enum object and made the necessary changes to incorporate it in the examples as well.

packages/terra-slide-panel/src/SlidePanel.jsx
```diff
@@ -5,6 +5,11 @@ import styles from './SlidePanel.module.scss';
 
 const cx = classNames.bind(styles);
 
+const position = {
+  start: 'start',
```
nramamurth: nit: I'd make the keys UPPERCASE. 

```
const position = {
   START: 'start',
   END: 'end',
}
```

supreethmr: +1 to make change to UpperCase.

we may also need to update examples and tests in doc with this new position constants to keep consistency between files.
here are the examples and tests which needs to update the position constants : 

- [DefaultSlidePanel.jsx](https://github.com/cerner/terra-framework/blob/c0ee6bbef747b0bba93a8359c58e250b333ba8d5/packages/terra-slide-panel/src/terra-dev-site/doc/example/DefaultSlidePanel.jsx#L136)

- [SlidePanelEnd.test.jsx](https://github.com/cerner/terra-framework/blob/4dcff701a4db4212eadd5f7e65b808b3872c11f2/packages/terra-slide-panel/src/terra-dev-site/test/slide-panel/SlidePanelEnd.test.jsx#L11)
- [SlidePanelFill.test.jsx](https://github.com/cerner/terra-framework/blob/4dcff701a4db4212eadd5f7e65b808b3872c11f2/packages/terra-slide-panel/src/terra-dev-site/test/slide-panel/SlidePanelFill.test.jsx#L12)
- [SlidePanelFullscreen.test.jsx](https://github.com/cerner/terra-framework/blob/4dcff701a4db4212eadd5f7e65b808b3872c11f2/packages/terra-slide-panel/src/terra-dev-site/test/slide-panel/SlidePanelFullscreen.test.jsx#L11)
- [SlidePanelLarge.test.jsx](https://github.com/cerner/terra-framework/blob/4dcff701a4db4212eadd5f7e65b808b3872c11f2/packages/terra-slide-panel/src/terra-dev-site/test/slide-panel/SlidePanelLarge.test.jsx#L11)
- [SlidePanelNoFill.test.jsx](https://github.com/cerner/terra-framework/blob/4dcff701a4db4212eadd5f7e65b808b3872c11f2/packages/terra-slide-panel/src/terra-dev-site/test/slide-panel/SlidePanelNoFill.test.jsx#L12)
- [SlidePanelOverlay.test.jsx](https://github.com/cerner/terra-framework/blob/4dcff701a4db4212eadd5f7e65b808b3872c11f2/packages/terra-slide-panel/src/terra-dev-site/test/slide-panel/SlidePanelOverlay.test.jsx#L11)
- [SlidePanelSmall.test.jsx](https://github.com/cerner/terra-framework/blob/4dcff701a4db4212eadd5f7e65b808b3872c11f2/packages/terra-slide-panel/src/terra-dev-site/test/slide-panel/SlidePanelSmall.test.jsx#L11)
- [SlidePanelSquish.test.jsx](https://github.com/cerner/terra-framework/blob/4dcff701a4db4212eadd5f7e65b808b3872c11f2/packages/terra-slide-panel/src/terra-dev-site/test/slide-panel/SlidePanelSquish.test.jsx#L11)

avinashg1994: The keys of the enum object have been changed to UpperCase.
Also, the enum values have been incorporated in the examples as well.

packages/terra-slide-panel/src/SlidePanel.jsx
```diff
@@ -5,6 +5,11 @@ import styles from './SlidePanel.module.scss';
 
 const cx = classNames.bind(styles);
 
+const position = {
```
mjhenkes: Leave this as strings. I don’t want to re introduced exporting enums.

packages/terra-slide-panel/src/SlidePanel.jsx
```diff
@@ -114,5 +138,6 @@ class SlidePanel extends React.Component {
 
 SlidePanel.propTypes = propTypes;
 SlidePanel.defaultProps = defaultProps;
+SlidePanel.position = position;
```
nramamurth: Why have you assigned it to the Class here?

packages/terra-slide-panel/src/SlidePanel.jsx
nramamurth: @mjhenkes 

I recommended creating an enum and exporting it assuming that it is a pattern [1] that is followed. May I know why you'd not want to re-introduce exporting enums?

[1] https://github.com/cerner/terra-core/pull/1923/

packages/terra-slide-panel/src/SlidePanel.jsx
avinashg1994: I have assigned it to the class for the purpose of exposing the enum.
Since the "position" enum will only be referenced for SlidePanel, assigning it to the class prevents us from explicitly importing the enum.
Reference : [https://jaketrent.com/post/expose-enum-props-in-react/](url)

packages/terra-slide-panel/src/SlidePanel.jsx
bjankord: We want to keep the propTypes as strings. You can export an enum that returns the expected strings, but we want the propTypes to stay accepting strings. To do this, we should keep the propTypes for panelPosition as `panelPosition: PropTypes.oneOf(['start', 'end']),`

Here is where we last talked about this: https://github.com/cerner/terra-core/issues/1044#issuecomment-386711960

bjankord: Let's update the `const` name to `const SlidePanelPositions` This aligns with what we talked about here: https://github.com/cerner/terra-core/issues/1044#issuecomment-386711960

packages/terra-slide-panel/src/SlidePanel.jsx
bjankord: Let's update to export slidepanel positions like so:

```jsx
export default SlidePanel;
export { SlidePanelPositions }
```
This aligns to the pattern we talked about here: https://github.com/cerner/terra-core/issues/1044#issuecomment-386711960

packages/terra-slide-panel/src/terra-dev-site/doc/example/DefaultSlidePanel.jsx
```diff
@@ -133,7 +133,7 @@ class DefaultSlidePanel extends React.Component {
 
 DefaultSlidePanel.propTypes = {
   panelBehavior: PropTypes.oneOf(['overlay', 'squish']),
-  panelPosition: PropTypes.oneOf(['start', 'end']),
+  panelPosition: PropTypes.oneOf([SlidePanel.position.START, SlidePanel.position.END]),
```
mjhenkes: Lets use the strings in the example.

packages/terra-slide-panel/src/SlidePanel.jsx
avinashg1994: Updated "const position" to "const SlidePanelPositions" and made the necessary changes to reflect the same, i.e. PropTypes now only use strings.

packages/terra-slide-panel/src/SlidePanel.jsx
avinashg1994: Updated to export the enum separately as suggested by @bjankord.

packages/terra-slide-panel/src/terra-dev-site/doc/example/DefaultSlidePanel.jsx
avinashg1994: Reverted the changes to use strings instead of enums

PR #678, opened 2019-05-17
Rebooting terra-application as a platform package

tbiethman:
### Summary

There are many applications built with Terra framework components. These applications are often made up of many separate components that are developed in isolation and composed together to build workflows. Due to this highly distributed environment, application and library developers have had a hard time managing their dependency versions and getting those versions to align. Under most circumstances this is okay (if not ideal), but a number of Context-based packages completely fail if versions are mismatched.

We want to reboot terra-application as a aggregate of Terra's framework components that we think make up the application "platform". Applications and libraries can unify on a single version of terra-application and set expectations around what functionality is provided and expected.

Along with this, we can provide an Application component that provides a one-stop shop for i18n, theming, and responsive APIs. This will help to define a baseline of functionality without requiring each app to compose all these pieces individually.

*This PR is a draft*

Everything here is functional and has been vetted with some of our work around navigation components. The big thing that needs to be improved here is the README documentation.

- We need to provide step by step guides for expected usage patterns.
- We need to provide detailed examples here for each piece of provided functionality.
- We need to consider exporting 'react-intl' from here (in anticipation of their impending Context-based release).
- We need to consider relocating the various packages within terra-application so that they cannot be consumed directly.


packages/terra-application/package.json
```diff
@@ -0,0 +1,50 @@
...
+  ],
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-framework/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-framework#readme",
+  "peerDependencies": {
+    "react": "^16.8.5",
+    "react-dom": "^16.48.5"
```
bjankord:
```suggestion
    "react-dom": "^16.8.5"
```

packages/terra-application/package.json
```diff
@@ -0,0 +1,50 @@
...
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "prop-types": "^15.5.8",
+    "terra-doc-template": "^2.0.0",
+    "terra-base": "^4.0.0",
```
bjankord:
```suggestion
    "terra-base": "^5.0.0",
```

packages/terra-application/src/breakpoints/_media-queries.scss
```diff
@@ -0,0 +1 @@
+@import '~terra-breakpoints/lib/media-queries';
```
bjankord: Curious, what is the intent of this file/import?

packages/terra-application/docs/README.md
```diff
@@ -0,0 +1,285 @@
...
+  - `npm install terra-application`
+
+## Usage
+
+The `terra-application` package should be used both by applications and libraries to define a baseline of expected features and functionality for their components.
+
+For application packages:
+
+1. `terra-application` must be specified as a dependency.
+2. A single `<Application />` component must be rendered as high as possible in the component tree.
```
bjankord: nit: We use the following verbiage on the i18n getting started page in Terra UI. `This component should be your root component and wrap all components in your app. This is to help ensure that the entire application is within the same configured i18n context provided by react-intl's <IntlProvider />.` 

It might make sense to copy/paste and reword that for the second bullet point here and remove the react-intl part, but note that this is needed for context provider concerns.

Something like.

```
This component should be your root component and wrap all components in your app. This is to help ensure that the entire React application is within the same configured React context providers. 
```

packages/terra-application/docs/README.md
```diff
@@ -0,0 +1,285 @@
...
+3. Components provided by `terra-application` should be consumed solely through `terra-application` and not through their own separate packages.
+
```
bjankord: Might be good to have a do / don't example of how we expect consumers to import packages provided by terra-application.

packages/terra-application/docs/README.md
```diff
@@ -0,0 +1,285 @@
...
+const Example = () => (
+  <Application
+    locale="en-US"
+  >
+    <ExampleContent />
+  </Application>
+)
+```
+
+Only a single `<Application />` component should be rendered within a given application. All Terra components should be rendered within the `<Application />` component to ensure they have access to the framework features.
```
bjankord: nit: should be rendered within a given **_React_** application
I think updating to this as many React applications may live inside of one of our Rails applications.

packages/terra-application/docs/README.md
```diff
@@ -0,0 +1,285 @@
...
+  >
+    <ExampleContent />
+  </Application>
+)
+```
+
+A Sass partial containing media query mixins based on the defined breakpoints is also available.
+
+```scss
+@import '~terra-application/lib/breakpoints/media-queries';
```
bjankord: Ah ok this makes sense why this was being imported.

mjhenkes:
I pulled this out of draft status to enable travis-ci builds.

packages/terra-application/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "classnames": "^2.2.5",
+    "prop-types": "^15.5.8",
+    "terra-doc-template": "^2.0.0",
+    "terra-base": "^5.0.0",
+    "terra-breakpoints": "^2.0.0",
+    "terra-disclosure-manager": "^4.0.0",
+    "terra-modal-manager": "^6.0.0",
+    "terra-navigation-prompt": "^1.0.0",
+    "terra-slide-panel-manager": "^5.0.0",
+    "terra-theme-provider": "^3.0.0"
```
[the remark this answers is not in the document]
mjhenkes: I verified these are all up to date

packages/terra-application/package.json
tbiethman: This has been updated.

packages/terra-application/docs/README.md
tbiethman: I've expanded this section pretty heavily. Please take a look and see what you think.

packages/terra-application/docs/README.md
tbiethman: I've expanded this section pretty heavily. Please take a look and see what you think.

packages/terra-application/docs/README.md
tbiethman: 👍 i've updated this a bit

packages/terra-application/docs/README.md
bjankord: Looks good to me.

packages/terra-application/docs/README.md
bjankord: LGTM

tbiethman:
Regarding my TODOs above:

> We need to provide step by step guides for expected usage patterns.
> We need to provide detailed examples here for each piece of provided functionality.
> We need to consider relocating the various packages within terra-application so that they cannot be consumed directly.

We do want to eventually want to move the various components into the terra-application and deprecate their separate packages. When this is done, we will reevaluate the state of the documentation and make changes as necessary.

> We need to consider exporting 'react-intl' from here (in anticipation of their impending Context-based release).

We decided against doing this now or in the future. react-intl will continue to be a dependency that must be provided by our consumers. Consistency will be enforced using peer dependencies within the Terra ecosystem.

PR #690, opened 2019-05-24
Issue-688 - replaced classList shim with polyfill

### LINKED ISSUE #688: terra-hookshot uses an incomplete shim for classList which is incompatible with d3js
# Bug Report

## Description

terra-hookshot uses a [shim](https://github.com/cerner/terra-framework/blob/2bc4173e557a6a56deb31ac690914f2f1087ee15/packages/terra-hookshot/src/HookshotContent.jsx#L41) that defines classList for SVGElements on IE. This shim only defines classList and does not implement it. d3js has a [classed()](https://github.com/d3/d3-selection/blob/4dcb3412318a6e2ecada79196b42336758d29133/src/selection/classed.js#L6) function for applying classes to elements, and this functions changes how it adds the class based on if classList is defined or not. Since the shim defined it, d3js tries to use it, but the functions it tries to use don't actually exist which leads to errors like "Object doesn't support property or method 'add'".

## Possible Solution

There are multiple classlist polyfills that can be used instead. This seems to be the most popular: https://www.npmjs.com/package/classlist-polyfill

## @ Mentions

@bjankord

Matthematic:
### Summary
fixes #688 Replaced the classList shim in hookshot with [classlist-polyfill](https://www.npmjs.com/package/classlist-polyfill)


packages/terra-hookshot/src/HookshotContent.jsx
```diff
@@ -5,6 +5,7 @@ import KeyCode from 'keycode-js';
 import onClickOutside from 'react-onclickoutside';
 import ResizeObserver from 'resize-observer-polyfill';
 import styles from './HookshotContent.module.scss';
+import 'classlist-polyfill'; // Full polyfill for browsers with no classList support
```
emilyrohrbough: Shouldn't we check if the polyfill is already included before importing it?

Matthematic: The polyfill handles checking if classList is already defined before doing anything, and i thought webpack took care of bundling repeated imports.

emilyrohrbough: True, webpack will take care of bundling.

I hate to be a pain, but since we are adding a new third party dependency to our library, I am wondering if we've scoped out the impact. Why did you choose `classlist-polyfill` over `classList.js`? Classlist-poyfill mentions they fork classList.js. Does `classlist-polyfill` have similar  issues to whats been logged in classList.js?

Additionally, why did we explore adding the [dom4 polyfill](https://github.com/WebReflection/dom4) suggested by react-outsideclick (the reason we have this shim) instead of classlist-polyfill?

Matthematic: classlist-polyfill is simply a published fork of classList.js. The owner of classList.js chose not to release on npm, but they are the exact same script, and the owner [suggests](https://github.com/eligrey/classList.js/pull/46#issuecomment-189782600) this package themselves. And the reason for not using dom4 was to try and change and little as necessary since dom4 includes polyfills for all sorts of things in addition to classList.

Matthematic:
@bjankord It says merging is blocked for me since I'm not an authorized user. Are you able to merge this?

bjankord:
@Matthematic Yeah, this looks good to merge. Thanks for the contribution @Matthematic! :+1:

PR #691, opened 2019-05-27
[terra-framework] Removed inline style for dev-site components Part 2

### LINKED ISSUE #657: Remove inline styles in dev site components
# Feature Request

## Description

We want to remove inline styles from our dev site components to set a better example for consumers of terra.  As a replacement, we should pass in classes and use css to style the components.

avinashg1994:
### Summary
Removed all inline styles from Dev-site components and replaced the same styles with external css for the following packages:
1. Brand Footer
2. Dialog Modal
3. Embedded Content Consumer
4. Menu
5. Modal Manager
6. Navigation Prompt

Partially resolves #657


packages/terra-brand-footer/src/terra-dev-site/doc/example/BrandFooterExampleCommon.scss
```diff
@@ -0,0 +1,30 @@
...
+    width: 40px; // setting this explicity for IE11
+  }
+
+  .content-left-terratext {
+    letter-spacing: 1px;
+  }
+
+  .content-left-uitext {
+    font-weight: 100;
+    letter-spacing: 1px;
```
ShettyAkarsh: we can also implement this using a variable like @content-letter-spacing: 1px .  and then assign the variable  to letter-spacing.

packages/terra-brand-footer/src/terra-dev-site/test/brand-footer/HorizontalWrapBrandFooter.test.jsx
```diff
@@ -1,10 +1,14 @@
 import React from 'react';
+import classNames from 'classnames/bind';
 
 import BrandFooter from '../../../BrandFooter';
+import styles from './HorizontalWrapBrandFooter.test.scss';
+
```
ShettyAkarsh: import missing for BrandFooterGTestCommon

avinashg1994: This is intentional, "HorizontalWrapBrandFooter.test.scss" already imports "BrandFooterTestCommon.scss".
The reason for having a separate scss file for this was because "HorizontalWrapBrandFooter.test.jsx" has an extra element with a  unique style which was not present in the other files.

avinashg1994:
> Would it be possible to break each change into its own PR so it's easier to review?

Yes, it is possible. But terra-framework has **32 Packages** and had **452 inline style warnings**.
Creating a separate PR for each component would result in 32 PR's which would make it difficult to keep track. So, it was decided to break down the 32 packages into sets based on the number of changes required for them.
#700, contains changes for only one package, i.e. "Popup", due to the large number of changes required.
@emilyrohrbough, Please share your thoughts on this.

packages/terra-brand-footer/src/terra-dev-site/doc/example/BrandFooterExampleCommon.scss
```diff
@@ -0,0 +1,30 @@
+:local {
+  .content-left-anchor {
+    background-color: transparent;
+    color: #fff;
+    text-decoration: none;
+  }
+
+  .content-left-graphics {
```
emilyrohrbough: Technically you can add more than one class name to an element. So this could be change to be something like:
```
:local {
  .anchor {
    background-color: transparent;
    color: #fff;
    text-decoration: none;
  }

  .graphics {
    border: 0;
    height: 40px; // setting this explicity for IE11
    width: 40px; // setting this explicity for IE11
  }

  .left-graphics {
    vertical-align: middle;
  }

  .right-graphics {
    vertical-align: top;
  }

  .text {
    letter-spacing: 1px;
  }

  .weighted-text {
    font-weight: 100;
  }
}
```

Then, add multiple where needed, like:

```diff
- <span className={cx('content-left-terratext')}>Terra </span>
+ <span className={cx('text')}>Terra </span>

- <span className={cx('content-left-uitext')}>UI</span>
+ <span className={cx(['text', 'weighted-text'])}>UI</span>
```

packages/terra-brand-footer/src/terra-dev-site/test/brand-footer/HorizontalWrapBrandFooter.test.jsx
emilyrohrbough: I think it is okay to reuse the same style sheet for multiple examples even if all styles are not used for a particular example. Overall it reduces the amount of code to maintain

packages/terra-modal-manager/src/terra-dev-site/test/modal-manager/DisclosureComponent.scss
```diff
@@ -0,0 +1,11 @@
+@import './ModalManagerCommon';
+
+:local {
+  .header {
+    @extend %common-header;
```
emilyrohrbough: Why not directly use the `common-header`  and `common-content-form` class names in the example?

packages/terra-modal-manager/src/terra-dev-site/test/modal-manager/_ModalManagerCommon.scss
```diff
@@ -0,0 +1,14 @@
...
+    margin: 0;
+  }
+
+  %common-content-form {
+    padding: 0.7rem;
+  }
+
+  %common-content-wrapper {
+    height: 100%;
+  }
```
emilyrohrbough: Lets add:
```
.select-field {
    max-width: 300px;
  }
``
Then then update the test example to use `_modalMangerCommon.scss` directly so we can remove `ModalManagerIntegration.test.scss`, `DisclosureIntegrationComponent.scss` and `DisclosureComponent.scss `. I think using the @extends syntax adds an extra complexity to these examples that aren't needed and these styles can be shared.

packages/terra-brand-footer/src/terra-dev-site/test/brand-footer/HorizontalWrapBrandFooter.test.jsx
avinashg1994: Removed "HorizontalWrapBrandFooter.test.scss" and using only "BrandFooterTestCommon.scss".

packages/terra-brand-footer/src/terra-dev-site/doc/example/BrandFooterExampleCommon.scss
avinashg1994: I implemented this by separating the styles as follows:
`.content-left-text {
    letter-spacing: 1px;
}

.weighted-text {
    font-weight: 100;
}`
and then by adding both these classnames to the element

packages/terra-brand-footer/src/terra-dev-site/doc/example/BrandFooterExampleCommon.scss
avinashg1994: I implemented `content-left-text` and `weighted-text` as per your suggestion.
But did not touch "left and right graphics" since the `width`and `vertical-align` held different values.

packages/terra-modal-manager/src/terra-dev-site/test/modal-manager/DisclosureComponent.scss
avinashg1994: Removed this file and moved all css to a common file.

packages/terra-modal-manager/src/terra-dev-site/test/modal-manager/_ModalManagerCommon.scss
avinashg1994: Done

avinashg1994:
> Let's get the comments @emilyrohrbough left on the review addressed and then this should be ready to merge. As @emilyrohrbough pointed out, we want to avoid using the `@extends` syntax adds an extra complexity to these examples.

All review comments have been addressed and avoided the use of `@extends`

PR #694, opened 2019-05-28
Dangerfile changelog checking improvements

Blackop778:
### Summary
Enforces that every package that has changes in the src folder has a changelog modification. Additionally even if only new files have been added a changelog modification is still needed. All packages that need a changelog modification will be listed as opposed to simply stating that a changelog entry need to be included.

### Additional Details
See cerner/terra-core#2418


dangerfile.js
```diff
@@ -1,31 +1,38 @@
...
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { danger, warn, fail } from 'danger';
 
-const newChangelog = danger.git.created_files.filter((filePath) => {
-  const srcFilePattern = /^packages\/([a-z-])*\/CHANGELOG.md/i;
-  return srcFilePattern.test(filePath);
-});
+const BIG_PR_THRESHOLD = 1000;
+const CHANGELOG_PATTERN = /^packages\/terra-([a-z-])*\/CHANGELOG\.md/i;
+const PACKAGE_SOURCE_PATTERN = /^packages\/terra-([a-z-])*\/src/i;
```
emilyrohrbough: We actually want this search pattern to be anything changes that wasn't the changelog. For example, if the package.json has new dependencies, the changelog needs to be updated for release. Currently if an entry isn't added, Ryan or Brett have to manually add an entry.

Blackop778: 5401977a5321608caab2a2b28b89139d368212ef

dangerfile.js
```diff
@@ -1,31 +1,37 @@
...
+  }
+
+  const packageName = file.split('packages/')[1].split('/')[0];
 
-const modifiedSrcFiles = danger.git.modified_files.filter((filePath) => {
-  const srcFilePattern = /^packages\/([a-z-])*\/src/i;
-  return srcFilePattern.test(filePath);
+  if (CHANGELOG_PATTERN.test(file)) {
+    changedChangelogs.add(packageName);
+  } else { // file is in a package and was changed - we need a changelog
```
StephenEsser: We may want to see what happens if any project specific file ( non-package file ) was change. Anything at the root level here: https://github.com/cerner/terra-framework

Blackop778: I've tested that by running `danger local`. This branch doesn't have any changes in a package so it doesn't need a changelog entry. I suppose additional verification would be Travis passing the build.

StephenEsser: I had forgotten you added that return statement above. That should be safe guarding any root level file changes from making here.

PR #699, opened 2019-05-29
Focus after trap

### LINKED ISSUE #698: [terra-date-time-picker] Focus is no longer put on the hour input after a date is selected
# Bug Report

## Description
After a date is selected from the picker the focus will automatically be put on the hour input. However, with the change in #608 the focus is being trapped and prevents the focus from being set on hour input after a date is selected.

With [returnFocusOnDeactivate](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/src/react-datepicker/index.jsx#L627) being set to `true` for the focus-trap, another issue this causes is that calling `this.hourInput.focus()` while focus is trapped would cause onBlur to trigger from the hour input since it will force to return the focus back on the calendar button which originally triggered the picker.

## Steps to Reproduce
1. Go to any date-time-picker [example](https://engineering.cerner.com/terra-ui/#/components/terra-date-time-picker/date-time-picker/date-time-picker).
2. Click on the calendar button and select a date from the picker.
3. Focus no longer automatically jumps to the hour input.

## Expected Behavior
For terra-date-picker, the focus should return to the calendar button after selecting a date from the picker. For terra-date-time-picker, the focus should be set on the hour input after a date is selected from the picker.

benbcai:
### Summary
The focus is now set on the hour input after selecting a date from the picker.

### Additional Details
- Fixes #698 
- `returnFocusOnDeactivate` to `true` for focus-trap-react to always return focus back to the calendar button. This prevents the focus to automatically be set on the hour input after selecting a date from the picker. The focus will need to handle at the component level since the date-picker by default will set focus back on the calendar button after selecting a date and the date-time-picker will auto focus on the hour input after a date is selected.

### Deployment Link
https://terra-framework-deploye-pr-699.herokuapp.com/#/components/terra-date-time-picker/date-time-picker/date-time-picker

@cerner/terra


packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -99,6 +103,7 @@ const defaultProps = {
   onFocus: undefined,
   onSelect: undefined,
   required: false,
+  returnFocusOnClose: true,
```
jmsv6d: Could we reverse the naming of this to something like `disableButtonFocusOnClose` and have the default be `false`?

benbcai: It makes sense to change it to disableButtonFocusOnClose and default to false.

https://github.com/cerner/terra-framework/pull/699/commits/9a581a1cf70c8c205f4826e2b32959ccc2e2aea6

mmalaker:
Functional verification completed. The date time examples now move focus correctly between date field/calendar control and hour field.

PR #700, opened 2019-05-30
[terra-framework] Removed inline style for dev-site components Part 4

### LINKED ISSUE #657: Remove inline styles in dev site components

avinashg1994:
### Summary
Removed all inline styles from Dev-site components and replaced the same styles with external css for the **Popup** package.

Partially resolves #657 

@cerner/terra


emilyrohrbough:
Can we align the styles in these examples and use a single stylesheet instead of adding a custom style sheet for each example?

packages/terra-popup/src/terra-dev-site/doc/example/PopupNoHeader.scss
```diff
@@ -0,0 +1,8 @@
+:local {
```
ShettyAkarsh: This file is same code as popupbounded.scss. can we resuse them by renaming the file ?

avinashg1994:
> Can we align the styles in these examples and use a single stylesheet instead of adding a custom style sheet for each example?

Yes, it is possible to have just a single file with all the styles. But due to the following reasons we decided with the current approach:
- For the purpose of modularity and readability.
- Multiple examples have elements, which serve the same purpose, have different styles, e.g. "content-wrappers".
- Having a common stylesheet will require unique class names for different elements, which will in-turn make them very long.

Even so, i have tried to combine styles from multiple files into a single stylesheet to reduce the number of redundant stylesheets where ever possible.
@emilyrohrbough, Please share your thoughts on this.

packages/terra-popup/src/terra-dev-site/doc/example/PopupNoHeader.scss
avinashg1994: I agree, i have renamed and reused PopupBounded.scss and removed PopupNoHeader.scss.

PR #701, opened 2019-05-30
Switch date picker from terra-hookshot to terra-popup

### LINKED ISSUE #527: [terra-date-picker] Unexpected behavior inside modals
# Bug Report

## Description
When used with [terra-disclosure-manager](https://github.com/cerner/terra-framework/tree/master/packages/terra-disclosure-manager) with the disclosure size set to 'small', the date-picker renders on top of the actual date field, and for mobile view as well as small desktop viewports, the datepicker in portal view does not render in the center of the modal. Instead, the bottom half of the picker gets cropped out and only the top half is visible.

## Steps to Reproduce
1. Create a form that uses terra-date-picker and render it in a modal using the ```disclose``` function of terra-disclosure-manager
2. Set the ```size``` argument to small in the ```disclose``` function mentioned above
3. Open the datepicker

screen shot 2019-02-25 at 2 53 17 pm
screen shot 2019-02-25 at 2 53 22 pm
screen shot 2019-02-25 at 2 53 33 pm

## Expected Behavior
It should behave as the modal example on the Terra UI website does.
screen shot 2019-02-26 at 10 41 41 am
screen shot 2019-02-26 at 10 42 02 am

## Possible Solution
These lines could potentially be the culprit : https://github.com/Hacker0x01/react-datepicker/blob/master/src/stylesheets/datepicker.scss#L511-L513

## Environment
* Component Name and Version: 2.39.0
* Browser Name and Version: Google Chrome Version 72.0.3626.109 (Official Build) (64-bit)
* Node/npm Version: Node v18.4.0/yarn v 1.13.0
* Operating System and version (desktop or mobile): mac OS High Sierra 10.13.6

## @ Mentions
@dkasper-was-taken

jmsv6d:
### Summary
Closes #527 The only part of this issue that is still a problem is the arrow still displaying when the date picker gets pushed and covers the input. Switching to popup resolves this issue and consolidates our arrow logic to one location. 
A few wdio screenshots needed to be updated as the popup has it's own borders defined which are slightly different that what was defined in react-datepicker. 

### Deployment Link
https://terra-framework-deploye-pr-701.herokuapp.com/#/components/terra-date-picker/date-picker/date-picker

@cerner/terra


packages/terra-date-picker/src/react-datepicker/index.jsx
```diff
@@ -470,59 +457,117 @@ export default class DatePicker extends React.Component {
...
+        dayClassName={this.props.dayClassName}
+        showTimeSelect={this.props.showTimeSelect}
+        onTimeChange={this.handleTimeChange}
+        timeFormat={this.props.timeFormat}
+        timeIntervals={this.props.timeIntervals}
+        minTime={this.props.minTime}
+        maxTime={this.props.maxTime}
+        excludeTimes={this.props.excludeTimes}
+        className={this.props.calendarClassName}
+        yearDropdownItemNumber={this.props.yearDropdownItemNumber}>
```
Blackop778: If the `>` was moved to it's own line it would be a lot easier to tell that `{this.props.children}` is a child and not another prop. Similar to how the `Calendar` below is formatted.

packages/terra-date-picker/package.json
```diff
@@ -40,9 +40,11 @@
     "terra-responsive-element": "^4.0.0"
   },
   "devDependencies": {
+    "terra-disclosure-manager": "^4.9.0",
```
Blackop778: It doesn't look like this, `terra-modal-manager`, or `terra-dialog` are used anywhere.

packages/terra-date-picker/src/react-datepicker/index.jsx
jmsv6d: Fixed

packages/terra-date-picker/package.json
jmsv6d: Oops, good catch, that was leftover from testing. Removed now.

jmsv6d:
> terra-date-picker seems to be missing terra-popup as a dependency and I don't believe needs terra-hookshot anymore.

Dependencies updated.

mmalaker:
Functional verification completed. The date picker examples display and function as expected using popup.

PR #705, opened 2019-06-03
[terra-navigation-side-menu] Default menu item font weight to normal

### LINKED ISSUE #704: [terra-navigation-side-menu] Font weight is bold in edge
# Bug Report
https://github.com/cerner/terra-framework/issues/594 introduced a bug in Edge, where navigation side menu items have bold font. The lesson is do not use `inherit` as a fallback value. The fix is to explicitly set this to normal or 400.

nav-item-font-weight

yuderekyu:
### Summary
Closes #704


packages/terra-navigation-side-menu/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Changelog
 Unreleased
 ----------
 ### Changed
+* Default --terra-navigation-side-menu-item-font-weight to normal.
```
emilyrohrbough: Can we update this to be under `### Fixed` heading and add a blrub to indicate this chagne was needed to fix a bug?

PR #712, opened 2019-06-04
Fix danger message formatting

Blackop778:
### Summary
The previous commit overhauling the danger file also updated danger's version. In some version danger stopped including two new lines before `warn` and `fail` messages. Removing those new lines causes the markdown to not be parsed correctly. Adding an empty line before the list starts causes the list to be displayed properly.

The change in effect: Screen Shot 2019-06-04 at 1 33 54 PM


emilyrohrbough:
can you fake trigger a failure ?

Issue #706, opened 2019-06-03
[terra-date-time-picker] Date Time Picker (12-Hour format) doesn't re-size itself to fit in mobile device screen


lulz010:
# Bug Report

## Description
The Date Time Picker component, 12-Hour format, doesn't resize itself to fit on mobile devices screens. It was confirmed to be happening on android and iOS devices. If the font size is adjusted to small on the CT-50 device's settings the component is displayed correctly. Modifications to the devices' display settings seems to help, but the component still doesn't fit entirely on the screen.

## Steps to Reproduce
1. Open CareAware Connect Messenger App's Set My Status View on a mobile device (iPhone 8, Galaxy S9, and Honeywell CT-50).
2. Select Unavailable or Do Not Disturb options on the 'Set your status to...' radio button
3. Select Custom on the 'Clear your status after...' drop down

## Expected Behavior
The Date Time Picker component (12-Hour Format) should resize itself to fit on a mobile device screen.

## Environment
* Component Version: 4.2.0
* Operating System and version (desktop or mobile): mobile android and iOS

## Example Screenshot
<img width="864" alt="Screen Shot 2019-06-11 at 1 23 53 PM" src="https://user-images.githubusercontent.com/1355485/59296714-41272b80-8c4c-11e9-89c3-58d74822706c.png">

## Mentions
The issue was mentioned on the Terra Component Support channel on Teams. It was addressed by Dave Kasper @dkasper-was-taken and Neil Pfeiffer @neilpfeiffer .

nramamurth:
@neilpfeiffer 

From the UX standpoint, what is the expected outcome? Should the Time Input wrap around like something below or should Date Picker and Time Input completely resize (get smaller (?)) accordingly and remain inline with each other?

![Screen Shot 2019-07-09 at 12 30 12 PM](https://user-images.githubusercontent.com/10090859/60866326-acf9b580-a245-11e9-8020-a457d0db8b27.png)

neilpfeiffer:
For an immediate fix - lets have the inputs wrap, but rework css to have an alignment similar to below. (Note: we'll also need to re-check with the enhancement that adds the Seconds input, if the entire time-input will fit all on one line):

<img width="480" alt="Screen Shot 2019-07-09 at 7 56 58 AM" src="https://user-images.githubusercontent.com/1355485/60889691-3944b200-a21f-11e9-9d6f-d0fe08295287.png">

nramamurth:
@neilpfeiffer 

[...]

PR #722, opened 2019-06-10
Add seconds field to time-input and date-time-picker

### LINKED ISSUE #660: [time-input]Add second precision in time-input and date-time-picker
# Feature Request

## Description
The time-input (and date-time-picker) currently supports hour and minute. This feature request is to add the option to support second in time-input and date-time-picker.

Certain applications (and engines) can process thousands of transactions per minute. There is a need to be able to query or filter these transactions within a 15-second range as this short time span could contain hundreds of records. Having the second precision in the time-input and date-time-picker component provides the ability to filter records in a more precise range.

Blackop778:
### Summary
Adds the option to show a field to enter seconds, but the field is not shown by default. If the field is shown then a valid number must be entered for the input to be considered valid.

### Deployment
- https://terra-framework-deploye-pr-722.herokuapp.com/#/components/terra-time-input/time-input/time-input
- https://terra-framework-deploye-pr-722.herokuapp.com/#/tests/terra-time-input/time-input/twelve-hour/mobile (make browser window thinner until component changes to mobile style)
- https://terra-framework-deploye-pr-722.herokuapp.com/#/components/terra-date-time-picker/date-time-picker/date-time-picker (way down at the bottom of the examples)
- https://terra-framework-deploye-pr-722.herokuapp.com/#/tests/terra-date-time-picker/date-time-picker/date-time-picker-12-hour-mobile (same procedure as other mobile page)

### Additional Details
Arabic translations have been delayed until a large request can be made.

Closes #660


packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -530,7 +681,14 @@ class TimeInput extends React.Component {
...
         || event.keyCode === KeyCode.KEY_DELETE
         || event.keyCode === KeyCode.KEY_BACK_SPACE) {
-      this.minuteInput.focus();
+      if (this.props.showSecond) {
+        this.secondInput.focus();
+        if (this.state.second) {
+          this.secondInput.setSelectionRange(this.state.second.length, this.state.second.length);
+        }
+      } else {
+        this.minuteInput.focus();
```
Blackop778: Not sure about this statement. This is how it was before, but it is somewhat inconsistent with other behavior. When keyboard navigating from the minute input to hour the cursor will be placed at the right of the input with nothing selected. When keyboard navigating from the meridiem directly to the minute the text inside the minute will be selected. The above represents how other keyboard navigation to text inputs is processed.

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -75,37 +75,52 @@ class DateTimeUtils {
...
    * Determines if the date and time are valid and conform to the given format.
    * @param {string} date - The date to validate.
    * @param {string} time - The time to validate.
    * @param {string} format - The date/time format to use for the validation.
    * @return {boolean} - True if both the date and time are valid and conform to the format.
    */
-  static isValidDateTime(date, time, format) {
-    return DateUtil.isValidDate(date, format) && DateTimeUtils.isValidTime(time);
+  static isValidDateTime(date, time, format, second) {
+    return DateUtil.isValidDate(date, format) && DateTimeUtils.isValidTime(time, second);
```
emilyrohrbough: we usually try to name boolean flags such that it is obvious it's a boolean. We update to be something like `includeSeconds`, `hasSeconds`, etc. 

Also, these inline docs need to be updated to include the new parameter addition

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -75,37 +75,52 @@ class DateTimeUtils {
...
 
   /**
    * Determines if the time is a valid time in the HH:mm format
    * @param {string} time - The time to validate.
    * @return {boolean} - True if the time is valid.
    */
-  static isValidTime(time) {
-    const timeMoment = moment(time, 'HH:mm', true);
+  static isValidTime(time, second) {
+    const timeMoment = moment(time, second ? 'HH:mm:ss' : 'HH:mm', true);
```
emilyrohrbough: lets break this check out of passing it to moment so it's a bit easier to read

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -75,37 +75,52 @@ class DateTimeUtils {
...
     return newDate.hour(date.get('hour')).minute(date.get('minute'));
   }
 
+  /**
+   * Gets the time from a date and time
+   * @param {string} time An ISO 8601 string to get the time of
+   * @param {boolean} second Whether or not seconds are enabled
+   * @return {string} The time from the date and time string
+   */
+  static getTime(time, second) {
```
emilyrohrbough: I am not seeing any new jest tests to test this new function or to test the additional options of these existing helpers.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -83,6 +83,10 @@ const propTypes = {
    * The first parameter is the event. The second parameter is the selected input value in ISO format.
    */
   onSelect: PropTypes.func,
+  /**
+   * Whether an input field for seconds should be shown or not
```
emilyrohrbough: Based on some discussion last week, I think this description need to have more information. 

It had sounded like when `showSecond` was used, a date-time format with `hh:mm:ss` time needed to be provided to be valid. Is that still the case?

Blackop778: That is still the case.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -136,7 +141,7 @@ class DateTimePicker extends React.Component {
     // It is used for date/time manipulation and used to calculate the missing/ambiguous hour.
     // The dateValue and timeValue are tracked outside of the react state to limit the number of renderings that occur.
     this.dateValue = DateUtil.formatMomentDate(this.state.dateTime, this.state.dateFormat) || '';
-    this.timeValue = DateTimeUtils.hasTime(this.props.value) ? DateUtil.formatISODate(this.props.value, 'HH:mm') : '';
```
emilyrohrbough: Is formatISODate still a used helper?

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -83,6 +83,10 @@ const propTypes = {
    * The first parameter is the event. The second parameter is the selected input value in ISO format.
    */
   onSelect: PropTypes.func,
+  /**
+   * Whether an input field for seconds should be shown or not
+   */
+  showSecond: PropTypes.bool,
```
emilyrohrbough: lets rename to `showSeconds` since we have a 2 second precision

packages/terra-date-time-picker/src/DateTimePicker.jsx
Blackop778: `formatISODate` is still used to format dates but is not used to format times anymore.

packages/terra-time-input/src/TimeInput.jsx
benbcai: I think the minute input here should also use setSelectRange to be consistent.

packages/terra-date-time-picker/src/DateTimePicker.jsx
emilyrohrbough: If someone adds the `showSeconds`  but don't update their intial date-time value doesn't include seconds, will this render with the pre-defined value?

Blackop778: If seconds are not included in the value, date time picker will assume 00 whether or not `showSeconds` is true. If `showSeconds` is true then the 00 will be rendered in the seconds input box.

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -75,37 +75,58 @@ class DateTimeUtils {
...
    * Determines if the date and time are valid and conform to the given format.
    * @param {string} date - The date to validate.
    * @param {string} time - The time to validate.
    * @param {string} format - The date/time format to use for the validation.
+   * @param {boolean} hasSeconds Whether or not the time should consider having seconds valid
    * @return {boolean} - True if both the date and time are valid and conform to the format.
    */
-  static isValidDateTime(date, time, format) {
-    return DateUtil.isValidDate(date, format) && DateTimeUtils.isValidTime(time);
+  static isValidDateTime(date, time, format, hasSeconds) {
```
benbcai: There are two places [[1](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-time-picker/src/DateTimePicker.jsx#L200)] [[2](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-time-picker/src/DateTimePicker.jsx#L215)] in DateTimePicker.jsx that calls isValidDateTime. The `showSeconds` prop needs to be passed down when making those calls. Otherwise, the ambiguous hour logic won't work when the second input field is shown.

Blackop778: Fixed in https://github.com/cerner/terra-framework/pull/722/commits/f0b8c47950943f019c0502eebc2439ea2f2f0921 and changes to the other file were reverted in https://github.com/cerner/terra-framework/pull/722/commits/7695bd247be326a7c12526d12767b514d75660e1

packages/terra-time-input/src/TimeInput.jsx
Blackop778: This was resolved in b18195c6cfa9959e7f7390c5aed3029b40147f5d

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -75,37 +75,58 @@ class DateTimeUtils {
    * Synchronize only the time for a given moment object.
    * @param {object} momentDate - The moment object to synchronize the date and time.
    * @param {string} time - The time to synchronize with the moment object.
+   * @param {boolean} hasSeconds - If true seconds will be processed
    * @return {object} - The synchronized moment object.
    */
-  static updateTime(momentDate, time) {
+  static updateTime(momentDate, time, hasSeconds) {
```
benbcai: There are a few places [[1](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-time-picker/src/DateTimeUtils.js#L52-L54)] [[2](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-time-picker/src/DateTimeUtils.js#L63-L65)] [[3](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-time-picker/src/DateTimeUtils.js#L173)] in DateTimeUtils that calls `updateTime`. Those need to pass down the hasSeconds parameter and need to check if the length of the time string is 5 or 8 depending if seconds is included.

Blackop778: Fixed in 873a928b83c481df3173645df2680d0e54f16fa0

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -474,30 +548,108 @@ class TimeInput extends React.Component {
...
       const hour = type === TimeUtil.inputType.HOUR ? timeValue : this.state.hour;
       const minute = type === TimeUtil.inputType.MINUTE ? timeValue : this.state.minute;
+      const second = type === TimeUtil.inputType.SECOND ? timeValue : this.state.second;
 
-      if (hour === '' && minute === '') {
+      if (hour === '' && minute === '' && second === '') {
         this.props.onChange(event, '');
       } else {
-        this.props.onChange(event, this.formatHour(hour, meridiem).concat(':', minute));
+        this.props.onChange(event, this.formatHour(hour, meridiem).concat(':', minute).concat(this.props.showSeconds ? ':'.concat(second) : ''));
```
benbcai: If only the hour is present and both the minute and second are empty, this would concatenate two `:` to the hour and ends up with something like `12::`

Blackop778: I'm not sure it should be changed. The util methods `splitHour`, `splitMinute`, `splitSecond` can be used to get the hour/minute/second if the colons are included. That is also the only place the passed in `onChange` callback is called so it could be considered a breaking change potentially to only call `onChange` if the time input is valid. Those are my thoughts.

benbcai: My comment was not about whether or not to call onChange. It should continue to call onChange just like it is now. My comment was about the formatted time value that is returned in the second parameter in onChange. If `showSeconds` is set, entering just the hour would return a formatted value consisting two colons with no minute and second (i.e. `12::`). Perhaps, this is fine as this indicates that a minute and second have not been entered.

Talked to @Blackop778 offline about this and decided to leave it as is for now unless anyone has other thoughts.

jeremyfuksa:
With respect to the alignment of time values with the delimiters (time-spacer), it's going to take some refactoring and rethinking in order to get things to look right.

In the examples, I was able to get "Seconds Displayed With Default Time" to look right by changing the minute input width to 1.1em, but as the screenshot shows, that breaks the appearance of the time inputs that have placeholder values.

Screen Shot 2019-06-19 04 24 24 PM

The inputs need to have flexible width in order to align properly.

Blackop778:
@jeremyfuksa testing `min-content` and `auto` width values seem to make the input even wider than `1.8em`. I did find a solution which I have pushed that makes the minute thinner when the input is not empty. It seemed that any other solution would require setting the width with JavaScript which would be less than optimal.

packages/terra-date-time-picker/src/terra-dev-site/doc/date-time-picker/DateTimePicker.1.doc.jsx
```diff
@@ -132,6 +136,16 @@ to see the starting and ending dates of daylight saving time for the current yea
         example: <DateTimePickerTimeVariant />,
         source: DateTimePickerTimeVariantSrc,
       },
+      {
+        title: 'With Seconds Field',
+        example: <DateTimePickerWithSeconds />,
+        source: DateTimePickerWithSecondsSrc,
+      },
+      {
+        title: 'With Seconds Field in  12-hour Time format',
```
sirfunkbutter: nit: extra space
```suggestion
        title: 'With Seconds Field in 12-hour Time format',
```

packages/terra-time-input/translations/de.json
```diff
@@ -3,8 +3,10 @@
   "Terra.timeInput.pm": "",
   "Terra.timeInput.hours": "Stunden",
   "Terra.timeInput.minutes": "Minuten",
+  "Terra.timeInput.seconds": "Sekunden",
   "Terra.timeInput.hh": "hh",
   "Terra.timeInput.mm": "mm",
+  "Terra.timeInput.ss": "Sek.",
```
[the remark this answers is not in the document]
Blackop778: I've just put what the translation team gave me.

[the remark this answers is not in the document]
sirfunkbutter: +1, may want to check with them though

Blackop778:
@bjankord another related and more specific to this problem issue is #706

mmalaker:
Functional verification completed. The time input examples with seconds display and function as expected.

neilpfeiffer:
Tested: #722
Resolves: #660 

**Reviewed:**
- Adds seconds input using current pattern as hours and minutes. 

**Resolve with Future Enhancement:**
- Will need to make design updates as needed to resolve for #706 - Date Time Picker (12-Hour format) doesn't re-size itself to fit in mobile device screen

\
\

**Pass** |
-----------|
UX Review|

PR #726, opened 2019-06-11
Replaced Terra-Form with Terra-Form-Checkbox in Collapsible Menu

### LINKED ISSUE #451: [terra-collapsible-view] import new terra-form-checkbox
# Feature Request

## Description
- Uplift the collapsible menu view to import the new terra-form-checkbox (currently uses old terra-form/lib/control, which is unthemeable). 
- Checkbox may need enhancement to have non-padded/margin option (needs investigation) #2197 
- will be major version bump

supreethmr:
### Summary
Replaced deprecated terra-form with terra-form-checkbox in CollapsibleMenuView.Toggle.

Resolves #451 

### Additional Details
Updated the snapshots of collapsible-menu-view since there was slight change in the way Checkbox displayed after changing it from terra-form to terra-form-checkbox.

Resolves below warning : 
WARNING in terra-base
  Multiple versions of terra-base found:
    3.33.2 ./~/terra-form/~/terra-base
    5.6.0 ./~/terra-base


ryanthemanuel:
@bjankord @mjhenkes @StephenEsser do we think this'll be a major version bump?  Previously the themeable variables for this component were reliant on terra-form and now they're reliant on terra-form-checkbox

mjhenkes:
I don't think it's sustainable to continue to treat themes as part of Terra's api. Except in special cases (consumer), themes should be owned and maintained by Terra. The major version bumps caused by theme variables are broad, difficult to communicate and respond to.

packages/terra-collapsible-menu-view/src/terra-dev-site/test/collapsible-menu-view/DefaultCollapsibleMenuView.test.jsx
```diff
@@ -51,6 +51,7 @@ class CollapsibleMenuViewDemo extends React.Component {
           shouldCloseOnClick={false}
           onChange={this.handleToggleTwoOnChange}
           isSelected={this.state.toggle2Selection}
+          isSelectable
```
emilyrohrbough: Out of curiosity, why did we need this change? Does `Toggle Item 1` also need this change?

bjankord:
@ryanthemanuel I'm fine treating this as a minor version bump.

neilpfeiffer:
This will need some minor style overrides until we have a standalone version of the checkbox (included margins/padding are for use in traditional forms and not inline with menus/toolbars. Taking a quick look to see what we need to add to reset and bring back to vertical center (attempting to not have screenshots update by much at all).

neilpfeiffer:
@supreethmr -- Adding this at the end of [CollapsibleMenuView.module.scss](https://github.com/cerner/terra-framework/blob/a641d9eca9e4f6fd4c58e4ab4f5b94a3acf2c7b4/packages/terra-collapsible-menu-view/src/CollapsibleMenuView.module.scss) should do the trick for the temporary style override:
```
 .control { /* temporary overrides until standalone (no additional spacing included) version of checkbox is added */
    margin-bottom: 0;
    margin-top: 0;

    label:first-child {
      padding-bottom: 0;
    }
  }
```

\
\
I'm checking it against the theme repos to see if any adjustments need to be added, but should be a close start.

packages/terra-collapsible-menu-view/src/terra-dev-site/test/collapsible-menu-view/DefaultCollapsibleMenuView.test.jsx
supreethmr: forgot to remove before commit 🙆‍♂updated : 6a5829b5401cbcb031978cb8b05a4b73acd13af3

supreethmr:
> @supreethmr -- Adding this at the end of [CollapsibleMenuView.module.scss](https://github.com/cerner/terra-framework/blob/a641d9eca9e4f6fd4c58e4ab4f5b94a3acf2c7b4/packages/terra-collapsible-menu-view/src/CollapsibleMenuView.module.scss) should do the trick for the temporary style override:
> 
> ```
>  .control { /* temporary overrides until standalone (no additional spacing included) version of checkbox is added */
>     margin-bottom: 0;
>     margin-top: 0;
> 
>     label:first-child {
>       padding-bottom: 0;
>     }
>   }
> ```
> 
> 
> 
> I'm checking it against the theme repos to see if any adjustments need to be added, but should be a close start.

Even after adding these styles there are still slight difference in the size of check box like shown below :
Enormous : 
Responsive default

Huge : 
Responsive default

Large : 
Responsive default

Medium : 
Responsive default

Small : 
Responsive default

bjankord:
There is a chance that this remaining difference is a different between the chrome version we used in terra-toolkit v4 and terra-toolkit v5.

neilpfeiffer:
After reviewing the screenshot diffs - the visual left/right space I think is due to the very slight difference in internal construction spacing between the old `terra-form/lib/Control` component and the new `terra-form-checkbox` component -- it is not enough to worry about and is an acceptable update.

But I also did notice the ellipsis icon inside the button having a different position, which was unexpected and needs taking a second look at -- it should not change.

The only theme-related issue I was finding is there may be additional spacing needed between multiple inline checkbox components (other themes have a circular focus glow ). Normally we would append the "isInline" prop, but that would be too much space here, so testing to see if we need an additional themeable value for this component, and the appropriate values.

packages/terra-collapsible-menu-view/CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ Unreleased
 * Removed node_modules from .npmignore
 
 
+### Changed
+* Replaced Terra-Form with Terra-Form-Checkbox in CollapsibleMenuViewToggle
+
```
StephenEsser: This entry must have moved into a released section after merging master, we should move it back under Unreleased

supreethmr: moved back to unreleased section : 846091737b673b34895379b71b994ce7ce6846ea

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewToggle.jsx
```diff
@@ -99,7 +98,7 @@ class CollapsibleMenuViewToggle extends React.Component {
           labelText={text}
```
jmsv6d: The `type` prop isn't needed anymore here: https://github.com/cerner/terra-framework/pull/726/files#diff-973cc791b5a115b2e4130226902dded6R97

supreethmr:
> After reviewing the screenshot diffs - the visual left/right space I think is due to the very slight difference in internal construction spacing between the old `terra-form/lib/Control` component and the new `terra-form-checkbox` component -- it is not enough to worry about and is an acceptable update.
> 
> But I also did notice the ellipsis icon inside the overflow menu button having a different position, which was unexpected and needs taking a second look at -- it should not have changed.
> 
> The only theme-related issue I was finding is there may be additional spacing needed between multiple inline checkbox components (other themes have a circular focus glow ). Normally we would append the "isInline" prop, but that would be too much space here, so testing to see if we need an additional themeable value for this component, and the appropriate values.

Sorry I am bit confused. Do I need to include this style in collapsible-menu-view : 
```css
 .control { /* temporary overrides until standalone (no additional spacing included) version of 
  checkbox is added */
    margin-bottom: 0;
    margin-top: 0;

    label:first-child {
      padding-bottom: 0;
    }
  }
```

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewToggle.jsx
supreethmr: Thanks..! correct we don't need this prop. Removed : f28770be24f893cbadc923f83b698207aba95d1b

bjankord:
> Sorry I am bit confused. Do I need to include this style in collapsible-menu-view :
I'd defer to @neilpfeiffer on that. It sounds like this may be a good solution until the enhancement can be done in terra-form-checkbox.

supreethmr:
Added Temporary styles for checkbox :  ff74d25cdea6df5c601f85e8eee9adb750c0d66e 
Thanks @bjankord , @neilpfeiffer

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewToggle.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
-import Control from 'terra-form/lib/Control';
+import Control from 'terra-form-checkbox';
```
nramamurth: Maybe it's better to call it `Checkbox` here instead of `Control`?

supreethmr: updated : 8e45afbd85e04adc7a369423f1105915413c71fc

neilpfeiffer:
Was able to discover the problems I noticed in the new screenshot replacement for: [../chrome_large/collapsible-menu-view-spec/Responsive_Hidden_Open[default].png](https://github.com/cerner/terra-framework/pull/726/files?short_path=e929e91#diff-e929e91a1f273b2a5b13b62acc1823c9)



Turns out to be three issues: 

- First was the issue that caused the different screenshots - the new checkbox and label is just slightly (1-2px) narrower width than previously, causing the screenshot image to change what was rolled-up in the menu. To fix, the collapsible-menu-view needs a new permanent themeing variable to add additional space, per theme, to offset each checkbox and label from each other (other themes have larger focus glows and will need additional space). Fix is to add a new margin-left on [line 42](https://github.com/cerner/terra-framework/pull/726/files#diff-51d1884e92ac7b6ce5c7d9317d2d2f36L42
): 
```
    .control {
      margin: 0;
      margin-left: var(--terra-collapsible-menu-view-control-margin-left, 0.10141rem);
    }
```

- Second issue I noticed (missed before) looking at the new screenshot was the height of the menu items changes depending on if there is a selectable menu item or just text in a menu - I logged https://github.com/cerner/terra-framework/issues/769 to resolve separately. 

- Third odd issue I noticed was the divider in the collapsible-menu-view transitions unexpectedly between the faceup/normal layout and when rolls up in the menu - the screenshot shows a divider with nothing above it inside the menu - I logged https://github.com/cerner/terra-framework/issues/770 to resolve separately.

This shows the combine things I noticed:
Screen Shot 2019-07-08 at 9 30 17 AM

supreethmr:
> margin-left: var(--terra-collapsible-menu-view-control-margin-left, 0.10141rem);

added margin-left to control : 1e33323ace8c2714d52730239ada9cb12d261640

PR #728, opened 2019-06-11
[terra-framework] Removing -ms-overflow-style css

BenBoersma:
### Summary
Removing the `-ms-overflow-style` css rule for the same reasons here https://github.com/cerner/terra-core/pull/2445.

@cerner/terra


BenBoersma:
Need to have a discussion on these removal, seems removing this in `terra-scroll` had unintended consequences

BenBoersma:
As per scrum discussion, we're okay with the changes we're seeing, so we're good to move forward with this.

packages/terra-application-menu-layout/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Unreleased
 3.5.0 - (June 12, 2019)
 ------------------
 ### Removed
+* Removed '-ms-overflow-style' style
```
StephenEsser: Looks like this entry ended up in the wrong section, we should move it back into Unreleased

BenBoersma: Fixed https://github.com/cerner/terra-framework/pull/728/commits/f6bc73acb8c9ff28ad329fa9f653d80f186a7e7f and https://github.com/cerner/terra-framework/pull/728/commits/5899a38012b4d48daae0e8d5690426fd3dbbc88b

PR #732, opened 2019-06-12
Add screen reader loading status to proggresive infinite list

bjankord:
### Summary
As you scroll through the progressive infinite list, as more items are added, we display a progressive loading indicator. This feature adds functionality to announce audibly that same loading state to screen readers.

Resolves: #732 

Screen Shot 2019-06-12 at 6 31 25 AM

When you scroll through the progressive list and it loads more items, it will announce the phrase "Loading..."


packages/terra-infinite-list/src/InfiniteList.jsx
```diff
@@ -140,9 +149,18 @@ class InfiniteList extends React.Component {
     if (!this.props.isFinishedLoading && !this.hasRequestedItems && !this.isRenderingNew && this.props.onRequestItems) {
       this.hasRequestedItems = true;
       this.props.onRequestItems();
+      this.updateAriaLiveLoadingStatus();
     }
   }
 
+  updateAriaLiveLoadingStatus() {
+    this.ariaLiveStatus = this.props.intl.formatMessage({ id: 'Terra.InfiniteList.loading' });
```
mjhenkes: Should this loading message be a prop set by users? ie. 'Loading allergies' etc.

mjhenkes: We could make it optional so even if unset the component says loading.

bjankord: Yeah we can make it prop based and default back to the translated string we provide.

bjankord: Updated: https://github.com/cerner/terra-framework/pull/732/commits/e2c43f48898947aead1be8f4526988159e07fbf1

PR #734, opened 2019-06-12
[terra-popup] Revamp Tests

emilyrohrbough:
### Summary
Revamped the wdio tests to use Terra.describeViewports. During this uplift discovered a few test pages that weren't used, a few tests that were missing, and a few integration tests that were better suited as a jest test.


packages/terra-popup/tests/wdio/popup-spec.js
```diff
@@ -1,499 +1,360 @@
...
     });
-
-    Terra.it.matchesScreenshot({ selector: '#test-popup-area' });
   });
 
-  describe('Applies content offset when target is smaller than the arrow for vertial-left attachment', () => {
-    before(() => browser.setViewportSize(Terra.viewports('medium')[0]));
-    beforeEach(() => {
+  describe('Popup with [Target Smaller Than Arrow]', () => {
+    before(() => browser.url('/#/raw/tests/terra-popup/popup/different-attachments-popup'));
```
yuderekyu: is this url necessary? i see we immediately hit a different url in the next it block

packages/terra-popup/tests/wdio/popup-spec.js
```diff
@@ -1,499 +1,360 @@
...
       browser.waitForVisible('.test-content');
-      browser
-        .moveToObject('#root', 300, 300)
-        .leftClick();
     });
 
-    Terra.it.matchesScreenshot({ selector: '#test-popup-area' });
-  });
+    it('remains open on height resize', () => {
+      browser.setViewportSize({ width: mediumViewport.width, height: 600 });
```
mjhenkes: I like to grab the current viewport width, then this would also work in the loops
```javascript
const viewport = browser.getViewportSize();
viewport.height = 600;
browser.setViewportSize(viewport);
```

emilyrohrbough: @mjhenkes Good point. Since we only tested one viewport I didn't think about this impact

packages/terra-popup/tests/wdio/popup-spec.js
emilyrohrbough: no. Looks like a copy & paste mistake. Good catch

emilyrohrbough: removed: https://github.com/cerner/terra-framework/pull/734/commits/0afcefd17e8a4c5e3f0b94be3d8fd1ad28cbd347#diff-75a22fd845592219495fcc77ea6104c3L190

packages/terra-popup/tests/wdio/popup-spec.js
emilyrohrbough: resolved: https://github.com/cerner/terra-framework/pull/734/commits/0afcefd17e8a4c5e3f0b94be3d8fd1ad28cbd347#diff-75a22fd845592219495fcc77ea6104c3R6

PR #737, opened 2019-06-14
[terra-collapsible-menu-view] to prevent it from breaking for Single Large Item.

### LINKED ISSUE #713: Terra-collapsible-menu-view breaks with a single large item.
# Bug Report
Terra-collapsible-menu-view breaks with a single large item.

## Description
If you have a single large item in the collapsible menu view, the item will not collapse and will overflow.

## Steps to Reproduce
1. Create a collapsible menu view
2. Add a large item group
3. Dispair

The following code should be close, I haven't tested it myself.
```javascript
const SettingsMenu = () => {
  return (
    <CollapsibleMenuView className={cx('flex-collapse')} key="menu">
      <CollapsibleMenuView.ItemGroup>
        <CollapsibleMenuView.Item
            text="long text"
            key="derp"
          />
        <CollapsibleMenuView.Item
            text="long text"
            key="derp1"
          />
        <CollapsibleMenuView.Item
            text="long text"
            key="derp2"
          />
        <CollapsibleMenuView.Item
            text="long text"
            key="derp3"
          />
        <CollapsibleMenuView.Item
            text="long text"
            key="derp4"
          />
        <CollapsibleMenuView.Item
            text="long text"
            key="derp5"
          />
      </CollapsibleMenuView.ItemGroup>,
    );
    </CollapsibleMenuView>
  );
```

Screen Shot 2019-06-04 at 1 09 06 PM

## Expected Behavior
If this first item is too big, it collapses.

## Possible Solution
A cheap work around is to always add an undefined item as a child. But it's gross.

## Environment
* Component Name and Version: terra-collapsible-menu-view@6.1.0
* Browser Name and Version:
* Webpack Version:

## @ Mentions
@mjhenkes

pranav300:
### Summary
Resolves #713


packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
```diff
@@ -107,7 +107,7 @@ class CollapsibleMenuView extends React.Component {
 
       if (calcWidth > availableWidth) {
         // If last child fits in the available space, leave it face up
-        if (i === this.props.children.length - 1 && calcWidth <= width) {
```
StephenEsser: I vaguely remember this `-1` being pretty important. Has this been tested in all scenarios? Specifically when the there are multiple items without showing the collapsible icon? 

I believe this check was for when the last button was roughly the same size as the ellipsis button. We should show the actual button instead of the ellipsis if it can fit.

mjhenkes: I think you're right, but It looks like we should find another way to accomplish this, as I'm not sure this is even fully fixing the original intent. Basically we're just exempting the last item in a list from length checking regardless of whether it fits or not. This won't affect single children only when the child is a single item in a list.

dkasper-was-taken: We should probably investigate this css rule:
https://github.com/cerner/terra-framework/blob/master/packages/terra-collapsible-menu-view/src/CollapsibleMenuView.module.scss#L36
Should we be constraining the width of our items/groups to the width of the menu view or not, that might affect the last item logic as well.

jmsv6d: So... removing the `-1` means that this `if` check never evaluates to `true`. The  index of the last item in the array is the array's `length`  - 1.

jmsv6d: The problem here is really that when react has one child, children is an element and not an array. So this probably just needs to be updated to
`if (i === React.Children.count(this.props.children) - 1 && calcWidth <= width) {`

pranav300: @StephenEsser, yes removing `-1` does create a problem with the ellipsis. So, now I have found that the bug is because of the function `getBoundingClientRect().width` as it only provides the size of the child in the current window size (length of visible window) and not for the overflown elements. I tried using `scrollWidth, offsetWidth, and getComputedStyle` properties but it didn't work out. I was getting the required result with `scrollWidth` but it was breaking for IE11 because of a bug in it but seemed to work fine with IE10.
 @jmsv6d, I tried this code `if (i === React.Children.count(this.props.children) - 1 && calcWidth <= width)`  but it didn't solve the issue. 
**I tried implementing it like this -:**
``` JAVASCRIPT
  handleResize(width) {
    const menuButtonWidth = this.menuButton.getBoundingClientRect().width;
    const availableWidth = width - menuButtonWidth;
    let hiddenStartIndex = -1;
    let calcWidth = 0;
    let menuHidden = true;

    for (let i = 0; i < React.Children.count(this.props.children); i += 1) {
      const child = this.container.children[i];
      if (this.props.children.length) === 1) {
        for (let j = 0; j < child.children.length; j += 1) {
          const childWidth = child.children[j].getBoundingClientRect().width;
          calcWidth += childWidth;
        }
      } else {
        const childWidth = child.getBoundingClientRect().width;
        calcWidth += childWidth;
      }
 
      if (calcWidth > availableWidth) {
        // If last child fits in the available space, leave it face up
        if (i === this.props.children.length - 1 && calcWidth <= width) {
          break;
        }

        hiddenStartIndex = i;
        menuHidden = false;
        break;
      }
    }
```

In this logic, when there is only 1 child with multiple elements or even 1 element it checks for the childwidth individually. I tried  
``` JAVASCRIPT
for (let j = 0; j < child.children.length; j += 1) {
          const childWidth = child.children[j].getBoundingClientRect().width;
          calcWidth += childWidth;
```
implementing it without the if-else condition will cause an error of width being undefined.

jmsv6d: @pranav300 Have you tried removing the style @dkasper-was-taken mentioned to see how that effects the width calculation?

pranav300: @jmsv6d I didn't remove the style but changed its value and it didn't seem to work at that time, But ya removing the style does solve the issue and doesn't affect the last item logic.

packages/terra-collapsible-menu-view/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ ChangeLog
 
 Unreleased
 ----------
+### Fixed
+* Fixed Collapsible issue for single item group
```
avinashg1994: Move this comment to the already existing "Fixed" section.

packages/terra-collapsible-menu-view/src/terra-dev-site/test/collapsible-menu-view/SingleItemGroup.test.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import CollapsibleMenuView from '../../../CollapsibleMenuView';
+
+const locales = ['en', 'en-AU', 'en-US', 'en-GB', 'es', 'es-US', 'es-ES', 'de', 'fi-FI', 'fr', 'fr-FR', 'nl', 'nl-BE', 'pt', 'pt-BR', 'sv', 'sv-SE', 'fr-FR', 'nl', 'nl-BE', 'pt', 'pt-BR', 'sv', 'sv-SE'];
```
avinashg1994: `'fr-FR', 'nl', 'nl-BE', 'pt', 'pt-BR', 'sv', 'sv-SE'` is repeated twice, might throw a warning for `key={locale}` at line 14.

packages/terra-collapsible-menu-view/tests/wdio/collapsible-menu-view-spec.js
```diff
@@ -31,6 +31,30 @@ Terra.describeViewports('Responsive Single Item', ['tiny', 'large'], () => {
     browser.moveToObject('#root', 0, 0);
   });
 
+  Terra.viewports('enormous', 'huge').forEach((viewport) => {
```
avinashg1994: Re-run this to update the screenshots after the changes in "SingleItemGroup.test.jsx"

PR #738, opened 2019-06-17
[terra-date-picker] Date picker as a controlled component

### LINKED ISSUE #467: Inconsistent behavior when setting date in terra-date-picker
# Issue Description
terra-date-picker allows to update the input value using the `selectedDate` prop on each render after selecting a date within the picker. However, it doesn't update the input value after manually editing the date value in the input. See example for this inconsistent behavior. 

The current selected date is 01/01/1018. Choose a different date using the picker. Click the custom blue button to reset to today’s date by passing in the current date to the `selectedDate` prop. The input value is updated. Now manually edit the date value in the input and click the blue button. Notice that it doesn’t update to today’s date.

setting date in date-picker

It is questionable which behavior is correct. terra-date-picker was intended to be an uncontrolled component in that it needs to manage all the states internally. It doesn't want consumers to manage the states since there is logic to handle things like time and DST. Regardless whether it should be controlled or uncontrolled, the above behavior is inconsistent.

We need to determine if terra-date-picker should always be uncontrolled. If so, it should not allow to externally change the date value. It currently can because of this [code](https://github.com/cerner/terra-core/blob/master/packages/terra-date-picker/src/DatePicker.jsx#L127-L135). If a controlled version of this component is desired, we need to make sure that it is still the date picker's responsibility to manage any state relating to DST and syncing with the time (especially in DateTimePicker). The controlled version would always use the value set in the prop. 

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [X] Bug
- [ ] Other

## Expected Behavior
If terra-date-picker is an uncontrolled component, it should not allow to update the date value on subsequent render. If it is a controlled component, it should always honor the passed in `selectedDate` prop regardless of workflow.

## Current Behavior
Allows to update the date value after selecting a date from the picker but does not allow to update the date value after manually editing the date value in the input.

## Steps to Reproduce
See above example.

benbcai:
### Summary
Allow terra-date-picker to be a controlled component.

### Additional Details
- Closes #467 
- Added `value` prop to be used in a controlled component.

### Deployment
- [Test example for controlled component](https://terra-framework-deploye-pr-738.herokuapp.com/#/tests/terra-date-picker/date-picker/date-picker-controlled)
- [Dev site example. The last example is the controlled component](http://localhost:8080/#/components/terra-date-picker/date-picker/date-picker)

@cerner/terra


packages/terra-date-picker/src/terra-dev-site/doc/example/DatePickerControlled.jsx
```diff
@@ -0,0 +1,65 @@
...
+  handleDateChangeRaw(event, date) {
+    this.setState({ date });
+  }
+
+  render() {
+    return (
+      <div>
+        <p>
+Selected ISO Date:
+          <span style={{ display: 'inline-block' }}>{this.state.date}</span>
```
tbiethman: Need a class for the inline style here

packages/terra-date-picker/docs/README.md
```diff
@@ -1,8 +1,18 @@
...
+
+Terra-date-picker can be composed either as a controlled or uncontrolled component. This allows you to decide the amount of control you need.
+
+### Uncontrolled Date Picker
+
+An uncontrolled date picker is the default mode. In the uncontrolled mode, a default date can be passed to the `selectedDate` prop to initially set as the default date. The date picker will manage all states internally. See the Default Date example below for implementation details.
+
+### Controlled Date Picker
+
+In a controlled date picker, the consumer will be responsible to manage the state of the date value. The date picker does not manage any of state internally. This mode give the consumer the flexibility to update the date value as needed. A controlled date picker can be composed by setting the `value` prop with a ISO date value. A handler is needed for both `onChange` and `onChangeRaw` to be notified when date selection and input change are made so the state can be updated. If both `value` and `selectedDate` are set, then `selectedDate` will have no effect. See the Default Component example below for implementation details.
```
yuderekyu:
```suggestion
In a controlled date picker, the consumer is responsible for managing the state of the date value. The date picker does not manage any of state internally. This mode give the consumer the flexibility to update the date value as needed. A controlled date picker can be composed by setting the `value` prop with a ISO date value. A handler is needed for both `onChange` and `onChangeRaw` to be notified when date selection and input change are made so the state can be updated. If both `value` and `selectedDate` are set, then `selectedDate` will have no effect. See the Default Component example below for implementation details.
```

packages/terra-date-picker/src/terra-dev-site/doc/example/DatePickerControlled.jsx
benbcai: Good call. https://github.com/cerner/terra-framework/pull/738/commits/d3647e3bcc31b6316300d192059c267647cbd3c3

packages/terra-date-picker/docs/README.md
benbcai: Updated: https://github.com/cerner/terra-framework/pull/738/commits/6102d72fa8349f66a4b50f93401428620959cbf3

packages/terra-date-picker/docs/README.md
```diff
@@ -1,8 +1,18 @@
 # Terra Date Picker
 
-DatePicker is a controlled input component that provides users a way to enter or select a date from the date picker. DatePicker is essentially a wrapper for [react-datepicker][1] and leverages many of its props. One important difference between DatePicker and [react-datepicker][1] is that all of the date props in [react-datepicker][1] must be a [moment][2] object whereas the date props in DatePicker are ISO 8601 representation of the date.
+DatePicker is an input component that provides users a way to enter or select a date from the date picker. DatePicker is essentially a wrapper for [react-datepicker][1] and leverages many of its props. One important difference between DatePicker and [react-datepicker][1] is that all of the date props in [react-datepicker][1] must be a [moment][2] object whereas the date props in DatePicker are ISO 8601 representation of the date.
```
jmsv6d: Not related to these changes, but we should probably remove the references to react-datepicker since we pulled that code into terra-date-picker

packages/terra-date-picker/docs/README.md
```diff
@@ -1,8 +1,18 @@
...
+
+Terra-date-picker can be composed either as a controlled or uncontrolled component. This allows you to decide the amount of control you need.
+
+### Uncontrolled Date Picker
+
+An uncontrolled date picker is the default mode. In the uncontrolled mode, a default date can be passed to the `selectedDate` prop to initially set as the default date. The date picker will manage all states internally. See the Default Date example below for implementation details.
+
+### Controlled Date Picker
+
+In a controlled date picker, the consumer is responsible for managing the state of the date value. The date picker does not manage any of state internally. This mode gives the consumer the flexibility to update the date value as needed. A controlled date picker can be composed by setting the `value` prop with an ISO date value. A handler is needed for both `onChange` and `onChangeRaw` to be notified when date selection or input change is made so the state can be updated. If both `value` and `selectedDate` are set, then `selectedDate` will have no effect. See the Default Component example below for implementation details.
```
jmsv6d:
```suggestion
In a controlled date picker, the consumer is responsible for managing the state of the date value. The date picker does not manage any of state internally. This mode gives the consumer the flexibility to update the date value as needed. A controlled date picker can be composed by setting the `value` prop with an ISO date value. A handler is needed for both `onChange` and `onChangeRaw` to be notified when date selection or input change is made so the state can be updated. If both `value` and `selectedDate` are set, then `selectedDate` will have no effect. See the Controlled Component example below for implementation details.
```

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -82,10 +82,14 @@ const propTypes = {
...
   required: PropTypes.bool,
   /**
-   * An ISO 8601 string representation of the initial value to show in the date input.
-   * This prop name is derived from react-datepicker but is analogous to value in a form input field.
+   * An ISO 8601 string representation of the default value to show in the date input.
+   * This prop name is derived from react-datepicker but is analogous to defaultvalue in a form input field.
    */
   selectedDate: PropTypes.string,
+  /**
+   * The date value.
```
jmsv6d: I know this is already in the docs above but we should also indicate here that the value prop should only be used for a controlled date picker, that the `selectedDate` prop will have no effect if this is set, and that both `onChange` and `onChangeRaw` must be set as well.

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -330,15 +344,36 @@ class DatePicker extends React.Component {
...
 
+    let formattedValue = DateUtil.strictFormatISODate(value, dateFormat);
+
+    if (!formattedValue) {
+      formattedValue = value;
+    }
+
+    let selectedDateInPicker;
+
+    // If using this as a controlled components
```
jmsv6d: nit:
```suggestion
    // If using this as a controlled component
```

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -330,15 +344,36 @@ class DatePicker extends React.Component {
...
+    if (value !== undefined) {
+      selectedDateInPicker = DateUtil.createSafeDate(value);
+
+      // If value is not a value date, keep the previous selected date in the picker.
```
jmsv6d:
```suggestion
      // If value is not a valid date, keep the previous selected date in the picker.
```

packages/terra-date-picker/docs/README.md
benbcai: Updated. https://github.com/cerner/terra-framework/pull/738/commits/216700ce9829692129df8c0b3853af7a24c2e4f5

packages/terra-date-picker/docs/README.md
benbcai: Updated: https://github.com/cerner/terra-framework/pull/738/commits/216700ce9829692129df8c0b3853af7a24c2e4f5

packages/terra-date-picker/src/DatePicker.jsx
benbcai: Updated: https://github.com/cerner/terra-framework/pull/738/commits/216700ce9829692129df8c0b3853af7a24c2e4f5

packages/terra-date-picker/src/DatePicker.jsx
benbcai: Updated: https://github.com/cerner/terra-framework/pull/738/commits/216700ce9829692129df8c0b3853af7a24c2e4f5

packages/terra-date-picker/src/DatePicker.jsx
benbcai: Updated: https://github.com/cerner/terra-framework/pull/738/commits/216700ce9829692129df8c0b3853af7a24c2e4f5

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -140,10 +146,19 @@ class DatePicker extends React.Component {
   }
 
   static getDerivedStateFromProps(nextProps, prevState) {
-    if (nextProps.selectedDate !== prevState.prevPropsSelectedDate) {
+    const selectedDate = nextProps.value || nextProps.selectedDate;
```
StephenEsser: Is an empty string a valid value? If so this will fall through and create an uncontrolled component. This may behave strangely and switch between controlled and uncontrolled within the same workflow. If an empty string is valid we may need to update the logic check. If it's not valid this should be fine.

benbcai: Yes, an empty is a valid value used to clear the date. Updated the logic to accept an empty string as the value: https://github.com/cerner/terra-framework/pull/738/commits/8a0d1a460950bd9fb1406f89ca4bdd73932bd2f0

mmalaker:
Functional verification completed. The controlled date example displays and functions as expected.

PR #740, opened 2019-06-17
Update react-final-form dependency version range

bjankord:
### Summary
Updates terra-form-validation to allow it to pull in react-final-form v6. The current code in terra-form-validation is compatible with both v5 and v6 versions of react-final-form.

This will currently pull in react-final-form v6.2.1

### DIFF
--- a/packages/terra-form-validation/package.json
+++ b/packages/terra-form-validation/package.json
@@ -25,12 +25,12 @@
     "final-form": "^4.6.0",
     "react": "^16.8.5",
     "react-dom": "^16.8.5",
-    "react-final-form": "^5.0.2"
+    "react-final-form": ">=5.0.2"
   },
   "devDependencies": {
     "final-form": "^4.6.0",
     "prop-types": "^15.5.8",
-    "react-final-form": "^5.0.2",
+    "react-final-form": ">=5.0.2",
     "terra-button": "^3.3.0",
     "terra-date-picker": "^3.0.0",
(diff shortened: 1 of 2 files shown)


packages/terra-form-validation/package.json
```diff
@@ -25,12 +25,12 @@
     "final-form": "^4.6.0",
     "react": "^16.8.5",
     "react-dom": "^16.8.5",
-    "react-final-form": "^5.0.2"
+    "react-final-form": ">=5.0.2"
```
emilyrohrbough: Should we cap this at v6? If v7 is released, this would accept it.

```suggestion
    "react-final-form": ">=5.0.2 <7.0.0"
```

bjankord: Yeah, that sounds good

packages/terra-form-validation/package.json
```diff
@@ -25,12 +25,12 @@
...
     "react": "^16.8.5",
     "react-dom": "^16.8.5",
-    "react-final-form": "^5.0.2"
+    "react-final-form": ">=5.0.2 <7.0.0"
   },
   "devDependencies": {
     "final-form": "^4.6.0",
     "prop-types": "^15.5.8",
-    "react-final-form": "^5.0.2",
+    "react-final-form": ">=5.0.2",
```
emilyrohrbough:
```suggestion
    "react-final-form": ">=5.0.2 <7.0.0"",
```

PR #745, opened 2019-06-18
Update packages to use terra-responsive-element v5

bjankord:
### Summary
* Updates elements to use responsive-element v5
* Removes responsive-element from packages that were not using it


packages/terra-application-header-layout/src/terra-dev-site/doc/example/HeaderWireframe.jsx
```diff
@@ -48,10 +48,10 @@ const HeaderWireframe = () => {
     <div className={cx('content-wrapper')}>
       <div className={cx('responsive-element-wrapper')}>
         <ResponsiveElement
-          defaultElement={tinyHeader}
           tiny={tinyHeader}
-          small={smallHeader}
-          medium={mediumHeader}
+          small={tinyHeader}
```
StephenEsser: We should be able to delete the `small` prop here, it will fallback to what's defined for `tiny`

packages/terra-application-header-layout/src/terra-dev-site/doc/example/HeaderWireframe.jsx
```diff
@@ -48,10 +48,10 @@ const HeaderWireframe = () => {
...
+          medium={smallHeader}
+          large={mediumHeader}
```
StephenEsser: Do we want to rename these elements to match the new breakpoints? Probably not a big deal either way

bjankord: Updated: https://github.com/cerner/terra-framework/pull/745/commits/d667a6c9214d2ba6c1d71e62da258d77776a9cb7

PR #749, opened 2019-06-21
[terra-date-picker] pre-selection date screen reader

### LINKED ISSUE #632: Using the date picker popup with a screen reader does not provide feedback on current state or selection
# Bug Report

## Description
Currently, the date picker provides no context to screen readers via descriptive text of what date is pre-selected as the user navigates the date picker popup. Along with this, once we enable the ability to navigate to the month and year select elements in the date picker popup, we would want to label those as well.

## Steps to Reproduce

1. Using JAWS or VoiceOver, open the date picker popup
2. Use keyboard to navigate through days
3. Notice that there is no indication of what day/month/year that your on as you navigate the date picker

## Expected Behavior
As you navigate the date picker with a screen reader, you should receive feedback of date you've preselected in the date picker.

## Possible Solution
Look into using aria-live to announce `this.state.preSelectedDate._d`. This returns the moment date object for the pre selected date in the calendar as the user navigates around via the keyboard. This value would need to be formatted into a localized value that is human readable.

benbcai:
### Summary
Added functionality to announce the pre-selected date to screen readers when navigating within the date picker.

### Additional Details
- Closes #632 

@cerner/terra


packages/terra-date-picker/src/react-datepicker/index.jsx
```diff
@@ -363,9 +373,30 @@ export default class DatePicker extends React.Component {
...
+      this.updateAriaLiveStatus(ariaLiveMessage);
     }
   }
 
+  updateAriaLiveStatus(message) {
+    this.ariaLiveStatus = message;
+
+    // Clears status so aria live announces correctly next time a date is preselected.
+    setTimeout(() => {
+      this.ariaLiveStatus = '';
```
ryanthemanuel: How did we arrive at 1 second for this?

benbcai: I'm using the same timeout as [select](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/_Menu.jsx#L253) and [infinite list](https://github.com/cerner/terra-framework/blob/master/packages/terra-infinite-list/src/InfiniteList.jsx#L166). 1 second is enough to allow enough time for the re-render and announce the text. Half a second might be a little too quick to reset it.

bjankord:
Tested the updates on iOS with VoiceOver and the dates are read as expected

mmalaker:
Functional verification completed. The preselected date picker displays and functions as expected with screenreaders. Note that there are navigation issues using JAWS within the calendar control. Those will be addressed separately.

PR #750, opened 2019-06-21
[terra-notification-dialog] Do not check a11y of elements behind overlay

yuderekyu:
### Summary
We have a test that does a11y checks on a notification dialog disclosed on top of a modal. This includes a11y for elements behind an overlay. 

Technically a user is not expected to interact or understand what is rendered behind an overlay. The open source theme happens to not have overlay color contrast issues with elements behind it, but other themes such as clinical do.

This updates the selector passed to `Validates` to capture a screenshot+ run a11y tests of the notification dialog *only*, not the notification dialog and modal.


packages/terra-notification-dialog/tests/wdio/notification-dialog-spec.js
```diff
@@ -122,7 +122,7 @@ Terra.describeViewports('NotificationDialog with additional focus trap sources w
 
     // it opens notification dialog
     browser.click('#trigger-notification-dialog');
-    Terra.validates.element('Open notification dialog', { selector: '[class*="abstract-modal"]' });
+    Terra.validates.element('Open notification dialog', { selector: '[aria-label="Notification Dialog"]' });
```
emilyrohrbough: This won't actually update the selector for accessibility. the accessibility selector will always be document.

see: https://github.com/cerner/terra-toolkit/blob/master/src/wdio/services/TerraCommands/axe-command.js#L37

yuderekyu:
@emilyrohrbough you're right, I should have remembered that.

Do we have other options besides disabled the color contrast rule here?

emilyrohrbough:
@yuderekyu Yeah, you can disable the rule directly with the  `axeRule` option.

packages/terra-notification-dialog/tests/wdio/notification-dialog-spec.js
```diff
@@ -120,9 +120,14 @@ Terra.describeViewports('NotificationDialog with additional focus trap sources w
 
     Terra.validates.element('Close popup', { selector: '[class*="abstract-modal"]' });
 
+    // Skip color contrast check for elements behind the notification dialog's overlay
+    const ignoredA11y = {
+      'color-contrast': { enabled: false },
```
emilyrohrbough: Lets scope this color-contrast check to the notification dialog itself, otherwise we drop this check which is commonly seen as an accessibility issue:
```
const scopeA11y = {
    'color-contract': {
          selector: '[class*="abstract-modal]'
    }
}

```

ref: https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#parameters-1

packages/terra-notification-dialog/tests/wdio/notification-dialog-spec.js
```diff
@@ -120,9 +120,14 @@ Terra.describeViewports('NotificationDialog with additional focus trap sources w
...
+    };
+
     // it opens notification dialog
     browser.click('#trigger-notification-dialog');
-    Terra.validates.element('Open notification dialog', { selector: '[aria-label="Notification Dialog"]' });
+    Terra.validates.element('Open notification dialog', { selector: '[class*="abstract-modal', axeRules: ignoredA11y });
```
emilyrohrbough: Missing `]` in selector

yuderekyu: ugh. really dropping the ball on this pr - updated

packages/terra-notification-dialog/tests/wdio/notification-dialog-spec.js
yuderekyu: good call. updated

yuderekyu: actually overlay is the component in question - updated the selector to reference that

bjankord:
Talked about logging an issue to toolkit to allow passing a selector for accessibility check.

PR #761, opened 2019-06-27
[terra-navigation-side-menu] Update selection a11y for menu items

### LINKED ISSUE #284: VoiceOver issues with navigation-side-menu
# Bug Report

## Description
When using VoiceOver on iOS, no indication was given of which item was selected in the [side navigation menu example](https://engineering.cerner.com/terra-framework/#/components/terra-navigation-side-menu/navigation-side-menu/navigation-side-menu). 
 VO also did not indicate the menu items where interactable (a button or a link).

## Steps to Reproduce

1. Navigate to https://engineering.cerner.com/terra-framework/#/components/terra-navigation-side-menu/navigation-side-menu/navigation-side-menu
2. Interact with side navigtion menu with VoiceOver

## Expected Behavior
* Selected side navigation menu item is communicated to screen reader users.
* Screen reader users are made aware of interactable elements in side navigation.

## Environment

* Component Name and Version: terra-navigation-side-menu

yuderekyu:
### Summary
Use an aria-label for what has been selected. 

Desktop safari does not detect aria-selected. Mobile safari does, resulting in Voiceover saying "selected" twice. Because of this, @bjankord and I have inserted browser detection for safari, similar to [form-select](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/_SharedUtil.js#L6-L13) to conditionally set aria-selected

### Deployment Link
https://terra-framework-deploye-pr-761.herokuapp.com/#/components/terra-navigation-side-menu/navigation-side-menu/navigation-side-menu

Closes #284.


yuderekyu:
Testing this on Jaws - `selected` is announced twice, once before the menu item title (expected), and once after the title (unexpected). If the visual hidden text is removed, no selection state is announced.

Voice over works as intended.

packages/terra-navigation-side-menu/src/_MenuItem.jsx
```diff
@@ -107,6 +108,7 @@ class MenuItem extends React.Component {
           onKeyUp={this.handleKeyUp}
           onBlur={this.handleOnBlur}
         >
+          <VisuallyHiddenText aria-atomic aria-live="assertive" text={isSelected ? 'Selected' : ''} />
```
emilyrohrbough: Need translations for this. Looks like form-select has this word translated already: https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/translations/en.json#L8

packages/terra-navigation-side-menu/src/_MenuItem.jsx
```diff
@@ -92,22 +101,26 @@ class MenuItem extends React.Component {
...
     /* eslint-disable jsx-a11y/no-static-element-interactions, jsx-a11y/no-noninteractive-tabindex */
     return (
       <li
         className={cx('list-item')}
-        aria-selected={isSelected}
-        role="option"
+        role="presentation"
       >
         <div
+          aria-selected={isSafari ? false : isSelected} // Mobile safari detects both aria-selected & aria-label. Desktop does not. This stops mobile from announcing selected twice.
```
bjankord: :+1:

packages/terra-navigation-side-menu/src/_MenuItem.jsx
bjankord: Updated: https://github.com/cerner/terra-framework/pull/761/commits/732e665965ecc0e5d9ed8a01c8c26a6cef892d1a

packages/terra-navigation-side-menu/src/_MenuItem.jsx
```diff
@@ -1,18 +1,26 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
+import { injectIntl, intlShape } from 'react-intl';
 import KeyCode from 'keycode-js';
 import ChevronRight from 'terra-icon/lib/icon/IconChevronRight';
 
 import styles from './MenuItem.module.scss';
 
+const isSafari = (navigator.userAgent.indexOf('Safari') !== -1 && navigator.userAgent.indexOf('Chrome') === -1);
```
ryanthemanuel: Let's add a comment that explains that we only want to do this type of thing sparingly and this is specifically for an accessibility issue between browsers.  This should not be the pattern that teams should use, etc.

yuderekyu: added https://github.com/cerner/terra-framework/pull/761/commits/3fc811e96c6bcc064e1bc211bf767438514c6cd9, take a look

benbcai:
Looks like the es.json translation file is missing.

yuderekyu:
Discussed offline - plan is to investigate switching menu items to links and use aria-current to announce selection

bjankord:
Looks like the `nl.json` file may be missing translations.

bjankord:
Talked with Derek offline about a potential update that may resolve issues with JAWS in Chrome. Want to test having a single aria-live container that we write updates to based on state changes in the navigation side menu similar to how we handle aria-live updates in the infinite list.

packages/terra-navigation-side-menu/src/NavigationSideMenu.jsx
```diff
@@ -223,14 +249,25 @@ class NavigationSideMenu extends Component {
     }
 
     return (
-      <ContentContainer {...customProps} header={header} fill className={sideMenuContentContainerClassNames}>
-        {this.buildListContent(currentItem)}
-      </ContentContainer>
+      <Fragment>
+        <span
```
yuderekyu: should we use `terra-visually-hidden-text` directly instead of copying it? we're already pulling it in the package.json

bjankord: We _should_ be using it, however it currently doesn't support ref forwarding which is needed to write the `innerText` which is needed for IE 11 support for aria-live to be read by JAWS correctly. I'm planning to log an issue to terra-core to add forward ref to visually hidden and then we can come back and update this and remove the associated styles.

mmalaker:
Functional verification completed. The navigation side menu example now displays and functions as expected with supported screen readers on desktop and mobile platforms.

yuderekyu:
oops, we need to update snapshots within application-layout

PR #767, opened 2019-07-05
Added Focus check for meridiem in onChange to Fix reset on Left and Right key press

### LINKED ISSUE #710: [time-input] Meridiem resets on left keyboard input in firefox
# Bug Report

## Description
When the meridiem is focused, if the user presses the left arrow key to navigate to the minute input then the meridiem will be set to a.m. This does not occur with backspace or delete keyboard inputs.

https://engineering.cerner.com/terra-ui/#/components/terra-time-input/time-input/time-input

## Steps to Reproduce
1. Use firefox
2. Visit the link above
3. Click or keyboard navigate to the meridiem in the "With Default Time - 12 Hour" example
4. Press the left arrow key

supreethmr:
### Summary
fixes #710 

It is default behaviour for left and right arrow keys in Firefox to navigate between options in html native select. In TimeInput case whenever we press left or right key it triggers onChange event of Meridiem which changes the value selected in Meridiem. 

onChange event of `<select>` is non-cancelable event : [MDN ref](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event) which cannot not be prevented with event.preventDefault(). hence event.preventDefault() in meridiem keydown event was not stopping onChange event on left key press.


nramamurth:
The original issue talks about left arrow key behavior. Although the left arrow keypress seems to be fixed, I noticed that similar behavior still persists in the following workflow on the 'time-input-meridiem' branch when the user uses right arrow to navigate: 

1. Use Firefox.
2. Start the server and access the local site - /components/terra-time-input/time-input/time-input
3. Click or keyboard navigate to the meridiem in the "With Default Time - 12 Hour" example - the initial value should read "09:25 p.m.".
4. Press the up arrow key and ensure that the value is changed to "09:25 a.m."
5. Press the right arrow key and notice that the value changes to "09:25 p.m."

ShettyAkarsh:
> The original issue talks about left arrow key behavior. Although the left arrow keypress seems to be fixed, I noticed that similar behavior still persists in the following workflow on the 'time-input-meridiem' branch when the user uses right arrow to navigate:
> 
> 1. Use Firefox.
> 2. Start the server and access the local site - /components/terra-time-input/time-input/time-input
> 3. Click or keyboard navigate to the meridiem in the "With Default Time - 12 Hour" example - the initial value should read "09:25 p.m.".
> 4. Press the up arrow key and ensure that the value is changed to "09:25 a.m."
> 5. Press the right arrow key and notice that the value changes to "09:25 p.m."
> 
> Is this an expected behavior or is it something that needs to be fixed?

I think the UP arrow one is the expected behavior as it similar across all the browsers. upon up and down arrow press it should toggle between AM , PM  only the right arrow one seems to be like similar to the issue reported.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -341,11 +341,13 @@ class TimeInput extends React.Component {
...
-    this.setState({
-      meridiem: event.target.value,
-    });
+    if (this.state.meridiemFocused) {
+      this.setState({
+        meridiem: event.target.value,
+      });
 
-    this.handleValueChange(event, TimeUtil.inputType.HOUR, this.state.hour.toString(), event.target.value);
+      this.handleValueChange(event, TimeUtil.inputType.HOUR, this.state.hour.toString(), event.target.value);
```
ShettyAkarsh: extra space

supreethmr: this space is due to new if block added around this line.

supreethmr:
> The original issue talks about left arrow key behavior. Although the left arrow keypress seems to be fixed, I noticed that similar behavior still persists in the following workflow on the 'time-input-meridiem' branch when the user uses right arrow to navigate:
> 
> 1. Use Firefox.
> 2. Start the server and access the local site - /components/terra-time-input/time-input/time-input
> 3. Click or keyboard navigate to the meridiem in the "With Default Time - 12 Hour" example - the initial value should read "09:25 p.m.".
> 4. Press the up arrow key and ensure that the value is changed to "09:25 a.m."
> 5. Press the right arrow key and notice that the value changes to "09:25 p.m."
> 
> Is this an expected behavior or is it something that needs to be fixed?

- this is expected behaviour for ```select ```in Mozilla Firefox. Firefox supports navigation for all four keys ( ⬆️ ⬅️ ⬇️ ➡️ ) unlike other browsers. 

- this behaviour of Firefox triggers non-cancellable on-change event of `select`(meridiem) on keypress. 

- on-change event of `<select>` is immune to methods like event.PreventDefault(), stopPropagation(), stopImmediatePropagation () and cannot be blocked from executing on key press event.  

- So Probably I can add condition for right key too in [meridiem_KeyDown_event() ](https://github.com/cerner/terra-framework/pull/767/commits/ac75751cabfc1daac4207b29bbb0ad4682bf3ddf) by doing this focus will move to minute(mm) and keeps meridiem unchanged.  

But these will change the existing behavior of time Input in other browsers. In other browsers focus remains on meridiem on press of both right arrow key. with these code changes focus will move to minute(mm) even with right key press.  Any suggestions on this will be helpful.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -528,6 +530,7 @@ class TimeInput extends React.Component {
 
   handleMeridiemInputKeyDown(event) {
     if (event.keyCode === KeyCode.KEY_LEFT
+      || event.keyCode === KeyCode.KEY_RIGHT
```
Blackop778: Moving _backwards_ to the minute input when right arrow is pressed seems like a bad idea. I think we should just say we can't/won't fix this.

supreethmr: Another work around would be moving focus out from meridiem to Time Input on right arrow key press : 
```
if (event.keyCode === KeyCode.KEY_RIGHT) {
  this.setState({ meridiemFocused: false });
}
```

this will prevent meridiem from changing on right key press, since focus will be on Time Input and not on meridiem.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -694,6 +696,8 @@ class TimeInput extends React.Component {
         }
       }
       event.preventDefault();
+    } else if (event.keyCode === KeyCode.KEY_RIGHT) {
+      this.setState({ meridiemFocused: false });
```
avinashg1994: This seems to have fixed the issue with the right arrow key changing the meridiem value.
But I noticed that after the right arrow key is pressed and the meridiem field no longer has focus, when the left arrow key is pressed, it skips the meridiem field and focuses the seconds/minutes field. This to me felt counter-intuitive, since I was expecting it to go back to the meridiem field, i.e. Left arrow key navigation sequence to be same as right arrow key navigation sequence.

benbcai: @avinashg1994 @supreethmr I agree that the right arrow key can change the meridiem value seems counter-intuitive. It makes to add logic to prevent that from happening. It does appear that the changes made above to set the meridiemFocused state to false when the key is KEY_RIGHT causes the focus to lose as @avinashg1994 mentioned. One option is instead of relying on using the meridiemFocused state to handle this, you could create a new variable (i.e. preventMeridiemChange) and set it to true on line 700. Then in handleMeridiemChange, you can skip the setting the meridiem value is preventMeridiemChange is true. This should fix the issue and does not cause it to lose focus.

Having said that, I heard from @neilpfeiffer that the design for #706 will be to get rid of the meridiem select that is used for the desktop breakpoint and change it to be the meridiem buttons just like on the mobile viewport. If that is the plan, then the changes made in PR will no longer be needed. I wonder if we should close this PR since the meridiem input/select will be removed in favor of the meridiem buttons.

neilpfeiffer: Thanks @benbcai - you are correct that the new design adjustments to solve for #706 will be to remove the difference between the "desktop" and "mobile", and will be removing the select, instead opting for the button-group that is used in the mobile version.

neilpfeiffer:
Closing as will be resolved by design changes being added to solve for #706

PR #768, opened 2019-07-08
[terra-framework] Updated components which interface with old React context API

### LINKED ISSUE #752: Update components which interface with old React context API
# Feature Request

## Description
In a lot of the terra components that interface with react-intl, [we've been interfacing with the old React context API](https://github.com/cerner/terra-core/blob/master/packages/terra-demographics-banner/src/DemographicsBanner.jsx#L67-L74) directly. We should update the components to use an abstraction like `FormattedMessage` or `injectIntl` to interface with react-intl's Intl context. This should allows us to more easily update to react-intl v3 when it is released.

More info on how we recommend to build components with `FormattedMessage` and/or `injectIntl`: https://engineering.cerner.com/terra-ui/#/getting-started/terra-ui/internationalization/building-components-which-include-translations

avinashg1994:
### Summary
Updated all the components in terra-framework which interface with old React Context API, which include:

1. Application Links
2. Collapsible Menu View
3. Date Picker
4. Date Time Picker
5. Notification Dialog
6. Tabs
7. Time Input

To use react-intl's context (FormattedMessage or injectIntl).

Resolves #752 

### Additional Details
Also updated the jest tests, which required changes because of the use of InjectIntl


packages/terra-application-links/src/tabs/_TabMenu.jsx
```diff
@@ -86,7 +82,7 @@ class TabMenu extends React.Component {
 
   createDisplay(popup) {
     const { location } = this.props;
```
nramamurth: nit: this could be a single line - 

```
const { location, intl } = this.props;
```

avinashg1994: Done, changed as part of : cc5007b

packages/terra-application-links/src/tabs/_TabMenu.jsx
```diff
@@ -11,6 +12,10 @@ const propTypes = {
    * Child tabs to be placed in the tab menu.
    */
   children: PropTypes.array,
+  /**
+   * intl object programmatically imported through injectIntl from react-intl.
+   * */
+  intl: intlShape.isRequired,
```
nramamurth: Please use the `@private`syntax wherever applicable. Reference - https://github.com/cerner/terra-core/pull/2507/files#r303638053

avinashg1994: Done, changed as part of : d5474e9

PR #771, opened 2019-07-08
Enable focus shift into modal dialogs when using VoiceOver on iOS

### LINKED ISSUE #736: Utility Button is Not Properly Highlighted in iOS
# Bug Report

## Description
This appears to be the same issue as: https://github.com/cerner/terra-framework/issues/277 but scoped to iOS

When using Terra Application layout in iOS view, clicking the utility button does not change focus to the utility menu. 

## Steps to Reproduce
1. Go to https://engineering.cerner.com/terra-framework/#/home on iOS with Screen Reader (VoiceOver) turned on
2. Select utility menu in small (hamburger) layout
3. Note that focus is not shifted, instead you are stuck on 'utility button - button' 

## Expected Behavior
Utility button will switch focus to utility menu, as it does in other layouts. 

## Environment
Although reproducible on https://engineering.cerner.com/terra-framework/#/home/terra-ui/index, here's information on our system
* Component Name and Version:  terra-application-layout, 4.8.0
* Browser Name and Version: Safari 12
* Node/npm Version: lts/carbon in our app
*  Webpack Version: 3.11.0
* Operating System and version (desktop or mobile): iOS 12

bjankord:
### Summary
This resolves an issue that occurred where the virtual indicator in VoiceOver on iOS was not being shifted into the modal dialog correctly when the modal was opened.  Prior to this PR, we had shifted focus to the div with the aria-modal="true" on it. This worked well with other screen readers but doesn't appear to work with VoiceOver on iOS. Prior to this behavior we had been using [focus-trap-react](https://github.com/davidtheclark/focus-trap-react) which uses [tabbable](https://github.com/davidtheclark/tabbable) to shift focus to the first interactive element within the modal.

The fix in this PR re-introduces behavior from focus-trap-react via [tabbable](https://github.com/davidtheclark/tabbable) where the first interactive DOM element inside the modal will receive focus when the modal is open.

In the case that there is no interactive content within the modal, it will fallback to reading screen reader help text that indicates the beginning of the modal dialog.

Resolves #736 

### Deployment URLs:
* https://terra-framework-deploye-pr-771.herokuapp.com/#/components/terra-abstract-modal/abstract-modal/abstract-modal
* https://terra-framework-deploye-pr-771.herokuapp.com/#/tests/terra-abstract-modal/abstract-modal/abstract-modal-content-overflow
* https://terra-framework-deploye-pr-771.herokuapp.com/#/components/terra-application-layout/application-layout/example


packages/terra-modal-manager/tests/wdio/modal-manager-spec.js
```diff
@@ -345,4 +325,18 @@ Terra.describeViewports('ModalManager - Behaviors', ['large'], () => {
       });
     });
   });
+
+  describe('Component Integration', () => {
```
bjankord: Moved this to be its own separate describe block outside of the focus management describe block.

packages/terra-abstract-modal/src/AbstractModal.jsx
```diff
@@ -117,11 +118,15 @@ class AbstractModal extends React.Component {
...
         mainDocumentElement.setAttribute('data-overlay-count', `${inert + 1}`);
-        this.modalElement.current.focus();
+      }
+      // Shift focus to first tabbable element in modal
+      if (tabbable(this.modalElement.current)[0]) {
+        tabbable(this.modalElement.current)[0].focus();
+      } else {
+        // If no focusable elements are withing the modal, shift to modal dialog begin help text
+        this.modalElement.current.querySelector('[data-terra-abstract-modal-begin]').focus();
       }
```
dev-cprice: I ran into this when working on the select's mobile modal as well, but I like this logic. Seems about as safe as we can get.

However, I think we'll still run into some issues on iOS VoiceOver, but they're gonna be really hard to combat. As in, the only thing we can really do is log a bug to Apple.

I believe you've seen this as well, but iOS VoiceOver doesn't always listen to you when you say

>"focus on this element, it has a tabIndex set".

It seems to only like to move the virtual indicator to interactable elements, like buttons, anchors, etc, when programmatically manipulating the focus..

dev-cprice: Looking below, I see that `data-terra-abstract-modal-begin` has a `tabIndex="-1"`. So, what happens here with VoiceOver when there's not a `tabbable` element in the modal (should be rare, hopefully)?

bjankord: You're correct, VoiceOver on iOS is very picky about which elements it shifts the virtual indicator to when you programmatically shifting focus. I've found from [testing this page](https://codepen.io/backwardok/pen/xpPZKL) that VoiceOver on iOS will shift focus to a `span` or `div` with `tabIndex="-1"` set on it when you programmatically shift focus to the element **if** the `span` / `div` only contains a string of text within it. 

If you nest other elements within the `span` or `div`, then VoiceOver on iOS doesn't seem to allow shifting the virtual indicator to a `span` or `div` with `tabIndex="-1"` on via programmatic focus shift. I agree that this seems like an Apple bug.

Based on this, I do have logic that will account for this without making the code to purpose built for VoiceOver. When the modal opens, we look for interactable elements via tabbable, and focus the first one we find within the modal, if in the rare case people don't include any interactable elements, focus will shift to visually hidden text with the tabIndex="-1" on it that indicates the beginning of the modal dialog. This allows the focus to move into the modal on iOS when using VoiceOver even if people forget or choose not to add interactable elements to the modal.

This can be tested on the following page: https://terra-framework-deploye-pr-771.herokuapp.com/#/tests/terra-abstract-modal/abstract-modal/abstract-modal-no-focusable-content

dev-cprice: Interesting.. I guess that would explain why when using iOS VoiceOver with the `terra-select`, it still navigates to the visually hidden text

image

Here it's focused (for lack of a better word) on this [hidden "Selected" text](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/_Frame.jsx#L229-L231)

bjankord: Yeah, I believe the virtual indicator will jump to DOM elements containing text as the user swipes left and right on iOS when using VoiceOver. Since that visually hidden text is in the DOM, VoiceOver will move the virtual indicator to it as the user navigates the page with left/right swipe gestures.

mmalaker:
Functional verification completed. The modal dialog examples display and function as expected. Focus moves into dialogs correctly on mobile and screen readers will now read the full content of alert modals.

PR #774, opened 2019-07-10
[application-layout] Ignore Axe Rules Removed

### LINKED ISSUE #715: [application-layout] Remove Ignore Axe Rules When Axe Bug is Resolved
# Feature Request

## Description
Application-layout adds axe-ignore rules to the wdio tests due to a bug in axe-core. Once a fix is released in axe-core, we should remove these.

Axe Issue: https://github.com/dequelabs/axe-core/issues/1583
Remove from: [application-layout-spec.js](https://github.com/cerner/terra-framework/blob/master/packages/terra-application-layout/tests/wdio/application-layout-spec.js#L1-L6)

saket2403:
### Summary
The ignore axe rules were removed from the wdio tests.
Resolves #715


packages/terra-application-layout/tests/wdio/application-layout-spec.js
```diff
@@ -1,17 +1,12 @@
 // This issue should be resolved once the following axe-core issue is resolved:
 // https://github.com/dequelabs/axe-core/issues/1583
-const ignoredA11y = {
```
tbiethman: You could remove the above comment at this point too

PR #779, opened 2019-07-11
Fixed Divider Transition between normal and in-menu view

### LINKED ISSUE #770: [terra-collapsible-menu-view] fix divider transition between normal and in-menu
## Description
the Divider in the collapsible menu unexpectedly transitions between the face-up/normal and in-menu like it is independent, so the overflow menu looks unexpected if the divider rolls-up into the menu, but nothing else with the preceding item-group does (screenshot below). 

Resolve by updating the transition of when the divider hides itself from the face-up normal view to when the following/after item-group is fully in the menu, but waits to be added into the menu until at least one item of the item-group preceding/before the divider is also rolled-up into the menu.

Screen Shot 2019-07-08 at 8 16 46 AM

supreethmr:
### Summary
resolves #770 

### Additional Details
Resolved by updating the transition of the divider  to hide itself from the face-up normal view only when there is at least one item of the item-group preceding/before the divider.


packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
```diff
@@ -111,7 +111,11 @@ class CollapsibleMenuView extends React.Component {
           break;
         }
 
-        hiddenStartIndex = i;
+        if (this.props.children.length !== undefined && this.props.children[i].type === CollapsibleMenuViewDivider) {
```
nramamurth: `this.props.children.length !== undefined` check isn't really needed.

supreethmr: this throws error on accessing children by index `this.props.children[i]` in few scenario's like CollapsibleMenuViewSingleItem where this.props.children will have single object.

nramamurth: Ah, gotcha.

packages/terra-collapsible-menu-view/CHANGELOG.md
```diff
@@ -3,11 +3,14 @@ ChangeLog
 
 Unreleased
 ----------
+### Fixed
+* Fixed divider transition between normal and in-menu view.
```
avinashg1994: This log should be moved under the other "Fixed" section.

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
supreethmr: Thank you,  yes React.Children.count() suits the purpose of this condition. Added comments above if :  c7ee6e81d451dbafd42cfcf8062038073eae9319

packages/terra-collapsible-menu-view/CHANGELOG.md
supreethmr: Can we do that..? Please let me know if i am wrong, I think while release it will be moved under single header until then this should be in the order changes were made.

supreethmr:
> Whoops. Nevermind earlier comment.
> Could a test be added to prevent regression?

supreethmr:
> Whoops. Nevermind earlier comment.
> Could a test be added to prevent regression?

do you mean to say test like this : 3a0e07713c677eec29979b4e62652447b4cc32e1

Blackop778:
It seems that the test passes with or without the changes to `CollapsibleMenuView.jsx`

supreethmr:
> It seems that the test passes with or without the changes to `CollapsibleMenuView.jsx`

yes, Change was just to stop divider from getting added to collapsible menu if divider was the last one to get hidden. Due to this I didn't considered adding WDIO Tests initially while opening PR.

neilpfeiffer:
Tested: PR #779
Resolves: Issue #770
 
**Reviewed:**
- resolves issue of divider moving into menu too early.
 
Marking as Verified, ready for Release.
 
**Pass** |
-----------|
UX Review|

PR #784, opened 2019-07-15
[terra-date-time-picker] Allow consumers' onBlur to be invoked after DST resolution.

### LINKED ISSUE #758: [terra-date-time-picker] onBlur called before/during DST resolution, selections lost
# Bug Report

## Description
This relates to https://github.com/cerner/terra-framework/issues/521 and https://github.com/cerner/terra-framework/issues/522. In short, with 521 requested for the ambiguous date resolution to occur before onChange, as onChange is what was causing our component to call our service to update the date. With 522 that was going to change to the newly consolidated onBlur, meaning we would need the ambiguous date resolution to occur before onBlur as well.

521 went through and worked well for our onChange implementation. Once 522 was completed and we updated our implementation (making it _much_ cleaner and easier for us, thank you again), DST issues reappeared.

## Steps to Reproduce
1. Have some external service called when onBlur is invoked from a date-time component.
2. Select an ambiguous hour (Nov 3, 1:15) and navigate away from the component.
3. Modal appears simultaneously with update going through, the value from the modal is essentially lost.

Below you can see that the time zone is not being honored.
image
image
image
image
These screenshots were generated by updating to the ambiguous hour, selecting and deselecting the component, which generates the modal (our wrapper will not call our service if the underlying time did not change), and picking one and then the other of the offsets. Despite changing the offset, the underlying time did not change.

## Expected Behavior
If the ambiguous hour is entered, onBlur is not called until after the modal is dismissed. The modal does not put focus back into the other parts of the component. To a consumer/user, interacting with the modal is still part of modifying the time, and onBlur (at least as requested in 522) is a signal that the user is done modifying the time.

## Possible Solution
Check if we're in the ambiguous hour prior to calling the onBlur prop, call the onBlur prop whenever the modal is dismissed.

## @ Mentions
@benbcai

nramamurth:
### Summary
Fixes #758

### Additional Details
I updated an existing _wdio_ test that was supposed to test the Standard offset click in the Time Clarification dialog.


packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -242,10 +253,8 @@ class DateTimePicker extends React.Component {
       value.trim();
```
Blackop778: Not exactly related to this PR but `trim()` doesn't modify the string it is called on so this needs to be changed to 
```suggestion
      value = value.trim();
```

packages/terra-date-time-picker/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Fixed
```
Blackop778: Could this be considered a breaking change?

nramamurth: Ah, that is a great question! 

I am not sure. @benbcai @ryanthemanuel should this be a MVB?

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -11,6 +11,7 @@ Terra.describeViewports('DateTimePicker', ['tiny', 'large'], () => {
   describe('Default', () => {
     before(() => {
       browser.url('/#/raw/tests/terra-date-time-picker/date-time-picker/date-time-picker-default');
+      browser.refresh();
```
nramamurth: For some weird reason, I had to add this to fix the following accessibility violation. I've no clue what suddenly set the `aria-hidden` attribute on the `root` element.

<details>
 <summary>WDIO Axe Error Report</summary>

```
{
  "description": "Ensures aria-hidden elements do not contain focusable elements",
  "help": "ARIA hidden element must not contain focusable elements",
  "helpUrl": "https://dequeuniversity.com/rules/axe/3.2/aria-hidden-focus?application=axeAPI",
  "id": "aria-hidden-focus",
  "impact": "serious",
  "nodes": [
    {
      "all": [
        {
          "data": null,
          "id": "focusable-disabled",
          "impact": "serious",
          "message": "Focusable content should be disabled or be removed from the DOM",
          "relatedNodes": [
            {
              "html": "<input class=\"Input-module__form-input___2z_wq DatePicker-module__input___p3mYV\" placeholder=\"MM/DD/YYYY\" name=\"terra-date-input\" type=\"text\" value=\"\">",
              "target": [
                ".DatePicker-module__input___p3mYV"
              ]
            },
            {
              "html": "<button class=\"Button-module__button___3afIE Button-module__neutral___ZyOh8 Button-module__compact___31Uah DatePicker-module__button___1KmcA\" type=\"button\" aria-disabled=\"false\" aria-label=\"Open Calendar\" title=\"Open Calendar\">",
              "target": [
                "button"
              ]
            },
            {
              "html": "<input aria-label=\"Hours\" class=\"Input-module__form-input___2z_wq TimeInput-module__time-input-hour___2VQ6_ TimeInput-module__desktop___1gtms\" placeholder=\"hh\" maxlength=\"2\" size=\"2\" name=\"terra-time-hour-input\" type=\"text\" pattern=\"\\d*\" value=\"\">",
              "target": [
                ".TimeInput-module__time-input-hour___2VQ6_"
              ]
            },
            {
              "html": "<input aria-label=\"Minutes\" class=\"Input-module__form-input___2z_wq TimeInput-module__time-input-minute___1elGu TimeInput-module__without-second___2G5o1 TimeInput-module__desktop___1gtms\" placeholder=\"mm\" maxlength=\"2\" size=\"2\" name=\"terra-time-minute-input\" type=\"text\" pattern=\"\\d*\" value=\"\">",
              "target": [
                ".TimeInput-module__time-input-minute___1elGu"
              ]
            }
          ]
        }
      ],
      "any": [],
      "failureSummary": "Fix all of the following:\n  Focusable content should be disabled or be removed from the DOM",
      "html": "<div id=\"root\" aria-hidden=\"true\">",
      "impact": "serious",
      "none": [],
      "target": [
        "#root"
      ]
    }
  ],
  "tags": [
    "cat.name-role-value",
    "wcag2a",
    "wcag412"
  ]
}
running chrome
AssertionError: expected no accessibility violations but got:
	{
  "description": "Ensures aria-hidden elements do not contain focusable elements",
  "help": "ARIA hidden element must not contain focusable elements",
  "helpUrl": "https://dequeuniversity.com/rules/axe/3.2/aria-hidden-focus?application=axeAPI",
  "id": "aria-hidden-focus",
  "impact": "serious",
  "nodes": [
    {
      "all": [
        {
          "data": null,
          "id": "focusable-disabled",
          "impact": "serious",
          "message": "Focusable content should be disabled or be removed from the DOM",
          "relatedNodes": [
            {
              "html": "<input class=\"Input-module__form-input___2z_wq DatePicker-module__input___p3mYV\" placeholder=\"MM/DD/YYYY\" name=\"terra-date-input\" type=\"text\" value=\"\">",
              "target": [
                ".DatePicker-module__input___p3mYV"
              ]
            },
            {
              "html": "<button class=\"Button-module__button___3afIE Button-module__neutral___ZyOh8 Button-module__compact___31Uah DatePicker-module__button___1KmcA\" type=\"button\" aria-disabled=\"false\" aria-label=\"Open Calendar\" title=\"Open Calendar\">",
              "target": [
                "button"
              ]
            },
            {
              "html": "<input aria-label=\"Hours\" class=\"Input-module__form-input___2z_wq TimeInput-module__time-input-hour___2VQ6_ TimeInput-module__desktop___1gtms\" placeholder=\"hh\" maxlength=\"2\" size=\"2\" name=\"terra-time-hour-input\" type=\"text\" pattern=\"\\d*\" value=\"\">",
              "target": [
                ".TimeInput-module__time-input-hour___2VQ6_"
              ]
            },
            {
              "html": "<input aria-label=\"Minutes\" class=\"Input-module__form-input___2z_wq TimeInput-module__time-input-minute___1elGu TimeInput-module__without-second___2G5o1 TimeInput-module__desktop___1gtms\" placeholder=\"mm\" maxlength=\"2\" size=\"2\" name=\"terra-time-minute-input\" type=\"text\" pattern=\"\\d*\" value=\"\">",
              "target": [
                ".TimeInput-module__time-input-minute___1elGu"
              ]
            }
          ]
        }
      ],
      "any": [],
      "failureSummary": "Fix all of the following:\n  Focusable content should be disabled or be removed from the DOM",
      "html": "<div id=\"root\" aria-hidden=\"true\">",
      "impact": "serious",
      "none": [],
      "target": [
        "#root"
      ]
    }
  ],
  "tags": [
    "cat.name-role-value",
    "wcag2a",
    "wcag412"
  ]
}

```

</details>

packages/terra-date-time-picker/CHANGELOG.md
ryanthemanuel: I think since this is a defect, while it's technically a change in behavior we are fixing the behavior to be "correct".  Let's leave it as a non MVB.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -483,6 +496,10 @@ class DateTimePicker extends React.Component {
       // again (unless they change the DateTime).
       this.props.onChange(event, newDateTime && newDateTime.isValid() ? newDateTime.format() : '');
     }
+
+    // Handle blur with the new selection upon dismiss of the Time Clarification dialog.
+    this.handleBlur(event, newDateTime);
```
benbcai: Did some testing using this deployed [example](https://terra-framework-deploye-pr-784.herokuapp.com/#/components/terra-date-time-picker/date-time-picker/events). onBlur is called once after closing the clarification dialog for the first time This appears correct. However, if I click on the offset button to open the clarification dialog and then close it, both onFocus and onBlur get called. When clicking anywhere else on the page to lose focus from the entire component, onBlur is called once again, which ends up getting called twice.

Here are the steps:
1. Go to [example](https://terra-framework-deploye-pr-784.herokuapp.com/#/components/terra-date-time-picker/date-time-picker/events) and keep track of the "onBlur Trigger Count"
2. Select 11/03/2019 as the date and 01:30 as the time to create the ambiguous hour.
3. Lose focus.
4. Select either button in the clarification dialog to close the dialog.
5. onBlur is called and the onBlur trigger count becomes 1.
6. Click outside of the entire date-time component to lose focus.
7.  onBlur is not called and the onBlur trigger count remains 1.
8. Everything up to this point appears correct.
9. Click on the offset button to relaunch the clarification dialog.
10. Select either button in the clarification dialog to close the dialog.
11. onBlur is called and the onBlur trigger count becomes 2.
12. Click outside of the entire date-time component to lose focus.
13. onBlur is called again and the onBlur trigger count remains 3. This is inconsistent with step 7 and ends up triggering onBlur twice when the consumer expects it to get triggered just once.

nramamurth: Ah, I see what is happening - Step 3 loses focus and that gets correctly handled once the Time Clarification dialog is dismissed (ambiguity is resolved in Step 4) and at this point there is no focus back in the date time picker component. 

However, when the clarification dialog is shown by explicitly clicking on the offset button, the focus is not really lost and blur should not be handled because the focus goes back to the offset button when the dialog is dismissed. 

I added a [focus outline](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-time-picker/src/_TimeClarification.module.scss#L61) for in-house testing and here's what I noticed - the button still has focus in Step 11. This makes me think that the blur should not handled when the Time Clarification dialog was opened by clicking on the offset button. 

What are your thoughts on doing something like below?

```
// Handle blur with the new selection upon dismiss of the Time Clarification dialog.
if (!this.wasOffsetButtonClicked) {
   this.handleBlur(event, newDateTime);
}
```

By doing this, the blur handler will get called (correctly) when the user clicks outside of the entire date-time component to lose focus in Step 12. 

onBlur 003

benbcai: @nramamurth Adding that check to both handleStandardTimeButtonClick and handleDaylightSavingButtonClick does fix the issue. I think it makes sense for adding this check since the wrapper date-time component hasn't lost focus as the focus is still on the offset button. And then onBlur would get called when the user manually clicks outside of the wrapper component. Could you also update the comment above the check for why this check is needed?

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -223,13 +228,19 @@ class DateTimePicker extends React.Component {
...
+          this.handleBlur(event, updatedDateTime);
+        }
+      });
     }
   }
 
-  handleBlur(event, isCompleteDateTime) {
+  handleBlur(event, momentDateTime) {
     if (this.props.onBlur) {
+      const isCompleteDateTime = DateTimeUtils.isValidDateTime(this.dateValue, this.timeValue, this.state.dateFormat);
```
benbcai: The `showSeconds` prop is need to be passed as the 4th parameter to `DateTimeUtils.isValidDateTime` to correctly validate the time value to see if the seconds precision is included.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -204,8 +204,13 @@ class DateTimePicker extends React.Component {
       const isDateTimeValid = DateTimeUtils.isValidDateTime(this.dateValue, this.timeValue, this.state.dateFormat, this.props.showSeconds);
       const enteredDateTime = isDateTimeValid ? this.state.dateTime : null;
 
-      this.checkAmbiguousTime(enteredDateTime);
-      this.handleBlur(event, isDateTimeValid);
+      this.checkAmbiguousTime(enteredDateTime, () => {
+        // If the entered time is ambiguous then do not handle blur just yet. It should be handled _after_
+        // the ambiguity is resolved (i.e., after dismissing the Time Clarification dialog).
+        if (!(this.state.isAmbiguousTime && this.state.isTimeClarificationOpen)) {
+          this.handleBlur(event, enteredDateTime);
```
benbcai: We should pass in `this.state.dateTime` as the 2nd parameter to `handleBlur` instead of `enteredDateTime`. The reason is that we want to use the date/time that reflects what the user selected (Daylight Savings or Standard Time) in the Time Clarification dialog. For example, in the Central timezone, if the user selected the `Before (Daylight Saving)` button, the `options.iSO` attribute set in `handleBlur` should have an offset of -500 where as if `After (Standard Time)` was selected, the `options.iSO` attribute should reflect an offset of -600. Passing in `this.state.dateTime` will allow `handleBlur` to use the moment object that reflects what the user selected.

Line 235 should also pass in `this.state.dateTime` as the 2nd parameter to `handleBlur`.

packages/terra-date-time-picker/src/DateTimePicker.jsx
nramamurth: Right, fixed it here: https://github.com/cerner/terra-framework/pull/784/commits/1eaae53d54c7058dd14309b33fd5e0879cea9dd4

packages/terra-date-time-picker/src/DateTimePicker.jsx
nramamurth: Ah, thanks for the heads-up. It looks like I missed the recent change that went into master. Fixed it here: https://github.com/cerner/terra-framework/pull/784/commits/1eaae53d54c7058dd14309b33fd5e0879cea9dd4

packages/terra-date-time-picker/src/DateTimePicker.jsx
nramamurth: That makes sense. Addressed it here: https://github.com/cerner/terra-framework/pull/784/commits/1eaae53d54c7058dd14309b33fd5e0879cea9dd4

PR #786, opened 2019-07-15
Update peer dependency docs

bjankord:
### Summary
Add documentation on expected handling of peer dependencies listed for terra packages.


scripts/peer-dependency-generator/findAndReplace.js
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable no-console */
```
mjhenkes: i assume we will move these to a common location in the future?

bjankord: Yeah, once we decided where we want these common scripts we use in the monorepos to live, we'll DRY them up in a single spot.

PR #791, opened 2019-07-18
Update and normalize dev dependencies

bjankord:
### Summary
This PR updates dev dependencies from terra-core that had new major versions.
It also normalizes dev dependency versions to mitigate lerna hoist warnings.

### DIFF
--- a/packages/terra-aggregator/src/terra-dev-site/doc/example/CustomRenderAggregatorExample.jsx
+++ b/packages/terra-aggregator/src/terra-dev-site/doc/example/CustomRenderAggregatorExample.jsx
@@ -27,18 +27,18 @@ const CustomRenderAggregatorExample = () => (
       render={renderData => (
         <Grid>
           <Grid.Row>
-            <Grid.Column col={6}>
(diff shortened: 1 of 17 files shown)


packages/terra-aggregator/src/terra-dev-site/doc/example/CustomRenderAggregatorExample.jsx
```diff
@@ -27,18 +27,18 @@ const CustomRenderAggregatorExample = () => (
       render={renderData => (
         <Grid>
           <Grid.Row>
-            <Grid.Column col={6}>
+            <Grid.Column tiny={6}>
```
mjhenkes: tiny={6}? What is going on here?

bjankord: The `col` prop changed to `tiny` in terra-grid v6. https://github.com/cerner/terra-core/blob/master/packages/terra-grid/docs/UPGRADEGUIDE.md#changed

PR #792, opened 2019-07-18
Added callback to handle Focus on Onchange event

### LINKED ISSUE #622: [terra-time-input] onChange triggered twice upon entering any value between 3 and 9 into the hour input
# Bug Report

## Description
The `onChange` callback prop is triggered twice when entering any value between 3 and 9 into the hour input.

## Steps to Reproduce
1. Pass in a callback to the onChange prop.
2. Go to the time-input.
3. Enter any value between 3 and 9. (A value between 3 and 9 would automatically prepend a 0 and move the focus to the minute input)
5. The onChange callback is triggered twice. It should only be triggered once upon completing the hour.

Stack trace for the first onChange trigger
1. `handleHourBlur`
2. `handleBlur`
3. `handleValueChange`
4. `onChange`

Stack trace for the second onChange trigger
1. `handleHourChange`
2. `handleValueChange`
3. `onChange`

## Expected Behavior
onChange is triggered only once.

## Possible Solution
Perhaps `handleValueChange` should not be called in handleBlur when there is a need to prepend a 0 to avoid the duplicated onChange event.

Screen Shot 2019-04-18 at 4 21 59 PM

supreethmr:
### Summary
Resolves #622 

### Additional Details
```javascript
handleHourChange() {
---------
  if (inputValue.length === 2) {
      this.minuteInput.focus();
  }

  this.handleValueChange(event, TimeUtil.inputType.HOUR, inputValue, this.state.meridiem);
}
``` 
this.minuteInput.focus() event triggers onBlur event here before this.handleValueChange() event. Since it has been triggered before setting state it will have single digit value as stateValue and event.target.value will return value without zero prepended.

```javascript
Onblur() {
-----
let stateValue = event.target.value;

if (stateValue.length === 1) {
    ------
    this.handleValueChange(event, type, stateValue, this.state.meridiem);
 }
```
this.handleValueChange() event will trigger onChange() event of time-input and onChange() will also be triggered again in handleHourChange() event (as shown above). 

Resolved this by adding callback to ensure that focus will be moved only after (hour/minute/second) state is updated.


packages/terra-time-input/src/terra-dev-site/doc/example/TimeInputSeconds.jsx
```diff
@@ -18,7 +22,7 @@ class timeInput extends React.Component {
       <div>
         <p>
           Time Provided:
-          <span style={{ display: 'inline-block' }}>{this.state.time}</span>
+          <span className={cx('time-wrapper')}>{this.state.time}</span>
```
benbcai: The changes made in here, TimeInputSecondsTwelveHour.jsx, and TimeInputSecondsWithDefault.jsx are being taken care of in #810. We can either get rid of these changes here or close #810 to keep these changes.

packages/terra-time-input/src/terra-dev-site/test/time-input/time-input/Second.test.jsx
```diff
@@ -1,8 +1,12 @@
 import React from 'react';
+import classNames from 'classnames/bind';
 import TimeInput from '../../../../TimeInput';
+import styles from './time-input.test.module.scss';
+
+const cx = classNames.bind(styles);
 
 const TimeInputDefault = () => (
-  <div style={{ caretColor: 'transparent' }}>
+  <div className={cx(['content-wrapper'])}>
```
benbcai: These changes here and in the next two files are being taken care of in #812.

supreethmr: will revert these changes here

PR #805, opened 2019-07-24
[terra-application-navigation] Fixed themable variable naming.

dkasper-was-taken:
### Summary
Corrected mislabeled themable variable that got past the lint rule.

@cerner/terra

### DIFF
--- a/packages/terra-application-navigation/src/tabs/Tab.module.scss
+++ b/packages/terra-application-navigation/src/tabs/Tab.module.scss
@@ -99,7 +99,7 @@
     &[aria-current='false']:active {
       .tab-inner {
         background-color: var(--terra-application-navigation-tab-active-background-color, rgba(3, 43, 70, 0.4));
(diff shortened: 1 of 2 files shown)


packages/terra-application-navigation/src/tabs/Tab.module.scss
```diff
@@ -99,7 +99,7 @@
     &[aria-current='false']:active {
       .tab-inner {
         background-color: var(--terra-application-navigation-tab-active-background-color, rgba(3, 43, 70, 0.4));
-        color: var(--terra-application-navigation-tab-active-background-color, rgba(255, 255, 255, 0.5));
+        color: var(--terra-application-navigation-tab-active-color, rgba(255, 255, 255, 0.5));
```
StephenEsser: This should have been flagged by the duplicate property rule. The rgba value might return an ast that needs special logic.

PR #810, opened 2019-07-26
[terra-time-input] Dev-site component inline style cleanup

### LINKED ISSUE #809: [terra-time-input] Inline Style Cleanup for dev-site  components
# Bug Report

## Description

Remove inline styles from terra-time-input dev-site components, which were introduced after the Inline Style cleanup of the terra-framework repo.

avinashg1994:
### Summary
Removed inline styles from dev-site components and replaced them using external css.

Resolves #809


avinashg1994:
> I'm going to be opening a PR later today that consumes `terra-toolkit`'s new `Terra.hideInputCaret` helper which removes these inline styles and the CSS file that hides the input caret on the `terra-time-input` so we can just close this I'd say.

Reverted the changes for the files containing only `carteColor` since addressed as part of #812. 
But keeping this PR open to address the inline styles which were not addressed as part of that PR, like the `display: 'inline-block'` as seen the "Files Changed".

packages/terra-time-input/src/terra-dev-site/doc/example/TimeInputSeconds.jsx
```diff
@@ -1,6 +1,10 @@
 import React from 'react';
 // eslint-disable-next-line import/no-unresolved, import/extensions
 import TimeInput from 'terra-time-input/lib/TimeInput';
+import classNames from 'classnames/bind';
+import styles from './TimeInput.module.scss';
```
tbiethman: I'd make a new stylesheet thats used exclusively for the example, rather than reuse the TimeInput's stylesheet. Consumers could get the wrong idea if the review the example source.

avinashg1994: My bad, I had created a new stylesheet for the example but named it the same as TimeInput's stylesheet.
I have now renamed the new stylesheet to be more appropriate.

PR #764, opened 2019-07-03
Ignore Accessibility Removed and accessibility issues fixed.

saket2403:
### Summary
Removed the Ignore Accessibility condition from terra-date -time-picker and terra-time input. Fixed duplicate Aria-Label issue for Time input elements. Added an Aria-Label for Date-Input . Updated the wdio screenshots for Date-Time-Picker.

### Additional Details
Translations will be required for the new label added in Terra- Date-Input.

### Translations JIRA Link
https://jira.cerner.com/browse/TRANS-6834

Resolves #744

PR #813, opened 2019-07-26
Update date time picker mobile tests in PR #764 to use hideInputCaret

Blackop778:
### Summary
Bridges the gap between changes in #812 and #764 (calling out @saket2403 to review this). I tried to make these changes in #812 but was having issues with not having `aria-label`s on `input`s and didn't want to merge #764 into #812 .

Notice that this PR is merging into branch `issue-744` and not master (I didn't want to commit directly to their branch without talking to them but time is limited).

With the `package.json` change #764 should probably get marked 'do not merge' after this is merged until Toolkit is released and `package.json` gets updated.


packages/terra-date-time-picker/tests/wdio/date-time-picker-mobile-spec.js
```diff
@@ -18,7 +18,7 @@ Terra.describeViewports('Date Time Picker Twelve Hour Mobile', ['tiny'], () => {
       browser.url('/#/raw/tests/terra-date-time-picker/date-time-picker/twelve-hour/date-time-picker-value-12-hour-mobile');
 
       browser.click('//*[@id="datetimeValueProvided"]/div[2]/div/div[3]/button[1]');
-      Terra.it.validatesElement();
+      Terra.validates.element();
```
saket2403: Is this change required?

Blackop778: Using `Terra.it.validatesElement();` inside an `it` block means that `Terra.it.validatesElement();` won't get ran, which is happening here. Didn't notice it until I was running the tests locally.

packages/terra-date-time-picker/src/terra-dev-site/test/date-time-picker/twelve-hour/DateTimePickerDisabled12HourMobile.test.jsx
```diff
@@ -36,15 +32,15 @@ export default class DateTimePicker12HourDisabledMobile extends React.Component
 
   render() {
     return (
-      <div className={cx('content-wrapper')}>
+      <>
```
emilyrohrbough: why was this `div` dropped? We don't normaally wrap with `<>`

Blackop778: `<>` is a shorthand for `<React.Fragment>`. Having some wrapper is needed since the fragment has two children but it doesn't need to be a div because the `div` was needed to apply the `className`.

PR #820, opened 2019-07-30
[terra-framework] Replaced old context API with react-intl's context

### LINKED ISSUE #752: Update components which interface with old React context API

avinashg1994:
### Summary
Replaced the old context API with react-intl's context (preferably FormattedMessage) for the following packages:

1. Application Links
2. Collapsible Menu View
3. Date Picker
4. Date Time Picker
5. Notification Dialog
6. Tabs
7. Time Input

Resolves #752


packages/terra-collapsible-menu-view/package.json
```diff
@@ -40,7 +41,7 @@
     "terra-icon": "^3.1.0"
   },
   "scripts": {
-    "compile": "babel --root-mode upward --out-dir lib --copy-files",
+    "compile": "babel --root-mode upward src --out-dir lib --copy-files",
```
tbiethman: Is this necessary?

avinashg1994: Yes it is required. Even though a repo level "compile" would run successfully, if we try to "compile" a single package without specifying "src" the following error is thrown `--out-dir requires filenames`.

mjhenkes: Does this need to be changed for all packages?

bjankord: It sounds that way, let's get an issue logged for this @avinashg1994 and then we can get it resolved for the other packages in a separate PR.

avinashg1994: I scanned the entire repo and found that only **Collapsible Menu View** and **Tabs** required this change, which has been addressed as part of this PR.
All the other packages do not require this change.

samuv:
Sorry for the sneaky comment but is this published?

PR #822, opened 2019-07-31
[terra-tabs] - Fix tab overflow

### LINKED ISSUE #793: Horizontal Scroll bar gets added for tabs at Medium viewport when base-font-size is 75%
# Bug Report

## Description
In Medium viewport min-width property exceeds width property of Tabs when terra-base-font-size is set to 75%. i.e. min-width gets computed to 85.716px which is greater then width of tab 85.7031px. (difference can be seen on screenshot below)

## Steps to Reproduce
1. Navigate to https://engineering.cerner.com/terra-framework/#/tests/terra-tabs/tabs/tabs/default-tabs
2. Toggle viewport to Medium (width: 768px IPad in chrome)
3. Change terra-base-font-size to 75%
4. Horizontal Scroll bar will shows up for the tabs.

In mpage-Fusion-Theme terra-base-font-size is set to 75% which makes more tabs to show up on screen as shown below : 
Screen Shot 2019-07-19 at 2 54 21 PM

Total 9 tabs shows-up in medium viewport each having min-width of 85.716 which sums up to 771.444 px which is greater than width of medium viewport (768 px). this causes tabs to overflow from available screen width of 768 px. Due to this overflow of tabs WDIO tests in mpage fusion theme were failing with **"endX is out of range"** error.

## Expected Behavior
Tabs should not overflow on given viewport

## Environment

* Component Name and Version: Terra-Tabs / 6.6.0
* Browser Name and Version: Chrome  ( Latest )
* Webpack Version:
* Operating System and version (desktop or mobile): Tab ( Medium Viewport 768px)

supreethmr:
### Summary
Resolves #793 

### Additional Details
Horizontal scroll bar was getting added due to tabs overflow in few viewports for base-font-size 75%. Hence updated the min-width calculations to get actual min-width of tabs without excluding decimal part. so that tabs overflow can be prevented for min-widths exceeding available width.

[Reason for Screenshots update ](https://github.com/cerner/terra-core/blob/master/packages/terra-icon/src/icon/IconConsultInstructionsForUse.jsx)


packages/terra-tabs/src/_CollapsibleTabs.jsx
```diff
@@ -120,7 +120,7 @@ class CollapsibleTabs extends React.Component {
       const tab = this.container.children[i];
       const tabMarginLeft = parseInt(window.getComputedStyle(this.menuRef, null).getPropertyValue('margin-left'), 10);
       const tabMarginRight = parseInt(window.getComputedStyle(this.menuRef, null).getPropertyValue('margin-right'), 10);
-      const minWidth = parseInt(window.getComputedStyle(tab, null).getPropertyValue('min-width'), 10);
+      const minWidth = parseFloat(window.getComputedStyle(tab, null).getPropertyValue('min-width'));
```
jmsv6d: Should we make this update on the margin values too?

supreethmr: to keep change as much minimal I have not considered changing parse function for margin values. Yes it makes sense to update those values too : 38e8074fbcc168091ead784f2ef7721d16437378

PR #827, opened 2019-08-07
[terra-date-picker] Remove time component from date value returned in callbacks.

### LINKED ISSUE #802: Date picker includes time
# Bug Report

## Description
On https://engineering.cerner.com/terra-ui/#/components/terra-date-picker/date-picker/date-picker, whenever I manually type in a date or select from the calendar, the date value returned in the `onChange` callback contains a time component. As a date only picker, we should only return the date portion of the ISO 8601 spec. It is perfectly fine for an ISO 8601 string to contain only the date portion. 

If some application needs to send the value back to an API with time value, then that is it's responsibility. It should not be expected of everyone using the component to chop off the time component.

## Steps to Reproduce
1. Go to https://engineering.cerner.com/terra-ui/#/components/terra-date-picker/date-picker/date-picker
2. Manually enter a date
3. Select a date from the calendar
4. Value returned by component is 2019-07-27T00:00:00-05:00 or 2019-07-22T15:02:27-05:00 if time value originally given in `value` prop.

## Expected Behavior
Date value only appears as YYYY-MM-DD. No time portion exists.

## Possible Solution
Remove the time component when returning the value in any callback. From talking with @mmalaker he agreed it was strange that the time value was being returned.

## Environment
* Component Name and Version: terra-date-picker
* Node/npm Version: 
* Webpack Version:

nramamurth:
### Summary
Fixes #802.


packages/terra-date-picker/src/terra-dev-site/doc/example/DatePickerControlled.jsx
```diff
@@ -58,11 +58,11 @@ Selected ISO Date:
 }
 
 DatePickerExample.propTypes = propTypes;
-DatePickerExample.defualtProps = defaultProps;
+DatePickerExample.defaultProps = defaultProps;
```
benbcai: 👍

Issue #665, opened 2019-05-14
[time-input] The height of the time-input increases as the container width becomes smaller.


benbcai:
# Bug Report

## Description
When the container width becomes smaller than the width of the time-input, the input to the right (minute/meridiem) gets pushed to the next line, causing the overall height of the time-input to increase.

## Additional Context / Screenshots
![time-input](https://user-images.githubusercontent.com/22618655/57733554-6765a580-7665-11e9-8041-cf8559dcaa11.gif)

## Possible Solution
Set the `width` on the time-input container to `max-content`

pranav300:
@benbcai Is this the expected behavior?
<img width="86" alt="time input" src="https://user-images.githubusercontent.com/29273628/60790806-f4b60980-a17f-11e9-9052-48fd909a96a6.PNG">

nramamurth:
Related: https://github.com/cerner/terra-framework/issues/706

@neilpfeiffer

benbcai:
This has been fixed by #837

PR #837, opened 2019-08-13
[terra-time-input] Revamp mobile variation and 12-hour format on desktop. 

### LINKED ISSUE #385: TimeInput not handling onInputFocus correctly for mobile/small browsers
# Bug Report

## Description
I'm getting the following error in the browser console when loading a page with a `DateTimePicker` on it that is less than 1024 pixels wide.
```
Warning: Unknown event handler property `onInputFocus`. It will be ignored.
    in div (created by TimeInput)
    in TimeInput (created by DateTimePicker)
    in div (created by DateTimePicker)
    in div (created by DateTimePicker)
    in DateTimePicker (created by ActionToolbarComponent)
    ...
```
It looks like the `TimeInput` has code to delete `onInputFocus` from the `customProps` passed to the div for the `desktopInput`, but not the `mobileInput`, so whenever the` DateTimePicker` is rendered on a page less than 1024 pixels wide this warning is logged.  I tried to reproduce the issue in Kaiju, but it looks like since Kaiju is running a prod build of React, the check that the handler is valid is skipped, and the `onInputFocus` handler is added to the div instead of logging a warning.

## Steps to Reproduce
1. Build or find a page containing a DateTimePicker
2. Serve that page with React in Development mode
3. Resize the browser window to less than 1024 pixels wide
4. Load the page in the browser, and check the console for the warning. (Make sure that the page is loaded after the browser is small.  You may need to refresh the page.)

image

## Expected Behavior
The `onInputFocus` handler should be extracted from the `customProps` and consumed where appropriate, instead of attempting to attach it to the `div` wrapping the `TimeInput`, where it does nothing.

## Possible Solution
Copy the rendering behavior of the desktop version, which handles the `onInputFocus` prop instead of passing it to the `div`

## Environment
* Component Name and Version: `terra-date-time-picker@2.30.1`/`terra-time-input@2.26.1`
* Webpack Version: 4.18.4

nramamurth:
### Summary
Closes #385, #665, #710, #706

Currently, there are two layout options for terra-time-input based on the viewport size, this change removes the two versions and there will only be a single UI for all situations meaning - it does not visually separate the hour/minute/second inputs on mobile and does not use the dropdown select for the AM/PM but instead only uses the button-group for meridiem on desktop. 

### Deployment Links:

* [Date Time Picker](https://terra-framework-deploye-pr-837.herokuapp.com/components/terra-date-time-picker/date-time-picker/date-time-picker)
* [Time Input](https://terra-framework-deploye-pr-837.herokuapp.com/components/terra-time-input/time-input/time-input)
* [Clinical Theme](https://pages.github.cerner.com/terra/cerner-clinical-theme/build/nramamurth-time-input-update/#/components/terra-time-input/time-input/time-input)
* [Fusion Theme](https://pages.github.cerner.com/terra/orion-fusion-theme/build/nramamurth-time-input-update/#/components/terra-time-input/time-input/time-input)
* [Consumer Theme](https://pages.github.cerner.com/healtheintent/cerner-consumer-theme/build/nramamurth-time-input-update/#/components/terra-time-input/time-input/time-input)
* [Lowlight Theme](https://pages.github.cerner.com/terra-incubator/ti-clinical-lowlight-theme/build/nramamurth-time-input-update/#/components/terra-time-input/time-input/time-input)

**TODO:**
- [x] Open PRs in all the downstream theming repositories.
- [x] Resolve deployment issue in Lowlight Theme.


avinashg1994:
In the following wdio snapshot: [Snapshot](https://github.com/cerner/terra-framework/blob/98e563f1d2328c996658576931f1d699d90b6229/packages/terra-date-time-picker/tests/wdio/__snapshots__/reference/en/chrome_tiny/date-time-picker-spec/Valid_date_entry_moves_focus_to_hour_input%5Bdefault%5D.png)
I see there is an input caret visible, this can be addressed as explained in https://github.com/cerner/terra-framework/pull/860#discussion_r319291514
We might also want to hide the input in other similar scenarios as well.

avinashg1994:
Looks like in [Snapshot](https://github.com/cerner/terra-framework/blob/98e563f1d2328c996658576931f1d699d90b6229/packages/terra-date-time-picker/tests/wdio/__snapshots__/reference/en/chrome_tiny/date-time-picker-mobile-spec/Date_Time_Picker_Twelve_Hour_Seconds_Mobile_-_Focus_Minute_Styles%5Bdefault%5D.png), in mobile view and when the focus is on the minutes field the value gets clipped.

avinashg1994:
@neilpfeiffer, I noticed some inconsistencies in the time input field as seen below:

Screenshot 2019-08-30 at 2 25 08 PM
Screenshot 2019-08-30 at 2 25 25 PM

In both the screenshots above, i.e. with and without the seconds field, the spacing/padding between the minute/seconds field w.r.t to the right border is not consistent. Further to add to that, neither of those spacings are the same as that between the left border and the hour field.
Is this something that should be looked at? since this is not introduced as part of this issue but was an existing behaviour.

nramamurth:
> In the following wdio snapshot: Snapshot
I see there is an input caret visible, this can be addressed as explained in #860 (comment)
We might also want to hide the input in other similar scenarios as well.

@avinashg1994, good one. I have fixed it: https://github.com/cerner/terra-framework/pull/837/commits/ea74ed2a8ae4d2e3fcc107287ef9a43b273fb7bc?short_path=2304874#diff-2304874555680f0615bb9685afa2793c

packages/terra-time-input/src/TimeInput.module.scss
```diff
@@ -26,7 +30,7 @@
     display: inline-block;
     font-size: var(--terra-time-input-font-size, 1.14286rem);
     line-height: var(--terra-time-input-line-height, 1.375);
-    margin: 0;
+    margin: 0 0.5rem 0.5rem 0;
```
avinashg1994: I am a bit sceptical about this change, not sure if this might impact the existing consumers of time input.
@neilpfeiffer, your thoughts on this?

nramamurth:
> Looks like in [Snapshot](https://github.com/cerner/terra-framework/blob/98e563f1d2328c996658576931f1d699d90b6229/packages/terra-date-time-picker/tests/wdio/__snapshots__/reference/en/chrome_tiny/date-time-picker-mobile-spec/Date_Time_Picker_Twelve_Hour_Seconds_Mobile_-_Focus_Minute_Styles%5Bdefault%5D.png), in mobile view and when the focus is on the minutes field the value gets clipped.

@avinashg1994 

This occurs both on the desktop and on mobile view. IMO, two things contribute to this - 
1. the existing styles toggle the minute field width depending on whether the value is empty or not. 
https://github.com/cerner/terra-framework/blob/03cdd6ec29120b617c115cc73e2c56585d24996e/packages/terra-time-input/src/TimeInput.module.scss#L207-L217
2. something is causing the input elements to behave strangely upon focus that might also contribute to this issue in addition to the one mentioned above. I am yet to find out what is causing this wobbly behavior, though 😞

sample

I might need UX to help validate the styles that have changed.

cc @neilpfeiffer

packages/terra-time-input/src/TimeInput.module.scss
nramamurth: It is done to accommodate the design requirement mentioned [here](https://github.com/cerner/terra-framework/issues/706#issuecomment-518269645). I will want the bottom margin also to be 7px to accommodate the meridiem buttons under the time-input component when the viewport squishes.

packages/terra-time-input/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
```
bjankord: I'm thinking we may want to treat this as a major version bump. Thoughts cc/ @ryanthemanuel @mjhenkes @tbiethman @neilpfeiffer

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -719,323 +644,136 @@ class TimeInput extends React.Component {
...
+            // Create a hidden input for storing the name and value attributes to use when submitting the form.
+            // The data stored in the value attribute will be the visible date in the date input but in ISO 8601 format.
+            type="hidden"
+            name={name}
+            value={timeValue}
+          />
           <Input
             {...inputAttributes}
-            {...instanceHoursAttrs}
+            {...minuteAttributes}
```
bjankord: This should be hourAttributes I believe.
```suggestion
            {...hourAttributes}
```

packages/terra-time-input/src/TimeInput.module.scss
```diff
@@ -43,11 +47,7 @@
       .time-input-hour,
       .time-input-minute,
       .time-input-second {
-        color: var(--terra-time-input-hover-color, #1c1f21);
-      }
-
-      .meridiem-display {
-        color: var(--terra-time-input-hover-color, #1c1f21);
+        color: var(--terra-time-input-time-component-hover-color, #1c1f21);
```
bjankord: Why the addition of the word `component` in the variable name?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -227,19 +223,25 @@ class TimeInput extends React.Component {
   handleSecondFocus(event) {
     this.handleFocus(event);
     this.setState({ secondInitialFocused: true });
-    this.secondInput.setSelectionRange(0, this.secondInput.value.length);
+    if (!TimeUtil.isConsideredMobileDevice()) {
```
bjankord: Whats the reason for wrapping these in the condition check?

packages/terra-time-input/CHANGELOG.md
benbcai: I agree since we're getting rid of the meridiem select and this would be a breaking change.

packages/terra-time-input/src/TimeInput.module.scss
```diff
@@ -216,76 +176,32 @@
...
   .meridiem-button {
-    background-color: var(--terra-time-input-meridiem-button-background-color, #dedfe0);
+    background-color: var(--terra-time-input-meridiem-button-background-color, transparent);
     border-color: var(--terra-time-input-meridiem-button-border-color, #c8cacb);
     color: var(--terra-time-input-meridiem-button-color, #1c1f21);
     font-size: var(--terra-time-input-meridiem-button-font-size, 1rem);
-    height: var(--terra-time-input-mobile-height, 2.1428571429rem);
-    line-height: var(--terra-time-input-mobile-line-height, 1.429);
+    height: var(--terra-time-input-meridiem-button-height, 2.1428571429rem);
+    line-height: var(--terra-time-input-meridiem-button-line-height, 1.67412);
```
benbcai: It looks like the a.m./p.m. text is not horizontally center aligned and the timezone button is not aligned with the meridiem buttons or time input.

Screen Shot 2019-08-30 at 4 37 11 PM

packages/terra-time-input/src/TimeInput.module.scss
nramamurth: At present, same variables are used for the sub-components (hour, minute, second) as well as their wrapper while they are technically different entities; although the variables are the same, the difference in how these entities are presented is currently brought out by the [explicit styles laid out under `.desktop` selector](https://github.com/cerner/terra-framework/blob/a9e915c6dee0e4cecf36eac1882e29282f13a342/packages/terra-time-input/src/TimeInput.module.scss#L161-L193). 

This gets confusing when there is no concept of mobile v/s desktop anymore (both hit the same code with this revamp), so I introduced the word 'time-component' into the variables that are meant to style the sub-components (hour, minute, second) _only_. This, imo, made it more readable and made it a lot easier to differentiate these components in the theming repositories downstream as well.

packages/terra-time-input/src/TimeInput.jsx
nramamurth: The `setSelectionRange` call sans this check brings up the contextual menu (see below) on iOS every time the focus shifts.

context-menu-ios

packages/terra-time-input/src/TimeInput.module.scss
nramamurth: You're right, Ben. I am waiting on UX to validate these new values and I shall follow-up.

@neilpfeiffer 👋

packages/terra-time-input/CHANGELOG.md
nramamurth: If this is a MVB, then I shall add the upgrade guide soon after the CSS changes are validated.

packages/terra-time-input/src/TimeInput.module.scss
bjankord: Sounds good, thanks for the clarification @nramamurth

packages/terra-time-input/src/TimeInput.jsx
bjankord: Ah, let's add a code comment about this before the conditional so we don't accidentally remove it without understanding its intended use.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -719,323 +650,136 @@ class TimeInput extends React.Component {
...
-              size={this.state.meridiem.length || 1}
-              readOnly
-              disabled={disabled}
-            />,
-            <div
-              className={cx('meridiem-select-wrapper')}
-              key="meridiem_select_box"
-            >
-              <select
-                aria-label={intl.formatMessage({ id: 'Terra.timeInput.select.meridiem' })} // value in translations set to 'Select Meridiem'
```
benbcai: The `Terra.timeInput.display.meridiem` and `Terra.timeInput.select.meridiem` [translation strings](https://github.com/cerner/terra-framework/blob/master/packages/terra-time-input/translations/en-US.json#L10-L11) can also be removed if they are no longer needed.

nramamurth: Good point! I have [removed](https://github.com/cerner/terra-framework/pull/837/commits/293524d4f0bc16b3cab27fa233240a6337d29d58) them.

benbcai: A new en-CA translation files just for merged today: https://github.com/cerner/terra-framework/blob/master/packages/terra-time-input/translations/en-CA.json#L10-L11
Can you also remove it en-CA?

packages/terra-time-input/src/TimeInput.module.scss
jeremyfuksa: The a.m./p.m.text is aligned with the baseline of the time input:
Screen Shot 2019-09-20 02 56 55 PM

To get the desired alignment, I set the `span` inside each button label to `line-height: 1;`
Screen Shot 2019-09-20 03 00 35 PM

I'm also seeing that there are some really funky alignment issues in Firefox:
Screen Shot 2019-09-17 10 51 07 AM

benbcai: @nramamurth @jeremyfuksa  Are we still waiting for UX feedback? Looks like the TZ and meridiem buttons are still misaligned.

jeremyfuksa: @benbcai @nramamurth setting `line-height: 1;` within each button label's `span` was the UX feedback. That will make everything align against centers rather than the text baseline.

nramamurth: @benbcai I’m sorry, I seem to have missed your comment over the weekend. I shall work on this and update the upgrade guides as well so it’s ready for MVB.

nramamurth: @benbcai @jeremyfuksa 

Here's a couple of findings: 
1. While setting `line-height: 1;` for the meridiem buttons solves the horizontal alignment issue in them, it shifts the button group down making it misalign with the time-input group. 

<details> <summary> <b> With `line-height: 1;` for the meridiem button </b> </summary> 

Screen Shot 2019-10-01 at 4 10 58 PM
</details>

--- 
2. To fix the alignment w.r.t. the TZ (offset) button, I introduced a line-height of 2.16285 to the [offset button](https://github.com/cerner/terra-framework/blob/7734543bf6304d7a4a4c6319fbdcc1b4f2ea4684/packages/terra-date-time-picker/src/_TimeClarification.module.scss#L55-L62). While it seemed to have fixed the issue on Safari and Chrome, the _wdio_ screenshot shows misalignment - notice that the TZ button has moved up a bit. I suspect the tests run against headless browser and I can't seem to find out the root cause. Do they? 

<details> <summary> <b> Chrome (current line-height) ❌ </summary> 

Screen Shot 2019-10-01 at 4 07 28 PM
</details>

<details> <summary> <b> Chrome (updated line-height 2.16285) ✅ </b></summary> 

Screen Shot 2019-10-01 at 4 06 52 PM
</details>

<details> <summary> <b> Headless (?) (current line-height) ✅ </b></summary> 

Displays_Offset_button_after_clicking_daylight_button 1  2
</details>

<details> <summary> <b> Headless (?) (updated line-height 2.16285) ❌ </b></summary> 

Displays_Offset_button_after_clicking_daylight_button 1
</details>

---

And Firefox is behaving totally funky.

jeremyfuksa: When you say "current line-height", I'm assuming you mean `line-height: 1`, correct?

nramamurth: @jeremyfuksa If you are talking about the TZ button, then no, its current line-height is the default line-height that is 1.429 that comes from terra-button.

packages/terra-time-input/CHANGELOG.md
benbcai: After further discussions with the Terra team, it was decided that the changes on this pull request will not be considered a breaking change because there are no changes to the API. Therefore, this does not require a major version bump. Since there are layout and visual changes to the meridiem buttons, we expect some WDIO tests to fail for consumers and those screenshots would need to be regenerated. In rare cases where consumer may have logic in their application to query for the old meridiem select or listen to any native events, that logic would need to be updated.

@nramamurth An upgrade guide would not be needed but it would be good to add additional documentation in both the terra-date-time-picker and terra-time-input changelogs to call out that the meridiem select normally shown on desktop is now changed to buttons. It would also be helpful for consumers by documenting on the changelogs that WDIO screenshots and any code in the consumer application that accesses the old meridiem select may need to be updated.

nramamurth:
I have removed the 'Major Version Bump' label based on [this discussion](https://github.com/cerner/terra-framework/pull/837#discussion_r319587785).

packages/terra-time-input/CHANGELOG.md
nramamurth: @benbcai Gotcha! I have [updated the Changelog](https://github.com/cerner/terra-framework/pull/837/commits/7dcf2192f7452dd9ec6905c0ce2ecfd2fe76d443). Lmk if this is not sufficient.

packages/terra-time-input/src/TimeInput.module.scss
jeremyfuksa: Using relative positioning on the meridiem button group and then offsetting the top by -1px fixes the styling, but that's not 100% ideal.

nramamurth: @jeremyfuksa It is already relative. I have however seemed to have fixed the alignment w.r.t. the TZ button. It'd be great if you could validate these changes in the latest deployment and this has resolved the funkiness in Firefox as well. 

Also, I think that the "a.m./p.m." text "misalignment" is probably because they are lowercase text. I changed it to uppercase just for internal testing and they seem to be aligned correctly to me. So, I am assuming that it is okay? 

Screen Shot 2019-10-07 at 11 12 20 AM

cc @benbcai

jeremyfuksa:
Tested: PR #837
Resolves: Issue #385, #665, #710, #706

**Reviewed:**
- Follows pattern considerations documented in the current Cerner Design Standards: [Date/Time Formatt](https://wiki.cerner.com/pages/viewpage.action?pageId=1661404488)
- Follows UX defined design implementation for terra-time-input, terra-date-time-picker
- Follows defined theme designs: Open Source, Clinical, Consumer, Fusion, Low Light
\
\
Verified by @jeremyfuksa, Marking as Verified, Ready for Release.

**Pass** |
-----------|
UX Review|

benbcai:
@nramamurth Great job in getting that last issue resolved! The changes in this PR will have conflicts with the changes in #923. There is some urgency in getting #923 merged and released early next week. I want to talk to the Terra team later today to determine whether or not we want to hold off on merging this PR to avoid the code conflicts with #923.

benbcai:
It was decided that we want to hold off in merging this PR until after #923 is merged to minimize the delay in completing #923.

PR #838, opened 2019-08-14
Add terra-date-input component

### LINKED ISSUE #808: Create date form field
# Feature Request

## Description
The date field would be a new form field specifically for entering known dates like Date of Birth. Date of Birth is a common form field that has been requested by a few teams now.
It is frequently displayed as 3 form fields grouped together.

Screen Shot 2019-07-22 at 4 47 55 PM

bjankord:
### Summary
Resolves #808 

### Deployment Links:
* [Date Input](https://terra-framework-deploye-pr-838.herokuapp.com/components/terra-date-input/date-input/date-input)
* [Date Input Field](https://terra-framework-deploye-pr-838.herokuapp.com/components/terra-date-input/date-input/date-input-field)
* [Date Input Form Validation](https://terra-framework-deploye-pr-838.herokuapp.com/components/terra-form-validation/form-validation/validation-components) (Last example on the page)


packages/terra-date-input/src/DateInput.module.scss
```diff
@@ -0,0 +1,57 @@
+:local {
```
bjankord: TODO: Clean up and add themable variables

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -0,0 +1,644 @@
...
+          className={cx('date-input-month', { 'is-focused': this.state.monthIsFocused })}
+          value={this.state.month}
+          name={'terra-date-month-'.concat(this.props.name)}
+          onChange={this.handleMonthChange}
+          onKeyDown={this.handleMonthInputKeyDown}
+          onFocus={this.handleMonthFocus}
+          onBlur={this.handleMonthBlur}
+          disabled={this.props.disabled}
+        >
+          <option value="" disabled>{this.props.intl.formatMessage({ id: 'Terra.date.input.monthPlaceholder' })}</option>
```
benbcai: Is it intentional that it only allows to use the Back/Delete keys to remove a selected month and not use the `allowClear` prop on Select to click on this option to clear it?

bjankord: This is using the native HTML select so the allowClear prop isn't available. There were some issues getting onKeyDown to work with the terra-select so I opted for a native-select.

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -0,0 +1,644 @@
...
+  dayAttributes: PropTypes.object,
+  /**
+   * Whether the date input should be disabled.
+   */
+  disabled: PropTypes.bool,
+  /**
+   * Can be used to set the display format. One of `month-day-year` or `day-month-year`.
+   * Overrides default locale display format.
+   */
+  displayFormat: PropTypes.string,
```
benbcai: I don't know how the date format for en-CA would apply for this component. It looks like the [official all-numeric date format](https://en.wikipedia.org/wiki/Date_and_time_notation_in_Canada) for en-CA is YYYY-MM-DD. However, other formats are still used depending on their context. According to [this](https://en.wikipedia.org/wiki/Date_format_by_country), the month still appears before the year when using the long month name. Since this component does not use an all-numeric date format, perhaps, we should can support just these two display formats for now and support other display formats in the future when new requests come in.

bjankord: Talked with @benbcai offline, planning to leave `en-CA` date rendering as month-day-year.

nramamurth:
Just curious, is it intentional for backspace keypress to shift focus to the input field on the left? For instance, the focus is being currently taken from 'Year' to 'Date' and 'Date' to 'Month' in the default example when pressing backspace in an empty field.

bjankord:
> Just curious, is it intentional for backspace keypress to shift focus to the input field on the left? 

@nramamurth Yes this is by design. It is intended to make clearing the combination of inputs easier similar to how backspace key works with the [time-input component](https://engineering.cerner.com/terra-ui/components/terra-time-input/time-input/time-input).

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -0,0 +1,662 @@
...
+
+    // Ignore the entry if the value did not change or it is invalid.
+    if (inputValue === stateValue || Number(inputValue) > maxValue) {
+      return;
+    }
+
+    this.handleValueChange(event, DateInputUtil.inputType.MONTH, inputValue);
+  }
+
+  handleDayChange(event) {
```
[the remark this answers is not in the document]
bjankord: I had coded in the auto-focus shift functionality whenever month / day values were filled in but ran into an issue when using only a keyboard to set the month value in the month select.

For example, if I type the character `J` while focused on the month input, it will switch the month to January, updating the value from emptry string to `01`. The month select value is now populated and so the auto-focus logic would kick in shifting the focus to the day input in the `month-day-year` formatted date input. 

This caused an issue as the user may not have wanted to select January when they typed in the letter `J` and maybe they were trying to type in June or July. I had considered adding a delay before auto-shifting focus from the month so they could type in `Jun` but thought introducing a delay on the month select would be weird. And only auto-shifting focus on the day input also felt like a weird possibility so I decided to remove auto-focus shift functionality instead.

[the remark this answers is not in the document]
benbcai: Thanks for the clarification. We can keep it as is and enhance it when we figure out a good way to handle it.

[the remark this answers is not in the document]
bjankord: Talked with Ben offline about possibly enabling auto-focus if the user is selecting a month by not typing in the name (the user clicks on the month select and then selects a month or the user press arrow keys up/down to select a month and then press enter) as a future enhancement if we think this is functionality that would be good to add.

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -0,0 +1,662 @@
...
+    }
+
+    const inputValue = event.target.value;
+    const stateValue = this.state.day;
+    const maxValue = 31;
+
+    // Ignore the entry if the value did not change or it is invalid.
+    // When 'Predictive text' is enabled on Android the maxLength attribute on the input is ignored so we have
+    // to check the length of inputValue to make sure that it is less then 2.
+    if (inputValue === stateValue || inputValue.length > 2 || Number(inputValue) > maxValue) {
```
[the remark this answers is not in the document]
bjankord: Good thinking. Updated in this commit: https://github.com/cerner/terra-framework/pull/838/commits/2126778e42e15c316526d85f8f4219fd69021f60

nramamurth:
This could be cosmetic and it could just be me but I notice a strange behavior on Safari with hover action over Day field in MM-DD-YYYY examples. Upon hovering over the Day field, the mouse pointer immediately changes to cursor on Chrome and Firefox while it does not on Safari. 

Note: This does not happen on DD-MM-YYYY example.

**Safari:**

safari-gif-test

**Chrome:** 
chrome

packages/terra-date-input/src/DateInputField.jsx
```diff
@@ -0,0 +1,240 @@
...
+  dayAttributes: PropTypes.object,
+  /**
+   * Whether the date input should be disabled.
+   */
+  disabled: PropTypes.bool,
+  /**
+   * Can be used to set the display format. One of `month-day-year` or `day-month-year`.
+   * Overrides default locale display format.
+   */
+  displayFormat: PropTypes.string,
```
nramamurth: Should this prop type be validated? 

```
PropTypes.oneOf(['month-day-year', 'day-month-year'])

Comment (bjankord) [reply packages/terra-date-input/src/DateInputField.jsx]: Updated in this commit: https://github.com/cerner/terra-framework/pull/838/commits/08c1db2e09a97f14f33f1cd64236a2d9940bf5a6

**Comment by bjankord** [comment]
@nramamurth Resolved issue with cursor in this commit: https://github.com/cerner/terra-framework/pull/838/commits/1160f0ab5d7513f2c7df49bd270fed73463c4fa6

File: packages/terra-date-input/src/DateInputField.jsx [hunk 7]
```diff
@@ -0,0 +1,240 @@
...
+  const legendClassNames = cx([
+    'legend',
+    legendAttributes.className,
+  ]);
+
+  /**
+   * aria-describedby ids
+   * Used to map legend, help text, and error messages with input/select elements
+   */
+  const helpAriaDescriptionId = help ? `terra-date-input-field-description-help-${uniqueid()}` : '';
```
tbiethman: Is uniqueId going to change with each render? Does that matter?

bjankord: Yeah, it looks like it is changing with each render. I'll update this to use uuid like we do in [terra-navigation-prompt](https://github.com/cerner/terra-framework/blob/7ec7a5589cb2dc17a0520d2fd4bd469b30c45f31/packages/terra-navigation-prompt/src/NavigationPrompt.jsx#L33) instead so its stays the same through renders.

Users don't need to interact with this ID, but we do want it to be unique.

bjankord: Updated to use uuid for unique ids in this commit: https://github.com/cerner/terra-framework/pull/838/commits/0549f6fb6ed0373bd06c8b80d005a2a983d4ccc1

mmalaker:
Functional verification completed. The examples date input examples display and function as expected. One issue noted with error text not being read by screen readers. This will be addressed in:
https://github.com/cerner/terra-framework/issues/873

neilpfeiffer:
Tested: PR #838
Resolves: Issue #808
 
\
**Reviewed:**
- Follows pattern considerations documented in the current Cerner Design Standards: [Date Time Selection](https://wiki.cerner.com/x/EWUVYQ)
- Follows UX defined design implementation for a basic Date Input
- Follows defined theme designs (5): terra-opensource-theme, orion-fusion-theme, cerner-clinical-theme, cerner-consumer-theme, ti-clinical-lowlight-theme
 
\
\
Marking as Verified by @neilpfeiffer, ready for release.
 
**Pass** |
-----------|
UX Review|

PR #844, opened 2019-08-15
[terra-date-time-picker] - Handle ambiguous hour that occurs in the s…

### LINKED ISSUE #460: DateTimePicker time clarification abbreviations don't account for southern hemisphere
# Bug Report

## Description

When a time during the ambiguous hour is entered, a dialog displays asking the user to chose Standard vs Daylight time, after which the timezone abbreviation corresponding to the selected time is displayed.  However, for users in the southern hemisphere, selecting Standard causes the Daylight abbreviation to be displayed, and vice versa

## Steps to Reproduce

1. Set your system time to Sydney Australia 
2. Go to the [`DateTimePicker` example page](https://engineering.cerner.com/terra-ui/#/components/terra-date-time-picker/date-time-picker/date-time-picker)
3. Enter April 1, 2018 02:30 (or another time in the ambiguous hour for [Sydney](https://www.australia.gov.au/about-australia/facts-and-figures/time-zones-and-daylight-saving)).
4. Select either Daylight or Standard
5. Observe that the timezone abbreviation displayed is opposite of what it should be (ie AEST when Daylight is selected, and AEDT when Standard is selected.

Currently, when determining what the timezone abbreviation should be, the code makes the incorrect assumption that the Daylight abbreviation can be found by looking at July 1, and that the Standard abbreviation can be gotten from January 1.  This is of course wrong for the southern hemisphere because January is during the summer, and July during the winter.

## Expected Behavior

The displayed timezone abbreviation correctly matches the value selected

## Possible Solution

Retrieve the abbreviation from the actual real date in the control, after it has been adjusted to the correct time based on the option selected.

## Environment

* Component Name and Version: `terra-date-time-picker@2.30.1`
* Webpack Version: 4.18.4

benbcai:
…outhern hemisphere.

### Summary
Fixes #460 
Handles the ambiguous hour that occurs in the southern hemisphere.

### Additional Details
In the U.S., daylight savings time starts in the Spring and ends in the Fall. However, in the southern hemisphere (i.e. Australia), it is the other way around where daylight savings time starts in the Fall and ends in the Spring. The current implementation assumes it is always the former and uses a constant date (2017-07-01) to retrieve the daylight savings time zone identifier and uses 2017-01-01 to retrieve the standard time zone identifier. The change is to instead of using a constant date to look up the time zone identifier, use the actual entered date/time to determine the time zone identifier.

@cerner/terra


mmalaker:
Functional verification completed. The date time picker now supports the ambiguous hour for time zones in the southern hemisphere.

Issue #835, opened 2019-08-12
[terra-infinite-list] Cannot read property 'scrollHeight' of null


dkasper-was-taken:
# Bug Report
"Cannot read property 'scrollHeight' of null" 

## Description
When resizing there is a debounce timeout set in the InfiniteList, if it is umounted during the timeout the reference to "this" or "this.contentNode" will become null and the subsequent call for it's scrollHeight will cause an error.

## Steps to Reproduce
1. Resize browser
2. Either programmatically remove the infinite list or use the dev-stie to select another tab

## Possible Solution
Adding additional clearTImeout to the cleaup should fix it:
clearTimeout(this.timer);

https://github.com/cerner/terra-framework/blob/master/packages/terra-infinite-list/src/InfiniteList.jsx#L233

ShettyAkarsh:
Closed #849

PR #849, opened 2019-08-19
[infinite-list] Fixed ScrollHeight null issue

lokesh-0813:
### Summary
Fixes `Cannot read property 'scrollHeight' of null` issue when infinite list is unmounted immediately after browser resize.

Solves : #835


packages/terra-infinite-list/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
```
avinashg1994: nit: i think this should be in the "FIxed" section.

packages/terra-infinite-list/src/InfiniteList.jsx
```diff
@@ -234,6 +234,7 @@ class InfiniteList extends React.Component {
     if (!this.contentNode) {
       return;
     }
+    clearTimeout(this.timer);
```
ShettyAkarsh: @lokesh-0813, could let me know how this callout to cleartimeout will make the scrollheight value not null ? or is this call prevent certain calls which was happening before ?

lokesh-0813: @ShettyAkarsh, scrollHeight is used in handleResize() which is called inside setTimeout() and when clearTimeout() is called, it terminates setTimeout().

packages/terra-infinite-list/CHANGELOG.md
lokesh-0813: Updated now. Thank you.

Issue #800, opened 2019-07-22
Controlled date picker misforms date entry as slashes are added to it


bjankord:
# Bug Report

## Description
When typing a date into the controlled date picker component, slashes will be added to it after typing in 5 numbers into the input. When this happens, the numbers that were originally entered are also updating malforming the entered data.

## Steps to Reproduce

1. Navigate to https://engineering.cerner.com/terra-ui/#/components/terra-date-picker/date-picker/date-picker
2. Scroll down to the Controlled Component example
3. Start typing in a date, `01011234`
4. When you press `3`, the date entered becomes malformed, changing to `05/03/0101`

## Additional Context / Screenshots

<img width="212" alt="Screen Shot 2019-07-22 at 11 38 17 AM" src="https://user-images.githubusercontent.com/633148/61648628-68beda00-ac75-11e9-9c10-2f9bd7afccff.png">

## Expected Behavior
Date values should not be malformed as users enter data into the date picker input.

## Environment

* Component Name and Version: terra-date-picker 4.7.0
* Browser Name and Version: Chrome Latests
* Node/npm Version: [e.g. Node 8/npm 5] node 8 / npm 5
* Webpack Version:
* Operating System and version (desktop or mobile): macOS 10.14

## @ Mentions
@benbcai

nramamurth:
Below is the line that causes it because apparently "0101123" is valid and proceeds to format it. 

https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/src/DatePicker.jsx#L361

The question however is - when exactly should it check for validity and format the date if not while the user is entering the text? 

@bjankord @benbcai @neilpfeiffer

bjankord:
@nickpith mentioned looking at removing the onRawChange handler in the example controlled function.

bjankord:
I tested removing the `handleDateChangeRaw` handler from updating the `<Datepicker />` value prop, but then I was unable to type data into the input.

```jsx
class DatePickerExample extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      date: this.props.value,
      rawDate: this.props.value,
    };
    this.handleDateChange = this.handleDateChange.bind(this);
    this.handleDateChangeRaw = this.handleDateChangeRaw.bind(this);
  }

  handleDateChange(event, date) {
    this.setState({ date });
  }

  handleDateChangeRaw(event, date) {
    this.setState({ rawDate: date });
  }

  render() {
    return (
      <div>
        <p>
[...]

PR #851, opened 2019-08-19
[terra-date-picker] Fixing Controlled date picker from misforming data entry

BenBoersma:
### Summary
Fixes https://github.com/cerner/terra-framework/issues/800
Mainly just a change in utility function used to determine if an entered date is in the right format. 

@cerner/terra


benbcai:
moment.ISO_8601 is also used [here](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/src/DateUtil.js#L16) and should be updated as well.

BenBoersma:
Added the same logic to replace ISO-8601 format usage
https://github.com/cerner/terra-framework/pull/851/commits/01f03905222b6f6b4c4a324745196bf657b44d60

BenBoersma:
Had to make some changes to date-time-picker since its using date-picker util functions https://github.com/cerner/terra-framework/pull/851/commits/981d0e668678c378a6c4899ed7ce48c416576eac

packages/terra-date-picker/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ Changelog
 
 Unreleased
 ----------
+### Breaking Changes
+* Now requires users to enter separators themselves such as '/' or '-' depending on format.
```
benbcai: Why would this be considered a breaking change? I was thinking that this would be a Fix to not accept values that do not conform to the provided date format.

BenBoersma: Forgot to change the changelog after you clarified it. oops. Fixed now https://github.com/cerner/terra-framework/pull/851/commits/97ec7aef04215c4ca8bc967cb46af0d6250d6d04

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -195,6 +195,22 @@ class DateTimeUtils {
   static convertDateTimeStringToMomentObject(date, time, dateformat, hasSeconds) {
     return DateTimeUtils.updateTime(DateUtil.createSafeDate(DateUtil.convertToISO8601(date, dateformat)), time, hasSeconds);
```
benbcai: There are a couple of occurrences of createSafeDate in here that need to be switched to use the one in DateTimeUtil.

BenBoersma: Should be replaced here https://github.com/cerner/terra-framework/pull/851/commits/fa0f2d77abb55b9a6d36ebec85eb3201ca1a3381

PR #855, opened 2019-08-23
[terra-application-navigation] Added empty check for Utilities and Nav Drawer

### LINKED ISSUE #829: [terrra-appilcation-naivation] No check for presence of utils
# Bug Report
No check for empty utilities

## Description
While rare, if no utilities and no settings/help/logout are provided to terra-application-navigation we still show the utility button, that should not be displayed.

## Possible Solution
In the header implementation check for the length of utils and the presence of settings/help/logout, and if nothing exists do not render the utility button.

## @ Mentions
@tbiethman

avinashg1994:
### Summary
Added check to prevent rendering of Utility Button or Nav Drawer Button if they are empty.

Resolves #829


avinashg1994:
> change-log is updated only for "preventing rendering of Utility Button or Nav Drawer Button". I think other changes should also have change-log entry.

Preventing the rendering of the Utility Button and the Nav Drawer was the only change, all the file changes are to implement the same behaviour. Besides that no other behaviour was changed, so i feel that the changelog entry is appropriate.

PR #860, opened 2019-08-29
[terra-date-time-picker] Fixed onBlur when ambiguous hour is resolved to Standard date time.

### LINKED ISSUE #856: [terra-date-time-picker] onBlur prop given old date when DST modal closes
# Bug Report

## Description
When using the terra-date-time-picker, we are seeing some unexpected behavior around updating when the user enters an ambiguous hour. [This video](https://cernerprod.sharepoint.com/:v:/s/OrdersandPlansDev/Projects/EXiLcCgMyS1IsQPu-KZ7nyQBJLgf8rSiMx6cUC09nRCBlg?e=sS57gY) focuses on that behavior.

You can see on the initial selection of CST (-6:00), the -5:00 of CDT is still provided to our onBlur method and logged in the console. Our component then immediately calls our service to update the detail to this time. Once the update finishes, the detail has a value in CDT. When we click into and blur on the component again, we again pick CST, but CDT is provided to onBlur, meaning it looks like no change has occurred to the value and update is not called. (Note: Both of these times the user is incorrectly lead to believe the value is in CST, as the time zone button appears with CST displayed).

On our final "focus and blur" we see that update is called, because this time the onBlur function is correctly provided the CST (-6:00) value, so it observes a change and calls our services.

## Steps to Reproduce
1. Enter an ambiguous hour
2. Pick CST (while the current timezone is CDT)
3. Observe via logging or a breakpoint what value is passed to the onBlur prop

--

## Expected Behavior
Value passed to the onBlur prop reflects the choice made in the DST modal

## Possible Solution
Looks like the issue is related to storing the dateTime in state when a modal button is clicked, and reading from state to pass to the consumers provided onBlur. This was evidenced by this change fixing the behavior:
image

By nesting the handleBlur call in a setState callback, we guarantee that all changes to state are propagated before calling handleBlur (which calls the consumers onBlur later). Rather than wrapping the handleBlur events in the two click handlers for the DST modal buttons, this could wrap where we read the state in handleBlur for a better guarantee of covering all workflows, but this approach could be seen as "smelly" and was really just used as a proof of concept.

What may be a better solution to this issue is no longer storing the dateTime in state, but passing it directly into the handleBlur event, so we don't have to worry about async state propagation.

## Environment

* Component Name and Version: terra-date-time-picker 4.10.0
* Browser Name and Version: Chrome, uh, recent

## @ Mentions
@benbcai

nramamurth:
### Summary
Fixes #856


packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -435,4 +435,30 @@ Terra.describeViewports('DateTimePicker', ['tiny', 'large'], () => {
 
     Terra.it.matchesScreenshot('after offset change', { selector: '#root' });
   });
+
+  describe('onBlur (CST to CDT)', () => {
+    before(() => {
+      browser.url('/#/raw/tests/terra-date-time-picker/date-time-picker/date-time-picker-dst-blur');
+      browser.click('input[name="terra-time-minute-input"]');
```
benbcai: Should call `Terra.hideInputCaret('input[name="terra-time-minute-input"]');` to hide the input caret?

nramamurth: Since I am not taking a screenshot immediately after clicking into the minute input field and instead I am tabbing out (next line) for the Time Clarification modal to appear, I think I shall keep this as-is.

PR #875, opened 2019-09-09
[terra-date-picker] Remove time-picker from react-datepicker.

### LINKED ISSUE #864: Remove all time picker functionality from terra-datepicker
# Feature Request

## Description
When we forked react-datepicker, it included a time picker component built into it. We don't currently support using the "react-datepicker" time picker, we've built our own date time picker component.

Knowing this, we should remove the time picker code from our fork of react-datepicker. It is unneeded code and adds unneeded complexity to the terra-datepicker source code.

nramamurth:
### Summary
Closes #864


packages/terra-date-picker/src/react-datepicker/calendar.jsx
```diff
@@ -226,10 +197,6 @@ export default class Calendar extends React.Component {
      * Prop to show short names of weekdays .
      */
     useWeekdaysShort: PropTypes.bool,
-    /**
-     * Prop to open a separate portal version of date picker .
-     */
-    withPortal: PropTypes.bool,
```
supreethmr: Don't we need this for date-picker...?

packages/terra-date-picker/src/react-datepicker/date_utils.js
```diff
@@ -319,30 +300,6 @@ export function isDayDisabled (day, { minDate, maxDate, excludeDates, includeDat
     false
```
supreethmr: [addMinutes() ](https://github.com/cerner/terra-framework/blob/6623edd64551db6f735b3854956149e90fc082f0/packages/terra-date-picker/src/react-datepicker/date_utils.js#L170) function can also be removed. since it was referred only in Time.jsx

packages/terra-date-picker/src/react-datepicker/calendar.jsx
nramamurth: Yes, we do but it is not needed in `Calendar.jsx` because it is being passed down _only_ to the `Time` component which will be irrelevant with this change.

packages/terra-date-picker/src/react-datepicker/date_utils.js
nramamurth: Done - https://github.com/cerner/terra-framework/pull/875/commits/dbf858ea4fd07d28f4d79988fa50e22a0f744054.

PR #878, opened 2019-09-11
[terra-menu] Screen reader prevented from moving focus out of popup

### LINKED ISSUE #833: [terra-menu] screen readers move focus out of popup
# Bug Report

## Description
When using the desktop versions of Voice Over on Safari and JAWS on Chrome, it is possible to move focus out of a menu while the popup is still open. The screen reader this starts reading other parts of the main page. You can also reverse navigate from the page back to the popup.

## Steps to Reproduce

1. Go to https://engineering.cerner.com/terra-ui/components/terra-menu/menu/menu
2. Navigate to the default menu Click Me button (or any of the examples).
3. Using the screen reader's navigation buttons (left/right arrow for VO, up/down for JAWS), navigate among the menu items.
4. Continue navigating up past the top menu item
5. The screen reader leaves the popup and starts reading the content on the main page.

Note that on VO, I've gotten the screen reader to leave focus both at the top and the bottom of the menu. Looks like JAWS only loses focus from the top.

## Expected Behavior
Screen readers should maintain focus within the menu popup until the user dismisses the popup.

saket2403:
### Summary
Fixes #833


packages/terra-menu/src/Menu.jsx
```diff
@@ -53,12 +57,21 @@ const propTypes = {
    * Otherwise, the menu will display without an arrow and right aligned.
    */
   isArrowDisplayed: PropTypes.bool,
+  /**
+   * Allows assigning of root element custom data attribute for easy selecting of document base component.
```
nramamurth: What does this mean for a consumer?

packages/terra-menu/src/Menu.jsx
```diff
@@ -89,6 +132,47 @@ class Menu extends React.Component {
     }
   }
 
+  disableContainerChildrenFocus() {
+    const mainDocumentElement = document.querySelector(this.props.rootSelector);
+
+    if (mainDocumentElement) {
+      const inert = +mainDocumentElement.getAttribute('data-overlay-count');
+
+      if (!mainDocumentElement.hasAttribute('data-overlay-count')) {
```
nramamurth: Could this be simplified? 

```
if (mainDocumentElement.hasAttribute('data-overlay-count')) {
   mainDocumentElement.setAttribute('data-overlay-count', `${inert + 1}`);
} else {
   mainDocumentElement.setAttribute('data-overlay-count', '1');
   mainDocumentElement.setAttribute('inert', '');
}
```

packages/terra-menu/src/Menu.jsx
```diff
@@ -53,12 +57,21 @@ const propTypes = {
...
    */
   isArrowDisplayed: PropTypes.bool,
+  /**
+   * Allows assigning of root element custom data attribute for easy selecting of document base component.
+   */
+  rootSelector: PropTypes.string,
+  /**
+   * Allows assigning of trigger element custom data attribute for easy selecting of triggering component.
+   */
+  triggerElement: PropTypes.object,
```
nramamurth: Why is it optional?

nramamurth: Why is it a prop as against [abstract modal implementation](https://github.com/cerner/terra-framework/blob/master/packages/terra-abstract-modal/src/AbstractModal.jsx#L118)?

packages/terra-menu/src/Menu.jsx
```diff
@@ -120,6 +216,8 @@ class Menu extends React.Component {
       targetRef,
       isArrowDisplayed,
       contentWidth,
+      rootSelector,
```
nramamurth: I'd not extract these if they ain't used within `render`.

packages/terra-menu/src/Menu.jsx
```diff
@@ -89,6 +132,47 @@ class Menu extends React.Component {
...
+
+    if (mainDocumentElement) {
+      const inert = +mainDocumentElement.getAttribute('data-overlay-count');
+
+      if (inert === 1) {
+        mainDocumentElement.removeAttribute('data-overlay-count');
+        mainDocumentElement.removeAttribute('inert');
+        if (this.props.triggerElement) {
+          // Shift focus back to element that was last focused prior to opening the menu
+          setTimeout(() => mainDocumentElement.removeAttribute('aria-hidden'), 0);
```
nramamurth: Why do we need to use `setTimeout` for this?

packages/terra-menu/src/Menu.jsx
saket2403: Fixed in  bad1e3c

packages/terra-menu/src/Menu.jsx
saket2403: Fixed in bad1e3c

packages/terra-menu/src/Menu.jsx
saket2403: This [line](https://github.com/cerner/terra-framework/blob/bed6ba818a71d660ae7416724fbd8a3293ecc0c6/packages/terra-abstract-modal/src/AbstractModal.jsx#L118) returns the menu module instead of the button.

packages/terra-menu/src/Menu.jsx
saket2403: Did it similar to [abstract modal](https://github.com/cerner/terra-framework/blob/bed6ba818a71d660ae7416724fbd8a3293ecc0c6/packages/terra-abstract-modal/src/AbstractModal.jsx#L150) to make sure aria-hidden is removed.

packages/terra-menu/src/Menu.jsx
saket2403: I had tried making it required but it gives an error saying the value cannot be null.

packages/terra-menu/src/Menu.jsx
saket2403: Used [abstract modal ](https://github.com/cerner/terra-framework/blob/bed6ba818a71d660ae7416724fbd8a3293ecc0c6/packages/terra-abstract-modal/src/AbstractModal.jsx#L54)as reference

packages/terra-menu/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ Unreleased
 * Cleaned up imports in examples and test files
 * Updated `keycode-js` from `v1.0.4` to `v2.0.1`
 
+### Changed
```
nramamurth: This should perhaps be 'fixed'?

packages/terra-menu/src/Menu.jsx
nramamurth: 🤔

packages/terra-menu/src/Menu.jsx
nramamurth: ..but why? 🤔Is it that the element might not be available on the DOM?

packages/terra-menu/src/Menu.jsx
nramamurth: I don't understand 'for easy selecting of document base component.' means for a consumer, tbh. When should the consumer set this and what should it be set to if not the root element?

packages/terra-tabs/src/_TabMenu.jsx
```diff
@@ -126,6 +127,7 @@ class TabMenu extends React.Component {
           onRequestClose={this.handleOnRequestClose}
           targetRef={this.getTargetRef}
           isOpen={this.state.isOpen}
+          triggerElement={document.getElementById('tab-button')}
```
supreethmr: you can remove this prop from here too.

packages/terra-tabs/src/_TabMenu.jsx
```diff
@@ -115,6 +115,7 @@ class TabMenu extends React.Component {
         onKeyDown={this.handleOnKeyDown}
         className={cx(['tab-menu', { 'is-active': menuActive }])}
         data-terra-tabs-menu
+        id="tab-button"
```
supreethmr: here too Id is not required. it has ref defined already.

packages/terra-menu/src/Menu.jsx
```diff
@@ -108,6 +192,18 @@ class Menu extends React.Component {
     });
   }
 
+  handleKeydown(e) {
```
supreethmr: Do we need this for menu..? Correct me if am wrong because menu has popup which uses hookshot-content and this is being handled in hookshot-content so I don't think we need to handle here.

packages/terra-menu/src/Menu.jsx
saket2403: Prop removed in 15a749c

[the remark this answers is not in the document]
saket2403: Changed in 15a749c

packages/terra-menu/src/Menu.jsx
```diff
@@ -53,12 +57,21 @@ const propTypes = {
    * Otherwise, the menu will display without an arrow and right aligned.
    */
   isArrowDisplayed: PropTypes.bool,
+  /**
+   * Allows assigning of root element custom data attribute for easy selecting of document base component.
+   */
+  rootSelector: PropTypes.string,
```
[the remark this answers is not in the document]
saket2403: Cannot use to `#root` directly in this line `const mainDocumentElement = document.querySelector(this.props.rootSelector);` . So had to declare it as a prop.

packages/terra-tabs/src/_TabMenu.jsx
saket2403: Removed in 15a749c

packages/terra-tabs/src/_TabMenu.jsx
saket2403: Removed in 15a749c

packages/terra-menu/src/Menu.jsx
saket2403: Removed in 15a749c

packages/terra-menu/src/Menu.jsx
saket2403: aria-hidden is not removed if setTimeout is not used.

packages/terra-menu/src/Menu.jsx
saket2403: Prop removed in 15a749c

packages/terra-menu/src/Menu.jsx
saket2403: Changed the description in 15a749c

packages/terra-menu/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+* Added inert/aria-hidden to prevent screen reader from moving focus out of popup.
```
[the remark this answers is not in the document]
dkasper-was-taken: I would imagine that the change could be moved to terra-popup for reuse within terra-menu and terra-date-picker, but I'm unsure about replacing the functionality with terra-overlay.

Unfortunately, while I don't think that it would hurt to use this functionality in terra-popup for a basic use-case, it isn't going to solve the problem when it's presented from a modal as this implementation only affects #root.

@bjankord How do we want to address layered portal'd elements? I imagine the most common use-case for this is going to be terra-collapsible-menu-view within headers of modals. The other likely layer scenario would be the terra-notification-dialog paired with modals and dirty data.

saket2403:
> We can close/backlog this issue. As the proper holistic solution to this involves the implementation of a stack management system to control inert across a combination of disclosures and popups.
> 
> This is also as mike mentioned it's also from a combination of assistive tech and browsers.

Closing the PR for now

PR #884, opened 2019-09-16
Removing details tag from doc-site.

### LINKED ISSUE #883: Remove problematic details tag from doc-site.
# Bug Report

## Description

The inert polyfill doesn't make `details` tag go out of `tab order` which causes an accessibility violation in the main doc-site. 

## @ Mentions

@bjankord

pranav300:
### Summary
Resolves #883


pranav300:
@bjankord Should I also update core and clinical?

bjankord:
@pranav300 Yes, let's apply the same changes to core and clinical repos.

PR #887, opened 2019-09-17
[terra-application-name] Fixed inconsistent failures of application header name tests

avinashg1994:
### Summary
The WDIO test for Application Header Name were inconsistent and during my tests after css-loader update to v3.2.0 (https://github.com/cerner/terra-toolkit/pull/338), the failures became more frequent.
Replacing `browser.pause` with `browser.waitForVisible` resolves this inconsistency.


packages/terra-application-name/tests/wdio/application-header-name-spec.js
```diff
@@ -1,7 +1,7 @@
 Terra.describeViewports('ApplicationHeaderName', ['tiny', 'medium'], () => {
   describe('Displays a default application header name', () => {
     before(() => browser.url('/#/raw/tests/terra-application-name/application-name/application-header-name-default'));
-    browser.pause(3000); // to display image
+    browser.waitForVisible('[class*="demo-image-container"]');
```
supreethmr: So does we need to make similar changes for tabs also instead of adding browser.pause() : [here](https://github.com/cerner/terra-framework/blob/56487c47a6fc41f973fbb5bf817b6bd73e963c74/packages/terra-tabs/tests/wdio/tabs-spec.js#L12)

avinashg1994: In the "Application Header Name" scenario the icon was not being loaded/displayed by the time the screenshot was taken which caused the failure. So by using `browser.waitForVisible` we wait for the icon to be made visible before we proceed further.

Whereas, in the case of the "Responsive Tabs" issue, the Tabs are **already visible** and the failure occurs because of a mismatch in the width of the Tabs by 1 or 2 pixels. So even if we use `browser.waitForVisible` it wouldn't resolve the inconsistency.

PR #888, opened 2019-09-17
Fixed missing hour not updated on re-entering same value

### LINKED ISSUE #458: DateTimePicker allows entering time in missing hour
# Bug Report

## Description

It is possible to enter a time from the missing hour (from the spring DST transition) into the `DateTimePicker`.  Normally when you do so, the time is automatically advanced by the size of the time shift.  However, if one enters a time in the missing hour, and then re-enters the missing time again, the time value saved in the control does not change (it remains the adjusted time from the first time the missing hour was entered), but the time in the actual input field does change, resulting in the-time-the-control-thinks-it-has, and the-time-the-control-is-showing-the-user-that-it-has to no longer be in sync with each other.  Once in this state, it is possible to change the date using the calendar to a different day, and the time field continues to display the value that was entered into it (from the missing hour), while the control continues to think that it has the adjusted time.

## Steps to Reproduce

1. Set your system time to US Central (don't have to be in Australia for this one 😃)
2. Go to the [`DateTimePicker` example page](https://engineering.cerner.com/terra-ui/#/components/terra-date-time-picker/date-time-picker/date-time-picker)
3. Enter March 11, 2018 02:30
4. Observe that the time is automatically adjusted to 03:30, and that the ISO time is `2018-03-11T03:30:00-05:00`
5. Change the hour back to 02
6. Observe that the hour is now 02, but the ISO time remains `2018-03-11T03:30:00-05:00`
7. Use the calendar to change the date to March 12, 2018.
8. Observer that the control now reads 03/12/2018 02:30, but the ISO time that the control thinks it has is `2018-03-12T03:30:00-05:00`

image

## Expected Behavior

If the time doesn't exist, I would expect to be corrected every time I attempt to enter it.  At the very least the time the component thinks it has should be consistent with the time displayed to the user.

## Possible Solution

I'm guessing there is some sort of comparison happening that skips the update if the times are "the same", which they might appear to be if you are using moment to compare a time in the missing hour to what that time gets shifted to when you turn in into a moment object, even though the two values aren't actually the same.

## Environment

* Component Name and Version: `terra-date-time-picker@2.30.1`
* Webpack Version: 4.18.4

supreethmr:
### Summary
Resolves #458 

### Additional Details
Date-Time Picker handles missing hour by replacing it with valid hour when entered, however the value in time-input was not updating on re-entering missing hour second time. This was due to the condition in componentDidUpdate() of [TimeInput](https://github.com/cerner/terra-framework/blob/7b1f5660175d02a77a860c81adac5faa962b30a2/packages/terra-time-input/src/TimeInput.jsx#L179)
```javascript
if (this.props.value === prevProps.value && this.props.variant === prevProps.variant) {
    return;
}
```
Since on initial input of missing hour props.value will get updated to valid time value and on re-entering the same missing hour date-time-picker again updates it to valid time value which will be same as previous hour this was preventing component from getting updated to display valid time value.

To fix this issue added prevState in componentDidUpdate() to handle missing hour changes. 

@cerner/terra

### DIFF
--- a/packages/terra-date-time-picker/src/DateTimePicker.jsx
+++ b/packages/terra-date-time-picker/src/DateTimePicker.jsx
@@ -3,9 +3,9 @@ import { injectIntl, intlShape } from 'react-intl';
 import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
 import DatePicker from 'terra-date-picker';
-import TimeInput from 'terra-time-input';
 import * as KeyCode from 'keycode-js';
 import DateUtil from 'terra-date-picker/lib/DateUtil';
+import TimeInput from 'terra-time-input';
 import styles from './DateTimePicker.module.scss';
 import DateTimeUtils from './DateTimeUtils';
 import TimeClarification from './_TimeClarification';
@@ -348,6 +348,7 @@ class DateTimePicker extends React.Component {
     // If both date and time are valid, check if the time is the missing hour and invoke onChange.
     // If the date is valid but time is invalid, the time in the dateTime state needs to be cleared and render.
     if (validDate && validTime) {
+      // this.handleChange(event, time);
       const updatedDateTime = DateTimeUtils.updateTime(previousDateTime, time, this.props.showSeconds);
 
       if (event.keyCode === KeyCode.KEY_DOWN
--- a/packages/terra-time-input/src/TimeInput.jsx
+++ b/packages/terra-time-input/src/TimeInput.jsx
@@ -133,6 +133,7 @@ class TimeInput extends React.Component {
     this.handleMeridiemInputFocus = this.handleMeridiemInputFocus.bind(this);
     this.handleMeridiemSelectFocus = this.handleMeridiemSelectFocus.bind(this);
     this.handleMeridiemButtonChange = this.handleMeridiemButtonChange.bind(this);
+    this.handleMissingHourChange = this.handleMissingHourChange.bind(this);
 
     let hour = TimeUtil.splitHour(value);
     let meridiem;
@@ -174,16 +175,23 @@ class TimeInput extends React.Component {
     };
   }
 
-  componentDidUpdate(prevProps) {
-    if (
-      this.props.value === prevProps.value
-      && this.props.variant === prevProps.variant
+  componentDidUpdate(prevProps, prevState) {
+    let skipUpdate = true;
+    const prevStateTimeValue = (this.props.showSeconds) ? `${prevState.hour}:${prevState.minute}:${prevState.second}` : `${prevState.hour}:${prevState.minute}`;
+    // Comparing partial time values on each update will result in incorrect comparisons of prevState and propsValue.
+    // Hence validating propsValue and prevState values to skip the comparison between them until time value is completely formed.
+    if (TimeUtil.isValidTime(this.props.value, this.props.showSeconds) && TimeUtil.isValidTime(prevStateTimeValue, this.props.showSeconds)) {
+      skipUpdate = this.handleMissingHourChange(prevState);
+    }
+
+    if (this.props.value === prevProps.value
+      && this.props.variant === prevProps.variant && skipUpdate
     ) {
       return;
     }
 
-    let hour = TimeUtil.splitHour(this.props.value);
     let { meridiem } = this.state;
+    let hour = TimeUtil.splitHour(this.props.value);
 
     if (this.props.variant === TimeUtil.FORMAT_12_HOUR) {
       if (!this.props.intl.messages['Terra.timeInput.am'] || !this.props.intl.messages['Terra.timeInput.pm']) {
@@ -214,6 +222,20 @@ class TimeInput extends React.Component {
       second: TimeUtil.splitSecond(this.props.value),
       meridiem,
     });
+  }
+
+  handleMissingHourChange(prevState) {
+    let hour = TimeUtil.splitHour(this.props.value);
+    if (this.props.variant === TimeUtil.FORMAT_12_HOUR) {
+      hour = TimeUtil.parseTwelveHourTime(hour, this.anteMeridiem, this.postMeridiem).hourString;
+    }
+    const minute = TimeUtil.splitMinute(this.props.value);
+    const second = TimeUtil.splitSecond(this.props.value);
+
+    if (hour !== prevState.hour || minute !== prevState.minute || second !== prevState.second) {
+      return false;
+    }
+    return true;
   }
 
   handleFocus(event) {
--- a/packages/terra-date-time-picker/src/DateTimeUtils.js
+++ b/packages/terra-date-time-picker/src/DateTimeUtils.js
@@ -126,9 +126,7 @@ class DateTimeUtils {
    * @return {boolean} - True if the time is valid.
    */
   static isValidTime(time, hasSeconds) {
-    const timeFormat = hasSeconds ? 'HH:mm:ss' : 'HH:mm';
-    const timeMoment = moment(time, timeFormat, true);
-    return timeMoment.isValid();
+    return TimeUtil.isValidTime(time, hasSeconds);
   }
 
   /**
--- a/packages/terra-time-input/CHANGELOG.md
+++ b/packages/terra-time-input/CHANGELOG.md
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+* Fixed time-input componentDidUpdate for updating time-input on re-entering the same missing hour in time-input.
 
 4.9.0 - (September 6, 2019)
 ------------------
--- a/packages/terra-time-input/package.json
+++ b/packages/terra-time-input/package.json
@@ -29,6 +29,8 @@
     "classnames": "^2.2.5",
     "keycode-js": "^2.0.1",
     "prop-types": "^15.5.8",
+    "moment": "^2.21.0",
+    "moment-timezone": "^0.5.13",
     "terra-button-group": "^3.0.0",
     "terra-form-input": "^2.3.0"
   },
--- a/packages/terra-time-input/src/TimeUtil.js
+++ b/packages/terra-time-input/src/TimeUtil.js
@@ -1,3 +1,5 @@
+import moment from 'moment-timezone';
+
 class TimeUtil {
   /**
    * Determines if a provided nuermic input value is valid.
@@ -7,6 +9,19 @@ class TimeUtil {
    */
   static validNumericInput(value) {
     return value.length === 0 || /^\d+$/.test(value);
+  }
+
+  /**
+   * Determines if the time is a valid time in the HH:mm (where hasSeconds is false) or
+   * HH:mm:ss (where hasSeconds is true) formats
+   * @param {string} time - The time to validate.
+   * @param {boolean} hasSeconds Whether or not the time should consider having seconds valid
+   * @return {boolean} - True if the time is valid.
+   */
+  static isValidTime(time, hasSeconds) {
+    const timeFormat = hasSeconds ? 'HH:mm:ss' : 'HH:mm';
+    const timeMoment = moment(time, timeFormat, true);
+    return timeMoment.isValid();
   }
 
   /**


packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -348,6 +348,7 @@ class DateTimePicker extends React.Component {
     // If both date and time are valid, check if the time is the missing hour and invoke onChange.
     // If the date is valid but time is invalid, the time in the dateTime state needs to be cleared and render.
     if (validDate && validTime) {
+      // this.handleChange(event, time);
```
benbcai: This can be removed.

benbcai:
Can you add a wdio test to test the scenario described on #458?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -216,6 +224,20 @@ class TimeInput extends React.Component {
     });
   }
 
+  handleMissingHourChange(prevState) {
+    let hour = TimeUtil.splitHour(this.props.value);
+    if (this.props.variant === TimeUtil.FORMAT_12_HOUR) {
+      hour = TimeUtil.parseTwelveHourTime(hour, this.anteMeridiem, this.postMeridiem).hourString;
+    }
+    const minute = TimeUtil.splitMinute(this.props.value);
```
nramamurth: This could be done incrementally as well without having to parse everything at once but maybe what you have now is more readable, tho.

```
let hour = TimeUtil.splitHour(this.props.value);
if (this.props.variant === TimeUtil.FORMAT_12_HOUR) {
   hour = TimeUtil.parseTwelveHourTime(hour, this.anteMeridiem, this.postMeridiem).hourString;
}

if (hour === prevState.hour) {
   const minute = TimeUtil.splitMinute(this.props.value);
   if (minute === prevState.minute) {
      const second = TimeUtil.splitSecond(this.props.value);
      return second !== prevState.second;
    }
} 

return true;
```

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -133,6 +133,7 @@ class TimeInput extends React.Component {
     this.handleMeridiemInputFocus = this.handleMeridiemInputFocus.bind(this);
     this.handleMeridiemSelectFocus = this.handleMeridiemSelectFocus.bind(this);
     this.handleMeridiemButtonChange = this.handleMeridiemButtonChange.bind(this);
+    this.doesComponentNeedsUpdate = this.doesComponentNeedsUpdate.bind(this);
```
nramamurth: Let's not call it that. Let's just move the logic inside `componentDidUpdate` and there's no harm, imo.

packages/terra-date-time-picker/src/DateTimePicker.jsx
supreethmr: removed : [commented line](https://github.com/cerner/terra-framework/blob/4ee766a28a07578ea5a76a52bf85cdfded0516f7/packages/terra-date-time-picker/src/DateTimePicker.jsx#L351)

supreethmr:
> Can you add a wdio test to test the scenario described on #458?

Added WDIO Tests for Missing Hour Scenario : [here](https://github.com/cerner/terra-framework/blob/4ee766a28a07578ea5a76a52bf85cdfded0516f7/packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js#L41-L69)

packages/terra-time-input/src/TimeInput.jsx
supreethmr: componentdidUpdate() looks complex if I move this in like below

```javascript
componentDidUpdate(prevProps, prevState) {
  let doesComponentNeedsUpdate = false;
  let { meridiem } = this.state;
  let hour = TimeUtil.splitHour(this.props.value);
  if (this.props.variant === TimeUtil.FORMAT_12_HOUR) {
    if (!this.props.intl.messages['Terra.timeInput.am'] || !this.props.intl.messages['Terra.timeInput.pm']) {
        if (process.env !== 'production') {
          // eslint-disable-next-line no-console
          console.warn('This locale only uses 24 hour clock. The ante meridiem and post meridiem will not be displayed');
        }

        this.anteMeridiem = '';
        this.postMeridiem = '';
      } else {
        this.anteMeridiem = this.props.intl.formatMessage({ id: 'Terra.timeInput.am' });
        this.postMeridiem = this.props.intl.formatMessage({ id: 'Terra.timeInput.pm' });
      }

      if (hour) {
      const parsedHour = TimeUtil.parseTwelveHourTime(hour, this.anteMeridiem, this.postMeridiem);
      hour = parsedHour.hourString;
      // eslint-disable-next-line prefer-destructuring
      meridiem = parsedHour.meridiem;
    }
  }
  const minute = TimeUtil.splitMinute(this.props.value);
  const second = TimeUtil.splitSecond(this.props.value);

  const prevStateTimeValue = (this.props.showSeconds) ? `${prevState.hour}:${prevState.minute}:${prevState.second}` : `${prevState.hour}:${prevState.minute}`;
    // Comparing partial time values on each update will result in incorrect comparisons of prevState and propsValue.
    // Hence validating propsValue and prevState values to skip the comparison between them until time value is completely formed.
    if (TimeUtil.validateTime(this.props.value, this.props.showSeconds) && TimeUtil.validateTime(prevStateTimeValue, this.props.showSeconds)) {
      if (hour !== prevState.hour || minute !== prevState.minute || second !== prevState.second) {
        doesComponentNeedsUpdate = true;
    }
  }

  if (this.props.value === prevProps.value
    && this.props.variant === prevProps.variant && !doesComponentNeedsUpdate
  ) {
    return;
  }

   // eslint-disable-next-line react/no-did-update-set-state
   this.setState({
     hour,
     minute,
     second,
     meridiem,
   });
}
```

all these value assignments, conversions can be skipped if we have this condition before them. hence created new method to compare props and prevState.

nramamurth: imo, this is not complex, and invoking a method named `doesComponentNeedsUpdate` inside this lifecycle method sounds a bit contradictory to me. As far as I understand from the updated code, it looks like you're checking if an update happened and not if an update needs to happen. This method name also confuses with the lifecycle method - `shouldComponentUpdate`.

supreethmr: I agree with your suggestion. the main reason to add method `handleMissinghour()` for this condition is to relate the reason for which these changes are added. I.e. [this](https://github.com/cerner/terra-framework/blob/4ee766a28a07578ea5a76a52bf85cdfded0516f7/packages/terra-time-input/src/TimeInput.jsx#L235) check is will serve it's purpose only when TimeInput is used with DateTimePicker and adds no value for TimeInput when used as individual component. 

Updated :  996f20d7bf9a9ecbab8a126deebe87d22cd72259

packages/terra-time-input/src/TimeInput.jsx
supreethmr: As you mentioned 😃  for readability purpose I would like to keep this condition the way it is now.

packages/terra-time-input/src/TimeInput.jsx
ShettyAkarsh: @nramamurth , @supreethmr  so with new componentupdate we are exiting after execution of code which is necessary only when we need to update the state. we could actually rename this old function **doesComponentNeedsUpdate** and use the previous logic of exiting this method at the beginning itself.

supreethmr: yes that was a one of the reason for having method earlier. I believe moving this code to method improves readability.
would like to know your thoughts on this : @nramamurth

nramamurth: I agree that this could return way sooner than it does now - I am not a big fan of adding a new method for it but maybe you could consider doing something similar. 

<details><summary>return sooner (also have a method (kinda))</summary>
<p>

```javascript
componentDidUpdate(prevProps, prevState) {
  let { meridiem } = this.state;
  let hour = TimeUtil.splitHour(this.props.value);
  const minute = TimeUtil.splitMinute(this.props.value);
  const second = TimeUtil.splitSecond(this.props.value);

  const shouldUpdateState = (() => {
    const prevStateTimeValue = (this.props.showSeconds) ? `${prevState.hour}:${prevState.minute}:${prevState.second}` : `${prevState.hour}:${prevState.minute}`;
    if (TimeUtil.validateTime(this.props.value, this.props.showSeconds) && TimeUtil.validateTime(prevStateTimeValue, this.props.showSeconds)) {
      return (hour !== prevState.hour || minute !== prevState.minute || second !== prevState.second);
    }
  })();

  if (!shouldUpdateState && this.props.value === prevProps.value && this.props.variant === prevProps.variant) {
    return;
  }

  if (this.props.variant === TimeUtil.FORMAT_12_HOUR) {
    if (!this.props.intl.messages['Terra.timeInput.am'] || !this.props.intl.messages['Terra.timeInput.pm']) {
        if (process.env !== 'production') {
          // eslint-disable-next-line no-console
          console.warn('This locale only uses 24 hour clock. The ante meridiem and post meridiem will not be displayed');
        }

        this.anteMeridiem = '';
        this.postMeridiem = '';
      } else {
        this.anteMeridiem = this.props.intl.formatMessage({ id: 'Terra.timeInput.am' });
        this.postMeridiem = this.props.intl.formatMessage({ id: 'Terra.timeInput.pm' });
      }

      if (hour) {
      const parsedHour = TimeUtil.parseTwelveHourTime(hour, this.anteMeridiem, this.postMeridiem);
      hour = parsedHour.hourString;
      // eslint-disable-next-line prefer-destructuring
      meridiem = parsedHour.meridiem;
    }
  }

   // eslint-disable-next-line react/no-did-update-set-state
   this.setState({
     hour,
     minute,
     second,
     meridiem,
   });
}
```
</p>
</details>

supreethmr: as you mentioned above : 
```javascript
 const shouldUpdateState = (() => {
    const prevStateTimeValue = (this.props.showSeconds) ? `${prevState.hour}:${prevState.minute}:${prevState.second}` : `${prevState.hour}:${prevState.minute}`;
    if (TimeUtil.validateTime(this.props.value, this.props.showSeconds) && TimeUtil.validateTime(prevStateTimeValue, this.props.showSeconds)) {
      return (hour !== prevState.hour || minute !== prevState.minute || second !== prevState.second);
    }
  })();
```
we still need ```parsedHour``` for comparison of 12-hour format Time values and parsedHour requires :
 ```Javascript
  if (this.props.variant === TimeUtil.FORMAT_12_HOUR) {
    if (!this.props.intl.messages['Terra.timeInput.am'] || !this.props.intl.messages['Terra.timeInput.pm']) {
        if (process.env !== 'production') {
          // eslint-disable-next-line no-console
          console.warn('This locale only uses 24 hour clock. The ante meridiem and post meridiem will not be displayed');
        }

        this.anteMeridiem = '';
        this.postMeridiem = '';
      } else {
        this.anteMeridiem = this.props.intl.formatMessage({ id: 'Terra.timeInput.am' });
        this.postMeridiem = this.props.intl.formatMessage({ id: 'Terra.timeInput.pm' });
      }

      if (hour) {
      const parsedHour = TimeUtil.parseTwelveHourTime(hour, this.anteMeridiem, this.postMeridiem);
      hour = parsedHour.hourString;
      // eslint-disable-next-line prefer-destructuring
      meridiem = parsedHour.meridiem;
    }
  }
```
So to reduces the changes to be made in componentDidUpdate() created method : [hasTimeValueUpdated](https://github.com/cerner/terra-framework/blob/ef448d8d9d629e0543cdc606e5e855e39df22666/packages/terra-time-input/src/TimeInput.jsx#L231). 

As of now i am not able to see any impact of creating new method will remove method if I found any.

File: packages/terra-time-input/src/TimeInput.jsx [hunk 4]
```diff
@@ -216,6 +224,24 @@ class TimeInput extends React.Component {
     });
   }
 
+  /* When Time-Input is used with date-time-picker time-input In missing hour scenario Time-Input skips state update due to
+     prevProps and props comparsion. To handle this scenario propValue is being validated with prevState values in below method
+     to ensure state update happens on time value change.
+  */
+  hasTimeValueUpdated(prevState) {
+    let hour = TimeUtil.splitHour(this.props.value);
+    if (this.props.variant === TimeUtil.FORMAT_12_HOUR) {
```
benbcai: Reminder: Depending on the merge order between this and #898, `this.props.variant` here needs to be updated to `TimeUtil.getVariantFromLocale(this.props.variant)`.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -216,6 +224,24 @@ class TimeInput extends React.Component {
     });
   }
 
+  /* When Time-Input is used with date-time-picker time-input In missing hour scenario Time-Input skips state update due to
+     prevProps and props comparsion. To handle this scenario propValue is being validated with prevState values in below method
```
benbcai: - "Time-Input" -> "time-input"
- "In" -> "in"
- "comparision" -> "comparison"
- "propValue" -> " the value prop"

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -174,16 +175,23 @@ class TimeInput extends React.Component {
...
-    if (
-      this.props.value === prevProps.value
-      && this.props.variant === prevProps.variant
+  componentDidUpdate(prevProps, prevState) {
+    let skipStateUpdate = true;
+    const prevStateTimeValue = (this.props.showSeconds) ? `${prevState.hour}:${prevState.minute}:${prevState.second}` : `${prevState.hour}:${prevState.minute}`;
+    // Comparing partial time values on each update will result in incorrect comparisons of prevState and propsValue.
+    // Hence validating propsValue and prevState values to skip the comparison between them until time value is completely formed.
+    if (TimeUtil.validateTime(this.props.value, this.props.showSeconds) && TimeUtil.validateTime(prevStateTimeValue, this.props.showSeconds)) {
+      skipStateUpdate = this.hasTimeValueUpdated(prevState);
```
benbcai: If the time-input passes in a valid `value` prop, it would always get in here and prevents manually changing the time. You can go to this [example](https://terra-framework-deploye-pr-888.herokuapp.com/tests/terra-time-input/time-input/time-input/default-time) and attempt to manually change the time.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -216,6 +224,24 @@ class TimeInput extends React.Component {
...
   }
 
+  /* When Time-Input is used with date-time-picker time-input In missing hour scenario Time-Input skips state update due to
+     prevProps and props comparsion. To handle this scenario propValue is being validated with prevState values in below method
+     to ensure state update happens on time value change.
+  */
+  hasTimeValueUpdated(prevState) {
+    let hour = TimeUtil.splitHour(this.props.value);
+    if (this.props.variant === TimeUtil.FORMAT_12_HOUR) {
+      hour = TimeUtil.parseTwelveHourTime(hour, this.anteMeridiem, this.postMeridiem).hourString;
```
benbcai: A check for a valid `hour` should be added here before calling `parseTwelveHourTime` as an empty or undefined hour would return NaN. See [here](https://github.com/cerner/terra-framework/blob/Update-Missing-hour/packages/terra-time-input/src/TimeInput.jsx#L155-L162) for an example.

packages/terra-time-input/src/TimeInput.jsx
supreethmr: Thank you for adding remainder here. will follow-up with #898 PR and will update these changes accordingly.

packages/terra-time-input/src/TimeInput.jsx
supreethmr: Since all this check will still happens even if we skip update. removed this method and moved this code to [componentDidUpdate() ](9c2a5b2eb31f2d3ca94033865f243793f5add4cc)

benbcai: @supreethmr I have two other ideas that could potentially handle this whole scenario but I wanted to test it first and will update you.

packages/terra-time-input/src/TimeInput.jsx
supreethmr: this issue has been addressed with : 7d1a4a89def749924df1b0ac06c1ae5887a71939

packages/terra-time-input/src/TimeInput.jsx
supreethmr: I have simplified the comparisons of props value to update state and also I have added condition to update state only when props gets updated : 7d1a4a89def749924df1b0ac06c1ae5887a71939

benbcai: @supreethmr I made some [changes](https://github.com/cerner/terra-framework/compare/master...force-update-missing-hour) here that you can take a look and test to see how this works. The changes in date-time-picker basically allows the time-input to render the missing hour. Then in `componentDidUpdate` if the time is the missing hour it will update the time value and force a render with the correct time.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -639,6 +639,7 @@ class DateTimePicker extends React.Component {
 
         <div className={cx('time-facade')}>
           <TimeInput
+            data-date-time-input
```
nramamurth: The way this is being used in `TimeInput` does not make sense to me. Do you intend this to be a data attribute on a DOM element? I see that this is used as a 'flag' to indicate that `DateTimePicker` created `TimeInput`, is it meant to be just that?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -221,11 +215,28 @@ class TimeInput extends React.Component {
       }
     }
 
+    let skipStateUpdate = true;
+    // Data Attribute `data-date-time-input` of date-time-picker is used to ensure state update happens only when props are updated in date-time-picker.
```
nramamurth: What does this comment mean? How is this used to ensure state update happens only when props are updated in `date-time-picker`?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -221,11 +215,28 @@ class TimeInput extends React.Component {
...
+    const dateTimePicker = 'data-date-time-input';
+    if (this.props[dateTimePicker]) {
+      const inputValue = (this.props.showSeconds) ? `${this.hourInput.value}:${this.minuteInput.value}:${this.secondInput.value}` : `${this.hourInput.value}:${this.minuteInput.value}`;
+      const propsValue = (this.props.showSeconds) ? `${hour}:${minute}:${second}` : `${hour}:${minute}`;
+      // Comparing partial time values on each update will result in incorrect comparisons of inputValue and propsValue. Hence validating propsValue and inputValue before comparison.
+      if (TimeUtil.validateTime(propsValue, this.props.showSeconds) && TimeUtil.validateTime(inputValue, this.props.showSeconds)) {
+        skipStateUpdate = (propsValue === inputValue);
```
nramamurth: I am certain I could be missing something here but do the member variables `hourInput`, `minuteInput`, `secondInput` reflect the values stored as the state? If so, what are we _actually_ comparing here to check whether or not the component has updated? We compare the current props and previous props values in Line 231 (which sounds right), are we comparing the current state value against current props value here?

packages/terra-time-input/src/TimeInput.jsx
supreethmr: this changes has been removed : 94624440855ecca7249f9c3cd35ae80cda292300

packages/terra-time-input/src/TimeInput.jsx
supreethmr: yes those variables actually reflects state values and Now instead of comparing states with props now I have updated Date-Time-Picker to call forceUpdate when timeValue and displayed are not same. : 94624440855ecca7249f9c3cd35ae80cda292300.

packages/terra-time-input/src/TimeInput.jsx
supreethmr: Thank you @benbcai  Changes works fine I have updated the code :  94624440855ecca7249f9c3cd35ae80cda292300

packages/terra-date-time-picker/src/DateTimePicker.jsx
supreethmr: yes it was meant to be just for that Until @benbcai suggested better way to fix this issue. 😄

benbcai: @supreethmr Is data-date-time-input still needed or can it be removed?

packages/terra-date-time-picker/src/terra-dev-site/doc/example/DateTimePickerDefaultDateTime.jsx
```diff
@@ -18,7 +19,14 @@ const defaultProps = {
...
     super(props);
-    this.state = { dateTime: this.props.value };
+    let dateTimeDisplay = props.value;
+    const dateTime = DateTimeUtils.createSafeDate(dateTimeDisplay);
+
+    if (dateTime && dateTime.isValid()) {
+      dateTimeDisplay = dateTime.format();
+    }
+
+    this.state = { dateTime: dateTimeDisplay };
```
benbcai: I don't think you need these changes in this specific example. The value prop set below using `moment().format()` would never be the missing hour.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -224,8 +223,8 @@ class TimeInput extends React.Component {
     // eslint-disable-next-line react/no-did-update-set-state
     this.setState({
       hour,
-      minute: TimeUtil.splitMinute(this.props.value),
-      second: TimeUtil.splitSecond(this.props.value),
+      minute,
+      second,
```
benbcai: Looks like these changes are just code formatting and moving the declarations up. Do you want to keep these changes or get rid of them? If keeping them, I suggest adding an actual entry in the change log.

packages/terra-date-time-picker/src/DateTimePicker.jsx
supreethmr: I don't we need data attribute in date-time-picker I will remove this.

packages/terra-date-time-picker/src/terra-dev-site/doc/example/DateTimePickerDefaultDateTime.jsx
supreethmr: yes, that's right. reverted the code changes : 4aaa072145241a931da717a591d0a5f9e644356f

packages/terra-time-input/src/TimeInput.jsx
supreethmr: Thank you I missed it, I would keep Terra-Time-Input to be unchanged for this Issue. since Missing hour will handled by date-time-picker. : 4aaa072145241a931da717a591d0a5f9e644356f

benbcai:
@PaulSolDev This is ready for functional review. Here is the deployment [link](https://terra-framework-deploye-pr-888.herokuapp.com/components/terra-date-time-picker/date-time-picker/date-time-picker). Please let @supreethmr or I know if you have questions on what needs to be verified.

PaulSolDev:
Functional verification complete.  No issues identified.

PR #891, opened 2019-09-18
Enhancements to terra-application (loading overlay, error boundary, navigation prompt integration)

tbiethman:
### Summary

Adding a couple new APIs to terra-application to support our ongoing application framework development.

Documentation is bare-bones at the minute. I'd like to give a larger overhaul once we get every new piece of functionality landed. What's added here is the bare-minimum.


packages/terra-application/translations/es.json
```diff
@@ -0,0 +1,3 @@
+{
```
emilyrohrbough: needs translations

benbcai:
Can you also add the en-CA translation file?

packages/terra-application/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ Changelog
 
 Unreleased
 ----------
+### Added
+* Added ApplicationLoadingOverlay and ApplicationErrorBoundary
```
emilyrohrbough: There are more enhancements that need added to this list

packages/terra-application/src/application-error-boundary/ApplicationErrorBoundary.jsx
```diff
@@ -0,0 +1,61 @@
...
+   * @private
+   * Object containing translation APIs
+   */
+  intl: intlShape,
+};
+
+class ApplicationErrorBoundary extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { hasError: false, reloadKey: false };
```
emilyrohrbough: can we rename `reloadKey` to a name that indicates it's a boolean?

packages/terra-application/src/application-loading-overlay/ApplicationLoadingOverlay.jsx
```diff
@@ -0,0 +1,41 @@
...
+    if (isOpen) {
+      applicationLoadingOverlay.show(overlayId, message);
+    } else {
+      applicationLoadingOverlay.hide(overlayId);
+    }
+
+    return () => {
+      applicationLoadingOverlay.hide(overlayId);
+    };
+  }, [isOpen, message, id, applicationLoadingOverlay]);
```
emilyrohrbough: When would you expect the `id` ref value to change?

packages/terra-application/src/application-loading-overlay/ApplicationLoadingOverlayProvider.jsx
```diff
@@ -0,0 +1,86 @@
...
+    },
+  }), []);
+
+  const registeredOverlayKeys = Object.keys(loadingIndicators);
+  const overlay = (
+    <LoadingOverlay
+      isRelativeToContainer
+      isAnimated
+      isOpen={!!registeredOverlayKeys.length}
+      message={Object.keys(loadingIndicators).map(key => (
```
emilyrohrbough: This feels a little weird to me. I seems like an overlay can be registered and then unregistered in state via the show/hide overlay functions, but why not just override the current message instead of join the messages together?

packages/terra-application/src/application-loading-overlay/ApplicationLoadingOverlay.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+const propTypes = {
+  /**
+   * A boolean value indicating whether the loading overlay should be visible or not.
+   */
+  isOpen: PropTypes.bool,
+  /**
+   * A string to be rendered within the loading overlay when open.
+   */
+  message: PropTypes.string,
```
emilyrohrbough: The say the loading overlay provider is implemented, it seems like this prop is required because the Loading Overlay's default message doesn't seem like it would be hit. Should this be required?

packages/terra-application/src/terra-dev-site/doc/application/ApplicationLoadingOverlay.md
```diff
@@ -0,0 +1,39 @@
+# Terra Application
+
+## Application Loading Overlay
+
+The ApplicationLoadingOverlay allows for the presentation of loading overlays within the Terra Application framework.
+
+Mounting an open ApplicationLoadingOverlay will result in the framework rendering a loading overlay. Unmounting the ApplicationLoadingOverlay (or removing the `isOpen` prop) will subsequently remove the loading overlay.
+
+### ApplicationLoadingOverlay Props
```
emilyrohrbough: could we render these props in the template so we don't have to worry about these get missed/out-of-date when code changes occur?

packages/terra-application/tests/wdio/application-base/application-base-spec.js
```diff
@@ -0,0 +1,73 @@
+Terra.describeViewports('ApplicationBase', ['small', 'large'], () => {
+  describe('breakpoint', () => {
```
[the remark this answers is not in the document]
tbiethman: They're validating that the correct breakpoint string and locale string are being rendered. The screenshots are the same because I'm using the same test example for both tests.

I just wanted a very high-level test to ensure that the context communication is happening. Should something fail in the future, the strings will either render something else (or not render anything at all).

packages/terra-application/src/terra-dev-site/doc/application/ApplicationLoadingOverlay.md
tbiethman: We could if react-docgen was smarter. Because its a functional component that returns null, react-docgen doesn't recognize the component and cannot generate a props-table for it. I didn't feel like rewriting the component to be a class to work around this, but maybe this will work better with a future version of react-docgen?

packages/terra-application/src/application-loading-overlay/ApplicationLoadingOverlayProvider.jsx
tbiethman: Good question.

```
   LOP
    |
    A
  /   \
B       C
```

If component B renders an overlay, and component C also renders an overlay at the same time, we wouldn't want to overwrite the message, right? Whoever renders last is going to erase the message of the first (even when the first is still loading).

But tbh I'm not super happy about joining the messages together either. I would rather remove the messages altogether if we can. @neilpfeiffer any insight into how we'd want to handle something like this?

packages/terra-application/src/application-loading-overlay/ApplicationLoadingOverlay.jsx
tbiethman: Hmm, that's just a bug. If no messages are provided, we should default back to the default string.

packages/terra-application/src/application-error-boundary/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './ApplicationErrorBoundary';
```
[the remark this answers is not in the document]
tbiethman: It allows consumers to import the ApplicationErrorBoundary like this:
`import ApplicationErrorBoundary from 'terra-application/lib/application-error-boundary';`

instead of

`import ApplicationErrorBoundary from 'terra-application/lib/application-error-boundary/ApplicationErrorBoundary';`

The other sub-directories have index files as well, but they're exporting multiple components. This one is a little weird since we only have a single component, but I put it in here to be consistent.

It might make more sense to just export everything from a single index file at the application root. I was trying to logically divide things a bit, but everything here is likely going to be included in the bundle anyway.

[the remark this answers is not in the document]
emilyrohrbough: I personally would prefer a root index export so you could do either:
```
import TerraApplication from 'terra-application';
// or 
import {  ApplicationErrorBoundary } from 'terra-application';
```

packages/terra-application/tests/jest/application-loading-overlay/ApplicationLoadingOverlay.test.jsx
```diff
@@ -0,0 +1,125 @@
...
+      return reactUseContext(contextValue);
+    };
+  });
+
+  afterAll(() => {
+    React.useContext = reactUseContext;
+  });
+
+  it('should render loading overlay as closed', () => {
+    loadingOverlayContextValue = {
```
yuderekyu: why don't we define this globally and return it in the beforeAll useContext mock, instead of setting it per it block?

packages/terra-application/src/application-loading-overlay/ApplicationLoadingOverlayProvider.module.scss
```diff
@@ -0,0 +1,5 @@
+:local {
+  .container {
+    height: 100%;
```
yuderekyu: semi-related, but shouldn't the terra overlay container always be at height 100%? it seems weird we have to set this

packages/terra-application/tests/jest/application-loading-overlay/ApplicationLoadingOverlay.test.jsx
yuderekyu: ah, we need to reset it - should we do a reset in the afterall then?

packages/terra-application/src/application-error-boundary/ApplicationErrorBoundary.jsx
```diff
@@ -0,0 +1,64 @@
...
+import StatusView from 'terra-status-view';
+
+const propTypes = {
+  /**
+   * Components to render within the context of the ApplicationErrorBoundary
+   */
+  children: PropTypes.node,
+};
+
+class ApplicationErrorBoundary extends React.Component {
```
mjhenkes: You need to add more comments to this guy explaining why and what you're doing

packages/terra-application/src/terra-dev-site/doc/application/ApplicationLoadingOverlay.md
mjhenkes: Could you log your props table enhancement requests to the terra-dev-site repo?

packages/terra-application/src/terra-dev-site/doc/example/ApplicationLoadingOverlayExample.jsx
```diff
@@ -0,0 +1,61 @@
+import React, { useState } from 'react';
+import classNames from 'classnames/bind';
+import ApplicationLoadingOverlay from '../../../application-loading-overlay/ApplicationLoadingOverlay';
```
mjhenkes:
```suggestion
import ApplicationLoadingOverlay from 'terra-application/lib/application-loading-overlay/ApplicationLoadingOverlay';
```

mjhenkes: You could change all of the relative paths in examples this way.

packages/terra-application/src/terra-dev-site/doc/application/ApplicationLoadingOverlay.md
emilyrohrbough: @tbiethman I got around this by moving my propTypes to it's own file and then importing it as a codeblock with the new codeblock loader: 
```
import PropsTable from '.../../applicationLoadingOverlayPropTypes?dev-site-codeblock';
```
Similar to what we do in terra-dev-site for it's config: https://engineering.cerner.com/terra-dev-site/getting-started/terra-dev-site/site-config/site-config

tbiethman: I'm going to be circling back around on the docs here in a future PR; I'll cover this scenario with the holistic update to mdx.

packages/terra-application/src/application-loading-overlay/ApplicationLoadingOverlay.jsx
emilyrohrbough: @tbiethman were changes made to fix this??

packages/terra-application/src/application-loading-overlay/ApplicationLoadingOverlayProvider.jsx
emilyrohrbough: How did we decide to handle this?

tbiethman:
We've decided to break terra-application out of the monorepo to head off concerns we have about maintainability long-term. Rather than merge this here, I'm going to lift this PR branch as the basis for the new package.

packages/terra-application/src/application-loading-overlay/ApplicationLoadingOverlayProvider.jsx
tbiethman: The messages were removed entirely due to future UX recommendations.

packages/terra-application/src/application-loading-overlay/ApplicationLoadingOverlay.jsx
tbiethman: The messages were removed entirely due to future UX recommendations.

packages/terra-application/src/application-loading-overlay/ApplicationLoadingOverlayProvider.module.scss
tbiethman: I can see it being required in most cases but not all. If the OverlayContainer decides to invert its height handling at some point, we can remove this.

PR #900, opened 2019-09-24
[terra-slide-panel] Added ariaLabel prop to slide panel container

### LINKED ISSUE #889: Add aria-label to the Slide-Panel container
# Bug Report

## Description
UX has indicated that when the slide panel opens, it should focus on the panel, not the first interact-able element.

@scottwilmarth mentioned:

> Focus should go to the panel after a user selects an option that opens a detail panel. Having focus go to the panel allows the assistive technology user (screen reader for instance) to understand where they are and allow them to experience all of the content in the panel. If we were to place focus on an element inside of the panel, they may not understand there is other information above the current focused element. The detail pane would have a title announced to the user to provide more context. And, when the user is done with the panel and closes it by clicking the close button or pressing escape, the focus should return to the element that originally opened the panel.

## Steps to Reproduce

1. Navigate to https://engineering.cerner.com/terra-ui/components/terra-slide-panel/slide-panel/slide-panel
2. Click on the "Main Toggle Panel" button in the first example 
3. Use the developer console to inspect the active element: `document.activeElement`

## Expected Behavior
Focus should go to the panel after a user selects an option that opens a detail panel. 

## Possible Solution
Look at adding a tabindex to the panel so focus can be shifted to it instead of the first interact-able element within the panel. May need to add an default aria-label to the container with a new prop that allows people to set their own aria-label to "allow the assistive technology user (screen reader for instance) to understand where they are"

Need to test this with VoiceOver on iOS and sometimes it has shifting focus programmatically to non-interactive elements.

## Environment

* Component Name and Version: terra-slide-panel

## @ Mentions
@scottwilmarth

avinashg1994:
### Summary
Added ariaLabel prop to the slide panel container with a default value as "Details Panel".
Also corrected the Doc test page to show default focus behaviour.

Resolves #889


packages/terra-slide-panel/src/SlidePanel.jsx
```diff
@@ -53,6 +58,7 @@ const propTypes = {
 };
 
 const defaultProps = {
+  ariaLabel: 'Details Panel',
```
avinashg1994: Do we require translations for the default value of `aria-label`?

neilpfeiffer: I don't think it should have a default prop value -- as the generic slide-panel containter can be used for a wide variety of uses well beyond the common Details Panel, include some layouts using the child panel as the "primary"/navigation panel content that launches pre-opened with the `panelBehavior=squish` prop value. 

I would prefer to leave the new `ariaLabel` prop with no default value - force consumers to add it if their situation needs it for accessibity and label it correctly for their usage.

packages/terra-slide-panel/src/SlidePanel.jsx
```diff
@@ -11,6 +11,11 @@ const SlidePanelPositions = {
 };
 
 const propTypes = {
+  /**
+   * String that labels the Details Panel for screen readers.
+   */
+  ariaLabel: PropTypes.string,
```
neilpfeiffer: prop name is too generic - not sure as a consumer whether this applies to the `main` content area or the `panel` content area. 

Would suggest changing this to `panelAriaLabel`, and if needed - also add a companion `mainAriaLabel`.

emilyrohrbough: IMO this should be marked as required in the next MVB of slide-panel since this is required for accessibility. Should we log an issue to make this change on the next MVB?

ShettyAkarsh: Yes i agree. will go ahead and log an issue.

avinashg1994:
Updated the following:

- Removed the default value
- Added `panelAriaLabel` and `mainAriaLabel` for the "Panel Content Area" and "Main Content area" respectively.
- Updated Jest tests and snapshots.

packages/terra-slide-panel/src/SlidePanel.jsx
avinashg1994: Issue link to update the props to be **Required** in the next MVB : https://github.com/cerner/terra-framework/issues/902

PR #911, opened 2019-10-01
[terra-date-picker] Aria-Disabled added.

### LINKED ISSUE #908: terra-date-picker with min/max dates fails WDIO tests because of color contrast issues
# Bug Report
`Terra-date-picker` with min/max dates fails WDIO tests because of color contrast issues 

## Description

Using `terra-date-picker` with min/max dates will cause accessibility issues when running WDIO tests because of color contrast issues between the disabled dates(`#cccccc`) and the background(`#ffffff`). 

## Steps to Reproduce

1. Use terra-date-picker component and set either minDate or maxDate.
2. Run wdio test with an action that opens up calendar picker (something like this `browser.click('.DatePicker-module__button__...'`);

Screen Shot 2019-09-26 at 3 47 25 PM
Screen Shot 2019-09-27 at 9 20 47 AM

## Expected Behavior

WDIO tests shouldn't fail because of color contrast issues.

## Possible Solution

Posted on the teams channel and it seems _the aria-disabled attribute should be implemented for dithered date values._

## Environment

* Component Name and Version: `terra-date-picker@4.13.0`
* Node/npm Version: Node 10.16.2, npm 6.9.0
* Operating System and version (desktop or mobile): MacOS

## @ Mentions

@dkasper-was-taken answered to my post on Teams for me to open the bug here.

saket2403:
### Summary
Fixes #908


benbcai:
I don't know that setting aria-disabled for the disabled days would necessarily fix the color-contrast issue. We initially [disabled color-contrast](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/tests/wdio/date-picker-spec.js#L2) in accessibility tests because we didn't have access to the [react-datepicker styles](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/src/react-datepicker/stylesheets/datepicker.scss). Now that we've forked it, my thought to fix this color contrast issues was to use a [darker color for the disabled days](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/src/react-datepicker/stylesheets/datepicker.scss#L223) and ultimately to remove [ignoredA11y](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/tests/wdio/date-picker-spec.js#L1-L4).

bjankord:
We'll want to check and see if setting this aria-attribute allows axe-core to pass on checking the contrast ratio for the text within it.

saket2403:
> We'll want to check and see if setting this aria-attribute allows axe-core to pass on checking the contrast ratio for the text within it.

I removed ignoredA11y and ran wdio looks like setting aria-disabled fixes the color contrast issue mentioned. But there is another color contrast issue (ratio 3.83) which pops up due to this https://github.com/cerner/terra-framework/blob/628ae60412da7756b66ede35619c9a14f41379c6/packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss#L202.
Although changing it to this:
https://github.com/cerner/terra-framework/blob/628ae60412da7756b66ede35619c9a14f41379c6/packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss#L197 fixes the issue.
Should I make this change and remove the ignore rules for color contrast. 
@benbcai @bjankord

ShettyAkarsh:
adding @neilpfeiffer

benbcai:
@saket2403 I will take your changes with removing ignoredA11y and test in the theme repos to see what the impacts are.

saket2403:
> @saket2403 I will take your changes with removing ignoredA11y and test in the theme repos to see what the impacts are.

I have tested the theme repos they are fine, will just wait for UX input on the color change @benbcai

benbcai:
@saket2403 I have also tested in both the clinical and fusion themes. Removing the color-contrast rule doesn't introduce any issue other than needing to update the wdio screenshots due to having to change the keyboard select background color. I think we should proceed to get a darker color from UX and remove the color-contrast rule.

jeremyfuksa:
**UX Input**

The first gray value that passes contrast ratio tests is `#757575`. But, if you look at that implemented in the component, that value doesn't yield enough contrast between disabled and enabled dates:

Screen Shot 2019-10-10 10 55 31 AM

saket2403:
> I'm seeing above what happens when ignoring this in WDIO tests, but did we try adding the`aria-disabled` attribute and seeing if that helped the issue?

`Aria-disabled` allows the test to pass without changing the gray color. The other contrast issue is for the blue color, for which we require the UX input.Primarily beacuse of these 2 lines:

https://github.com/cerner/terra-framework/blob/628ae60412da7756b66ede35619c9a14f41379c6/packages/terra-date-picker/src/react-datepicker/stylesheets/variables.scss#L5

https://github.com/cerner/terra-framework/blob/628ae60412da7756b66ede35619c9a14f41379c6/packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss#L202

So I need a new color for this. @jeremyfuksa

benbcai:
@jeremyfuksa  Sounds like adding `aria-disabled` to the disabled days will fix the color-contrast violation for the disabled days so a different gray color is not needed. The issue @saket2403 runs into now is the color contrast between with the background-color and the text color when a day in the calendar is preselected using the keyboard. Should we make the background-color for the pre-selected day (as well as when hovered) match what we use as the background-color for the [emphasis terra-button](https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/_variables.scss#L28), which is `#0079be `?

jeremyfuksa:
> Should we make the background-color for the pre-selected day (as well as when hovered) match what we use as the background-color for the [emphasis terra-button](https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/_variables.scss#L28), which is `#0079be ` or do you have a different background-color we should use?

That color will work.

saket2403:
Color updated and ignoreA11y removed in 271216a

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
```diff
@@ -198,22 +198,22 @@
...
 
     &--keyboard-selected {
-      background-color: lighten($datepicker-selected-color, 10%);
+      background-color: $datepicker-selected-color;
       border-radius: $datepicker-border-radius;
       color: #fff;
 
       &:hover {
-        background-color: darken($datepicker-selected-color, 5%);
+        background-color: $datepicker-selected-color;
```
benbcai: The background-color for hover should not be the same as `$datepicker-selected-color`. It should be `#0065a3` to match [hover-background-color for the emphasis button](https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/_variables.scss#L36). A new variable may be needed for this.

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
```diff
@@ -198,22 +198,22 @@
...
       }
     }
 
     &--in-selecting-range:not(&--in-range) {
-      background-color: rgba($datepicker-selected-color, 0.5);
+      background-color: $datepicker-selected-color;
```
benbcai: Selecting a range of date is currently not supported but this is not being used today. Maybe we should leave it the way it was instead of changing it since updating it would have the same background color as a single selected date. We can come back to this if we ever allow range selection.

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
saket2403: Changed in 7802d23

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
saket2403: Reverted in 7802d23

packages/terra-date-picker/src/react-datepicker/stylesheets/variables.scss
```diff
@@ -2,7 +2,8 @@ $datepicker-background-color: #f0f0f0 !default;
 $datepicker-border-color: #aeaeae !default;
 $datepicker-highlighted-color: #3dcc4a !default;
 $datepicker-muted-color: #ccc !default;
-$datepicker-selected-color: #216ba5 !default;
+$datepicker-selected-color: #0079be !default;
+$datepicker-hover-color: #0065a3 !default;
```
benbcai: Can we rename `$datepicker-hover-color` to `$datepicker-selected-hover-color` to better represent that this is the hover colored used only when the day is selected.

saket2403: Changed in eee3598

saket2403:
@benbcai. If this PR looks good to go can we remove the Functional Review label and merge?

benbcai:
@saket2403 Engineering review is complete but we need @PaulSolDev to do a quick functional review and then we should be able to merge.

PaulSolDev:

PR #915, opened 2019-10-02
Fix wdio test url.

benbcai:
### Summary
- Fix wdio test url in terra-application-navigation.
- Remove extra comma in scss files.

### Additional Details
- The following error occurs in the clinical theme repo. The cause appears to be having an incorrect test url. I would expect the same test in terra-frame to fail but that's not the case. Not sure why it is only happening for the clinical theme. The header toggle button does appear in the clinical theme dev site.

```
1) [small] ApplicationNavigation - Small Nav drawer button should be displayed when only help, settings or logout utilities are present open drawer:
element ("[data-compact-header-toggle="true"]") still not visible after 3000ms
running chrome
Error: element ("[data-compact-header-toggle="true"]") still not visible after 3000ms
    at elements("[data-compact-header-toggle="true"]") - isVisible.js:54:17
    at isVisible("[data-compact-header-toggle="true"]") - waitForVisible.js:73:22
```

@cerner/terra

### DIFF
--- a/packages/terra-menu/src/MenuItem.module.scss
+++ b/packages/terra-menu/src/MenuItem.module.scss
@@ -56,7 +56,7 @@
     }
 
     .start-icon {
-      color: var(--terra-menu-item-disabled-start-icon-color, #9b9fa1);;
+      color: var(--terra-menu-item-disabled-start-icon-color, #9b9fa1);
     }
   }
 
--- a/packages/terra-application-navigation/CHANGELOG.md
+++ b/packages/terra-application-navigation/CHANGELOG.md
@@ -7,6 +7,9 @@ Unreleased
 * Adjusted title margin and padding
 * Update tests and screenshots to support theme tests
 
+### Fixed
+* Fix wdio test url.
+
 1.6.0 - (September 26, 2019)
 ------------------
 ### Changed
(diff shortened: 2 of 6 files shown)


packages/terra-menu/src/MenuItem.module.scss
```diff
@@ -56,7 +56,7 @@
     }
 
     .start-icon {
-      color: var(--terra-menu-item-disabled-start-icon-color, #9b9fa1);;
+      color: var(--terra-menu-item-disabled-start-icon-color, #9b9fa1);
```
supreethmr: I am not sure why this was not caught by style-lint. style lint should have caught this while building the code. if we don't have rule specified for extra-semicolon error then we might need to add rule 
``` "no-extra-semicolons":true, ``` in [stylelint-config-terra](https://github.com/cerner/stylelint-config-terra/blob/f2b8bc0f908df3ada150d077e2964dd6996063f4/stylelint.config.js#L14) to catch these (copy-paste) errors in earlier stage.

benbcai: @supreethmr Good suggestion. PR to add the `no-extra-semicolons` rule: https://github.com/cerner/stylelint-config-terra/pull/31

PR #917, opened 2019-10-03
[terra-abstract-modal] Place focus onto the HTML body element upon modal dismissal in IE in special cases.

### LINKED ISSUE #916: [terra-abstract-modal] Place focus onto the HTML body element upon dismissal in IE.
## Description

This is a follow-up to #901 and here's the rationale for this change request: 

When a modal is disclosed by clicking on a SVG element, since IE considers the SVG element as the `modalTrigger` while the other major browsers consider the HTML body element as the `modalTrigger`, the focus goes back to the `body` element on Chrome, Safari, etc. while focus does not get restored on IE on any element with the fix that went in. In such a scenario, to maintain the same behavior across all browsers as of today, then I think that we may have to explicitly set the focus back to the 'body' element - 

```diff
if (this.state.modalTrigger && this.state.modalTrigger.focus) {
   // Shift focus back to element that was last focused prior to opening the modal
   this.state.modalTrigger.focus();
- }
+ } else {
+  document.getElementsByTagName('body')[0].focus();
}

```

## Steps to Reproduce
Run the reduced test sample below and check the active element on the page on Chrome and on IE after dismissing the modal. 

<details> <summary> Reduced Test Case </summary>

```
import React from 'react';
import { withDisclosureManager, disclosureManagerShape } from 'terra-disclosure-manager';
import ModalManager from 'terra-modal-manager';

class DisclosureComponent extends React.Component {
  constructor(props) {
    super(props);

    this.disclose = this.disclose.bind(this);
    this.closeDisclosure = this.closeDisclosure.bind(this);
  }

  disclose(size, dimensions) {
    const { disclosureType, nestedIndex } = this.props;

    const newIndex = nestedIndex + 1;
    return () => {
      this.props.disclosureManager.disclose({
        preferredType: disclosureType,
        size,
        dimensions,
        content: {
          key: `DemoContainer-${newIndex}`,
          component: <WrappedDisclosureComponent identifier={`DemoContainer-${newIndex}`} nestedIndex={newIndex} />,
        },
      });
    };
  }

  closeDisclosure() {
    this.props.disclosureManager.closeDisclosure();
  }

  render() {
    const { disclosureManager, identifier } = this.props;

    return (
      <ContentContainer id={identifier} className="nested-component" fill header={<h2>IE Focus Issue</h2>}>
        <br />
        <div>
          <svg width="75" height="25">
            <g>
              <rect x="0" y="0" width="100%" height="100%" fill="#da552f" />
              <text x="10" y="17" onClick={this.disclose()}>Disclose</text>
            </g>
          </svg>
        </div>
        {disclosureManager && disclosureManager.closeDisclosure ? <button type="button" className="close-disclosure" onClick={this.closeDisclosure}>Close Disclosure</button> : null}
      </ContentContainer>
    );
  }
}

DisclosureComponent.propTypes = {
  disclosureManager: disclosureManagerShape,
  identifier: PropTypes.string,
  disclosureType: PropTypes.string,
  nestedIndex: PropTypes.number,
};

DisclosureComponent.defaultProps = {
  nestedIndex: 0,
};

const WrappedDisclosureComponent = withDisclosureManager(DisclosureComponent);

const ModalManagerDefault = () => (
  <div role="main">
    <ModalManager>
      <WrappedDisclosureComponent identifier="root-component" disclosureType="modal" />
    </ModalManager>
  </div>
);

export default ModalManagerDefault;

```
</details>

Refer https://github.com/cerner/terra-framework/issues/901#issuecomment-536937334 for additional details.

Also, https://github.com/davidtheclark/focus-trap-react/pull/19 illustrates another scenario.

## @ Mentions
@bjankord

nramamurth:
### Summary
Closes #916.


packages/terra-abstract-modal/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
```
nramamurth: I ain't sure if this can technically be treated as a bug but comments are welcome whether this should prolly be under 'Changed'.

PR #923, opened 2019-10-09
[terra-date-input][terra-date-picker][terra-date-time-picker][terra-time-input] adding isincomplete prop and isinvalid css stylings

BenBoersma:
### Summary

Resolves https://github.com/cerner/terra-core/issues/2659

Added isIncomplete prop to date-input, date-picker, date-time-picker, and time-input
If any of the above has inComplete and required then it will change the background color to match incomplete stylings.
Added isInvalid css stylings
Consumes the changes in https://github.com/cerner/terra-core/pull/2670

@cerner/terra


packages/terra-date-input/src/DateInput.jsx
```diff
@@ -570,6 +582,8 @@ class DateInput extends React.Component {
         autoComplete="off"
         disabled={this.props.disabled}
         isInvalid={this.props.isInvalid}
+        isIncomplete={this.props.isIncomplete}
+        required={this.props.required}
       />
```
benbcai: The `isIncomplete` and `required` props should be extracted out of the customProps in the [render](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-input/src/DateInput.jsx#L619) function.

packages/terra-date-input/src/terra-dev-site/test/date-input/InvalidDateInputField.test.jsx
```diff
@@ -4,8 +4,8 @@ import React, { useState } from 'react';
 import DateInputField from 'terra-date-input/lib/DateInputField';
 
 const Example = () => {
-  const [value, setValue] = useState('$5%g-ad-!f');
-  const [isInvalid, setIsInvalid] = useState(false);
+  const [value, setValue] = useState('1999-03-10');
```
benbcai: This test is intended to test passing down invalid value. What's the reason for changing to a valid value?

BenBoersma: I renamed the test for this to [InvalidProvidedInputField](https://github.com/cerner/terra-framework/pull/923/files/25d76f83c6c7e1986883293c7b219aad195e5f7c#diff-952c2da42252c1edf25365b7ce031dd5R1) since its a test of what happens when the component is provided an invalid date. I am now using this test's name to signify a test of the `isInvalid` css styling.

Currently InvalidProvidedInputField test is disabled as the contrast ratio for the placeholder is causing the test to fail

packages/terra-date-input/src/DateInput.jsx
BenBoersma: Fixed here https://github.com/cerner/terra-framework/pull/923/commits/b43c2261fa209a9bbca6d8202315a4cd7754df28?file-filters%5B%5D=.jsx#diff-9620f8593dd0f7ccffbf6a626cc100f9R627

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -703,18 +718,29 @@ class TimeInput extends React.Component {
...
       hourAttributes,
       intl,
+      isIncomplete,
+      isInvalid,
       onBlur,
       onChange,
       onFocus,
       name,
       refCallback,
+      required,
```
benbcai: The `required` prop should be passed down to the `Input` component.

benbcai:
@neilpfeiffer @BenBoersma Should the meridiem ButtonGroup also get red borders when inValid since it is a subcomponent?

Screen Shot 2019-10-14 at 2 24 39 PM

neilpfeiffer:
@benbcai - No to adding the red borders on the meridiem button-group as a unit, since it technically cannot be invalid with one option always being chosen, and both options are always valid for entering general dates. 

However-- I think we need to add additional opt-in API/prop for if the currently selected AM/PM value is invalid, where if there is a date/time range, and the preselected date/time chosen falls outside of the range, the currently selected AM (or PM) would be the issue -- so we'll need to have the ability to highlight the currently selected button (in the group) only and highlight it's border/box-shadow as red.

benbcai:
The height on the month select in DateInput does not match the height on the day/year input. In talking with @BenBoersma , it appears that the existing input already has a border to use but new borders are needed to be added to the month select to get the red borders. Hence making the height 2 pixels (top and bottom borders) taller.

Screen Shot 2019-10-15 at 11 41 46 AM

packages/terra-date-input/src/DateInput.module.scss
```diff
@@ -64,29 +64,51 @@
     &.error {
       background-color: var(--terra-date-input-month-error-background-color, #fff);
       background-image: var(--terra-date-input-month-error-background-image);
+      border: var(--terra-date-input-month-error-border, 1px solid #e50000);
+      box-shadow: var(--terra-date-input-month-error-box-shadow, 0 0 0 1px #e50000);
```
neilpfeiffer: remove from here -- is added to the wrong location.

\
Go down to line 177-178 and update for `.month-select`:
```
&.error {
      border: var(--terra-date-input-month-select-error-border, 1px solid #e50000);
      box-shadow: var(--terra-date-input-month-select-error-box-shadow, 0 0 0 1px #e50000);
```

BenBoersma: Should be fixed here https://github.com/cerner/terra-framework/pull/923/commits/e1b103a65ed2a1fbc8cfcd4fb595a68c5307b269

benbcai:
> However-- I think we need to add additional opt-in API/prop for if the currently selected AM/PM value is invalid, where if there is a date/time range, and the preselected date/time chosen falls outside of the range, the currently selected AM (or PM) would be the issue -- so we'll need to have the ability to highlight the currently selected button (in the group) only and highlight it's border/box-shadow as red. The extra opt-in prop would only be used if the time input has a value.

Addressed on this commit: https://github.com/cerner/terra-framework/pull/923/commits/9f0ea7f2816d0be256fe9a033a170036b2a53419

packages/terra-date-input/tests/wdio/date-input-field-spec.js
```diff
@@ -1,30 +1,50 @@
 Terra.describeViewports('Date Input', ['medium'], () => {
-  describe('Date Input Field no date provide', () => {
+  // Temporary disabled test, since placeholder causes contrast accessibility failure.
```
nramamurth: Do we have a follow-up issue to fix this?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -31,6 +31,18 @@ const propTypes = {
...
+  */
+  isIncomplete: PropTypes.bool,
+  /**
+  * Whether the input displays as Invalid. Use when value does not meet validation pattern.
+  */
+  isInvalid: PropTypes.bool,
+  /**
+  * Whether the selected meridiem displays as Invalid. Use when value does not meet validation pattern.
+  */
+  isInvalidMeridiem: PropTypes.bool,
```
nramamurth: Do we have a test or an example to use this?

packages/terra-time-input/src/TimeInput.module.scss
```diff
@@ -92,6 +92,46 @@
...
+        border: var(--terra-time-input-invalid-disabled-border, 1px solid #e50000);
+        color: var(--terra-time-input-invalid-disabled-color, #1c1f21);
+        opacity: var(--terra-time-input-invalid-disabled-opacity, 0.25);
+      }
+    }
+
+    &.is-incomplete {
+      background-color: var(--terra-time-input-incomplete-background-color, #ffc);
+
+      &:hover {
```
neilpfeiffer: Need to also include higher specificity hover for desktop (so doesn't show as white on hover):
```suggestion
      &:hover,
      &.desktop:hover {
```

packages/terra-time-input/src/TimeInput.module.scss
```diff
@@ -92,6 +92,46 @@
...
+        background-color: var(--terra-time-input-incomplete-hover-background-color, #ffc);
+      }
+
+      &:focus {
+        background-color: var(--terra-time-input-incomplete-focus-background-color, #ffc);
+      }
+
```
neilpfeiffer: add solid color for desktop initial-focus:
```suggestion
      .desktop.initial-focus {
        background-color: var(--terra-time-input-desktop-incomplete-focused-input-background-color,  rgb(185, 215, 233));
       }
       
```

nramamurth:
@BenBoersma @benbcai 

Should there be changes made to the downstream theme repos? Is there a plan to make sure that the tests in the downstream theme repos do not fail because of the mismatch in these screenshots? Fusion theme tests should already be failing because https://github.com/cerner/terra-core/pull/2670 was merged and released.

I am trying to avoid a scenario wherein the conflicts with https://github.com/cerner/terra-framework/pull/837 grow further in theme repos.

benbcai:
@nramamurth Yes, there will be effort to update clinical-theme and fusion-theme for the changes made in [#2670](https://github.com/cerner/terra-core/pull/2670) and this PR.

packages/terra-date-input/tests/wdio/date-input-field-spec.js
neilpfeiffer: @nramamurth [Placeholders generally are problematic](https://developer.paciellogroup.com/blog/2011/02/html5-accessibility-chops-the-placeholder-attribute/) when trying to pass color contrast, as needing to pass the minimum 4.5:1 ratio if provided to properly meet the WCAG requirements, but doing so and setting a color that dark results in sighted users (normal acuity and low-vision alike) mistaking the placeholder text as an actual value. So for now - Terra and UX have been generally following older internal (not WCAG) color rules we used in the past (3:1 for placeholders). 

Going forward, UX is figuring out how to transition away from using placeholders, opting instead for better labels and help text, and instead leave empty form elements actually empty. There are quite a few usability articles that suggest this is a better way to go: [Nielsen Norman Group](https://www.nngroup.com/articles/form-design-placeholders/) and [UXmatters](https://www.uxmatters.com/mt/archives/2010/03/dont-put-hints-inside-text-boxes-in-web-forms.php), UX just needs to decide when to make the change and how to do so as the new Cerner standard.

neilpfeiffer:
Tested: PR #923
Resolves: Issue https://github.com/cerner/terra-core/issues/2659


**Reviewed:**
- Follows pattern considerations documented in the current Cerner Design Standards: [Forms](https://wiki.cerner.com/display/UserExperience/Forms)
- Follows UX defined design implementation for date-picker, date-input, date-time-picker, time-input
- Follows defined theme designs: Open Source
\
\
Verified by @neilpfeiffer, Marking as Verified, Ready for Release.

**Pass** |
-----------|
UX Review|

Related PR in Terra-Core: https://github.com/cerner/terra-core/pull/2670

packages/terra-date-input/tests/wdio/date-input-field-spec.js
neilpfeiffer: Added this issue to track the future work: https://github.com/cerner/terra-core/issues/2712

Issue #929, opened 2019-10-14
[date-picker] Buttons theme incorrectly in the date-picker


hboege:
# Bug Report

## Description
In PR #866, changes were made to date-picker to use terra-buttons in the popup content, instead of its own buttons.  This means the buttons will change theme, but the reset of the popup content (calendar) will not; this results in the user has mixed themes/styling in the popup.

## Steps to Reproduce
1.  Import/create a theme, where terra-buttons are different colors.
2. Create a page using date-picker.
3. Run this page in the new theme.  The popup content will appear in the "default" theme, except for the buttons.

## Expected Behavior
Controls need to maintain a consistent theme.

## Possible Solution
Revert the changes back to how it was.  Or implement new buttons with in the date-picker that can be styled differently than terra-buttons.

## Environment
* Component Name and Version: terra-date-picker 4.14.0
* Browser Name and Version: 
* Node/npm Version: [e.g. Node 8/npm 5]
* Webpack Version:
* Operating System and version (desktop or mobile):

## @ Mentions
@bjankord @ryanthemanuel  @neilpfeiffer

nramamurth:
@hboege 

Just to confirm, are you talking about the 'Previous' and 'Next' navigation buttons here? 

![chrome-focus-issue](https://user-images.githubusercontent.com/10090859/66823279-88f3f180-ef63-11e9-9d5d-7a091ba972c4.gif)

PR #932, opened 2019-10-16
Update date picker buttons

bjankord:
### Summary
This update changes the buttons for previous and next month back to rendering an HTML button element instead of using the terra-button react component for rendering these buttons. There were some issues with the hover state in the low-light theme where we found this styling is more favorable for the buttons at this time compared to using the terra-button component.

Resolves https://github.com/cerner/terra-framework/issues/929


benbcai:
- Some of the screenshots need to be updated.
- This [line](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/src/react-datepicker/calendar.jsx#L282-L283) is throwing an error when clicking on the Previous/Next month button. Also I think one of them on that line should be `nextMonthBtnRef` instead of both being `previousMonthBtnRef`.

`Uncaught TypeError: a.previousMonthBtnRef.contains is not a function`

bjankord:
@benbcai I've updated the code to address the comments in the following commits:
* https://github.com/cerner/terra-framework/pull/932/commits/a48f08ddf31213bf668a0be2e7d2e11b59b8638a
* https://github.com/cerner/terra-framework/pull/932/commits/d58d77a8fe1a849e6d6bdb3fb4062290ae7fc9fa
* https://github.com/cerner/terra-framework/pull/932/commits/b5edae3124ab8645813f7f361b752041d668e158
* https://github.com/cerner/terra-framework/pull/932/commits/908ce1fde4d332586d8e52feb476d376638a7f23
* https://github.com/cerner/terra-framework/pull/932/commits/2d3cff84ae755c9b9789285f1a8f84df905cfc6f

benbcai:
In the [example](https://terra-framework-deploye-pr-932.herokuapp.com/components/terra-date-picker/date-picker/date-picker) with the min/max date set, the previous and next month buttons are not rendered. There is still an error if you just select a date that is within the range in the calendar.

`Uncaught TypeError: Cannot read property 'contains' of null`

I don't remember whether or not the next and previous month button were shown or disabled when there are no valid dates past the currently shown month. I need to go back and check.

bjankord:
@benbcai Resolved the `Uncaught TypeError: Cannot read property 'contains' of null` error in this commit: https://github.com/cerner/terra-framework/pull/932/commits/8afff4dcdcc9b0a98cdd038055eacc2108377d20

PR #934, opened 2019-10-18
[terra-aggregator]Focus not released from item after item removed from view

### LINKED ISSUE #931: [terra-aggregator] Ensure the focus is released forcefully, when the corresponding item is removed from view
# Bug Report

## Description
When a component with auto/manual refresh capability uses a terra-aggregator, the disclosed component continues to remain in view even after the item that invoked it is removed from the parent component. And when a second refresh happens on the parent component (with or without the item in focus) the disclosed component is removed as an unintended effect

## Steps to Reproduce
1.  Load a component with Aggregator having a bunch of items that have disclosure capabilities
2. Click on an element within one of those items say (item A) and allow the disclosure component to open
3. On a timeout or probably a button action update the aggregator with a new set of items (without item A)
4. The expectation is to see the disclosed component close out. However it would continue to persist
5. Now if the component reloads with initial data set. The disclosed component will close on you.

## Expected Behavior
Focus should release as expected when item is dynamically removed based on user activity or a refresh interval from aggregator

## Possible Solution
The reasoning for this behavior is because of this line [Aggregator.jsx#L54](https://github.com/cerner/terra-framework/blob/master/packages/terra-aggregator/src/Aggregator.jsx#L54). When checking if focused item is present in the view or not, the check should be made on the incoming props.However, it is being made on prevProps resulting in this behavior

## @ Mentions
@tbiethman

lokesh-0813:
### Summary
Fixes #931


packages/terra-aggregator/src/terra-dev-site/test/aggregator/AggregatorWithDisclosure.test.jsx
```diff
@@ -23,19 +24,48 @@ const items = Object.freeze([{
...
-    <SlidePanelManager>
-      <Wrapper />
-    </SlidePanelManager>
-  </div>
-);
+class AggregatorWithDisclosure extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
```
tbiethman: Want to try and implement the example using hooks? No pressure, but it'd shrink this example back down a bit.

lokesh-0813: Implemente here c0a69e9625a61825b7aec1215decefc2bfff1a8d

VivekChittamuri:
@lokesh-0813 @ShettyAkarsh I just wanted to reach out and see if we have a tentative release date for this fix ? One of our Cerner internal jira's is blocked by #931. 

cc: @mjhenkes @tbiethman

PR #946, opened 2019-10-22
Axe Violations fixed by adding missing/wrong roles.

### LINKED ISSUE #945: Missing/Wrong Aria-Roles giving Axe Violation.
## Description
Certain components have no roles or wrong role-attribute combination assigned which are giving violations on the axe plugin.

## Error Report
` Certain ARIA roles must contain particular children`

`Elements must only use allowed ARIA attributes`

## Components

- Aggregator
- Collapsible Menu View
- Tabs

saket2403:
### Summary
Fixes #945


emilyrohrbough:
What version of the axe plugin are you using? Is it the same version we are testing with in wdio? I am curious why these errors were not see in the tests? Should we add tests for these states to ensure the correct aria labels are used?

packages/terra-tabs/src/Tabs.jsx
```diff
@@ -161,7 +161,6 @@ class Tabs extends React.Component {
 
       clonedPanes.push(React.cloneElement(child, {
         className: cx([{ 'is-active': isActive }, child.props.className]),
-        'aria-selected': isActive,
```
emilyrohrbough: Does removing this aria role affect how the screen readers read this value?

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItemGroup.jsx
```diff
@@ -80,7 +80,7 @@ class CollapsibleMenuViewItemGroup extends React.Component {
     } if (isCollapsibleMenuItem) {
       return (
         <li role="none">
-          <List {...customProps} role="listbox">
+          <List {...customProps} role="menu">
```
emilyrohrbough: I think listbox is the role we want here. According to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role: 
> The listbox role is used to identify an element that creates a list from which a user may select one or more static item

This example even uses the role `listbox` for a collapsible menu:
https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html

saket2403:
> What version of the axe plugin are you using? Is it the same version we are testing with in wdio? I am curious why these errors were not see in the tests? Should we add tests for these states to ensure the correct aria labels are used?

I used axe core version 3.3.2 and got these violations. But I am seeing we are using v3.2.2 in the toolkit.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItemGroup.jsx
jmsv6d: Actually, I think the `li` and `List` can just be removed entirely for this scenario and only return children. This code block is to build out an un-selectable CollapsibleMenuViewItemGroup there's functionally no need to have the items wrapped by anything when displayed in the Menu.

saket2403: Removed in 8dd92b0

packages/terra-tabs/src/Tabs.jsx
saket2403: Added  aria-selected only for tab panes in 8dd92b0

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItemGroup.jsx
```diff
@@ -78,11 +77,9 @@ class CollapsibleMenuViewItemGroup extends React.Component {
       );
     } if (isCollapsibleMenuItem) {
       return (
-        <li role="none">
-          <List {...customProps} role="listbox">
```
benbcai: Would this change break anything since it is no longer spreading the custom props on the list?

saket2403: As per this [comment](https://github.com/cerner/terra-framework/pull/946#discussion_r339127165) the `List` and `li`  are not providing any functionalities, but there could be a case where user might not be able to apply custom css classnames to the children.
cc @jmsv6d @benbcai

mjhenkes: I think we should be spreading the custom props on the div now

packages/terra-tabs/src/TabPane.jsx
```diff
@@ -35,11 +35,16 @@ const propTypes = {
    * Indicates if the pane label should only display the icon. When tab collapses into menu the label text will be used.
    */
   isIconOnly: PropTypes.bool,
+  /**
+   * Indicates if the pane has active styles or not.
```
mjhenkes: This doesn't explain why a consumer would use the isActive prop. Maybe something like

```suggestion
   * If enabled, this prop will apply the `aria-selected` style to the pane.
```

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItemGroup.jsx
saket2403: Custom props added cb40173

benbcai:
After discussing the implications of the changes on this PR for addressing [this rule](https://dequeuniversity.com/rules/axe/3.4/scrollable-region-focusable) introduced in axe-core-3.3.0, we decided to hold off on these changes and disable this new rule for now in [terra-toolikit](https://github.com/cerner/terra-toolkit/issues/343). We will revisit this rule once we have a better plan on how to address it. I will close this PR after I disable the rule to ensure the violations no longer appear.

saket2403:
This PR is for a different axe issue in `v3.4` regarding aria-roles,  not the keyboard accessibility one. If this one looks good to be merged, it can be merged, or [this rule](https://dequeuniversity.com/rules/axe/3.4/aria-required-children?application=AxeChrome) can be disabled in toolkit @benbcai .

benbcai:
@saket2403 Yes, this is an existing rule and we will not disable it and should be fixed before bumping to axe-core v3.4.

I found one other place that has this violation. This [test](https://github.com/cerner/terra-framework/blob/master/packages/terra-application-navigation/tests/wdio/application-navigation-spec.js#L14-L25) also shows that the popup menu in terra-application-navigation uses an incorrect role. This is the violation. 

```
[large] ApplicationNavigation - Large Tabs roll up properly [default] is accessible and is within the mismatch tolerance:
expected no accessibility violations but got:
	{
  "description": "Ensures elements with an ARIA role that require child roles contain them",
  "help": "Certain ARIA roles must contain particular children",
  "helpUrl": "https://dequeuniversity.com/rules/axe/3.4/aria-required-children?application=axeAPI",
  "id": "aria-required-children",
  "impact": "critical",
  "nodes": [
    {
      "all": [],
      "any": [
        {
          "data": [
            "listitem"
          ],
          "id": "aria-required-children",
          "impact": "critical",
          "message": "Required ARIA child role not present: listitem",
          "relatedNodes": []
        }
      ],
      "failureSummary": "Fix any of the following:\n  Required ARIA child role not present: listitem",
      "html": "<ul class=\"PopupMenu-module__utility-list___2rhJk\" aria-label=\"More Navigation\" role=\"list\" tabindex=\"0\">",
      "impact": "critical",
      "none": [],
      "target": [
        "ul"
      ]
    }
  ],
  "tags": [
    "cat.aria",
    "wcag2a",
    "wcag131"
  ]
}
```

It shows that the [popmenu](https://github.com/cerner/terra-framework/blob/7981071c6f84a0cfdfbd21b81b36cd00ee0756c0/packages/terra-application-navigation/src/tabs/_Tabs.jsx#L281) has a role of `list` and expects its children to have a role of `listitem` but its children's [role](https://github.com/cerner/terra-framework/blob/master/packages/terra-application-navigation/src/common/_PopupMenuListItem.jsx#L89) is `link`.
Screen Shot 2019-12-12 at 9 20 16 AM

In talking with @dkasper-was-taken he thinks the `link` role should be set in a `div` contained inside an `<li>` similar to the navigation side menu. 
Screen Shot 2019-12-12 at 9 22 34 AM

saket2403:
> I found one other place that has this violation. This [test](https://github.com/cerner/terra-framework/blob/master/packages/terra-application-navigation/tests/wdio/application-navigation-spec.js#L14-L25) also shows that the popup menu in terra-application-navigation uses an incorrect role. 

Fixed in 95f43ac

benbcai:
@saket2403 The role for the children should remain `link` instead of changing to `listitem`. The content should be wrapped in a `div` inside an `li` similar to how the navigation side menu in the above screenshot.

saket2403:
> @saket2403 The role for the children should remain `link` instead of changing to `listitem`. The content should be wrapped in a `div` inside an `li` similar to how the navigation side menu in the above screenshot.

changed in 662d37c

PR #962, opened 2019-11-05
[terra-date-picker] Fixing test logic for incomplete and invalid tests

BenBoersma:
### Summary
Test logic for the incomplete and invalid wdio tests was causing them to fail in theming repos

@cerner/terra

### DIFF
--- a/packages/terra-date-picker/CHANGELOG.md
+++ b/packages/terra-date-picker/CHANGELOG.md
@@ -3,10 +3,12 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+* Fixed Invalid and Incomplete wdio test logic to ensure states are off after tests runs.
+
 ### Changed
 * Resolved lint warnings for multiple empty lines
 * Regenerate wdio screenshots
-* Fixed some wdio test logic
 
 4.18.0 - (October 30, 2019)
 ------------------
--- a/packages/terra-date-picker/CHANGELOG.md
+++ b/packages/terra-date-picker/CHANGELOG.md
@@ -6,6 +6,7 @@ Unreleased
 ### Changed
 * Resolved lint warnings for multiple empty lines
 * Regenerate wdio screenshots
+* Fixed some wdio test logic
 
 4.18.0 - (October 30, 2019)
 ------------------
--- a/packages/terra-date-picker/tests/wdio/date-picker-spec.js
+++ b/packages/terra-date-picker/tests/wdio/date-picker-spec.js
@@ -326,7 +326,7 @@ Terra.describeViewports('Date Picker', ['medium'], () => {
 
     describe('Invalid DatePickerField', () => {
       before(() => {
-        browser.url('/#/raw/tests/terra-date-picker/date-picker/date-picker-field');
+        browser.refresh();
         browser.click('#validity-toggle');
       });
       after(() => {
@@ -338,7 +338,7 @@ Terra.describeViewports('Date Picker', ['medium'], () => {
 
     describe('Incomplete DatePickerField', () => {
       before(() => {
-        browser.url('/#/raw/tests/terra-date-picker/date-picker/date-picker-field');
+        browser.refresh();
         browser.click('#incomplete-toggle');
       });
       after(() => {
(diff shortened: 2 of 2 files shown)


packages/terra-date-picker/tests/wdio/date-picker-spec.js
```diff
@@ -329,6 +329,9 @@ Terra.describeViewports('Date Picker', ['medium'], () => {
         browser.url('/#/raw/tests/terra-date-picker/date-picker/date-picker-field');
```
yuderekyu: This is personal preference, but I notice date picker specs use `browser.refresh` to reset state. Because these tests manipulate the same test component, these repeated `browser.url` calls are unnecessary - we can update the repeated `browser.url` calls to `browser.refresh()`, which eliminates the after blocks. This removes extra code and stays consistent with the other tests.
```suggestion
         browser.refresh()
```

BenBoersma: Fixed here https://github.com/cerner/terra-framework/pull/962/commits/018cda40ca19f0ca53ec7faff8475c064edc0930

yuderekyu: sorry if it wasn't clear, but if we use refresh we can remove the after blocks

BenBoersma: Further Fixed here https://github.com/cerner/terra-framework/pull/962/commits/0e36b9e40797140ba28864a7d6c87d38f776f3be

packages/terra-date-picker/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Unreleased
 ### Changed
 * Resolved lint warnings for multiple empty lines
 * Regenerate wdio screenshots
+* Fixed some wdio test logic
```
emilyrohrbough: This should be under `### Fixed`. Also can you be a bit more descriptive?

packages/terra-date-picker/tests/wdio/date-picker-spec.js
```diff
@@ -326,7 +326,7 @@ Terra.describeViewports('Date Picker', ['medium'], () => {
 
     describe('Invalid DatePickerField', () => {
       before(() => {
-        browser.url('/#/raw/tests/terra-date-picker/date-picker/date-picker-field');
+        browser.refresh();
         browser.click('#validity-toggle');
       });
```
emilyrohrbough: We should avoid using refresh when possible. What are these changes fixing with the theme repo?

emilyrohrbough: Looking through the commit history, if we can re-click the `#validity-toggle` button to adjust the state, it would be cheaper from a testing perspective. Using `refresh` kills the current session, sends a new session request, (when using the selenium grid, this can cause tests to be pending till a node becomes available) and then refreshes the page to reset the state.

BenBoersma: In the current way the fusion theme repo run the tests, the pages are keeping the toggle states between screenshots so the later tests (incomplete and incomplete invalid) are displaying wrong.

I've only tested the fusion theme repo with this so far.

packages/terra-date-picker/CHANGELOG.md
BenBoersma: Fixed Here https://github.com/cerner/terra-framework/pull/962/commits/78ec9821e8de85d34678c4514e45c0b21db37209

packages/terra-date-picker/tests/wdio/date-picker-spec.js
BenBoersma: Fixed here https://github.com/cerner/terra-framework/pull/962/commits/78ec9821e8de85d34678c4514e45c0b21db37209

packages/terra-date-picker/tests/wdio/date-picker-spec.js
BenBoersma: Unfurther Fixed here https://github.com/cerner/terra-framework/pull/962/commits/78ec9821e8de85d34678c4514e45c0b21db37209

as per discussion here https://github.com/cerner/terra-framework/pull/962#discussion_r342785207

PR #964, opened 2019-11-11
[terra-tabs] Handle visibility of tabs added/removed post initial mount.

### LINKED ISSUE #486: [terra-tabs] Allow additional tabs to be added after initial mount of Tabs component
# Issue Description

The Tabs component has some internal state that manages the visible tab panes. If additional panes are added after the initial mounting, that state is not updated and the added panes are not visible. 

We should make changes to ensure that componentWillReceiveProps is implemented and keeping the internal component state in sync with the props.

## Issue Type

- [X] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

nramamurth:
### Summary
Closes #486.


packages/terra-tabs/tests/wdio/tabs-spec.js
```diff
@@ -82,4 +83,17 @@ Terra.describeViewports('Tabs - Large screen', ['large'], () => {
...
+
+  describe('Additional Tabs after mount', () => {
+    before(() => {
+      browser.url('/#/raw/tests/terra-tabs/tabs/tabs/additional-tabs');
+    });
+
+    Terra.it.matchesScreenshot('before');
+
+    it('Additional Tabs', () => {
+      $('#tabsWrapper-20').waitForExist(4000);
```
jmsv6d: This could be updated to 
```suggestion
      browser.waitForExist('#tabsWrapper-20', 4000);
```
Removed the need for the added global and matches our other tests.

packages/terra-tabs/src/terra-dev-site/test/tabs/Tabs/AdditionalTabs.test.jsx
```diff
@@ -0,0 +1,42 @@
...
+      tabKeys: [...Array(5).keys()],
+    };
+  }
+
+  componentDidMount() {
+    setTimeout(() => {
+      this.setState({
+        tabKeys: [...Array(20).keys()],
+      });
+    }, 3000);
```
jmsv6d: Could we test this without adding a timeout? Maybe utilize a button to add tabs? Or at least try to minimize the timeout to minimize the time added to the test run?

nramamurth: Done.

packages/terra-tabs/tests/wdio/tabs-spec.js
nramamurth: Done.

packages/terra-tabs/src/_CollapsibleTabs.jsx
```diff
@@ -73,10 +73,13 @@ class CollapsibleTabs extends React.Component {
     this.handleSelectionAnimation();
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps) {
     if (this.isCalculating) {
       this.isCalculating = false;
       this.handleResize(this.contentWidth);
+    } else if (React.Children.count(this.props.children) !== React.Children.count(prevProps.children)) {
```
StephenEsser: One thing to note is that I don't think this will resolve issues with tab reordering or tab swaps (removing 1 tab and adding 1 tab at the same time) because the count will remain the same, but until a user requests this functionality I would not make code changes for it.

PaulSolDev:
Functional review complete.  No issues identified.

PR #943, opened 2019-10-21
[terra-collapsible-menu-view] Add ability for 'More Button' to switch variants.

nramamurth:
### Summary
Closes #942.

### Additional Details
Related: https://github.com/cerner/terra-core/pull/2714
Context: https://github.com/cerner/terra-core/issues/1835#issuecomment-543815207

PR #965, opened 2019-11-12
Revert "[terra-collapsible-menu-view] Add ability for 'More Button' to switch variants."

neilpfeiffer:
Reverts cerner/terra-framework#943

Change, Prop is no longer needed will just be updating base styling and icon.


packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
```diff
@@ -3,7 +3,6 @@ import ResizeObserver from 'resize-observer-polyfill';
 import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
 import { FormattedMessage } from 'react-intl';
-import Button from 'terra-button';
```
mjhenkes: Should button be removed from CollapsibleMenuView's package.json? or is it used elsewhere in the package?

neilpfeiffer: No, assuming it shouldn’t be removed. (I just pressed the Github Revert button on merged PR.) 
Looking at previous version and will modify, or just may cancel this PR and only change what is needed rather than full revert, since just needs to remove the extra prop that was added.

nramamurth: It’s used [here](https://github.com/cerner/terra-framework/blob/master/packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItem.jsx#L140).

neilpfeiffer: Replacing this PR with #966 instead

PR #967, opened 2019-11-13
[terra-time-input] Remove references to 'desktop' class in CSS.

nramamurth:
### Summary
All references to 'desktop' and 'mobile' were **supposed** to be removed when https://github.com/cerner/terra-framework/pull/837 was merged on top of https://github.com/cerner/terra-framework/pull/923 but a couple of them sneaked in 😔 

This PR removes those references and adds a test to avoid regression in future.


packages/terra-time-input/src/TimeInput.module.scss
```diff
@@ -114,17 +114,16 @@
...
       }
 
       &:focus {
         background-color: var(--terra-time-input-incomplete-focus-background-color, #ffc);
       }
 
-      .desktop.initial-focus {
-        background-color: var(--terra-time-input-desktop-incomplete-focused-input-background-color, rgb(185, 215, 233));
+      .initial-focus {
+        background-color: var(--terra-time-input-incomplete-focused-input-background-color, rgb(185, 215, 233));
```
mjhenkes: Don't forget to update the various themes this variable change impacts.

nramamurth: @avinashg1994 This needs to be accounted for as part of https://github.com/cerner/terra-framework/pull/978 and https://github.com/cerner/terra-framework/pull/975.

PR #968, opened 2019-11-13
[terra-tabs] Display tabs inline on smaller viewports. 

### LINKED ISSUE #484: Allow Tabs to display inline when placed in smaller containers
# Bug Report

## Description
terra-tabs gives drop-down list when the component is placed in small containers.

>
tabsbug

## Expected Behavior
 Tabs appear inline instead of drop down list
## Possible Solution
The Tabs uses Responsive Element component which has responsiveTo props which default value is set to 'parent' changing it window would solve the issue. 
## Environment

* Component Name and Version: terra-tabs v3.16.0
* Browser Name and Version: Internet Explorer v11.1206.15063.0
* Node/npm Version: npm v5.7
* Webpack Version: v3.6.0
* Operating System and version (desktop or mobile): windows 10 Enterprise v1703

### LINKED ISSUE #485: [terra-tabs] - Menu dropdown is not displayed near dropdown text
# Issue Description

On small viewports the Tab component collapses each tab into a drop down. The options are left aligned but the menu is displayed to the right. 

image

image

This may be related to: https://github.com/cerner/terra-core/issues/1056

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Cosmetic
- [ ] Other

## Expected Behavior

The Menu dropdown should display near the dropdown text.

## Current Behavior

The Menu and drop down text are not displayed near each other.

## Steps to Reproduce

1. Create a Tabs component
2. Shrink the screen down to a small view port
3. Click the Tab text to open the drop down
4. Observe the drop down is not displayed near the dropdown text

## Environment

* Component Version:
* Browser Name and Version:

nramamurth:
### Summary
* Added `responsiveTo` prop to provide ability to consumers to decide whether or not tabs need to completely collapse into a menu on smaller viewports or when tabs are placed within a very small container.
* Center-align text label in the single dropdown tab when all tabs completely collapse into a menu.

**Note**: "More" button looks off in the screenshot [here](https://github.com/cerner/terra-framework/pull/968/files?short_path=50d1105#diff-50d11059da76d878d7ed9c04aff27810) and I have logged a separate issue: https://github.com/cerner/terra-framework/issues/970. 

Closes #485, Closes #484


packages/terra-tabs/src/Tabs.jsx
```diff
@@ -169,35 +178,44 @@ class Tabs extends React.Component {
...
+          <CollapsedTabs activeKey={activeKey || this.state.activeKey} onTruncationChange={this.handleTruncationChange}>
+            {clonedPanes}
+          </CollapsedTabs>
+        );
+
+        return (
+          <ResponsiveElement
+            tiny={collapsedTabs}
+            small={collapsibleTabs}
+            responsiveTo={responsiveTo}
```
emilyrohrbough: Will these changes be valid after https://github.com/cerner/terra-framework/pull/972 is merged?

nramamurth: I don't think so; the core logic should probably remain as-is. I think what gets rendered now should be based off of both `breakpoint` value and `responsiveTo` value - something like —

```
if ((current breakpoint is "tiny") AND (responsiveTo is either "parent" or "window")) {
    <---- show collapsed tabs ---->
 } else {
    <---- show collapsible tabs ---->
}
```

packages/terra-tabs/src/Tabs.jsx
```diff
@@ -169,35 +178,44 @@ class Tabs extends React.Component {
...
-        activeKey={activeKey || this.state.activeKey}
-        activeIndex={this.getActiveTabIndex()}
-        onChange={this.handleOnChange}
-        onTruncationChange={this.handleTruncationChange}
-        variant={variant}
-      >
-        {clonedPanes}
-      </CollapsibleTabs>
-    );
+    const header = (() => {
```
emilyrohrbough: Can we rename this func to something like getTabs() to be representative of what it is returning?

emilyrohrbough: or could we convert this to be something like:
```js
...
let tabContent = collapsibleTabs;

 if ([DependentViewport.PARENT, DependentViewport.WINDOW].includes(responsiveTo)) {
    ...
     tabContent (
          <ResponsiveElement
            tiny={collapsedTabs}
            small={collapsibleTabs}
            responsiveTo={responsiveTo}
      />
    )
 }

....

return (
      <ContentContainer
        {...customProps}
        className={tabsClassNames}
        fill={fill}
        header={tabContent}
      >
...
```

nramamurth: I have renamed `header` ➡️ `tabContent` - https://github.com/cerner/terra-framework/pull/968/commits/e293226fe48790135e5279fb37bfa4a6a0f4fb35.

nramamurth:
@emilyrohrbough 

> We should had a window-responsive test example and generate wdio screenshots

Now that you mention it, I realized that my default parent-responsive spec wasn't really showing the difference between parent resize v/s window resize, so I have updated it and have also added a window-responsive spec: https://github.com/cerner/terra-framework/pull/968/commits/e293226fe48790135e5279fb37bfa4a6a0f4fb35

> Should we add add an example of these variants to the face-up doc examples?

That will surely help the consumers, me thinks. I shall add it.

packages/terra-tabs/src/Tabs.jsx
```diff
@@ -58,11 +58,19 @@ const propTypes = {
...
+   * Note: If the responsive viewport is set to `none`, then tabs never _completely_
+   * collapse into a menu.
+   */
+  responsiveTo: PropTypes.oneOf(['window', 'parent', 'none']),
 };
 
 const defaultProps = {
   tabFill: false,
   fill: false,
+  responsiveTo: DependentViewport.PARENT,
```
mjhenkes:
```suggestion
  responsiveTo: 'parent',
```
This will look better in the generated props table.

packages/terra-tabs/src/Tabs.jsx
```diff
@@ -169,35 +178,44 @@ class Tabs extends React.Component {
...
+
+      if ([DependentViewport.PARENT, DependentViewport.WINDOW].includes(responsiveTo)) {
+        const collapsedTabs = (
+          <CollapsedTabs activeKey={activeKey || this.state.activeKey} onTruncationChange={this.handleTruncationChange}>
+            {clonedPanes}
+          </CollapsedTabs>
+        );
+
+        return (
+          <ResponsiveElement
```
mjhenkes: If tabs are responsive to none, we should not wrap the tabs in a responsive element and just return `CollapsibleTabs`

mjhenkes: Ah, i see you are doing this already, Disregard.

packages/terra-tabs/src/Tabs.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
 import ContentContainer from 'terra-content-container';
-import ResponsiveElement from 'terra-responsive-element';
+import ResponsiveElement, { DependentViewport } from 'terra-responsive-element';
```
StephenEsser: I think the `DependentViewport` constant makes the code more challenging to understand and doesn't follow a consistent export strategy. There are only two options `window` and `parent`. I'd recommend just using the string literals. `window`, `parent`. 

I don't personally think keying objects in this context lowers the chances of typos/errors or adds enough reusability to adjust for the added complexity.

nramamurth: > I don't personally think keying objects in this context lowers the chances of typos/errors or adds enough reusability to adjust for the added complexity.

It's been a grey area for me as to when to use it and when not to but this makes sense. I have [removed](https://github.com/cerner/terra-framework/pull/968/commits/4264fffd83d8119e411ddf2b7d1f7d303bf63c76) it.

nramamurth:
@emilyrohrbough 

Examples have been added to the site - https://github.com/cerner/terra-framework/pull/968/commits/d4333c8e2986f4fd631c694c0709461a4bbe5928

PR #977, opened 2019-11-26
[terra-menu] New headerTitle prop

### LINKED ISSUE #474: [terra-menu] Add new features (header, optgroup titles, icons)
# Feature Request

## Description
From #1849:
- [ ] ability to add a grouping/section sub-header titles, similar to [OptionGroups](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/_OptGroup.jsx)

#

From #1864:
When bounded, Menu renders header text within the back button. So, the click target is the back button + header text. It is now expected for the click target to only encompass the back button.

#### Potential Solution
- [ ] Render back button with `fitStart`
- [ ] Render header text with `fill`

#

Additional new features & adjustments:
- [ ] ability to add an icon before a menu item (other than checkmark which is existing) 
- [x] ~~ability to add "Electronic Instructions For Use" icon before a menu item first class (Moved to #818)~~
- [x] ~~add a mandatory top header with a title for any menu that has a drill-in nested menu on the first-tier of the menu (currently only adds the header on the second-tier and higher). Fixed on #977~~
- [ ] spike on markup structure to factor and best match construction with theme designs (separate, non-overlapping background areas for the header, menu-body)
- [ ] either refactor top header(and submenu nested menu headers) so the height will flex based on title content and correctly wrap, or import action-header which already has intended layout.
- [ ] menu content gets focus glow in Chrome when focus trapped (noticeable when zoomed in) so may need to add extra focus styling (or remove outline) for the outer menu container when automatically put in focus as menu opens.
- [ ] need to refactor the disabled menu-item so the themable opacity can apply to the entire list-item (icons/text/and chevrons), and then themeable individual colors for the icons/text/chevrons.
- [ ] Ability to define the menu attachment behavior (center, and left-to-right, in addition to the current right-to-left) #473

#

lokesh-0813:
### Summary
New headerTitle prop added.

### Additional Details
Menu now takes an additional prop `headerTitle`. The title is set to first tier menu page. This new header title is only introduced to menus which have at least one sub-menu.

fixes #474 
PS: partially fixes the mentioned issue.


lokesh-0813:
The default header can be seen in [this](https://terra-framework-deploye-pr-977.herokuapp.com/tests/terra-menu/menu/menu/sub-menu) test example.

jeremyfuksa:
Tests are still failing in `terra-collapsible-menu-view` and `terra-notification-dialog`. 

Visually, this looks OK.  I'm not removing the block or giving +1 yet. Accessibility input from @neilpfeiffer and @benbcai need to weigh in before final approval.

nramamurth:
Please generate/update fusion theme screenshots as well.

lokesh-0813:
> Please generate/update fusion theme screenshots as well.

Sure. But this is still 'Work In Progress'. I am waiting to know if UX requires any style modifications or provide some style changes to the header. Will update the screenshots once UX approves the new header styles.

packages/terra-menu/src/Menu.jsx
```diff
@@ -53,12 +53,17 @@ const propTypes = {
    * Otherwise, the menu will display without an arrow and right aligned.
    */
   isArrowDisplayed: PropTypes.bool,
+  /**
+   * Header Title for menu if menu contains drill-in sub-menus
```
jmsv6d: Could we add to this to say that if this prop is provided, but the menu doesn't contain at least one submenu, the header won't be applied.

lokesh-0813: Updated here 11c1defc99dc75b240409155cea00a864efe808a

packages/terra-menu/src/Menu.jsx
```diff
@@ -53,12 +53,18 @@ const propTypes = {
    * Otherwise, the menu will display without an arrow and right aligned.
    */
   isArrowDisplayed: PropTypes.bool,
+  /**
+   * Header Title for menu.
+   * Header Title will only be visible if the menu contains at least one sub-menu.
+   */
+  headerTitle: PropTypes.string,
```
StephenEsser: Does every submenu display the same title? Should each submenu be able to supply a different title or is a single shared titled the requested behavior?

packages/terra-menu/src/Menu.module.scss
```diff
@@ -46,6 +46,10 @@
       word-wrap: normal;
     }
 
+    .main-header-title {
+      margin-left: var(--terra-menu-mainmenu-text-margin-left, 10px);
```
StephenEsser: We'll want a consistent approach to naming these custom properties. Following other conventions in this file I would expect this to be something similar to:

```
--terra-menu-header-title-margin-left
```

There is another header-title property in this file, we'll want to make sure these don't collide.

StephenEsser: We'll want to get UX feedback on the units being used here. Generally margins are represented with `rem` units, but there seems to be some inconsistency in this file.

StephenEsser:
A few of the screenshots without the new title have changed, do we know what is causing that?

packages/terra-menu/src/Menu.module.scss
lokesh-0813: > We'll want a consistent approach to naming these custom properties. Following other conventions in this file I would expect this to be something similar to:
> 
> ```
> --terra-menu-header-title-margin-left
> ```
> 
> There is another header-title property in this file, we'll want to make sure these don't collide.

Updater here d0fb28990d97c747f039a532a4922640215c05a7

lokesh-0813: > We'll want to get UX feedback on the units being used here. Generally margins are represented with `rem` units, but there seems to be some inconsistency in this file.

@neilpfeiffer @jeremyfuksa

packages/terra-menu/src/Menu.jsx
lokesh-0813: The new `headerTitle` is specific to main-menu(first-tier). `sub-menus` header presentation title remains the same as it was, i,e same as its drilled in parents menu-item text.

Here is a minor change to the prop description 925d3978aa5c7b49ac4a66c4fd06ce07761016b9

lokesh-0813:
> A few of the screenshots without the new title have changed, do we know what is causing that?


I missed a condition check and this was generating a empty header `<DIV>`
Updated here 925d3978aa5c7b49ac4a66c4fd06ce07761016b9

lokesh-0813:
> Tests are still failing in `terra-collapsible-menu-view` and `terra-notification-dialog`.


Fix for `terra-collapsible-menu-view` and `terra-notification-dialog` tests 925d3978aa5c7b49ac4a66c4fd06ce07761016b9

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -269,18 +273,27 @@ class MenuContent extends React.Component {
...
+        fill={<h1 className={cx(['header-title', 'main-header-title'])}>{this.props.headerTitle}</h1>}
+      />
+    ) : undefined;
+
     return (
       <Arrange
         className={cx('header')}
         fitEnd={closeButton}
-        fill={backButton}
+        fill={this.props.index > 0 ? subMenuHeader : defaultHeader}
```
emilyrohrbough: nit: can we update this be something like:
````jsx
let subMenuHeader = (
      <div
        role="button"
        onClick={this.props.onRequestBack}
        onKeyDown={this.onKeyDownBackButton}
        tabIndex="0"
        aria-label={backBtnText}
      >
        <Arrange
          align="center"
          fitStart={(
            <div className={cx('header-button')}>
              {backIcon}
            </div>
          )}
          fill={<h1 className={cx('header-title')}>{this.props.title}</h1>}
        />
      </div>
);

let header = <div />

if (this.props.index > 0) {
    header = subMenuHeader;
} else if (this.props.headerTitle.length > 0) {
    header = (
       <Arrange
          align="center"
          fill={<h1 className={cx(['header-title', 'main-header-title'])}>{this.props.headerTitle}</h1>}
         />
     );
}
...

eturn (
      <Arrange
        className={cx('header')}
        fitEnd={closeButton}
        fill={header}
```

emilyrohrbough:
@lokesh-0813  Is this PR ready for a full review? I notice it has been out for a while, but it's still in draft mode.

packages/terra-menu/src/_MenuContent.jsx
lokesh-0813: Updated here 423a343d8fa324eeab15d11a2cf2c87e4e36e6f7

lokesh-0813:
> @lokesh-0813 Is this PR ready for a full review? I notice it has been out for a while, but it's still in draft mode.

@emilyrohrbough , I was waiting for @neilpfeiffer to pitch in. I had assumed there would be more changes to make. But I think most of it is covered and will be moving it to 'ready for review'.

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -79,6 +79,10 @@ const propTypes = {
    * Ref callback function to be applied to content container.
    */
   refCallback: PropTypes.func,
+  /**
+   * Header Title for menu if menu contains drill-in sub-menus
+   */
+  headerTitle: PropTypes.string,
```
nramamurth: maybe, add a default value to this prop?

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -267,20 +270,35 @@ class MenuContent extends React.Component {
           fill={<h1 className={cx('header-title')}>{this.props.title}</h1>}
         />
       </div>
-    ) : <div />;
+    );
+
+    let header = <div />;
+
+    if (this.props.index > 0) {
+      header = subMenuHeader;
```
nramamurth: This [chunk of code](https://github.com/cerner/terra-framework/pull/977/files#diff-8821ddb6189a1accf05f7aef26338373R255-R272) can directly move in here.

packages/terra-menu/src/_MenuContent.jsx
lokesh-0813: Default value is defined here [Menu](https://github.com/cerner/terra-framework/blob/814b2bf3b5e65676e29562789da76819cefba73b/packages/terra-menu/src/Menu.jsx#L67)

packages/terra-menu/src/_MenuContent.jsx
lokesh-0813: updated here dcae9eedc320f50ade9e1b88b8d942181019b385

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -297,7 +313,11 @@ class MenuContent extends React.Component {
           onKeyDown,
           isActive,
         });
-      // If the child has children then it is an item group, so iterate through it's children
+
+        if (hasHeaderTitle && item.props.subMenuItems && item.props.subMenuItems.length > 0) {
```
nramamurth: Why should `hasHeaderTitle` decide whether or not there are sub menu items?

lokesh-0813: If the user does not provide the `headerTitle` prop and the menu contains sub-menu-items, `hasSubmenuItem` will be true and trigger building a header and will create an empty `<div/>`. `hasHeaderTitle` makes sure to set `hasSubmenuItem` true only upon providing a valid `headerTitle` prop content.

packages/terra-menu/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Unreleased
 6.23.0 - (January 7, 2020)
 ------------------
 ### Added
+* Added new `headertTitle` prop for menus which have drill-in submenus
```
emilyrohrbough: whoops, looks like this got moved under a released section on merge

packages/terra-menu/src/Menu.module.scss
```diff
@@ -46,6 +46,10 @@
       word-wrap: normal;
     }
 
+    .main-header-title {
+      margin-left: var(--terra-menu-header-title-margin-left, 10px);
```
emilyrohrbough: What happens if the heading is long? Should this have padding on both sides?

emilyrohrbough: Actually, why does the main header have unique styling? Should it not be the same as the sub menus?

packages/terra-menu/src/_MenuContent.jsx
emilyrohrbough: We should define a default here as well

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -79,6 +79,10 @@ const propTypes = {
    * Ref callback function to be applied to content container.
    */
   refCallback: PropTypes.func,
+  /**
+   * Header Title for menu if menu contains drill-in sub-menus
```
emilyrohrbough: I though this was specific to the root menu, not sub menus?

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -333,7 +353,7 @@ class MenuContent extends React.Component {
     ]);
 
     let header;
-    if (isFullScreen || isSubMenu) {
+    if (isFullScreen || isSubMenu || hasSubMenuItems) {
```
emilyrohrbough:
```suggestion
    if (isFullScreen || hasHeaderTitle  || isSubMenu) {
        header = this.buildHeader(isFullScreen);
    }
```

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -297,7 +313,11 @@ class MenuContent extends React.Component {
           onKeyDown,
           isActive,
         });
-      // If the child has children then it is an item group, so iterate through it's children
+
+        if (hasHeaderTitle && item.props.subMenuItems && item.props.subMenuItems.length > 0) {
+          hasSubMenuItems = true;
```
emilyrohrbough: I don't think the `hasSubMenuItems` bool is needed, the `isSubMenu` bool below will render sub menu titles

packages/terra-menu/src/_MenuContent.jsx
emilyrohrbough: I don't think the `hasSubMenuItems` bool is needed, the `isSubMenu` bool below will render sub menu titles

packages/terra-menu/CHANGELOG.md
lokesh-0813: oops my bad. Updated here 4b0525184b23469cb9fe417ee47bbff1a30e034b

packages/terra-menu/src/Menu.module.scss
lokesh-0813: > What happens if the header title is long? Should this have padding on both sides?

That's a good question. Currently ,if the sub-menu header title is longer than the width of the menu, an ellipsis is shown at the end of the title(a non-functioning ellipsis ). And the same logic applies to main menu header-title as well at this point. 

@jeremyfuksa what are your thoughts for longer header-titles ? Should we display the complete text in the header-title section or should we enable a functioning ellipsis ?

packages/terra-menu/src/_MenuContent.jsx
lokesh-0813: `isSubMenu` check is true only for second and higher-tier menus.

To render the header for main-menu(first-tier) we need the `hasSubMenuItems` check.

packages/terra-menu/src/_MenuContent.jsx
lokesh-0813: According to the requirement `add a mandatory top header with a title for any menu that has a drill-in nested menu on the first-tier of the menu (currently only adds the header on the second-tier and higher)` from issue #474 .

Above _suggested changes_ does not meet the issue requirement and fail in a scenario->

When Consumer provides headerTitle to a menu without any sub-menu.
Now `isSubMenu=false`, `hasHeaderTitle=true`
And the header is rendered which does not satisfy the requirement.

The `hasSubMenuItems` check assures to build main-header only when menu has sub-menu items and valid `headerTitle` content.

packages/terra-menu/src/_MenuContent.jsx
lokesh-0813: Updated here 277939ef97fccdb445d4a11850b83f430e56c73b

packages/terra-menu/src/_MenuContent.jsx
lokesh-0813: Updated here 277939ef97fccdb445d4a11850b83f430e56c73b

packages/terra-menu/src/Menu.module.scss
supreethmr: Since fusion theme is ported to terra-menu, I think you can add this variable in [orion-fusion-theme.scss](https://github.com/cerner/terra-framework/blob/master/packages/terra-menu/themes/orion-fusion-theme/orion-fusion-theme.scss) and might need to plan for add in other themes as well.

emilyrohrbough: @lokesh-0813 Looking at the issue you linked, bullet number five says:
>  either refactor top header(and submenu nested menu headers) so the height will flex based on title content and correctly wrap, or import action-header which already has intended layout.

packages/terra-menu/src/Menu.jsx
```diff
@@ -53,12 +53,18 @@ const propTypes = {
...
+   * Header Title will only be visible if the main-menu contains at least one sub-menu.
+   */
+  headerTitle: PropTypes.string,
 };
 
 const defaultProps = {
   isArrowDisplayed: false,
   isOpen: false,
   contentWidth: '240',
+  headerTitle: '',
```
ShettyAkarsh: we could set this as Undefined

packages/terra-menu/src/_MenuContent.jsx
ShettyAkarsh: > If the user does not provide the `headerTitle` prop and the menu contains sub-menu-items, `hasSubmenuItem` will be true and trigger building a header and will create an empty `<div/>`. `hasHeaderTitle` makes sure to set `hasSubmenuItem` true only upon providing a valid `headerTitle` prop content.

@lokesh-0813 ,  in this case we might have to add the prop level validation to trigger the warning mentioning that. Set the headertitle prop value as the submenutiems are present. and also update the doc mentioning that the the value is required when submenu items are present.

cc @neilpfeiffer

packages/terra-menu/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Added
+* Added new `headertTitle` prop for menus which have drill-in submenus
```
nramamurth: headerTitle*

packages/terra-menu/src/_MenuContent.jsx
nramamurth: +1 to @emilyrohrbough's comment and I still stand by my original comment. I think this logic could be simplified to not have new booleans - isn't it as simple as checking the value of  `this.props.headerTitle.length > 0` **AND** `item.props.subMenuItems && item.props.subMenuItems.length > 0` directly? If this condition evaluates to true, then build the header with title, otherwise do not build the header with title.

lokesh-0813:
@jeremyfuksa To accommodate longer header titles I have made some minor style adjustments.


Here are the changes a102a41f0af6205b178605b7f1ba66ca1df26a9f

Test for long menu item texts -> https://terra-framework-deploye-pr-977.herokuapp.com/tests/terra-menu/menu/menu/sub-menu-long-title


But, In Fusion theme, header has a fixed height

`  --terra-menu-submenu-header-height: 2.5rem`

and is truncating long header texts. Below is a gif

fusionfail

Let me know your thoughts.

packages/terra-menu/src/_MenuContent.jsx
supreethmr: I think if you modify line number : 301 like below. above suggested changes can be included.
```javascript
const hasHeaderTitle = this.props.headerTitle.length > 0 && isSubMenu;
```

packages/terra-menu/src/_MenuContent.jsx
lokesh-0813: @emilyrohrbough @nramamurth @ShettyAkarsh updated here b1254c430078f5a32a9c1de79d9402410ecc4df2

packages/terra-menu/src/_MenuContent.jsx
lokesh-0813: updated here b1254c430078f5a32a9c1de79d9402410ecc4df2

packages/terra-menu/src/Menu.jsx
lokesh-0813: There is a length check on headerTitle, setting the default value to `undefined` will result in a error.

packages/terra-menu/CHANGELOG.md
lokesh-0813: updated here a102a41f0af6205b178605b7f1ba66ca1df26a9f

jeremyfuksa:
Really close. In the test, there are some padding issues in the nested menu.

Screen Shot 2020-01-23 05 03 13 PM

lokesh-0813:
> Really close. In the test, there are some padding issues in the nested menu.
> 
> Screen Shot 2020-01-23 05 03 13 PM

I will try fixing the padding issues for the default theme. In the meantime any comments for terra-menu long titles for fusion-theme ?

jeremyfuksa:
Sorry. I missed looking at Fusion last week.

Unfortunately, it looks pretty broken. I pulled it and built it locally, too, just to make sure. Broken there too.

Screen Shot 2020-01-27 12 57 20 AM

lokesh-0813:
> Sorry. I missed looking at Fusion last week.
> 
> Unfortunately, it looks pretty broken. I pulled it and built it locally, too, just to make sure. Broken there too.
> 
> Screen Shot 2020-01-27 12 57 20 AM

Yes that's because fusion-theme has a fixed-height for headerTitle.

jeremyfuksa:
> Yes that's because fusion-theme has a fixed-height for headerTitle.
We need to fix that then, please.

lokesh-0813:
> > Yes that's because fusion-theme has a fixed-height for headerTitle.
> 
> We need to fix that then, please.

Updated here ef3239cd78e5d657c848f92d318f790c6a31f47a

lokesh-0813:
> Really close. In the test, there are some padding issues in the nested menu.
> 
> Screen Shot 2020-01-23 05 03 13 PM

Updated here 99331a764be4076d201c71839dccffec508acd52

jeremyfuksa:
Wrapped header in the Fusion theme is not working in IE 11.

Screen Shot 2020-02-03 10 30 40 AM

ryanthemanuel:
@lokesh-0813 with the force push is this ready to review again?

lokesh-0813:
> @lokesh-0813 with the force push is this ready to review again?

@ryanthemanuel Apparently the `word-break` styles are not getting applied in IE.

@jmsv6d I was wondering if you would have any thoughts here. 
When building header for menus, there is a scenario where we use nested `Arrange`.
 [Here](https://github.com/cerner/terra-framework/pull/977/files#diff-8821ddb6189a1accf05f7aef26338373R281) and then [here](https://github.com/cerner/terra-framework/pull/977/files#diff-8821ddb6189a1accf05f7aef26338373R289). I am suspecting this could be causing the words not to break in IE(works fine in other browsers).

So i removed the `Arrange` [here](https://github.com/cerner/terra-framework/pull/977/files#diff-8821ddb6189a1accf05f7aef26338373R281) and directly returned the `h1` element, and it seem to work fine. Although this cannot be done for nested-menu's since we need `fitStart` for back-icon and `fill` to fill the text content.

jmsv6d:
@lokesh-0813 I'm not sure why the `word-break` style isn't applying appropriately, but the `Arrange` element you referenced should definitely be removed regardless as it doesn't provide any value if you're only using the fill prop. I wouldn't be opposed to removing the other `Arrange` as well and just using flex-box styling.

packages/terra-menu/src/Menu.module.scss
lokesh-0813: updated here ef3239cd78e5d657c848f92d318f790c6a31f47a

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -248,39 +254,44 @@ class MenuContent extends React.Component {
...
-        fill={backButton}
+        fill={header}
         align="center"
       />
     );
   }
 
   render() {
     let index = -1;
+    let isMainMenuHeader;
```
jmsv6d: I think this variable would make more sense as `shouldDisplayMainMenuHeader`

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -297,7 +308,10 @@ class MenuContent extends React.Component {
           onKeyDown,
           isActive,
         });
-      // If the child has children then it is an item group, so iterate through it's children
+        if (this.props.headerTitle.length > 0 && item.props.subMenuItems && item.props.subMenuItems.length > 0) {
```
jmsv6d: Could you add a comment explaining why this check is being done.

packages/terra-menu/src/_MenuContent.jsx
lokesh-0813: updated here 16ed11f47288f7ab7e65f9410e955857d99c80f1

packages/terra-menu/src/_MenuContent.jsx
lokesh-0813: updated here 16ed11f47288f7ab7e65f9410e955857d99c80f1

PR #973, opened 2019-11-20
Add Rendered Prop to Enabled Application-Level Content Wrapping of Disclosed Content

emilyrohrbough:
### Summary
Added support for application-level disclosure content wrapping

These changes are related to: https://github.com/cerner/terra-application/pull/2

PR #982, opened 2019-12-06
Fix fusion theme modal manager tests

benbcai:
### Summary
To fix fusion theme test failures.

### Additional Details
- Some fusion theme tests are getting an insufficient color contrast between the black font color and the purple background color used in the modal manger disclosure container test example. The background color needs to be lighter.

- Include all the other modal manager tests that were excluded.

@cerner/terra


emilyrohrbough:
@benbcai  wasn't the fusion theme port just merged? Are there changes that need pulled forward that we're missed?

avinashg1994:
> @benbcai wasn't the fusion theme port just merged? Are there changes that need pulled forward that we're missed?

The orion-fusion-theme ported to terra-framework has the latest approved changes from the orion-fusion-theme internal repo. So that is not the issue.
From the looks of it, i guess the modal manager screen shots were not updated due to the "describe.only" in the spec file which came in from the master branch.

benbcai:
@emilyrohrbough @avinashg1994 I saw this insufficient color contrast failure in the orion-fusion-theme repo after #973 got merged. However, I would except that the same insufficient color contrast failure would be seen on #975 but apparently it passed. I don't know if there is any difference between the test runs in the orion-fusion-theme repo versus the ones here but the fusion theme screenshots look the same.

PR #984, opened 2019-12-10
[terra-tabs] Fix overflow issue in icon-only tabs. 

### LINKED ISSUE #970: [terra-tabs] Overflow in icon-only collapsible tabs.
# Bug Report

## Description
This issue occurs _only_ when all tabs are icon-only. In a scenario when some tabs need to be collapsed into a menu due to unavailability of space, container supposedly appears to overflow — few tabs that needed to be _actually_ hidden are visible pushing the "More" button to the right and causing an overflow. 

## Steps to Reproduce
1. Display, let's say, 50 icon-only tabs.

icon-only

## Repro

```
import React from 'react';
import IconSearch from 'terra-icon/lib/icon/IconSearch';
import Tabs from 'terra-tabs';

const createTabPanes = () => {
  const tabPanes = [];
  for (let i = 0; i < 50; i += 1) {
    const tabPane = (
      <Tabs.Pane label={`Search${i}`} icon={<IconSearch />} isIconOnly key={`Search${i}`} id={`search${i}`} />
    );

    tabPanes.push(tabPane);
  }
  return tabPanes;
};

const IconOnlyTabs = () => (
  <Tabs id="icononlytabs" responsiveTo="none">
    {createTabPanes()}
  </Tabs>
);

export default IconOnlyTabs;

```
## Expected Behavior
Horizontal scroll should not appear, 'More' button should not look cut off. 

## Possible Solution
The minimum width set to icon-only tab affects the [logic that we have to calculate the `hideStateIndex`](https://github.com/cerner/terra-framework/blob/6cf7f8722f57836915c06589297b30efd09fb1df/packages/terra-tabs/src/_CollapsibleTabs.jsx#L109-L130) that decides how many tabs to show and how many to collapse into the menu. **Increasing the min-width fixes this issue**. 

https://github.com/cerner/terra-framework/blob/6cf7f8722f57836915c06589297b30efd09fb1df/packages/terra-tabs/src/Tabs.module.scss#L58-L62

## Environment

* Component Name and Version: terra-tabs v6.18.0
* Browser Name and Version: All browsers.

## @ Mentions
@neilpfeiffer

nramamurth:
### Summary
Fixes #970


jmsv6d:
I don't think we should make this change until we understand the reason behind the issue. I suspect this would be caused by the tabs not actually shrinking down to the min width for some reason. There could be another style change that would fix this issue.

jmsv6d:
I think the size of the icon + padding is larger than the current min-width, so the icon-only tabs aren't actually resizing down to the min-width.

nramamurth:
> the size of the icon + padding is larger than the current min-width. 

@jmsv6d I'm curious, how could that be an issue? Also, what is the bad impact of increasing the min-width here? 

Here's what I see - [the calculation that decides the index from which the rest of the tabs need to be hidden depends on the min-width of each tab set.](https://github.com/cerner/terra-framework/blob/316117abf1e13c11c72172c5f555fa8257ba384b/packages/terra-tabs/src/_CollapsibleTabs.jsx#L123-L130) Below is the information when the min-width is set to 2.571rem v/s 2.771 rem on the test example that I linked on the original issue.

**min-width - 2.571rem:** 

Screen Shot 2019-12-23 at 12 46 08 PM

**min-width - 2.771rem:** 
Screen Shot 2019-12-23 at 12 44 26 PM

1. Margin does not affect this behaviour because there is no margin at all. 
2. Calculation loop breaks at the right index (38) with increased min-width (2.771rem) because the accumulated min-width at that point is greater than the available width. 

cc / @jeremyfuksa @neilpfeiffer

jmsv6d:
@nramamurth The width calculation depends on the tabs actually resizing down to the min width. That assumption is wrong in iconOnly currently because the size of the icon + padding is larger than the minimum width. Increasing the min width does fix the issue and is fine, but with the knowledge of what is actually happening there are other styles that could also be changed to fix it (padding or icon size). So that would be a UX decision. 
We could also investigate if there is a better way to determine the smallest width that the tab can be when doing calculations. Perhaps getting the width of the icon + padding instead of min-width so it's less likely to get out of sync again, particularly during theme development.

jeremyfuksa:
Changing the left/right padding of the icon-only variant to 0.788rem also fixes this issue.

ryanthemanuel:
+1 to @jeremyfuksa's suggestion

nramamurth:
> Changing the left/right padding of the icon-only variant to 0.788rem also fixes this issue.

@jeremyfuksa, having 0.688 did it and I have updated both the fusion theme and the default theme.

emilyrohrbough:
@jeremyfuksa if the padding affects the tab calculation, should we remove the option to theme it or wha happens if this theme value is altered?

jeremyfuksa:
If the value is >= 0.8rem we get the overflow problem. I changed the value all the way down to 0 for left/right padding and it had the same appearance as any value <= 0.688rem.

My concern about removing theming was with the Fusion theme and what it would do with a different base font size, but it was OK when I tweaked values. 

So, I'd suggest removing the theme variable here. We could also make left/right padding = 0 to take out another potential point of error.

emilyrohrbough:
Is this screenshot correct or outdated? It's labeled icon-only but I am only seeing text in the tab:
Screen Shot 2020-01-13 at 11 53 36 AM

PR #987, opened 2019-12-13
[terra-notification-dialog] Custom Validation added for action props and marked as optional.

### LINKED ISSUE #974: [terra-notification-dialog] 'acceptAction' prop marked as required even when given 'primaryAction'
# Bug Report

## Description
If `primaryAction` is given a value and `acceptAction` is not, notification-dialog will throw a prop warning since `acceptAction` is marked as required.

## Steps to Reproduce
```
import React, { useState } from 'react';
import NotificationDialog, { NotificationDialogVariants } from 'terra-notification-dialog';

const clickConfirm = () => {
  console.log('You clicked confirm'); // eslint-disable-line no-console
};

const CompleteNotificationDialog = () => {
  const [isOpen, setIsOpen] = useState(false);

  const handleOpenModal = () => {
    setIsOpen(true);
  };

  const handleCloseModal = () => {
    setIsOpen(false);
  };

  return (
    <>
      <NotificationDialog
        variant={NotificationDialogVariants.ALERT}
        isOpen={isOpen}
        title="Make sure that the title relates directly to the choices."
        startMessage="The Main Instruction is text used to provide more detail or define terminology. Don’t repeat the title verbatim."
        primaryAction={{
          text: 'Confirm',
          onClick: clickConfirm,
        }}
        secondaryAction={{
          text: 'Close',
          onClick: handleCloseModal,
        }}
        buttonOrder="acceptFirst"
        emphasizedAction="accept"
      />
      <Button text="Trigger NotificationDialog" onClick={handleOpenModal} />
    </>
  );
};

export default CompleteNotificationDialog;
```

Error Message:
> Warning: Failed prop type: The prop `acceptAction` is marked as required in `NotificationDialog`, but its value is `undefined`.

## Expected Behavior
Should not throw a warning, should just accept the `primaryAction` prop as the `acceptAction` prop.

## Environment
- Component Name and Version: terra-notification-dialog v3.18.0
- Browser Name and Version: All browsers.

saket2403:
### Summary
Fixes #974


packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -223,6 +223,14 @@ class NotificationDialog extends React.Component {
       message,
       ...customProps
     } = this.props;
+    if (acceptAction === undefined && primaryAction === undefined) {
+      // eslint-disable-next-line no-console
+      console.warn('acceptAction or primaryAction prop must be given.');
```
emilyrohrbough: We don't want to log these messages in production mode. Lets add:
```
if (process.env.NODE_ENV !== 'production') {
     console.warn(....)
}
```

packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -223,6 +223,14 @@ class NotificationDialog extends React.Component {
...
+    }
+    if (rejectAction === undefined && secondaryAction === undefined) {
+      // eslint-disable-next-line no-console
+      console.warn('rejectAction or secondaryAction prop must be given.');
+    }
```
BenBoersma: I think we only need a warning if none of `acceptAction`, `rejectAction`, `primaryAction` or `secondaryAction` is defined.  

As long as one of those four is defined we can exit the notification-dialog which was the main reason `primaryAction` and `acceptAction` were marked as required in previous versions.

packages/terra-notification-dialog/src/NotificationDialog.jsx
emilyrohrbough: Also can we add some context to the message. A consumer that didn't set these props may not know these messages are meant for notification-dialog

packages/terra-notification-dialog/src/NotificationDialog.jsx
saket2403: Resolved in 2957b32

packages/terra-notification-dialog/src/NotificationDialog.jsx
saket2403: Changed in 2957b32

packages/terra-notification-dialog/src/NotificationDialog.jsx
```diff
@@ -223,6 +223,10 @@ class NotificationDialog extends React.Component {
       message,
       ...customProps
     } = this.props;
+    if (acceptAction === undefined && primaryAction === undefined && rejectAction === undefined && secondaryAction === undefined && process.env.NODE_ENV !== 'production') {
```
ShettyAkarsh: You can add this `process.env.NODE_ENV !== 'production'  as the first condition. as this should satisfy first to validate the rest.

saket2403: Changed in 9e2431a

nramamurth:
I'd like to suggest a few documentation updates - 

1. The description of a deprecated prop does not emphasize that the prop is deprecated. It kinda blends with the rest of the description - maybe enclose the word in square brackets or make it appear bold if you can : **[Deprecated]** followed by original description.
2. Generated prop table does not specify the values that the `buttonOrder` and `emphasizedAction` can take - please add them in the respective comments so the prop table shows the accepted values for the consumers. 
3. AFAIK, this is a temporary "fix" until MVB but I'd expect whatever is being logged as a warning [here](https://github.com/cerner/terra-framework/pull/987/files#diff-cb62dc7ce0197bc569a9952450bb9fe5R228) to be mentioned as a note in the component description and shown on the site until MVB. Without a note, the consumers will have to find out about which props are in fact required/optional the hard way. 
4. Please enclose the code references in the prop description within a pair back-ticks wherever applicable [1][2].

[1] https://github.com/cerner/terra-framework/blob/master/packages/terra-notification-dialog/src/NotificationDialog.jsx#L32
[2] https://github.com/cerner/terra-framework/blob/master/packages/terra-notification-dialog/src/NotificationDialog.jsx#L48

saket2403:
> I'd like to suggest a few documentation updates -

Added in e25ca0f

packages/terra-notification-dialog/docs/README.md
```diff
@@ -1,6 +1,8 @@
 # Terra NotificationDialog
 
 `terra-notification-dialog` is a notification dialog component built over the `terra-abstract-modal`. It has the highest z-index of 9001. It is a common component to be used for confirmation/acceptance criteria style dialogs.
+<br />**Note:** Use at least one of `acceptAction`, `rejectAction`, `primaryAction` or `secondaryAction` props to avoid getting a warning.
```
nramamurth: > to avoid getting a warning

maybe stress the real problem that the dialog will not function as expected.

saket2403: Updated in 567c907

PR #989, opened 2019-12-13
[terra-modal-manager] Reenabling wdio tests

BenBoersma:
### Summary
Enabling wdio tests and adding in missing screenshots for orion-fusion-theme

@cerner/terra


packages/terra-modal-manager/tests/wdio/modal-manager-spec.js
```diff
@@ -492,7 +492,7 @@ Terra.describeViewports('ModalManager - Disclosure Accessory', ['large'], () =>
 
     it('renders each disclosed component in a disclosure container', () => {
       expect(browser.isExisting('[data-disclosure-container="true"]')).to.be.true;
-      Terra.validates.element('nested disclosure', { selector });
+      Terra.validates.screenshot('nested disclosure', { selector });
```
emilyrohrbough: Can you remove the `only` on this describe block? It was something I accidently merged in

BenBoersma: removed here https://github.com/cerner/terra-framework/pull/989/commits/dc3742c06e6b79e7055729172d004837736b891a

BenBoersma:
I ran master with the new axe-core update changes and am no longer getting the wdio failure this PR fixes. Since the error was intermittent, I ran the tests locally about 4+ times, and it succeeded on every one. 

However this PR still has the fix for removing the `describe.only` so I can undo the tests changes from `.element` to `.screenshot` and keep the the change to `describe.only` and the screenshots that produces.  @neilpfeiffer @ryanthemanuel thoughts?

BenBoersma:
As per offline discussion, Changing this PR to just reenabling some wdio tests and adding the missing orion-fusion-theme screenshots for those tests.

Issue #952, opened 2019-10-24
[Spike] [terra-menu] - Investigate Allowing Submenu Resizing


StephenEsser:
# Feature Request

Investigate allowing the terra-menu to resize for submenu navigation.

## Description

When the menu opens the initial height of the menu establishes the static height the menu will be for the lifetime of that popup.

Teams are starting to uncover workflows where the initial main menu height is relatively small and the submenu height is much larger. We should spike on the feasibility/usability of allowing menus to resize. 

## @ Mentions

@jmsv6d @neilpfeiffer

lokesh-0813:
@StephenEsser  Are we looking at exposing a prop to set the height of the menu? Or we could reset the menu height based on subMenu items(My initial opinion and still need to dig more). Any thoughts on this?

StephenEsser:
@lokesh-0813 

When logging this issue originally I was thinking something similar to a prop that sets an auto  dynamic height of the menu as necessary. Similar to the terra-popop auto height API.

This request came through from a UX designer and a few other teams. We should get more confirmation about the expected functionality. 

cc: @neilpfeiffer

jmsv6d:
Currently the menu will auto size the height based on the height of the main menu, but maintains that height for submenus. There is code in the menu right now that forces this to happen. If the submenu is smaller than the main menu, this is fine, but there are some scenarios where the submenu is longer than the main menu and the user may end up with a small popup with a scrollbar. I was thinking we have a couple of options (or some combination of these options)
1) Bool prop that allows for dynamic sizing of submenus. The downside to this approach is that resizing of menu could cause the popup to need to be repositioned, causing a jarring user experience.
2) Open up the prop with all the popup menu sizing options. Then consumers could pick the set height for the menu, users wouldn't have the potentially jarring experience of the popup resizing during menu navigation and they could choose a size that would be more appropriate for all the menu's "pages".
But I think we need input from @neilpfeiffer on what approach we want to take and if we want to make a change at all.

neilpfeiffer:
Should also pair with a fix missing headers and text wrapping/flexing header height:
https://github.com/cerner/terra-framework/issues/474

Changes needed were discussed during the UX Audit:
https://github.com/cerner/terra-framework/issues/1297

PR #992, opened 2019-12-18
[terra-menu] Dynamic resizing of sub-menu

lokesh-0813:
fixes : #952


lokesh-0813:
Dynamic resizing can be seen in [this](https://terra-framework-deploye-pr-992.herokuapp.com/tests/terra-menu/menu/menu/sub-menu) test example.

Let me know your thoughts @jmsv6d @StephenEsser

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -146,6 +143,13 @@ class MenuContent extends React.Component {
     }
   }
 
+  setPageHeight(node) {
```
jmsv6d: You shouldn't need to do this calculation. Popup should handle this without setting an explicit height on the content. You can change this https://github.com/cerner/terra-framework/pull/992/files#diff-8821ddb6189a1accf05f7aef26338373R344 to be 
`const contentHeight = this.props.isHeightBounded && '100%' ;
And update the content container's `fill` prop to be just `this.props.isHeightBounded`

jmsv6d:
SubmenuResize

Is this acceptable behavior? @neilpfeiffer

packages/terra-menu/src/_MenuContent.jsx
lokesh-0813: updated here b91205212bdd630b10eb4d5f5748cd63e33ac4a0

packages/terra-menu/src/terra-dev-site/test/menu/menu/SubMenu.test.jsx
```diff
@@ -50,6 +50,18 @@ class SubMenu extends React.Component {
...
+              <Menu.Item text="Default 2.4" key="2.4" className="TestNestedMenuContent" />,
+              <Menu.Item text="Default 2.5" key="2.5" className="TestNestedMenuContent" />,
+              <Menu.Item text="Default 2.6" key="2.6" className="TestNestedMenuContent" />,
+              <Menu.Item text="Default 2.7" key="2.7" className="TestNestedMenuContent" />,
+              <Menu.Item text="Default 2.8" key="2.8" className="TestNestedMenuContent" />,
+              <Menu.Item text="Default 2.9" key="2.9" className="TestNestedMenuContent" />,
+              <Menu.Item text="Default 3.0" key="3.0" className="TestNestedMenuContent" />,
+              <Menu.Item text="Default 3.1" key="3.1" className="TestNestedMenuContent" />,
+              <Menu.Item text="Default 3.2" key="3.2" className="TestNestedMenuContent" />,
+              <Menu.Item text="Default 3.1" key="3.3" className="TestNestedMenuContent" />,
```
saket2403: Nit: It should be `Default 3.3` instead of 3.1.

nramamurth:
@lokesh-0813 Are there no _wdio_ screenshots that need to be updated with this change? If not, then I'd recommend adding one.

packages/terra-menu/src/terra-dev-site/test/menu/menu/SubMenu.test.jsx
lokesh-0813: Fixed here 9515f20110ed39ea657e076965e3ed0e74cfd39c

lokesh-0813:
> @lokesh-0813 Are there no _wdio_ screenshots that need to be updated with this change? If not, then I'd recommend adding one.

Updated here 9515f20110ed39ea657e076965e3ed0e74cfd39c and here a87334b66361e2c03b09cbaff34af6611959f615

packages/terra-menu/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Removed fixedHeight logic from Menu.
```
emilyrohrbough: Can we update this descriptions o it's meaningful to users? Maybe something like we removed the fix height in sub-menus or we allow sub-menu's height to be determined by it's content?

emilyrohrbough:
@lokesh-0813  I am noticing different behavior of bound menus when the sub-menu is clicked. It seems the sub-menu is not longer bound. [Here is a gif](https://share.getcloudapp.com/xQu0xone) of the behavior seen in the deployment vs the existing behaviors found on https://engineering.cerner.com/terra-framework/tests/terra-menu/menu/menu/bounded-menu.

lokesh-0813:
> @lokesh-0813 I am noticing different behavior of bound menus when the sub-menu is clicked. It seems the sub-menu is not longer bound. [Here is a gif](https://share.getcloudapp.com/xQu0xone) of the behavior seen in the deployment vs the existing behaviors found on https://engineering.cerner.com/terra-framework/tests/terra-menu/menu/menu/bounded-menu.

It's expected, since sub-menu has lesser menu-items compared to main-menu menu-items. Now sub-menu determines its height based on the number of menu-items it possess.

I have tweaked the Bounded Menu example(removed main-menu items and added more sub-menu items) to give you an idea of what's going on ->

Existing behaviour(master Branch)
currenttweak

Deployment behaviour
deploymentTweak

emilyrohrbough:
> @emilyrohrbough It's expected, since sub-menu has lesser menu-items compared to main-menu menu-items. Now sub-menu determines its height based on the number of menu-items it possess.

I thought a bounded menu's height was determined by i's parent container, not its contents. Is that not the case?

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -350,7 +346,7 @@ class MenuContent extends React.Component {
         role="dialog"
         onKeyDown={this.onKeyDown}
       >
-        <ContentContainer header={header} fill={this.props.isHeightBounded || this.props.index > 0}>
+        <ContentContainer header={header} fill={this.props.isHeightBounded}>
```
ShettyAkarsh: why was the index prop used here before ?

lokesh-0813: index > 0 would imply that it is sub-menu that is being rendered.

lokesh-0813:
> > @emilyrohrbough It's expected, since sub-menu has lesser menu-items compared to main-menu menu-items. Now sub-menu determines its height based on the number of menu-items it possess.
> 
> I thought a bounded menu's height was determined by i's parent container, not its contents. Is that not the case?

cc : @neilpfeiffer

packages/terra-menu/src/_MenuContent.jsx
ShettyAkarsh: so how we are handling this scenario now ? after removing this ?

jmsv6d:
@emilyrohrbough A menu/popup is "bounded" depending on if the height or width is greater than that of the bounding container. With the menu having a dynamic height, the state of if it's bounded or not can change. This is how I would expect this to work for this situation, I don't think there is an efficient way to get around it, especially for the reverse situation that @lokesh-0813 showed. But it's one of the visual oddities that we should have ux verify if it's acceptable, to determine if this is the actual approach we want to take to fix the original issue of if the submenu is much larger than the main menu, the submenu can end up with a very small scrollable area.

mjhenkes:
Closing until we identify the appropriate solution on the parent issue: #952

PR #993, opened 2019-12-19
[terra-modal-manager] Fixing axe contrast reliability in wdio test

BenBoersma:
### Summary
WDIO test was failing intermittently. To improve reliability, I've changed the test to ignore anything under the `slide-shadow` class to help only check contrast where we need to check.

This test was fixed with the Axe-core upgrade but since we downgraded back, we need this fix again.

@cerner/terra


packages/terra-modal-manager/tests/wdio/modal-manager-spec.js
```diff
@@ -475,14 +475,22 @@ Terra.describeViewports('ModalManager - Disclosure Accessory', ['large'], () =>
 
   describe('Disclosure Container', () => {
     before(() => browser.url('/#/raw/tests/terra-modal-manager/modal-manager/with-disclosure-container').refresh());
+    // Skip color contrast check for elements behind a slide-shadow
+    const ignoredA11y = {
+      'color-contrast': {
```
emilyrohrbough: Can we just remove the color in the test example?

BenBoersma: It still might occasionally fail due to the dark grey of the transition from `slide-shadow` I think. 

We can still remove the color if we need, but I'm starting to view this to how we deal with notification-dialog where we do a similar thing and ignore contrast tests under overlay.

emilyrohrbough: Can you log an issue to terra-toolkit. If we need to define a strategy for these types of test behaviors, I'd be nice to point consumers to it as they are also probably see these same issues.

packages/terra-modal-manager/tests/wdio/modal-manager-spec.js
```diff
@@ -492,7 +500,7 @@ Terra.describeViewports('ModalManager - Disclosure Accessory', ['large'], () =>
 
     it('renders each disclosed component in a disclosure container', () => {
       expect(browser.isExisting('[data-disclosure-container="true"]')).to.be.true;
-      Terra.validates.element('nested disclosure', { selector });
+      Terra.validates.element('nested disclosure', { selector, axeRules: ignoredA11y });
```
nramamurth: This may be unrelated but recently I noticed that this [modal manager test page](https://engineering.cerner.com/terra-framework/tests/terra-modal-manager/modal-manager/modal-manager-integration) has a select field that does not seem to do anything. I'm curious, what is it for?

packages/terra-modal-manager/tests/wdio/modal-manager-spec.js
BenBoersma: issue logged https://github.com/cerner/terra-toolkit/issues/373

packages/terra-modal-manager/tests/wdio/modal-manager-spec.js
StephenEsser: This is/was used previously to validate that the Select hookshot implementation worked as expected within Modals. The Select dropdown is using hookshot to position itself. The z-layer ordering and positioning changes are really important to validate are working as intended.

PR #866, opened 2019-09-04
Improve datepicker popup accessibility

bjankord:
### Summary
Resolves #631
Resolves #633

### Deployment Link
https://terra-framework-deploye-pr-866.herokuapp.com/components/terra-date-picker/date-picker/date-picker

PR #994, opened 2019-12-23
[terra-date-picker] Update keyboard focus styling to not appear as two dates selected simultaneously

### LINKED ISSUE #988: [Terra-date-picker] Update date selection keyboard focus styling
# Bug Report

## Description

When a date is selected in the first few days of the month, when you navigate to the previous month, it highlights both the selected one and the same in the previous month. This happens in both `Terra-date-picker` and `Terra-date-time-picker`

## Steps to Reproduce

1. Default datepicker value to a day between 1-5 of the month.
2. Open datepicker
3. Navigate to previous month
4. Vouila

duplicate-date

## Expected Behavior

Only show one date at a time. Either the previous month or the current selected one.

## Possible Solution

Only show one date at a time. Either the previous month or the current selected one.

## Environment

* Component Name and Version: Terra-date-time-picker 4.23.0
* Node/npm Version: [e.g. Node 8/npm 5] Node 10.16.2, npm 6.9
* Webpack Version: 
* Operating System and version (desktop or mobile): Mac

saket2403:
### Summary
Fixes #988


nramamurth:
When I have 12/01/2019 selected first and change the month, then it should still have 12/01/2019 selected (if visible) but I see that the first date of the month (navigated to) is selected. 

**Current behaviour (deployed):** 
date-picker

**react-datepicker:**
date-picker-2

saket2403:
> **react-datepicker:**
> date-picker-2

Implemented in 770b958

packages/terra-date-picker/src/react-datepicker/calendar.jsx
```diff
@@ -601,7 +606,8 @@ export default class Calendar extends React.Component {
             utcOffset={this.props.utcOffset}
             handleCalendarKeyDown={this.props.handleCalendarKeyDown}
             locale={this.props.locale}
-            intl={this.props.intl} />
+            intl={this.props.intl}
+            disableKeyboardSelected={this.state.disableKeyboardSelected} />
```
emilyrohrbough: When is this ever set to false after the initial render?

emilyrohrbough:
Can you add a description of how this change is fixing the issue and add a test case to test against?

Does this fix solve issues when an initial date is used and then a date is updated by keyboard and/or mouse selection?

packages/terra-date-picker/src/react-datepicker/calendar.jsx
```diff
@@ -241,6 +241,7 @@ export default class Calendar extends React.Component {
       date: this.localizeDate(this.getDateInView()),
       selectingDate: null,
       calendarIsKeyboardFocused: false,
+      disableKeyboardSelected: false,
```
nramamurth: This issue is reproducible sans keyboard - via mouse clicks. How does this fix the issue?

nramamurth:
With respect to https://github.com/cerner/terra-framework/pull/994/commits/770b958d042290765d65913b556d85beb9854346, I can't seem to navigate via keyboard. 

1. Select a date in the first week of a month. 
2. Click on the month's dropdown and choose the month before.
3. Click within the calendar so that focus is now moved from month dropdown to days. 
4. Use keyboard to change days.

benbcai:
I think there are a couple of changes that contributed to this behavior.

1. A change was made on #866 to intentionally pre-select the date as you navigate to a different month to assist in keyboard navigation.  See these [comments](https://github.com/cerner/terra-framework/pull/866#issuecomment-532420840). One of these two highlighted dates is the pre-selected date whereas the other one is the date that was selected. I agree that the background color looks the same and it's hard to differentiate the difference. The background color for the pre-select date used to be a lighter blue. Maybe it should not update the pre-select date if the selected date is visible on the picker.

2. I think it is this [change](https://github.com/cerner/terra-framework/pull/911/files#diff-ea25495629d17dc8e1e6c545f7101273) that updated the background color of the pre-selected date which now uses the same color as the selected date. Perhaps, we should choose a different color but one that has sufficient color contrast.

saket2403:
> I think there are a couple of changes that contributed to this behavior.
In that case I will just darken the `selected` colour and keep the `keyboard selected` colour as it is.

packages/terra-date-picker/src/react-datepicker/calendar.jsx
saket2403: Reolved in 15987a0

packages/terra-date-picker/src/react-datepicker/calendar.jsx
saket2403: Resolved in 15987a0

saket2403:
> With respect to [770b958](https://github.com/cerner/terra-framework/commit/770b958d042290765d65913b556d85beb9854346), I can't seem to navigate via keyboard.
Changed in 15987a0

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
```diff
@@ -193,7 +193,7 @@
     &--selected,
     &--in-selecting-range,
     &--in-range {
-      background-color: $datepicker-selected-color;
+      background-color: darken($datepicker-selected-color, 10%);
```
benbcai: We probably want to check with UX to see if we want to update the background color of the _selected_ date or the _pre-selected_ date.

ShettyAkarsh:
@benbcai , @neilpfeiffer i had a question here in datetimepicker like why we are allowing user to select dates of previous or next month in the current month view as shown below .

DateSelect

benbcai:
@ShettyAkarsh Most date pickers in other applications also display and allow selecting a date that falls in the previous or next month if the picker has enough slots to fill those dates. I think this makes it easier to visually see the flow of the dates that connect from the previous month to the current month and the current month to the next month. Displaying but disabling those outside dates would make it appear as if they can never be selected but if you then navigate to previous/month month, those dates would then become enabled. This would seem confusing.

Ideally, those dates that are outside of the current month should probably have a slightly lighter color to differentiate them from dates that are in the current month.

@neilpfeiffer might have other better explanations.

neilpfeiffer:
So after full investigation - per conversation with @benbcai and comments posted above, date-picker is not actually showing or selecting two dates simultaneously, but instead this is due to the recent intentional addition to show a visual indication, both for the currently selected day, and also as you keyboard/arrow navigate on the calendar forward and board, showing the day that has focus-styling applied, indicating it is the potential day to be chosen and it would be the new date selected if pressing the enter key. Fixes needed are to update the visual treatment, change when the styling is applied/visible and hidden, and update some additional behavior details to make it perform more inline with user expectations. 

### Design Input

**1. Update when the calendar day Keyboard Focus treatment is applied/visible:**
- [ ] Month-body currently has visual toggle for keyboard-focus treatment by setting the `calendarIsKeyboardFocused` (keep these and add to it with new items listed next):
https://github.com/cerner/terra-framework/blob/1755128d321afb244dcfaa7f4714f6a3f1093d43/packages/terra-date-picker/src/react-datepicker/calendar.jsx#L316-L329
- [ ] Pressing direction (left, right, up, down) arrow keys while the month-body & calendar-day-buttons are in focus sets it to true (don't apply it when pressing arrow keys when the other buttons have focus like the month and year dropdown)
- [ ] Pressing the calendar-icon-button before the calendar picker appears by mouse/touch would set style to false, selecting the calendar-icon-button by pressing the enter key would set it to true (or assume for keyboard first and turn off with mouse clicks of any button).

\
\
**2. Update visual styling:**
- [ ] Remove this day styling entirely from here:
https://github.com/cerner/terra-framework/blob/1755128d321afb244dcfaa7f4714f6a3f1093d43/packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss#L205-L213
- [ ] remove the month styling (convert it to the day styling)
https://github.com/cerner/terra-framework/blob/1755128d321afb244dcfaa7f4714f6a3f1093d43/packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss#L491-L492
- [ ] Add the day styling, but only apply it when `calendarIsKeyboardFocused`/`.is-calendar-focused` is also being applied. Should apply to all days, both non-selected days plus the current selected day. Will be something like this: 
```
  .is-calendar-focused {
    .react-datepicker-day {
      &--keyboard-focus {
        outline: var(--terra-date-picker-day-focus-outline, 2px dashed #000);
        outline-offset: var(--terra-date-day-month-focus-outline-offset, 0);
      }
    }
  }
```
- [ ] make sure keyboard focus style class is addictive to the selected day and other day styling (currently overridden/competes with specificity of `&--selected, &--in-selecting-range, &--in-range {`)
 

\
\
**3. Update/Reset the day that is focused**
- [ ] When the Previous Month or Next Month arrow buttons are pressed, don't keep the same day of month, reset to the 1st day of that month, (So if today is February 5th, pressing previous month button puts focus on January 1st, not January 5th).

- [ ] When a new Month is selected from the dropdown, don't keep the same day of month, reset to the 1st day of that month, (So if today is February 5th, setting the month dropdown to June puts focus on June 1st, not June 5th).

- [ ] When a new Year is selected from the dropdown, don't keep the same day of month, reset to the 1st day of that month for that year, (So if today is February 5th, 2020 setting the year dropdown to 2019 puts focus on February 1st, 2019, not February 5th, 2019).




\
\
thanks @saket2403

neilpfeiffer:
> @benbcai , @neilpfeiffer i had a question here in datetimepicker like why we are allowing user to select dates of previous or next month in the current month view as shown below .

@akar - I understand your question now, why technically the days before and after the current month that because they are visible in the calendar and are active, why users can select them even though they are technically outside the listed month/year in the header and users could perhaps make a mistake and misperceive it as a date in the current month.

Without changing the underlying approach to the inherited date picker where we would need to hide any date that is not part of the current month (possible, but would create some new signifiant whitespace + visual gaps that would then need to be resolved), I think the average user understands the rolling nature of calendars and dates and is not a dramatically dangerous practice, since it appears to be common in other single month date-picker patterns. 

Going forward as we decide to make major advancements to our Date-Picker, I would agree that we should examine and Human Factors test the overall usability relative to UX best practices being recommended in the industry like the  [Nielsen Norman Group](https://www.nngroup.com/articles/date-input/) as a guide for how to make advancements.

saket2403:
I have made the changes here 0406a6d. Can you have a look at the deployment and review the functionality cc @neilpfeiffer .

packages/terra-date-picker/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Unreleased
 * Added clinical-lowlight-theme theming files
 
 ### Changed
+* Fixed issue due to two dates getting selected simultaneously.
```
benbcai: Can we elaborate the other changes listed [here](https://github.com/cerner/terra-framework/pull/994#issuecomment-587482809)? Multiple log entries can be added if doing so makes it easier to call out these sets of changes.

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
benbcai: Do we still need this change given the UX feedback on the design input?

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
```diff
@@ -487,8 +477,11 @@
...
   }
 
-  .is-calendar-focused {
-    outline: var(--terra-date-picker-month-focus-outline, 2px dashed #000);
-    outline-offset: var(--terra-date-picker-month-focus-outline-offset, 0);
+  .is-calendar-focused,
+  .react-datepicker-day {
+    &--keyboard-focus {
+      outline: var(--terra-date-picker-day-focus-outline, 2px dashed #000);
+      outline-offset: var(--terra-date-picker-day-month-focus-outline-offset, 0);
```
benbcai: These styles should also be applies to the selected day (the `.react-datepicker-day--selected` class).

packages/terra-date-picker/src/react-datepicker/calendar.jsx
```diff
@@ -397,16 +403,16 @@ export default class Calendar extends React.Component {
 
   changeYear = (year) => {
     this.setState({
-      date: setYear(cloneDate(this.state.date), year)
+      date: setYear(cloneDate(this.state.date), year).startOf('month')
     })
-    this.props.setPreSelection(setYear(cloneDate(this.state.date), year));
+    this.props.setPreSelection(setYear(cloneDate(this.state.date), year).startOf('month'));
```
benbcai: Looks like the preselected date is not getting the dotted line focus style.

benbcai: I was using the mouse click to navigate to the previous/next month and after consulting @neilpfeiffer the preselected date of the first of the month should _not_ have the focus style when navigating using the mouse. When I use the keyboard to interact with the picker, then the preselected date gets the focus. So this part is working as expected.

This scenario doesn't appear to be working as expected.
When I launched the picker by clicking on the calendar button, the preselected date is not focused (which is expected). However, when I start using the arrow keys to navigate within the calendar, the preselected date should display but it is not showing. If I hit the right arrow twice, it would jump to the next month but then hitting the left or right arrow again doesn't do anything.

benbcai:
The preselected date now has the dotted black line as the focus style. However, focus style for the previous-month/next-month buttons and the month/year drop-downs is still the blue focus glow. Is it possible to also update the focus style for these elements to use the dotted line to make the styling more consistent? The today button doesn't currently have a focus style. Can the same focus style be added by making an inset border?

Screen Shot 2020-02-28 at 5 07 35 PM

packages/terra-date-picker/CHANGELOG.md
saket2403: Elaborated here f91e565

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
saket2403: Applied here 251400f

packages/terra-date-picker/src/react-datepicker/calendar.jsx
saket2403: Fixed 251400f

saket2403:
>  Is it possible to also update the focus style for these elements to use the dotted line to make the styling more consistent? The today button doesn't currently have a focus style. Can the same focus style be added by making an inset border?

Yes, applied focus styles 251400f

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
benbcai: We would probably need to flex this for the selected day. We do want to be able to apply the dashed outline to the selected day when it is focused but the styles should be removed once it focus is lost. Currently, if there is a selected day, the focus style is always there and doesn't get removed when the focus is moved to a different day.

Screen Shot 2020-03-02 at 2 56 07 PM

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
```diff
@@ -368,7 +358,8 @@
 
     &:focus {
       background-color: #e4e4e4;
-      outline: none;
+      outline: 2px inset #000;
```
benbcai: Should this have the `dashed` style as well?

jeremyfuksa: For consistency's sake, yes.

packages/terra-date-picker/src/react-datepicker/index.jsx
```diff
@@ -593,34 +601,42 @@ class DatePicker extends React.Component {
       let newSelection
       switch (eventKey) {
         case 'ArrowLeft':
+          this.setState({ calendarOpenedViaKeyboard: true })
```
benbcai: We should rename `calendarOpenedViaKeyboard` to something that better represents what its purpose it. It can cause confusion to set this to true in `handleCalendarKeyDown` to indicate the calendar was opened using a keyboard when it may have been opened by clicking on the calendar button.

The way I understand how the focus should work (after talking to @neilpfeiffer) is that the preselected day should show the focus style whenever the user interacts with the picker using the keyboard. Whenever the user interacts by clicking, then the focus on the preselected day should go away. So it is not just at the time when the picker is launched. Therefore, we should name this something along the line of `isCalendarKeyboardFocused` similar to what's in day.jsx.

packages/terra-date-picker/src/react-datepicker/calendar.jsx
```diff
@@ -363,16 +369,16 @@ export default class Calendar extends React.Component {
 
   increaseMonth = () => {
     this.setState({
-      date: addMonths(cloneDate(this.state.date), 1)
+      date: addMonths(cloneDate(this.state.date), 1).startOf('month')
     }, () => this.handleMonthChange(this.state.date))
-    this.props.setPreSelection(addMonths(cloneDate(this.state.date), 1));
+    this.props.setPreSelection(addMonths(cloneDate(this.state.date), 1).startOf('month'));
```
benbcai: `increaseMonth` is called when the next Month button is clicked. Whenever the user interacts using the mouse (instead of keyboard) the focus on the preselected day should not display. See this [comment](https://github.com/cerner/terra-framework/pull/994#discussion_r386692700). So should we update `this.state.calendarIsKeyboardFocused` to false or at least `this.props.calendarOpenedViaKeyboard` should be false so that the preselected day should not get the focus?

packages/terra-date-time-picker/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* Updated wdio screenshots due to changes in react-datepicker colors.
+
+4.27.0 - (February 4, 2020)
```
benbcai: Looks like this line should not be here.

packages/terra-date-time-picker/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* Updated wdio screenshots due to changes in react-datepicker colors.
```
benbcai:
```suggestion
* Updated wdio screenshots due to focus style changes in react-datepicker.
```

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
saket2403: Changed e4930e6

packages/terra-date-picker/src/react-datepicker/index.jsx
saket2403: Changed e4930e6

packages/terra-date-time-picker/CHANGELOG.md
saket2403: Oops, removed e4930e6

packages/terra-date-time-picker/CHANGELOG.md
saket2403: Changed e4930e6

packages/terra-date-picker/src/react-datepicker/calendar.jsx
saket2403: Changed e4930e6

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
benbcai: @saket2403 I'm still seeing the selected day keeping the dashed border when moving to a different preselected day.

benbcai:
@jeremyfuksa @saket2403 I'm seeing some inconsistent behaviors around how the preselected day gets the dash border. This is my understanding on how it should behave and correct me if I'm wrong.
- Whenever clicking (using a mouse) on the previous/next month buttons or month/year dropdown, there should _not_ be a dash border on the preselected day because I'm using the mouse to interact with the picker.
- Whenever I'm using the keyboard to interact with the picker, then the dash border should show on the preselected day. 

  - For example, if focus is on the next month button and I hit the `return` key on the keyboard, there should be a dash border on the first of the month. If I then click on the next month button, the dash border should no longer show since I'm using a mouse. Hitting the `return` key again should bring back the dash border.

The way that it currently behaves is that when the picker launches and there is no preselected day, no matter if I use the mouse to click on the previous/next month or pressing the return key, the preselected day on the first of the month does not have the dash border.
I then tab to the calendar body and use the arrow keys to preselect a day. Now that there is a preselected day and has the dash border, no matter if I use click on the previous/next month or pressing the return key, the preselected day always shows the dash border. I would expect that it should show/hide based on mouse click or keyboard interaction.

cc/ @neilpfeiffer

packages/terra-date-picker/src/react-datepicker/index.jsx
```diff
@@ -562,6 +565,12 @@ class DatePicker extends React.Component {
     }
   }
 
+  onInputKeyDown = (event) => {
+    if(event.key === 'Enter') {
+      this.setState({ isCalendarOpenedViaKeyboard: true })
```
benbcai: If the picker was launched using the `Enter` key once, `this.state.isCalendarOpenedViaKeyboard` always remains true even when you click to open it. As a result, the preselected day will get the dash border even when you open it via clicking on the calendar button.

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
benbcai: @saket2403 Are we able to get the dash border with inset? This looks more like a dash divider than a border around the Today button. If we can't do that with the `outline` style, perhaps try using an inset `box-shadow`?

@jeremyfuksa Any suggestions?

Screen Shot 2020-03-04 at 12 09 03 PM

benbcai: Looks like you can't combine an outline style of `inset` and `dashed`. I added `outline-offset: -3px` and that seems to make an inset dash border. An offset of 2px seems not enough but you can test it out.

Screen Shot 2020-03-04 at 12 30 30 PM

packages/terra-date-picker/src/react-datepicker/index.jsx
saket2403: Fixed cbabd6c

saket2403:
> @jeremyfuksa @saket2403 I'm seeing some inconsistent behaviors around how the preselected day gets the dash border. 

Changes cbabd6c

packages/terra-date-picker/src/react-datepicker/index.jsx
```diff
@@ -394,6 +395,7 @@ class DatePicker extends React.Component {
   }
 
   handleOnRequestClose() {
+    this.setState({ isCalendarKeyboardFocused: false, isCalendarOpenedViaKeyboard: true });
```
benbcai: Should isCalendarOpenedViaKeyboard be set to false here? When you launch the picker using the keyboard, close it and the open it with a mouse click, the preselected day still has the dash border when it should not.

saket2403: Fixed 72febfc

benbcai:
@ShettyAkarsh @saket2403 Two of my previous comments probably got buried since they are not yet addressed. Let me know if you're still working on addressing these or if you have questions about them.

- https://github.com/cerner/terra-framework/pull/994#discussion_r386645515
- https://github.com/cerner/terra-framework/pull/994#discussion_r387844878

saket2403:
> @ShettyAkarsh @saket2403 Two of my previous comments probably got buried since they are not yet addressed. Let me know if you're still working on addressing these or if you have questions about them.
> 
> * [#994 (comment)](https://github.com/cerner/terra-framework/pull/994#discussion_r386645515)
> * [#994 (comment)](https://github.com/cerner/terra-framework/pull/994#discussion_r387844878)

Ooops, Looks like I missed those, Changed here 3c6324f

benbcai:
@saket2403 I'm still seeing some inconsistency with the preselected date getting focus.
- **Scenario 1** - Click to open the picker. Tab thru each element. Notice that when the focus gets to the calendar that the focus is not on the preselected day (should be the current day in this case)

date-picker1

- **Scenario 2** - Click to open the picker. Use the arrow key to preselect any day. Tab thru each element. When the calendar loses focus and the next element gain focus, the preselected day should no longer have the dash border focus. This looks as if there are two elements having focus at the same time.

date-picker2

- **Scenario 3** - There was probably some confusion on my part when I explained how the focus works for the preselect day when navigating to a different month. When navigating to a different month, regardless by pressing the Enter key or clicking on the Next/Previous month button or selecting an option from the month/year dropdown. The preselected day on the first of the month should not show the focus immediately when the month is rendered. The focus on the preselected day should show _only_ when the calendar receives focus (by tabbing to it).

date-picker3

packages/terra-date-picker/src/react-datepicker/calendar.jsx
```diff
@@ -361,18 +388,26 @@ export default class Calendar extends React.Component {
...
-  decreaseMonth = () => {
+  decreaseMonth = (event) => {
     this.setState({
-      date: subtractMonths(cloneDate(this.state.date), 1)
+      isMonthChanged: true,
+      date: subtractMonths(cloneDate(this.state.date), 1).startOf('month')
     }, () => this.handleMonthChange(this.state.date))
-    this.props.setPreSelection(subtractMonths(cloneDate(this.state.date), 1));
+    this.props.setPreSelection(subtractMonths(cloneDate(this.state.date), 1).startOf('month'));
+    if(event.target.type === undefined) {
```
benbcai: Is this check for determining whether the button was clicked or hitting the Enter key? Can you add a brief comment what this does as it wasn't obvious at first glance? Same in `increaseMonth`.

saket2403: Added c436d1a

saket2403:
> @saket2403 I'm still seeing some inconsistency with the preselected date getting focus.

Fixed c436d1a

PaulSolDev:
**Lost Tab Focus:**

- When I tab past the day selection focus in the calendar, and I reverse tab (shift + tab) back to day selection, the focus outline is not shown.  Focus style returns when I use the arrow keys to navigate within the day/calendar view.

benbcai:
@PaulSolDev 

- For your 2nd bullet point, have your already navigated to a different month when you use the arrow keys to navigate to a different day? If you haven't moved to a different month and are still on the same month that was the initially opened, the preselected day would be based on the current day. If there is no default date and you initially open the picker, should the preselected day be the today's day or should it be the 1st of the month?

- For your 3rd bullet point, I think it makes sense to put the preselected day back on the selected day (if there is one) instead of the 1st of the month.

@saket2403 I noticed that when you open the picker and start to hover over the calendar, the preselected day will automatically show the focus on the preselected day.

date-picker4

PaulSolDev:
@benbcai although I was able to consistently produce the Firefox issue described in the second bullet comment (text shown again below) I am unable to get the same results today.  

- (Firefox) When I reverse tab (shift + tab) back to the day selection the focus outline is not shown. When I navigate using the arrows the focused date appears on a different date (does not appear consistent). **Can't reproduce today.**

**New Firefox issue:**
-(Firefox) focus border not shown on in calendar/day view with both standard and reverse tab order.  In both scenarios the focus border reappears when navigation continues using arrow keys.

benbcai:
> @benbcai although I was able to consistently produce the Firefox issue described in the second bullet comment (text shown again below) I am unable to get the same results today.
> 
> * (Firefox) When I reverse tab (shift + tab) back to the day selection the focus outline is not shown. When I navigate using the arrows the focused date appears on a different date (does not appear consistent). **Can't reproduce today.**
> 
> **New Firefox issue:**
> -(Firefox) focus border not shown on in calendar/day view with both standard and reverse tab order. In both scenarios the focus border reappears when navigation continues using arrow keys.

@PaulSolDev @saket2403 Looks like the 'New Firefox issue' issue you mentioned above is not just in Firefox. It is also in Chrome. I think it is this [change](https://github.com/cerner/terra-framework/pull/994/commits/495c64129c3188f46c83dc11b822a24edc67c059) that is causing this.

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
ryanthemanuel: Was this question ever answered?

packages/terra-date-picker/src/react-datepicker/calendar.jsx
```diff
@@ -1,5 +1,6 @@
 import { FormattedMessage, intlShape } from 'react-intl';
 import Button from 'terra-button';
+import moment from 'moment';
```
ryanthemanuel: Where is this being used?

packages/terra-date-picker/src/react-datepicker/index.jsx
```diff
@@ -8,6 +8,7 @@ import Popup from 'terra-popup';
 import classNames from 'classnames/bind';
 import { injectIntl, intlShape } from 'react-intl';
 import VisuallyHiddenText from 'terra-visually-hidden-text';
+import moment from 'moment';
```
ryanthemanuel: Where is this being used?

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
benbcai: @saket2403 Is it correct that this change is no longer needed given the new dashed focus style from UX?

packages/terra-date-picker/src/react-datepicker/calendar.jsx
benbcai: This is being used by `startOf` [here](https://github.com/cerner/terra-framework/pull/994/files#diff-f0779fb2b55a5a8b2bef87c374f19eceR396).

@saket2403 Would you be able to use `getStartOfMonth` from [date_util](https://github.com/cerner/terra-framework/blob/ed759f73d487259b057b34818cc3b352b1098e3f/packages/terra-date-picker/src/react-datepicker/date_utils.js#L148)?

packages/terra-date-picker/src/react-datepicker/index.jsx
benbcai: Looks like this is no longer needed.

jeremyfuksa:
> * (Firefox) The day/calendar view appears to have a double focus border
> 
> image

Still seeing this, but when invoking focus styles using the inspector, the double border doesn't show up. Are we dealing with peculiarities in native focus rendering?

Aside from the "new Firefox issue," stylistically this is getting super close. 

I did find an accessibility problem when I tested in IE 11 with JAWS turned on. JAWS instructs the user to press the spacebar to select, but enter/return is the key that needs to be pressed.

packages/terra-date-picker/src/react-datepicker/index.jsx
saket2403: Removed 76967e7

packages/terra-date-picker/src/react-datepicker/calendar.jsx
saket2403: Changed 76967e7

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
saket2403: Yes I have reverted it 76967e7

saket2403:
Fixed Firefox Double border issue and reverse tabbing issue 76967e7


> Tab border cut off in IE on today option:
> image

For the above issue in IE, `outline-offset` seems to be not working in IE, is there any alternative that can be used @jeremyfuksa ?

benbcai:
@jeremyfuksa Is the JAWS issue in IE 11 broken with this change or was it always like that even without this change?

jeremyfuksa:
I'm going to assume it was always there. This was the first time I looked at it with JAWS on.

benbcai:
@saket2403 
- Not sure if this [recent commit](https://github.com/cerner/terra-framework/pull/994/commits/000fd7daba12a76c22748cc1fe5416ceeeead26f) might have this jumpy behavior when the today buttons gets focus. 
- When hitting enter on the Previous month, the focus on the 1st day of the month becomes visible.
- I don't see we have any blue focus styles added for the month/year dropdown when they are in focus. Not sure if that is a native Firefox behavior.

date-picker5

saket2403:
> @saket2403
> 
> * Not sure if this [recent commit](https://github.com/cerner/terra-framework/pull/994/commits/000fd7daba12a76c22748cc1fe5416ceeeead26f) might have this jumpy behavior when the today buttons gets focus.

The jumpy behaviour is because the button slightly resizes for the border to fit in which was getting cut off in IE.

> * When hitting enter on the Previous month, the focus on the 1st day of the month becomes visible.

Fixed

> * I don't see we have any blue focus styles added for the month/year dropdown when they are in focus. Not sure if that is a native Firefox behavior.

Yes the blue focus glow is indeed browser specific in Firefox.
Any css-workaround for this or we can leave it? @jeremyfuksa

supreethmr:
@saket2403  I believe that jumping behaviour ( IE browser) on Today button could be due to the border that is being added on focus [here](https://github.com/cerner/terra-framework/blob/1bd124186d839a664ae76d336b1391672cecab51/packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss#L361) . `today` button border has been [set to 0 initially](https://github.com/cerner/terra-framework/blob/1bd124186d839a664ae76d336b1391672cecab51/packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss#L349) and 2px border is being added on Focus hence that border transition causing the jump behaviour. 

So adding default border and having color as transparent to today button might resolve this IE issue. 

I found similar behaviour when we focus on next month button as well. I suspect this could be the reason for [this](https://github.com/cerner/terra-framework/blob/1bd124186d839a664ae76d336b1391672cecab51/packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss#L447) behaviour as well.

benbcai:
> @saket2403 I believe that jumping behaviour ( IE browser) on Today button could be due to the border that is being added on focus [here](https://github.com/cerner/terra-framework/blob/1bd124186d839a664ae76d336b1391672cecab51/packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss#L361) . `today` button border has been [set to 0 initially](https://github.com/cerner/terra-framework/blob/1bd124186d839a664ae76d336b1391672cecab51/packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss#L349) and 2px border is being added on Focus hence that border transition causing the jump behaviour.
> 
> So adding default border same as border added in focus state and having color as transparent to today button might resolve this IE issue.
> 
> I found similar behaviour when we focus on next month button as well. I suspect this could be the reason for [this](https://github.com/cerner/terra-framework/blob/1bd124186d839a664ae76d336b1391672cecab51/packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss#L447) behaviour as well.

@saket2403 @supreethmr I'm seeing the jumpy behavior in all browsers, not just IE.

supreethmr:
> > @saket2403 I believe that jumping behaviour ( IE browser) on Today button could be due to the border that is being added on focus [here](https://github.com/cerner/terra-framework/blob/1bd124186d839a664ae76d336b1391672cecab51/packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss#L361) . `today` button border has been [set to 0 initially](https://github.com/cerner/terra-framework/blob/1bd124186d839a664ae76d336b1391672cecab51/packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss#L349) and 2px border is being added on Focus hence that border transition causing the jump behaviour.
> > So adding default border same as border added in focus state and having color as transparent to today button might resolve this IE issue.
> > I found similar behaviour when we focus on next month button as well. I suspect this could be the reason for [this](https://github.com/cerner/terra-framework/blob/1bd124186d839a664ae76d336b1391672cecab51/packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss#L447) behaviour as well.
> 
> @saket2403 @supreethmr I'm seeing the jumpy behavior in all browsers, not just IE.

Oh, Yes I missed that. this is happening on all browsers. I still feel it is because of changing border styles on focus.

saket2403:
> Oh, Yes I missed that. this is happening on all browsers. I still believe it is because of changing border styles on focus.

It worked Fixed here 6797a7f

lokesh-0813:
> > Oh, Yes I missed that. this is happening on all browsers. I still believe it is because of changing border styles on focus.
> 
> It worked Fixed here [6797a7f](https://github.com/cerner/terra-framework/commit/6797a7fabac0b1362c304632714d3c4ecb429013). cc @benbcai @supreethmr

From 6797a7fabac0b1362c304632714d3c4ecb429013. Shouldn't all focus be of same weight? I find `Today` button much bolder comparitively.

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
```diff
@@ -488,7 +496,27 @@
...
+    &--keyboard-focus {
+      outline: var(--terra-date-picker-day-focus-outline, 2px dashed #000);
+      outline-offset: var(--terra-date-picker-day-month-focus-outline-offset, 0);
+    }
+  }
+
+  .react-datepicker-day {
+    &--selected-border {
+      outline: var(--terra-date-picker-day-focus-outline, 2px dashed #000);
+      outline-offset: var(--terra-date-picker-day-month-focus-outline-offset, 0);
```
benbcai: Should `month` be removed from `--terra-date-picker-day-month-focus-outline-offset` to be consistent with `--terra-date-picker-day-focus-outline`? Same on line 501.

benbcai:
> > > Oh, Yes I missed that. this is happening on all browsers. I still believe it is because of changing border styles on focus.
> > 
> > 
> > It worked Fixed here [6797a7f](https://github.com/cerner/terra-framework/commit/6797a7fabac0b1362c304632714d3c4ecb429013). cc @benbcai @supreethmr
> 
> From [6797a7f](https://github.com/cerner/terra-framework/commit/6797a7fabac0b1362c304632714d3c4ecb429013). Shouldn't all focus be of same weight? I find `Today` button much bolder comparitively.

@lokesh-0813 Are you referring to the dashed focus style or the top border on the today button . The dashed focus style looks consistent to me on all browsers.

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
saket2403: Fixed 6797a7f

lokesh-0813:
> @lokesh-0813 Are you referring to the dashed focus style or the top border on the today button . The dashed focus style looks consistent to me on all browsers.

Yes the dashed focus style @benbcai. My browser(Chrome) was set at 90% zoom. It looks fine at 100%.

packages/terra-date-picker/CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ Unreleased
 ### Changed
 * Changed `onChange` and `onChangeRaw` callbacks to return the following metadata as part of the third parameter: iso, inputValue, isCompleteValue, isValidValue
 * Update the placeholder for the `es-US`, `nl`, `sv-SE`, and `sv` locales to match the actual date format.
+* Applied new border styles for keyboard navigation.
+* Changed pre-selected date to 1st of month when changing year/month.
```
avinashg1994: Move these changelog entries to the unreleased section

packages/terra-date-time-picker/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Unreleased
 4.32.0 - (March 24, 2020)
 ------------------
 ### Changed
+* Updated wdio screenshots due to focus style changes in react-datepicker.
```
avinashg1994: Move this entry to the unreleased section

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
```diff
@@ -488,7 +496,27 @@
   }
 
   .is-calendar-focused {
-    outline: var(--terra-date-picker-month-focus-outline, 2px dashed #000);
-    outline-offset: var(--terra-date-picker-month-focus-outline-offset, 0);
```
avinashg1994: The removal/renaming of this theme variable should be documented in the changelog

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
```diff
@@ -488,7 +496,27 @@
...
+    &--keyboard-focus {
+      outline: var(--terra-date-picker-day-focus-outline, 2px dashed #000);
+      outline-offset: var(--terra-date-picker-day-focus-outline-offset, 0);
+    }
+  }
+
+  .react-datepicker-day {
+    &--selected-border {
+      outline: var(--terra-date-picker-day-focus-outline, 2px dashed #000);
+      outline-offset: var(--terra-date-picker-day-focus-outline-offset, 0);
```
avinashg1994: Should these theme variables be provided with values for fusion or lowlight themes?

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
```diff
@@ -356,8 +346,9 @@
 
   .react-datepicker-today-button {
     background: $datepicker-background-color;
-    border: 0;
-    border-top: 1px solid $datepicker-border-color;
+    border: 2px dashed transparent;
+    border-top: 2px solid $datepicker-border-color;
```
avinashg1994: Was changing the `border-top` to 2px an intentional change?
image

Because in doing so, the border for the month body becomes inconsistent, i.e. the border above the month body is 1px and the border below the month body is 2px.

packages/terra-date-picker/CHANGELOG.md
saket2403: Moved 64aef1f

packages/terra-date-time-picker/CHANGELOG.md
saket2403: Moved 64aef1f

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
saket2403: Added 64aef1f

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
saket2403: I did not see the values for the previous variable also which was renamed in the theme files.

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
saket2403: Yes this was an intentional change to fix the jump issue as suggested in this [comment](https://github.com/cerner/terra-framework/pull/994#issuecomment-606011322)

avinashg1994: In that case, we should consider changing the top border of the month body as well, to keep the styling consistent.

benbcai: @jeremyfuksa What are you thoughts on this. Should the weight on the divider between the days of week and the month match the one below?

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
benbcai: @jeremyfuksa Now that we have dashed borders for the focus style, should the fusion and lowlight themes use different colors for the border?

jeremyfuksa: I had to dig around Zeplin to find examples of what the focus style should be.

I didn't find any examples for Fusion, so I'm going to make the call that it's OK to have focus color on open source and Fusion to be the same. Low Light shows that the border color is the same as its text color - #1078 B7B7B8

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
jeremyfuksa: Let's make them consistent.

jeremyfuksa:
> Yes the blue focus glow is indeed browser specific in Firefox.
> Any css-workaround for this or we can leave it? @jeremyfuksa

There's always [:-moz-focusring](https://developer.mozilla.org/en-US/docs/Web/CSS/:-moz-focusring).

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
saket2403: Done d882aad

saket2403:
> :-moz-focusring
I had tried it initially but the glow is still there.

packages/terra-date-picker/src/react-datepicker/stylesheets/react_datepicker.module.scss
saket2403: I am seeing this [comment](https://github.com/cerner/terra-framework/blob/66db72147831e2093b7a876a9db3e1b9d3d42276/packages/terra-date-picker/themes/clinical-lowlight-theme/clinical-lowlight-theme.scss#L29) in the theme file so added the day outline variable in this list.

PaulSolDev:
Functional review complete.  The only variance that i noted was that in IE when using the arrows on the month and year dropdown options the month/year change but the dropdown menu doesn't open.  I don't see any harm from this since the control is still allowing keyboard control of the component but I did want to point it out as it was a different experience in this browser.

jeremyfuksa:
> > :-moz-focusring
> I had tried it initially but the glow is still there.

I was able to use `:-moz-focusring` to get rid of the glow, but it did far worse things to the focus styling. So, let's leave it alone for now.

PR #999, opened 2020-01-10
[terra-infinite-list] ariaLabel prop added

### LINKED ISSUE #997: [terra-infinite-list] Missing aria-label prop for infinite list
# Bug Report

## Description

Currently, there isn't any prop available to set the aria-label for the Infinite list. This causes an accessibility violation when the Infinite List is used as `listbox`. 
## Steps to Reproduce

1. Create an Infinite List with role set to **lisbox**.
2. Analyzing the component through axe plugin on chrome should throw an accessibility violation
`Ensures every ARIA input field has an accessible name `

## Expected Behavior

There should be no accessibility violations.
## Possible Solution

Expose ariaLabel prop to set the aria-label, similar to other Terra components.
## Environment

* Component Name and Version: terra-infinite-list  version-3.0.0
* Browser Name and Version: Chrome version-79.0.3945.88
* Node/npm Version: 8.9.2
* Webpack Version: 4.41.2

## @ Mentions

@dkasper-was-taken

avinashg1994:
### Summary
Fixed accessibility issue when role was specified as "listbox", by adding `aria-label` prop.

Resolves #997


packages/terra-infinite-list/src/InfiniteList.jsx
```diff
@@ -11,6 +11,10 @@ import styles from './InfiniteList.module.scss';
 const cx = classNames.bind(styles);
 
 const propTypes = {
+  /**
+   * String that labels the modal for screen readers.
+   */
+  ariaLabel: PropTypes.string.isRequired,
```
emilyrohrbough: making this required makes this a MVB. @ryanthemanuel do we want to make this a MVB or just add the prop so teams have the ability to make this accessible?

mjhenkes: Keep this minor.

packages/terra-infinite-list/src/terra-dev-site/test/infinite-list/InfiniteListListbox.test.jsx
```diff
@@ -0,0 +1,23 @@
...
+import styles from './InfiniteListTestCommon.module.scss';
+
+const cx = classNames.bind(styles);
+
+const InfiniteListListbox = () => (
+  <InfiniteList
+    id="test-infinite-list"
+    isFinishedLoading
+    role="listbox"
+    ariaLabel="Listbox"
```
emilyrohrbough: If this is fixing accessibility, should all infinite list examples be updated to use it? plus this was made a required prop; are the other examples throwing console errors for missing required prop?

avinashg1994: Yes, since `ariaLabel` is a required prop, the other examples were throwing console errors.
I have addressed that here : https://github.com/cerner/terra-framework/pull/999/commits/a1ebbff90d1499551ed4e6a5432370509814edaf

packages/terra-infinite-list/src/InfiniteList.jsx
supreethmr: Curious to know why do we need this prop for adding aria-label to infinite-list.
When we can directly add 'aria-label' as custom prop to `infinite-list`  like : 
```javascript
  <InfiniteList
    role="listbox"
    aria-label="label"
  >
```
by Updating [List](https://github.com/cerner/terra-framework/blob/d660b07885ef6d8389390c00f057b4194f9ea135/packages/terra-infinite-list/src/InfiniteList.jsx#L525) tag in render of InfiniteList.jsx to have `{...customProps} ` like : 
```Javascript
<List {...customProps} {...attrSpread} refCallback={refCallback}>
```

ryanthemanuel: By making it a prop we can make it required.

ryanthemanuel: We like to call out props that are more first class concepts (even if they can be passed in).  It makes the API a little more clearer.  In the future, we could also make this required if we wanted.

avinashg1994:
Updated `ariaLabel` to be a non-required prop.
Also, Logged an issue to update the prop to be Required in the next MVB : https://github.com/cerner/terra-framework/issues/1003

PaulSolDev:
Functional review complete.  No issues identified.

PaulSolDev:

Issue #468, opened 2019-01-29
[terra-date-picker] Automatically add a separator when typing in a date


hs025828:
# Issue Description
It would be helpful if the date picker automatically inserted the separator character as the user is typing the date.

## Issue Type

- [x] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The separator character should automatically be inserted by the date picker as the user enters the date.

Here is an example in `en-US` locale:  After the month is entered, the forward slash should automatically be inserted and the cursor should be moved past it ready for the user to input the day.  Once the day has been entered, then a forward slash should automatically be inserted and the cursor moved past it ready for the user to input the year.

## Current Behavior
The separator character must be manually typed by the user.

## Steps to Reproduce

1. Navigate to the date picker examples.
2. Try entering a date manually into the date picker.
3. Observe no separators are automatically inserted.

SharathAlur:
Even the date pickers accepts string. Can we mask the field and accept only numbers?

benbcai:
The changes made on #595 now limits the date input field to accept only numeric values and the "/" and "." separators (depending on the locale).

nickpith:
From leadership sync today, we are going to split the day, month and year into 3 separate inputs instead of a single one and require the user to type the "/".

cc: @amichaelparker, @neilpfeiffer

rockhold:
@ryanthemanuel / @avinashg1994, I was just talking with a solution designer about the timeline for this change. She mentioned that Ryan had indicated it would be available in Q3. I'm curious if this change is expected to be passive. We have some components we'd like to get it pulled into for Q3, but without knowing what it might take to uplift, it's hard to know whether this would be possible. Any info you can provide here would be appreciated. Thanks.

avinashg1994:
@rockhold, the intention is to keep things passive, but that being said this is still in progress and splitting the date input requires quite a lot of changes.
cc: @ryanthemanuel

saigorantla:
Will internationalization be supported? Not every county uses '/' in their date format, few countries use '-' and few use '.'  [WIKI](https://en.wikipedia.org/wiki/Date_format_by_country). How will the new date picker work in that case?

avinashg1994:
@saigorantla,
Yes, Internationalisation will be supported.
[...]

PR #1004, opened 2020-01-16
[terra-date-picker] Split Date Input

amichaelparker:
### Summary
Ref - [#468](https://github.com/cerner/terra-framework/issues/468). This splits date input into 3 separate inputs, and moves between inputs when each input is the correct length.

Closes [#468](https://github.com/cerner/terra-framework/issues/468).


packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -13,6 +13,21 @@ const cx = classNames.bind(styles);
...
 
+const dateFormats = [
+  'س‌س‌س‌س/ي‌ي/ش‌ش',
+  'TT.MM.JJJJ',
+  'DD/MM/YYYY',
+  'DD/MM/AAAA',
+  'DD.MM.YYYY',
+  'JJ/MM/AAAA',
+  'DD/MM/JJJJ',
+  'DD/MM/ÅÅÅÅ',
```
benbcai: Can we just get these date formats from the [translations files](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/translations/sv.json#L2)?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -180,27 +314,69 @@ class DatePickerInput extends React.Component {
...
+            type="hidden"
+            name={name}
+            value={dateValue}
+          />
+          <Input
+            {...additionalInputProps}
+            className={cx('date-input-month')}
+            type="text"
+            name={'terra-date-month-'.concat(name)}
+            value={this.state.month}
```
benbcai: The date format for our supported locales can be in any of these formats:
`MM/DD/YYYY`
`DD/MM/YYYY`
`YYYY/MM/DD`

The ordering of these inputs would need to flex based on the locale's date format instead of always being MM/DD/YYYY

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -180,27 +314,69 @@ class DatePickerInput extends React.Component {
...
+            onChange={this.handleMonthChange}
+            placeholder={dateMonth || 'MM'}
+            onFocus={this.handleOnFocus}
+            onBlur={this.handleOnBlur}
+            maxLength="2"
+            size="2"
+            pattern="\d*"
+            aria-label={intl.formatMessage({ id: 'Terra.datePicker.monthLabel' })}
+          />
+          <span className={cx('date-spacer')}>/</span>
```
benbcai: The date separator for all locales is a `/` with the exception of us-CA there the values are separated by `-`.

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Now that I look over it, this became unnecessary after some other changes, so this can be removed.

amichaelparker: Removed! [847d4f0](https://github.com/cerner/terra-framework/pull/1004/commits/847d4f05fecc8074996a2d504e3cc216973b6eef)

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Changed! [847d4f0](https://github.com/cerner/terra-framework/pull/1004/commits/847d4f05fecc8074996a2d504e3cc216973b6eef)

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Added: [847d4f0](https://github.com/cerner/terra-framework/pull/1004/commits/847d4f05fecc8074996a2d504e3cc216973b6eef). It's rather extensive considering the various changes down the line needed to support it, so I'm currently looking into moving some stuff into a separate file to clean things up a bit.

benbcai: +1 for composing each of these inputs in a separate function.

One option is we could add a new attribute (i.e. `Terra.datePicker.dateFormatOrder`)  in each of the translation files to indicate the ordering of the date format. For examples, for MM/DD/YYYY it would be set to 1, for DD/MM/YYYY -> 2, and YYYY/MM/DD -> 3.

It can then call these functions to compose the inputs in the order based on the new attribute.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -104,9 +108,27 @@ class DatePickerInput extends React.Component {
   constructor(props) {
     super(props);
 
+    this.state = {};
+
+    if (this.props.value) {
+      this.state = {
+        day: this.props.value.substr(3, 2),
+        month: this.props.value.substr(0, 2),
```
benbcai: We can't guarantee that the first two digits are the month due to the different formats.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -115,6 +137,44 @@ class DatePickerInput extends React.Component {
...
+        year: props.value.substr(0, 4),
+        prevDate: props.value,
+      };
+    }
+    return null;
+  }
+
+  moveFocusOnChange(inputValue, type, dateVariant) {
+    // Move focus based on input selected and date variant.
+    if (inputValue.length === 2) {
```
benbcai: Not sure if we want to make the behavior for the auto focus while entering value to match the time-input. For the time-input if you enter the first value between 3-9 for the hour, it will prepend a 0 and then auto focus to the next input. It does not require two digits since an hour that starts with 3 to 9 is not acceptable. See https://github.com/cerner/terra-framework/blob/master/packages/terra-time-input/src/TimeInput.jsx#L335-L351

cc/ @neilpfeiffer

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -231,22 +231,21 @@ class DatePicker extends React.Component {
...
 
-        if (this.dateValue === '' || (isCompleteDate && this.isDateWithinRange(DateUtil.createSafeDate(iSOString)))) {
+        if ((isCompleteDate && this.isDateWithinRange(DateUtil.createSafeDate(iSOString)))) {
           isValidDate = true;
         }
 
         const options = {
           iSO: iSOString,
-          inputValue: this.dateValue,
+          inputValue: iSOString,
```
benbcai: What's the reason for changing this to be the ISO format, which is a duplicate value returned in the `iSO` key? This used to be what ever value the user entered. This would break consumers since it will no longer be what they expect.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -311,9 +311,10 @@ class DateTimePicker extends React.Component {
         isValid = true;
       }
 
+      const inputDateTime = `${iSOString.substr(0, 10)} ${this.timeValue}`;
       const options = {
         iSO: iSOString,
-        inputValue: value,
+        inputValue: inputDateTime,
```
benbcai: This used to be what ever value the user entered. This would break consumers since it will now be in the iSO format.

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Not having the auto focus behavior would actually make a lot of this significantly easier, and I could trim out quite a bit. It makes sense not to have it with regard to date formats, but part of the original issue was about inserting slashes, so I tried to implement it here. @neilpfeiffer if it isn't needed, let me know!

packages/terra-date-time-picker/src/DateTimePicker.jsx
amichaelparker: This and the above are temporary. I'm currently working around the formatting issues, as well as composing the total input value for use in these functions (as it stands, it really only recognizes the year input).

packages/terra-date-picker/src/DateInput.jsx
benbcai: Sorry for the confusion on my previous comment. What I intended to say is I think we do want the auto focus but I wonder if we want to auto prepend a 0 to the entered value similar to how the time-input works. 

For example,  if the user enters "3" for the month, this would imply that the month is "03" since a two digit month that starts with "3" is invalid. In this [example](https://engineering.cerner.com/terra-framework/tests/terra-time-input/time-input/time-input/default), as soon as you enter a value from 3-9, it will prepend a 0 and auto focus to the next field since you can't enter an hour greater than 24.

amichaelparker: Understood! Yeah, if that is desired, I can definitely throw that in.

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Updated: [452f10b](https://github.com/cerner/terra-framework/pull/1004/commits/452f10bd2fac12ea3c50aff9d7b90d0cff1d3b01). I rearranged things and composed the varying layouts better. I'm currently looking into doing this via the suggested attribute.

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Updated: [452f10b](https://github.com/cerner/terra-framework/pull/1004/commits/452f10bd2fac12ea3c50aff9d7b90d0cff1d3b01). Since the render changes based on the format, I formatted whatever the initial input value is to something consistent.

packages/terra-date-picker/src/DatePicker.jsx
amichaelparker: Changed back! [452f10b](https://github.com/cerner/terra-framework/pull/1004/commits/452f10bd2fac12ea3c50aff9d7b90d0cff1d3b01)

packages/terra-date-time-picker/src/DateTimePicker.jsx
amichaelparker: Changed back! [452f10b](https://github.com/cerner/terra-framework/pull/1004/commits/452f10bd2fac12ea3c50aff9d7b90d0cff1d3b01)

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Made some updated to this logic that... actually make sense.

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Added in the attribute to make formatting decisions easier to reason about.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -104,9 +103,29 @@ class DatePickerInput extends React.Component {
   constructor(props) {
     super(props);
 
+    const format = DateUtil.getDateFormat(props.intl.formatMessage({ id: 'Terra.datePicker.dateFormatOrder' }));
```
nramamurth: Linter rule on Line 101 can now be removed.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -104,9 +103,29 @@ class DatePickerInput extends React.Component {
...
+    const tempDate = DateUtil.convertToISO8601(props.value, format);
+
+    this.state = {};
+
+    if (tempDate) {
+      this.state = {
+        day: tempDate.substr(8, 2),
```
nramamurth: > [`String.prototype.substr`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is considered a legacy function and should be avoided when possible. 

You may want to use [`substring()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) method instead.

nramamurth: I could be wrong but if the return value of [`DateUtil.convertToISO8601()`](https://github.com/cerner/terra-framework/blob/809ece57b4400499b0200f7bd9a34445f886dded/packages/terra-date-picker/src/DateUtil.js#L121-L127) is "invalid", then aren't we parsing and sticking to incorrect values here?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -115,6 +134,48 @@ class DatePickerInput extends React.Component {
...
+        month: tempDate.substr(5, 2),
+        year: tempDate.substr(0, 4),
+        prevDate: props.value,
+      };
+    }
+    return null;
+  }
+
+  moveFocusOnChange(inputValue, type, dateVariant) {
```
nramamurth: Please add documentation for the arguments here so that it's easier for the readers to understand in the future.

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -266,17 +265,19 @@ class DatePicker extends React.Component {
...
-  handleChangeRaw(event) {
+  handleChangeRaw(event, date) {
+    if (date) {
+      this.handleChange(DateUtil.createSafeDate(date), event);
+    }
     this.dateValue = event.target.value;
 
     if (this.props.onChangeRaw) {
-      this.props.onChangeRaw(event, event.target.value);
+      this.props.onChangeRaw(event);
```
nramamurth: Breaking change?

packages/terra-date-picker/translations/ar.json
```diff
@@ -1,5 +1,6 @@
 {
   "Terra.datePicker.dateFormat": "س‌س‌س‌س/ي‌ي/ش‌ش",
   "Terra.datePicker.today": "اليوم",
-  "Terra.datePicker.openCalendar": "فتح التقويم"
+  "Terra.datePicker.openCalendar": "فتح التقويم",
+  "Terra.datePicker.dateFormatOrder": "2"
```
nramamurth: What do "0", "1" and "2" signify? Can they be more meaningful - maybe throw in the actual formats ('MM-DD-YYYY', 'YYYY-MM-DD' and 'DD-MM-YYYY' respectively) in here?

nramamurth:
@amichaelparker 

1. Should the left-arrow and right-arrow keys help users navigate between the three input fields?
2. Is it a UX decision to not let users select the date and maybe clear it off in one go? Now it looks like one has to individually select each of the three small input components to clear the whole date. Also, I see that the styles do not tell me if I have the selection made or not (Chrome & Safari). cc / @neilpfeiffer 
3. I do not see the cursor blinking on Chrome when I press Shift + Tab which makes it harder for me to understand where the current focus is at.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -179,28 +312,89 @@ class DatePickerInput extends React.Component {
...
+        name={'terra-date-day-'.concat(name)}
+        value={this.state.day}
+        onChange={this.handleDayChange}
+        placeholder={placeholderValues.day || 'DD'}
+        onFocus={this.handleOnFocus}
+        onBlur={this.handleOnBlur}
+        maxLength="2"
+        size="2"
+        pattern="\d*"
+        aria-label={intl.formatMessage({ id: 'Terra.datePicker.today' })}
```
nramamurth: Why should it read "today"?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -123,19 +184,81 @@ class DatePickerInput extends React.Component {
...
+      tempDate = DateUtil.formatISODate(year.concat('-', month).concat('-', day), 'YYYY-MM-DD');
+      formattedDate = DateUtil.strictFormatISODate(tempDate, dateFormat);
+    }
+
+    if (day === '' && month === '' && year === '') {
+      this.props.onChange(event, '');
+    }
+
+    if (DateUtil.isValidDate(formattedDate, dateFormat)) {
+      event.target.value = formattedDate; // eslint-disable-line no-param-reassign
```
nramamurth: 😲 This pattern looks strange to me. cc / @benbcai

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -123,19 +184,81 @@ class DatePickerInput extends React.Component {
...
+      this.setState({
+        month: value,
+      }, this.moveFocusOnChange(value, type, this.variant));
+    } else if (type === DateUtil.inputType.YEAR) {
+      this.setState({
+        year: value,
+      }, this.moveFocusOnChange(value, type, this.variant));
+    }
+
+    const day = type === DateUtil.inputType.DAY ? value : this.state.day;
```
nramamurth: Please add a brief code comment for this blob of code (Lines 237 to 261).

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -123,19 +184,81 @@ class DatePickerInput extends React.Component {
...
+    const month = type === DateUtil.inputType.MONTH ? value : this.state.month;
+    const year = type === DateUtil.inputType.YEAR ? value : this.state.year;
+    const dateFormat = DateUtil.getFormatByLocale(this.props.intl.locale);
+
+    let tempDate;
+    let formattedDate;
+
+    if (this.variant === 'YYYY-MM-DD') {
+      if (month && day && year && day.length === 2) {
+        tempDate = DateUtil.formatISODate(year.concat('-', month).concat('-', day), 'YYYY-MM-DD');
```
nramamurth: Logic in Lines 247 and 248 seem redundant with Lines 251 and 252. Can these conditionals be made simpler then?

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Removed! [89c3693](https://github.com/cerner/terra-framework/pull/1004/commits/89c3693f31162adfc53f90f681cdfda21588b6b0)

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Changed! [89c3693](https://github.com/cerner/terra-framework/pull/1004/commits/89c3693f31162adfc53f90f681cdfda21588b6b0)

amichaelparker: This has been mostly removed and adjusted in various ways: [89c3693](https://github.com/cerner/terra-framework/pull/1004/commits/89c3693f31162adfc53f90f681cdfda21588b6b0)

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Added! [89c3693](https://github.com/cerner/terra-framework/pull/1004/commits/89c3693f31162adfc53f90f681cdfda21588b6b0)

packages/terra-date-picker/src/DatePicker.jsx
amichaelparker: No, but it also is unnecessary, so I reverted it: [89c3693](https://github.com/cerner/terra-framework/pull/1004/commits/89c3693f31162adfc53f90f681cdfda21588b6b0)

packages/terra-date-picker/translations/ar.json
amichaelparker: This is probably easier, but I'd be interested to get @benbcai 's input on it.

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: It needs a new translation string, this is just a temp to get things working until the PR is at the point where it is worth sending out for translations.

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: This has been changed significantly: [89c3693](https://github.com/cerner/terra-framework/pull/1004/commits/89c3693f31162adfc53f90f681cdfda21588b6b0)

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Added! [89c3693](https://github.com/cerner/terra-framework/pull/1004/commits/89c3693f31162adfc53f90f681cdfda21588b6b0)

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: I tried to clean this up a little bit! [89c3693](https://github.com/cerner/terra-framework/pull/1004/commits/89c3693f31162adfc53f90f681cdfda21588b6b0)

amichaelparker:
I corrected the issue regarding shift-tabbing in order to allow highlighting, however clearing / arrow key input I'll leave up to @neilpfeiffer 's judgment.

benbcai:
After entering a complete date (e.g.; `12/24/2020`), if I go back to the month input and hit the delete key in an attempt to delete and re-enter the month, it would automatically change the month to `01` and automatically moves the focus to the time input. If I go back to it again, delete the value, and enter the value `1`, it automatically prepends a 0 but my intention is to enter `11`.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -115,6 +121,54 @@ class DatePickerInput extends React.Component {
     }
   }
 
+  static getDerivedStateFromProps(props, state) {
+    // Formats the selected date into a consistent value and sets input state.
+    const format = DateUtil.getDateFormat(props.intl.formatMessage({ id: 'Terra.datePicker.dateFormatOrder' }));
```
benbcai: This can be moved inside the if block so that it is not created unnecessarily.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -115,6 +121,54 @@ class DatePickerInput extends React.Component {
...
+          this.dayInput.focus();
+        } else {
+          this.yearInput.focus();
+        }
+      }
+    } else if (dateVariant === 'DD-MM-YYYY') {
+      if (inputValue.length === 2) {
+        if (type === DateUtil.inputType.MONTH) {
+          this.yearInput.focus();
+        } else if (type === DateUtil.inputType.DAY) {
```
benbcai: Is there a reason why this needs the check for `type === DateUtil.inputType.DAY` but line 151 above doesn't?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -115,6 +121,54 @@ class DatePickerInput extends React.Component {
...
+          this.monthInput.focus();
+        }
+      }
+    } else if (dateVariant === 'YYYY-MM-DD') {
+      if (inputValue.length > 1 && type === DateUtil.inputType.MONTH) {
```
benbcai: Does this need to be `inputValue.length > 1` or can it be `inputValue.length === 2` to be consistent with the logics above?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -123,19 +177,92 @@ class DatePickerInput extends React.Component {
...
+  }
+
+  handleMonthChange(event) {
+    const inputValue = event.target.value.replace(/\D/gm, '');
+    this.handleDateChange(event, DateUtil.inputType.MONTH, inputValue);
+  }
+
+  handleYearChange(event) {
+    const inputValue = event.target.value.replace(/\D/gm, '');
+    this.handleDateChange(event, DateUtil.inputType.YEAR, inputValue);
```
benbcai: nit: Since `handleDayChange`, `handleMonthChange`, and `handleYearChange` all do the same except for the type passed to `handleDateChange`, the line to get the inputValue from the event can be moved to `handleDateChange`.

packages/terra-date-picker/src/react-datepicker/index.jsx
```diff
@@ -480,15 +480,15 @@ class DatePicker extends React.Component {
...
-  handleChange = (event) => {
+  handleChange = (event, dateValue) => {
     if (this.props.onChangeRaw) {
-      this.props.onChangeRaw(event)
+      this.props.onChangeRaw(event, dateValue)
       if (event.isDefaultPrevented()) {
         return
       }
     }
     this.setState({ inputValue: event.target.value })
```
benbcai: Does `event.target.value` here also need to be changed to `dateValue`? Same as in line 492.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -123,19 +177,92 @@ class DatePickerInput extends React.Component {
...
+      if (this.variant === 'YYYY-MM-DD') {
+        if (day.length === 2) {
+          tempDate = DateUtil.formatISODate(year.concat('-', month)
+            .concat('-', day), 'YYYY-MM-DD');
+          formattedDate = DateUtil.strictFormatISODate(tempDate, dateFormat);
+        }
+      } else if (year.length === 4) {
+        tempDate = DateUtil.formatISODate(year.concat('-', month)
+          .concat('-', day), 'YYYY-MM-DD');
+        formattedDate = DateUtil.strictFormatISODate(tempDate, dateFormat);
```
benbcai: Looks like the if and else blocks do the same thing so I'm not following what the checks (`day.length === 2` and `year.length === 4`) are for. Are we trying to convert to inputs into a formatted date. If so, can we just do this and let `DateUtil.isValidDate` handle whether or not it is a valid date?

`formattedDate = DateUtil.formatISODate(year.concat('-', month).concat('-', day), 'YYYY-MM-DD');`

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -123,19 +177,92 @@ class DatePickerInput extends React.Component {
...
+      }
+    }
+
+    // Allows easy deletion of input values.
+    if (day === '' && month === '' && year === '') {
+      this.props.onChange(event, '');
```
benbcai: At first glance, I thought this would call onChange twice (once here and another on line 265). Could we move this to an `else if` block for readability?

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: If you're in the month input, it needs to move to the day input. If you're in the year input, it doesn't matter. The only other option is that you're in the day input (for that date variant). The variant above could be rewritten the same way, since it is just swapping day and month.

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Changed! [0b496eb](https://github.com/cerner/terra-framework/pull/1004/commits/0b496eb729d1b0fc5851cc50d820b24a267a67ec)

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Changed! [0b496eb](https://github.com/cerner/terra-framework/pull/1004/commits/0b496eb729d1b0fc5851cc50d820b24a267a67ec)

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Changed! [0b496eb](https://github.com/cerner/terra-framework/pull/1004/commits/0b496eb729d1b0fc5851cc50d820b24a267a67ec)

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Cleaned this up a bit - [0b496eb](https://github.com/cerner/terra-framework/pull/1004/commits/0b496eb729d1b0fc5851cc50d820b24a267a67ec)

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: Done! [0b496eb](https://github.com/cerner/terra-framework/pull/1004/commits/0b496eb729d1b0fc5851cc50d820b24a267a67ec)

packages/terra-date-picker/src/react-datepicker/index.jsx
amichaelparker: Changed! [0b496eb](https://github.com/cerner/terra-framework/pull/1004/commits/0b496eb729d1b0fc5851cc50d820b24a267a67ec)

packages/terra-date-picker/CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ Unreleased
 * Added clinical-lowlight-theme theming files
 
 ### Changed
+* DateInput split into 3 distinct inputs
```
emilyrohrbough: this need to be moved under the unreleased section

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -115,6 +121,58 @@ class DatePickerInput extends React.Component {
...
+   * disabled if a complete date has been entered in order to make single input
+   * corrections easier, and is re-enabled if the whole date is erased.
+   * @param {string} inputValue - The value from the current input.
+   * @param {int} type - The input type, based on DateUtil.inputType.
+   * @param {string} dateVariant - Date variant based on the currently supported date
+   *                      orders: 'MM-DD-YYYY', 'DD-MM-YYYY', or 'YYYY-MM-DD'.
+   */
+  moveFocusOnChange(inputValue, type, dateVariant) {
+    if (!this.state.prevDate) {
+      if (dateVariant === 'MM-DD-YYYY') {
```
emilyrohrbough: These checks seem bad. How do we handle all the local formats we provide? What about dutch? https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/translations/nl.json#L2

packages/terra-date-picker/src/DateUtil.jsx
```diff
@@ -142,6 +143,113 @@ class DateUtil {
...
   }
 
+  /**
+   * Returns an easily parsable date format string.
+   * @param {string} locale - The locale variant from dateFormatOrder.
+   */
+  static getDateFormat(locale) {
+    let dateVariant;
+
+    if (locale === '0') {
```
emilyrohrbough: When will locale ever be 0? Shouldn't this be a string value representing the translation and locale data loaded by terra-i18n?

packages/terra-date-picker/src/react-datepicker/date_utils.js
```diff
@@ -1,5 +1,7 @@
 import moment from 'moment'
 
+moment.suppressDeprecationWarnings = true;
```
emilyrohrbough: Why has this been included in src code? Is this something we need to disabled in tests instead? Do we need to log an issue to remove this later on?

packages/terra-date-picker/src/react-datepicker/index.jsx
```diff
@@ -768,7 +768,6 @@ class DatePicker extends React.Component {
       onClick: this.onInputClick,
       onFocus: this.handleFocus,
       onKeyDown: this.onInputKeyDown,
-      id: this.props.id,
```
emilyrohrbough: Why did we remove the id from the input being rendered?

packages/terra-date-picker/CHANGELOG.md
ryanthemanuel: Maybe update it as well to say something about how they're masking inputs.

packages/terra-date-picker/translations/es-US.json
```diff
@@ -5,9 +5,11 @@
...
   "Terra.datePicker.today": "Hoy",
   "Terra.datePicker.monthLabel": "Mes",
+  "Terra.datePicker.dayLabel": "Día",
   "Terra.datePicker.yearLabel": "Año",
   "Terra.datePicker.date": "Fecha",
   "Terra.datePicker.nextMonth": "Mes siguiente",
   "Terra.datePicker.previousMonth": "Mes anterior",
-  "Terra.datePicker.calendarInstructions": "Para cambiar la selección, use las teclas de flecha. Presione Entrar para seleccionar una fecha. Presione Esc para cerrar el selector de fecha."
+  "Terra.datePicker.calendarInstructions": "Para cambiar la selección, use las teclas de flecha. Presione Entrar para seleccionar una fecha. Presione Esc para cerrar el selector de fecha.",
+  "Terra.datePicker.dateFormatOrder": "2"
```
emilyrohrbough: This is really confusing to me. Why aren't we utilizing the date format already included in these translations files?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -115,6 +121,58 @@ class DatePickerInput extends React.Component {
     }
   }
 
+  static getDerivedStateFromProps(props, state) {
+    // Formats the selected date into a consistent value and sets input state.
+    if (props.value !== state.prevDate) {
+      const format = DateUtil.getDateFormat(props.intl.formatMessage({ id: 'Terra.datePicker.dateFormatOrder' }));
```
emilyrohrbough: Why can't we use the date format already supplied for the locale? https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/translations/nl.json#L2

If anything I could see adding a data format separator so we can update the below to substring parsing

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -277,26 +277,42 @@ class DatePicker extends React.Component {
...
   }
 
-  handleChangeRaw(event) {
-    this.dateValue = event.target.value;
+  handleChangeRaw(event, date) {
+    let validDate;
+    if (date) {
+      validDate = DateUtil.isValidDate(date, DateUtil.getFormatByLocale(this.props.intl.locale));
+      if (validDate) {
+        this.dateValue = date;
```
ryanthemanuel: Does dateValue need to be set if !date?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -179,28 +305,89 @@ class DatePickerInput extends React.Component {
...
+    const dateYearInput = (
+      <Input
+        {...additionalInputProps}
+        refCallback={(inputRef) => { this.yearInput = inputRef; }}
+        className={cx('date-input-year')}
+        type="text"
+        name={'terra-date-year-'.concat(name)}
+        value={this.state.year}
+        onChange={this.handleYearChange}
+        placeholder={placeholderValues.year || 'YYYY'}
```
emilyrohrbough: Shouldn't this fallback be translated?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -179,28 +305,89 @@ class DatePickerInput extends React.Component {
...
-          name={'terra-date-'.concat(name)}
-          value={value}
-          onChange={this.handleOnChange}
-          placeholder={placeholder}
-          onFocus={onFocus}
-          onBlur={onBlur}
-          aria-label={intl.formatMessage({ id: 'Terra.datePicker.date' })}
-        />
+      <div
+        {...customProps}
```
emilyrohrbough: we didn't spread the custom props previously so I'm not suer we should now

packages/terra-date-picker/src/DateUtil.jsx
```diff
@@ -142,6 +143,113 @@ class DateUtil {
...
+      dateYear = ph.substr(0, 4);
+      dateDelimiter = '-';
+    } else {
+      dateMonth = ph.substr(3, 2);
+      dateDay = ph.substr(0, 2);
+      dateYear = ph.substr(6, 4);
+    }
+
+    if (!dateDelimiter) {
+      dateDelimiter = '/';
```
emilyrohrbough: What about when the deliminator should be `.`? https://github.com/cerner/terra-framework/blob/master/packages/terra-date-picker/translations/de.json#L2

packages/terra-date-picker/src/DateUtil.jsx
```diff
@@ -155,7 +263,7 @@ class DateUtil {
 
   /**
    * Converts an ISO string to the given format.
-   * @param {string} iSODate - The ISO string to convert.
+   * @param {T[]} iSODate - The ISO string to convert.
```
emilyrohrbough: What does `T` stand for?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -115,6 +121,58 @@ class DatePickerInput extends React.Component {
...
+  /**
+   * Moves focus to the correct input depending on date ordering. Focus changing is
+   * disabled if a complete date has been entered in order to make single input
+   * corrections easier, and is re-enabled if the whole date is erased.
+   * @param {string} inputValue - The value from the current input.
+   * @param {int} type - The input type, based on DateUtil.inputType.
+   * @param {string} dateVariant - Date variant based on the currently supported date
+   *                      orders: 'MM-DD-YYYY', 'DD-MM-YYYY', or 'YYYY-MM-DD'.
+   */
+  moveFocusOnChange(inputValue, type, dateVariant) {
```
emilyrohrbough: Should the internals of this method be broken out into the DateUtils so this can be unit tested?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -99,14 +98,21 @@ const defaultProps = {
   value: undefined,
 };
 
-// eslint-disable-next-line react/prefer-stateless-function
 class DatePickerInput extends React.Component {
   constructor(props) {
     super(props);
 
+    this.state = {};
```
emilyrohrbough: Can we set all the state values used in the component to undefined so we know what to expected face-up?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -115,6 +121,58 @@ class DatePickerInput extends React.Component {
     }
   }
 
+  static getDerivedStateFromProps(props, state) {
+    // Formats the selected date into a consistent value and sets input state.
```
emilyrohrbough: Should this be moved to DateUtils so we can unit test this?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -123,19 +181,78 @@ class DatePickerInput extends React.Component {
...
+    if (type === DateUtil.inputType.DAY) {
+      this.setState({
+        day: inputValue,
+      });
+    } else if (type === DateUtil.inputType.MONTH) {
+      this.setState({
+        month: inputValue,
+      });
+    } else if (type === DateUtil.inputType.YEAR) {
+      this.setState({
```
emilyrohrbough: Is there any worry on update this state before the onChange value is called? The state update will cause a re-render

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -179,28 +305,89 @@ class DatePickerInput extends React.Component {
     ]);
     const buttonText = intl.formatMessage({ id: 'Terra.datePicker.openCalendar' });
 
+    const dateMonthInput = (
+      <Input
+        {...additionalInputProps}
+        refCallback={(inputRef) => { this.monthInput = inputRef; }}
+        className={cx('date-input-month')}
+        type="text"
+        name={'terra-date-month-'.concat(name)}
```
emilyrohrbough:
```suggestion
        name={`terra-date-month-${name}`}
```

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -179,28 +305,89 @@ class DatePickerInput extends React.Component {
...
+      />
+    );
+
+    const dateDayInput = (
+      <Input
+        {...additionalInputProps}
+        refCallback={(inputRef) => { this.dayInput = inputRef; }}
+        className={cx('date-input-day')}
+        type="text"
+        name={'terra-date-day-'.concat(name)}
```
emilyrohrbough:
```suggestion
        name={`terra-date-day-${name}`}
```

packages/terra-date-picker/src/DateUtil.jsx
```diff
@@ -142,6 +143,113 @@ class DateUtil {
...
+    let dateDay;
+    let dateYear;
+    let dateDelimiter;
+    let ph = placeholder;
+
+    if (!ph) {
+      ph = 'MM-DD-YYYY';
+    }
+
+    if (variant === 'MM-DD-YYYY') {
```
emilyrohrbough: Can we drop the variant logic and use the translation to parse the month, day, year?

webpack.config.js
```diff
@@ -1,4 +1,9 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
+const merge = require('webpack-merge');
 const terraDevSiteWebpackConfig = require('terra-dev-site/config/webpack/webpack.config');
 
-module.exports = terraDevSiteWebpackConfig;
+const mergedConfig = (env, argv) => (
+  merge(terraDevSiteWebpackConfig(env, argv), { devServer: { clientLogLevel: 'debug', stats: { assets: false } }, watchOptions: { ignored: /aggregated-translations/ } })
```
emilyrohrbough: Can we revert these config changes?

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -562,7 +607,8 @@ Terra.describeViewports('DateTimePicker', ['tiny', 'large'], () => {
     before(() => {
       browser.url('/#/raw/tests/terra-date-time-picker/date-time-picker/date-time-picker-default-date-out-of-range');
       browser.refresh();
-      browser.click('input[name="terra-date-input"]');
+      Terra.hideInputCaret('input[name="terra-date-month-input"]');
```
emilyrohrbough: This happens by default on refresh: https://github.com/cerner/terra-toolkit/blob/master/src/wdio/services/TerraService.js#L82

so this can be removed

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -87,7 +90,10 @@ Terra.describeViewports('DateTimePicker', ['tiny', 'large'], () => {
       browser.url('/#/raw/tests/terra-date-time-picker/date-time-picker/date-time-picker-default');
       browser.refresh();
       Terra.hideInputCaret('input[name="terra-time-minute-input"]');
-      browser.setValue('input[name="terra-date-input"]', '03/11/2018');
+      browser.click('input[name="terra-date-month-input"]');
```
emilyrohrbough: If our tests need changed in this way, I am worried about the passivity of these changes for our consumers that have written tests for this component.

packages/terra-date-picker/src/terra-dev-site/test/date-picker/DatePickerOnChangeRaw.test.jsx
```diff
@@ -13,7 +13,7 @@ class DatePickerOnChange extends React.Component {
       iSO: '',
       inputValue: '',
       isCompleteValue: 'No',
-      isValidValue: 'Yes',
+      isValidValue: 'No',
```
emilyrohrbough: why this change?

packages/terra-date-picker/src/DateUtil.jsx
```diff
@@ -142,6 +143,113 @@ class DateUtil {
     return undefined;
   }
 
+  /**
+   * Returns an easily parsable date format string.
+   * @param {string} locale - The locale variant from dateFormatOrder.
+   */
+  static getDateFormat(locale) {
```
benbcai: Perhaps, rename this to getDateFormatVariant or the like. The parameter passed in should be the format order instead of the locale so perhaps rename it something other than locale.

packages/terra-date-picker/src/DatePicker.module.scss
```diff
@@ -1,82 +1,240 @@
...
+// stylelint-disable scss/at-mixin-pattern
+
 :local {
-  .date-picker {
+  @mixin focus-styles {
+    border-color: var(--terra-date-picker-focus-border-color, #26a2e5);
+    border-style: var(--terra-date-picker-focus-border-style, solid);
+    border-width: var(--terra-date-picker-focus-border-width, 1px);
+    opacity: var(--terra-date-picker-focus-opacity, 1);
+    outline: none;
```
emilyrohrbough: We need to update both themes for these changes: https://github.com/cerner/terra-framework/tree/master/packages/terra-date-picker/themes

packages/terra-date-picker/src/DateUtil.jsx
benbcai: This is not the locale. This is Terra.datePicker.dateFormatOrder from the translation files. I think we should give it a better name.

packages/terra-date-picker/src/DatePicker.module.scss
```diff
@@ -1,82 +1,240 @@
...
+  }
+
+  .date-input-container {
+    display: inline-block;
+  }
+
+  .date-input {
+    background-color: var(--terra-date-picker-background-color, #fff);
```
emilyrohrbough: @neilpfeiffer @jeremyfuksa  Should we be adding new theme variables with these changes? Are there callouts to reference. I'm worried we are adding in vars too early and increasing the complexity of the themes.

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: `nl` would fall under [this](https://github.com/cerner/terra-framework/blob/0e4f6a7773905b689fc5e94a2b0587f6e732ff3c/packages/terra-date-picker/src/DateUtil.jsx#L190). There are 3 actual date variations, but the input itself doesn't need to know whether it is `es` or `fr`, just what order they should be in.

packages/terra-date-picker/src/DateUtil.jsx
amichaelparker: I can rename this to something easier to reason about. :)

packages/terra-date-picker/src/react-datepicker/date_utils.js
amichaelparker: This can be removed. I missed taking that out from diagnosing a previous issue. :/

packages/terra-date-picker/src/react-datepicker/index.jsx
amichaelparker: I will likely have to swap that over to being used by the input itself. Used here, it attempts to apply the same id to all inputs, of which does not make React happy.

packages/terra-date-picker/translations/es-US.json
amichaelparker: Essentially this allows me to parse the date format without having to worry about localization string variances. Ex: it doesn't matter that the format is `DD/MM/AAAA`, just that it is in day/month/year format. If anyone has any other suggestions I'm not attached to it, so I'm happy to accomplish this another way.

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: It should, though I'm thinking maybe I could use the locale format to provide it.

packages/terra-date-picker/src/DateUtil.jsx
amichaelparker: Oh, good catch! I can add that in. :)

packages/terra-date-picker/src/DateUtil.jsx
amichaelparker: This is a code format correction from the ide I've been using that I didn't notice. :/ I'll swap it back!

packages/terra-date-picker/src/DateUtil.jsx
amichaelparker: Parsing the translation seems like it would have variations to actually parse? I can get rid of the variant logic, and just the locale format from the date format order to cut some stuff down. That said, if there would be a relatively concise way to parse all of the date translations, I'd be happy to go that route!

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
amichaelparker: It definitely wouldn't be passive in that case, just by virtue of the one input being split into 3.

packages/terra-date-picker/src/DatePicker.module.scss
jeremyfuksa: Callouts haven't updated since 2017: https://zpl.io/ag0J0lO

@neilpfeiffer, does Seth have something mocked up?

packages/terra-date-picker/src/DateInput.jsx
amichaelparker: I'm updating things to remove this. The overhaul should take care of quite a bit of stuff like this!

packages/terra-date-picker/src/terra-dev-site/test/date-picker/DatePickerOnChangeRaw.test.jsx
amichaelparker: I'll revert it. It resulted from some confusion as to why a test would be checking to verify that `04/01/2019` is an invalid date. There's a little bit of ambiguity in referring to an excluded date as an invalid date, so I had things mixed around a bit. :)

packages/terra-date-picker/src/DateUtil.jsx
amichaelparker: I updated the translation files to include the actual variant instead of a number. This should be less confusing, and it also means that the above util method has been removed! :)

amichaelparker:
Whoops. I can't reopen this at the moment, so I'm working in the above PR until it can be!

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -99,124 +107,330 @@ const defaultProps = {
...
+    ...customProps
+  } = props;
 
-  componentDidUpdate(prevProps) {
-    if (this.shouldShowPicker && !prevProps.shouldShowPicker && this.props.onClick) {
-      this.props.onClick();
+  const { onCalendarButtonClick, shouldShowPicker } = customProps;
+  delete customProps.onCalendarButtonClick;
+  delete customProps.shouldShowPicker;
+  const additionalInputProps = { ...customProps, ...inputAttributes };
```
emilyrohrbough: Why are you spreading the custom props on the input props?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -99,124 +107,330 @@ const defaultProps = {
...
 
-    if (!attributes.readOnly && this.onCalendarButtonClick && this.props.onClick) {
-      this.onCalendarButtonClick(event, this.props.onClick);
+  /**
+   * Moves focus to the correct input depending on date ordering. Focus changing is
+   * disabled if a complete date has been entered in order to make single input
+   * corrections easier, and is re-enabled if the whole date is erased.
+   * @param {string} inputValue - The value from the current input.
+   * @param {int} type - The input type, based on DateUtil.inputType.
+   * @param {string} dateVariant - Date variant based on the currently supported date
```
emilyrohrbough: @benbcai may be able to provide more insights. I think you just updated some logic based on my previous comments. I was confused because I didn't realize you were trying to standardize the formatting order for parsing & composing the date value from the three inputs.

What are your thoughts on updating some of the naming to align with the concept naming associated with this?

I'm thinking `dateOrder` might be the clearest.

The search google returns https://en.wikipedia.org/wiki/Date_format_by_country which states:
> Order of the basic components [components of calendar date]:
.
    B – big-endian (year, month, day), e.g. 2006-04-22 or 2006.04.22 or 2006/04/22 or 2006 April 22
    L – little-endian (day, month, year), e.g. 22.04.2006 or 22/4/2006 or 22-04-2006 or 22 April 2006
    M – middle-endian (month, day, year), e.g. 04/22/2006 or April 22, 2006 

Then according to this doc: https://en.wikipedia.org/wiki/Calendar_date, there are four potential formats:
- DMY
- YMD
- MDY
- YMD
- YDM

packages/terra-date-picker/src/DateUtil.jsx
```diff
@@ -120,7 +121,7 @@ class DateUtil {
    */
   static convertToISO8601(date, format) {
     if (date && format) {
-      const momentDate = moment(date, format, true);
+      const momentDate = moment(date, format);
```
benbcai: What is the reason for removing the strict formatting?

packages/terra-date-picker/src/DateUtil.jsx
```diff
@@ -142,6 +143,98 @@ class DateUtil {
     return undefined;
   }
 
+  /**
+   * Returns an object consisting of placeholder strings and a delimiter,
+   * based on the date format.
+   * @param {string} variant - Variant string obtained from getDateFormat.
+   * @param {string} placeholder - Passed on placeholder prop string.
+   */
+  static getPlaceholderValues(variant, placeholder) {
```
emilyrohrbough: My thoughts - This seem like it could be easily pass in the date separator that we load from translations and we can use it to split the placeholder with that value:

```
// call like
getPlaceholderValues(dateOrder, placeholder, separator)

// use like
static getPlaceholderValues(dateOrder, placeholder, intl.formatMessage({ id: 'Terra.datePicker.seperator' })) {
     const placeholderParts = placeholder.split(separator);

      let day;
      let mouth;
      let year;

      switch(dateOrder) {
          case 'DMY':
               [day, month, year] = placeholderParts;
                break;
          case 'MDY':
               [month, day, year] = placeholderParts;
                break;
          case 'YMD':
               [year, month, day] = placeholderParts;
                break;
         default':
               [year, day, month] = placeholderParts;
      }
      return { year, day, month }
}

```

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -99,124 +107,330 @@ const defaultProps = {
...
+  ]);
+
+  const buttonClasses = cx([
+    'button',
+    { 'is-invalid': isInvalid },
+  ]);
+
+  const dateMonthInput = (
+    <Input
+      {...additionalInputProps}
```
emilyrohrbough: This is a bit tricky to pass the input props on each input. We aren't sure what these could tied to and usually these are accessibly attributes that are added for screen readers that are passed.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -99,124 +107,330 @@ const defaultProps = {
...
-  handleOnButtonClick(event) {
-    const attributes = { ...this.props.inputAttributes };
+  // Sets the date state based on the passed in value prop, or if it changes via a calendar click.
+  useEffect(() => {
+    if (DateUtil.isValidDate(value, dateFormat)) {
+      const parseDate = {
+        dayValue: dateValue.substring(8, 10),
+        monthValue: dateValue.substring(5, 7),
+        yearValue: dateValue.substring(0, 4),
+      };
```
emilyrohrbough: This feels like it should be using the same parsing logic that was used in `getPlaceholderValues`

packages/terra-date-picker/src/DateInput.jsx
benbcai: Looks like some places in DateUtil calls it `order` while other places have it `variant` so I think it would be easier to follow when they are more consistent. `dateOrder` or dateFormatOrder` would make sense. Among the locales we currently support, we would have 3 format orders - `DMY`, `YMD`, and `MDY`. We currently don't have a locale that uses the `YDM` order.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -99,124 +107,330 @@ const defaultProps = {
...
+      formattedDate = DateUtil.strictFormatISODate(inputDate, dateFormat);
     }
-  }
 
-  handleOnKeyDown(event) {
-    if (this.props.onKeyDown) {
-      this.props.onKeyDown(event);
+    /* eslint-disable no-param-reassign */
+    if (DateUtil.isValidDate(formattedDate, dateFormat)) {
+      event.target.value = formattedDate;
```
benbcai: Why are we setting the formatted date back on the input?

amichaelparker: The actual value is only the most recently entered part of the date. The logic in the `DatePicker.jsx` is looking for a single input that eventually turns into a whole date (and really, react-datepicker itself is also built entirely around a single input). This way, I can leave the `DatePicker.jsx` (and react-datepicker itself) as-is without major changes, still allow raw input, and also pass it a full date when it expects it. This was (mostly) done to try to contain the various changes needed into a single place as possible.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -99,124 +107,330 @@ const defaultProps = {
...
+      // Handles raw input change updates by combining inputs.
+      event.preventDefault();
+      if (localeFormat === 'MM-DD-YYYY') {
+        event.target.value = month + day + year;
+      } else if (localeFormat === 'DD-MM-YYYY') {
+        event.target.value = day + month + year;
+      } else {
+        event.target.value = year + month + day;
+      }
+      onChange(event);
```
benbcai, line 288: - If I entered a complete date in the MM/DD/YYYY format but it is invalid (For example, `12/99/2020`) it would not satisfy any of the above statements and as a result, it would not trigger the `onChangeRaw` event. The `onChangeRaw` event would continue to get called for each entry I make up to the last digit because it gets in the 3rd `else if` but once I enter the last digit to complete a 4-digit year, it falls thru these checks so none of the `onChange` above gets hit.

- Previously the value returned in the onChangeRaw callback is whatever the user enters. For example, if they enter "01/01/2020" they would get back "01/01/2020. Now with the separator provided, they would just enter "01" followed by "01" followed by "2020". However, the value in onChangeRaw they now get is "01012020" without the separator. I think that is the right value to return since technically those are the only values they entered. However, from a consumer standpoint it would probably be hard for them to parse or get anything out of "01012020" if they need to convert it back to an ISO-8601 format. Should we inject the separators in the value to make easier for parsing?

I'm interested to hear what other people think.

Screen Shot 2020-04-07 at 3 23 23 PM

benbcai:
The time-input allows deletion of values to flow thru the hour/minute/second inputs. For the date inputs, after deleting the year I would have to manually put focus on the month input to continue deleting.

amichaelparker:
Hey all! Sorry, this one has fallen a bit by the wayside while attending to various quarterly commits. I'm going to try to get back to this as soon as I can, but if anyone wants to pick this up please do!

neilpfeiffer:
Removed `UX Reviewed` label to re-review and add feedback for updates.

benbcai:
In addition to the above unaddressed comments, the following are additional changes to consider:

- A date can be in 3 different format (`MDY`, `DMY`, `YMD`) and can have different separators (`/`, `.`, `-`). This makes the parsing logic complicated and hard to follow and understand. It would be great if we could simply the parsing logic or structure it in a way that is easier to follow. The least is to add code comment around the parsing logic.
- Strict entry should be enforced. For example, only allow a month to be from 01-12 and year from 1900-2100. It’s tricky to enforce the day so maybe always allow 01-31.
- Unable to use `backspace` to go back to previous inputs.
- Unable to use left/right arrow keys to navigate between inputs.
- Add auto complete so that it is consistent with time-input. For example, when entering `5` as the month, it should automatically populate `05` in the month field and move focus to the day field. 
- Cursor jumps to the end of the string after deleting a value that is in the middle of the string. For example, if I delete 1 from `2019`, the cursor jumps to the right of `209`. The cursor is expected to stay between 0 and 9.

PR #1014, opened 2020-01-23
[terra-disclosure-manager] Allow Disclosing Content to Define Title Value

### LINKED ISSUE #1013: [disclosure-manager] Allow Disclosing Content to Define Title Value
# Feature Request

## Description
Some implementation of the Disclosure Manager require a header to be rendered.

Currently the DisclosureManagerHeaderAdapter is used to register the header information with to the disclosed content, such that the DisclosureManager's render can manage rendering a header.

There are instances where this header 'pops-in' / has a delayed render (noticeable in IE)
Speaking with @tbiethman, this is due to the DisclosureManagerHeaderAdapter registering its data and causing the headers to be loaded in after the initial mount and render.

To keep things passive (for use-cases where is still no default header, i.e. SlidePanelManager/ModalManager), we want to provide a mechanism for the header to render on mount, instead of after the initial mount and render.

emilyrohrbough:
### Summary
 Fix delayed disclosure headers by providing a mechanisms for disclosing content to define the disclosed content's title. Worse case, this allows user to define an empty string such that the header is rendered with the back/close buttons, etc.

This can be used in conjunction with existing DisclosureManagerHeaderAdapter in both the Slide Panel Manager and Model Manger.

Closes #1013


packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -298,10 +298,12 @@ class DisclosureManager extends React.Component {
           name: data.content.name,
           props: data.content.props,
           component: data.content.component,
+          ...(data.content.title) && { headerAdapterData: { title: data.content.title } },
```
StephenEsser: > define an empty string

The spread operators will evaluate to false for an empty string. The check will need to account for `''`. 

Possible solution:
```jsx
...(data.content.title !== undefined) 
```

emilyrohrbough: I didn't realize that. Thanks!! 👍

yuderekyu: How granular do we want this check to be? Do we want to also allow null titles?

mjhenkes:
In general I'm against adding more responsibilities to the discloser. The disclosed content should be the primary driver of the title for the disclosed content. I realize this does not fix the issue you're working on but this is a slippery slope to having the discloser add items to the headers toolbar etc, then the discloser is responsible for composing the disclosed content which will have a detrimental effect on our content ownership model.

packages/terra-disclosure-manager/src/terra-dev-site/doc/disclosure-manager/DisclosingContentWithAHeader.4.doc.mdx
```diff
@@ -44,3 +44,29 @@ const DisclosedComponent = () => {
 ```
 
 This will register the header information on mount.
+
+## Providing a Title on Disclose
+It has been observed that there can be a 'pop-in' affect with the header render in slow domains and using the `DisclosureManagerHeaderAdapter`. It is recommended to provide a `title` parameter to the disclose function that will render the disclosed content to ensure the header is rendered immediately. This will render the header with the title, and then the `DisclosureManagerHeaderAdapter` can be used to add additional button content into the header.
```
tbiethman: Probably worth it to throw in a note here saying that the provided title will supersede any title value defined by the header adapter.

emilyrohrbough: Updated: https://github.com/cerner/terra-framework/pull/1014/commits/717d01e3e9e8b41248cadb72d34c6a5763948e31

Issue #1010, opened 2020-01-22
Opening a new window when selecting items from ApplicationNavigation triggers pop-up blockers


regimus:
# Bug Report

## Description
Attempting to open a new window from the `ApplicationNavigation`'s `onSelectUtilityItem` or `onSelectHelp` callback functions triggers Firefox/Safari/IE11/Edge pop-up blocker.

## Steps to Reproduce
Example code:
```
      <ApplicationNavigation
        titleConfig={titleConfig}
        onSelectHelp={() => window.open('', '_blank')}
        utilityItems={utilityItems}
        onSelectUtilityItem={() => window.open('https://github.com/', '_blank')}
      >
        {grid}
      </ApplicationNavigation>
```

## Additional Context / Screenshots
Screenshots and additional details can be found in:
https://jira.cerner.com/browse/IDFED-3465

## Expected Behavior
Opening a new window from a user's click shouldn't trigger pop up blockers.

## Possible Solution
Opening a window a part of the event handler generally won't trigger pop-up blockers. It looks like the `ApplicationNavigation` component is [storing the supplied callback function](https://github.com/cerner/terra-framework/blob/0290f7bf5907bde4f8f57e598a79bfdb3f2a418f/packages/terra-application-navigation/src/ApplicationNavigation.jsx#L157-L172) and [calling the supplied callback function later](https://github.com/cerner/terra-framework/blob/0290f7bf5907bde4f8f57e598a79bfdb3f2a418f/packages/terra-application-navigation/src/ApplicationNavigation.jsx#L325-L330).

If, instead, the supplied callback function was called as part of the click (or keypress, etc) handler, the pop-up blocker shouldn't be triggered.

## Environment
* Component Name and Version: ApplicationNavigation v1.15.0
* Browser Name and Version: Firefox v72.0.2, Edge v42, IE11, Safari v13.0.4
* Node/npm Version: Node v10.15.3, npm v6.4.1
* Webpack Version: v4.41.4
* Operating System and version (desktop or mobile): MacOS Mojave 10.14.6, Windows 10, desktop

neilpfeiffer:
Per original UX design intent, we had not planned to support using external links for utility menu items in application-navigation, for concerns that they could be problematic (due to behavior consistency/expectations, human factors + design standards guidelines, and accessibility). 

We are working on creating a pattern of helpers and managers to launch internal app modals that can display rich content including external links. I have reached out to the requester and following up with internal suggestions.

[...]

PR #1017, opened 2020-01-24
[terra-application-navigation] Fixes pop-up blocker on clicking items.

lokesh-0813:
fixes : #1010


tbiethman:
@neilpfeiffer @scottwilmarth  Before we get too much farther into this, do we know whether or not it we want to support link-like behavior from these utility menu item selections? I don't think we've defined our expectations very well at this point. It seems as though direct linking from these items (or linking from some but not all) would have accessibility concerns?

neilpfeiffer:
Per original UX design intent, we had not planned to support using external links for utility menu items in application-navigation, for concerns that they could be problematic (due to behavior consistency/expectations, human factors + design standards guidelines, and accessibility). 

We are working on creating a pattern of helpers and managers to launch internal app modals that can display rich content including external links, which should resolve the immediate needs from #1010 and similar. I have reached out to the requester and following up with internal suggestions.

Closing as Not an Issue at this time. We can revisit if design direction is updated in the future.

cc: @lokesh-0813 @tbiethman

Issue #1016, opened 2020-01-24
[terra-application-navigation] Clicking on 'help' in drawer-menu fires 'settings' api.


lokesh-0813:
# Bug Report

## Description
In smaller viewports when consumers interact on drawer menu, clicking on `help` calls `settings` api.

## Steps to Reproduce

1.Navigate to [application-navigation](https://engineering.cerner.com/terra-ui/components/terra-application-navigation/application-navigation/example)
2.Reset the browser to smaller window until drawerMenu is rendered.
3.Open drawer-menu and scroll down to help
4.Notice the text above header when clicked on `help`. Its set to 'settings'

## Expected Behavior
"Help" should fire `onSelectHelp` instead of `onSelectSettings`.

## Environment

* Component Name and Version: terra-application-navigaiton, current
* Browser Name and Version: any
* Node/npm Version: [e.g. Node 8/npm 5]

PR #1018, opened 2020-01-24
[terra-application-navigation] Fixes `settings` called on clicking `help` issue

lokesh-0813:
fixes : #1016


lokesh-0813:
@tbiethman @mjhenkes Added tests for help/settings/logout

commit link 22455e1a89e1f50a1b209e01b21d00104845f29a

lokesh-0813:
@tbiethman Was this included by mistake ??

https://github.com/cerner/terra-framework/blob/e6cfe4c7f23e7836799bd707ce1f539b942ce976/packages/terra-application-navigation/tests/jest/drawer-menu/DrawerMenuListItem.test.jsx#L23

tbiethman:
@lokesh-0813 Yeah I imagine that is just a holdover from a previous implementation. The DrawerMenuListItem doesn't do anything with unknown props, so this `hasCount` hasn't caused any problems. We can remove it.

lokesh-0813:
> @lokesh-0813 Yeah I imagine that is just a holdover from a previous implementation. The DrawerMenuListItem doesn't do anything with unknown props, so this `hasCount` hasn't caused any problems. We can remove it.

Updated here 756733e8c63c289cf05e5bec295575edc1a20d4a

packages/terra-application-navigation/src/drawer-menu/_DrawerMenuFooterButton.jsx
```diff
@@ -19,14 +19,15 @@ const propTypes = {
...
-const DrawerMenuFooterButton = ({ text, onClick }) => (
+const DrawerMenuFooterButton = ({ text, onClick, ...customProps }) => (
   <button
     className={cx('drawer-menu-footer-button')}
     type="button"
     onClick={onClick}
     onBlur={enableFocusStyles}
     onMouseDown={disableFocusStyles}
     data-focus-styles-enabled
+    {...customProps}
```
StephenEsser: Generally custom props should always be spread first so they don't overwrite the props the component relies upon.  Here for example, although unlikely in this case,  the custom props could theoretically overwrite the click, blur, and mouse down callbacks.

packages/terra-application-navigation/src/utility-menu/_UtilityMenu.jsx
```diff
@@ -72,6 +72,7 @@ const UtilityMenu = ({
       key: utilityMenuSettingsKey,
       text: intl.formatMessage({ id: 'Terra.applicationNavigation.utilityMenu.settings' }),
       icon: <IconSettings />,
+      dataAttrs: { 'data-navigation-utility-item-settings': 'true' },
```
StephenEsser: Are dataAttrs a part of the previous API? Could the attributes be spread through custom props without the new prop? If dataAttrs stays as a first class props, which component should have documentation for it in the prop types?

lokesh-0813: @StephenEsser 
> Are dataAttrs a part of the previous API?

No.

Please correct me if I am wrong, I parse and pass down `dataAttrs` as a custom prop [here](https://github.com/cerner/terra-framework/pull/1018/files#diff-1c8abd4cb6c3f757361be6e3f68c4646R110). Do you mean to say this is a first class prop at this [point](https://github.com/cerner/terra-framework/pull/1018/files#diff-d8360db5f441f93b4f8126e278dc25caR201).

StephenEsser: It becomes part of the API contract between the component that sets the `dataAttrs` and the component that reads it. These appear to be a private implementations using objects so it may not be too big of a deal if it's not exposed to consumers.

packages/terra-application-navigation/src/drawer-menu/_DrawerMenuFooterButton.jsx
lokesh-0813: That makes sense, Updated here aa934352b61a81e007fbb26d35fefff310f16d02

packages/terra-application-navigation/src/utility-menu/_UtilityMenu.jsx
tbiethman: @lokesh-0813 I'd at least add `dataAttrs` to the shape defined here for the menuItems: https://github.com/cerner/terra-framework/pull/1018/files#diff-d8360db5f441f93b4f8126e278dc25caR49

I don't think the line Stephen called out will be a problem (unless `item` itself is undefined). But please double check that execution without a dataAttrs prop for an item works appropriately, we don't want it to be a required property.

lokesh-0813: updated here a6f2c8a982409414faaefd171e7462ffb03dcd17

PR #1020, opened 2020-01-24
[terra-disclosure-manager] Remove improper dismiss check registration chaining

tbiethman:
### Summary

The DisclosureManager exposes an API (`registerDismissCheck`) that allows disclosure content to register a function that can allow/prevent disclosure dismissal.

This registration function was originally implemented to register dismiss checks to every single DisclosureManager is the component tree. There is two problems with this, though:

1. By chaining the registration calls together in this way, dismissal logic can be overwritten unexpectedly by a nested DisclosureManager instance.
2. The chaining logic, as implemented, doesn't even function appropriately and will always return a resolved Promise, which isn't intended.

To resolve the present issues, I have removed the ancestor chaining logic altogether. The end result of this is that a DisclosureManager rendered within a disclosure (e.g. a SlidePanelManager rendered within a ModalManager's modal component) will not communicate with its parent DisclosureManager when a dismiss check is registered to it. 

Seeing as how the existing logic has been broken for some time without complaint, I feel it's safe to assume this logic was handling the edgiest of cases already and axe it. The NavigationPrompt should be used for these cross-component concerns at this point, anyway.


packages/terra-disclosure-manager/src/DisclosureManager.jsx
```diff
@@ -240,17 +240,8 @@ class DisclosureManager extends React.Component {
      * Allows a component to register a function with the DisclosureManager that will be called before the component is dismissed for any reason.
      */
     delegate.registerDismissCheck = (checkFunc) => {
-      const { disclosureManager } = this.props;
-
       this.dismissChecks[componentData.key] = checkFunc;
 
-      if (disclosureManager && disclosureManager.registerDismissCheck) {
```
emilyrohrbough: Are people still allowed to register a custom dismiss check?

tbiethman: They can. The dismiss check just never leaves the scope of the DisclosureManager it is directly registered to. 

The removed logic would take the provided function and register it to an ancestor DisclosureManager, if one exists. This would have happened if, for example, a SlidePanelManager was rendered within a ModalManager modal. Not exactly an expected or UX desired configuration, but one that exposed these issues.

tbiethman: I will say that the NavigationPrompt is the recommended and future-facing solution to dismissal prevention. registerDismissCheck is on the chopping block whenever we decide to revisit our disclosure patterns.

mjhenkes:
Is there documentation that needs to be cleaned up to reflect this change?

tbiethman:
@mjhenkes No, I don't believe this logic was ever communicated through technical documentation. The existing registerDismissCheck documentation should still be relevant.

PR #1021, opened 2020-01-27
[terra-application-layout] Update docs to use MDX syntax

saket2403:
### Summary
Updated terra-application-layout docs to use MDX syntax


packages/terra-application-layout/src/terra-dev-site/doc/example/ExampleApplication.jsx
```diff
@@ -330,10 +330,12 @@ ExampleApplication.propTypes = {
 const WrappedApplication = withRouter(injectIntl((ExampleApplication)));
 
 const AppRouter = () => (
-  <div className={cx('app-router')}>
-    <MemoryRouter>
-      <WrappedApplication />
-    </MemoryRouter>
+  <div className={cx('height-tag')}>
```
emilyrohrbough: Since these example's code can be toggled, can we add a descriptive class name to indicate this additional wrapper is only necessary for the example? Something like `height-needed-for-docs`? though probably a better name than that

saket2403: Changed

PR #983, opened 2019-12-09
[terra-time-input] Remove 'desktop' reference in Fusion theme.

nramamurth:
### Summary
#975 was merged before https://github.com/cerner/terra-framework/pull/967#discussion_r354509144 was addressed. So, here it is.

PR #1031, opened 2020-01-31
[terra-theme-provider] Removed isGlobalTheme prob and changed default behaviour

### LINKED ISSUE #1002: Terra Theme Provider cannot provide proper theme encapsulation.
# Bug Report

## Description
Terra theme provider does not provide theme encapsulation for all theme-able values. By default theme provider places the theme class on a div that the component renders. This will only apply the css theme variables to selectors below this div. Terra base places theme-able variables on both `html` and `body` which will not be themed by a default theme provider.

Theme provider has a `isGlobalTheme` prop which will place the theme class on the html div which will theme the html and body theme-able variables but does not allow nested theme providers.

## Steps to Reproduce
1. Setup theme provider to not use `isGlobalTheme`.
2. Theme a variable on `html` or `body`.

## Expected Behavior
All theme-able variables are themed.

## Possible Solution
I investigated possibly moving the theme variables from `html` and `body` to being applied the theme provider div, which works for most them-able variables except `font-size` which must remain on `html` due to REM calculation.

My recommendation as of now is to remove the optional `isGlobalTheme` prop and to make that behavior the default. This does not allow for theme encapsulation and cannot as long as `font-size` remains a theme-able variable.

avinashg1994:
### Summary
Removed the `isGlobalTheme` prop and implemented that behaviour by default, i.e. applying the theme to the `html` element.

Resolves #1002


packages/terra-time-input/CHANGELOG.md
```diff
@@ -13,7 +15,6 @@ Unreleased
 ------------------
 ### Added
 * Added new WDIO screenshots
-* Fixed background color issue of time components in incomplete `time-input` in Fusion theme.
```
emilyrohrbough: Whoops. Think this change wasn't intended

packages/terra-theme-provider/CHANGELOG.md
```diff
@@ -3,6 +3,11 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Changed default behavior to always apply theme globally
```
emilyrohrbough: Can you update this comment to be a bit more descriptive. I'm not sure what this means without looking at the code changes

packages/terra-theme-provider/src/ThemeProvider.jsx
```diff
@@ -18,52 +18,42 @@ const propTypes = {
...
       document.documentElement.classList.remove(this.props.themeName);
     }
   }
 
   render() {
     const {
-      themeName, isGlobalTheme, children, ...customProps
+      themeName, children, ...customProps
     } = this.props;
     return (<div {...customProps} className={cx(themeName, customProps.className)}>{children}</div>);
```
mjhenkes: Theoretically, you could remove the themeName from this div, but you can't remove the div.

packages/terra-time-input/CHANGELOG.md
supreethmr: looks like came in while updating the branch with latest changes.

packages/terra-theme-provider/CHANGELOG.md
avinashg1994: Updated the changlog entry to be more descriptive: https://github.com/cerner/terra-framework/pull/1031/commits/0c05483821789068554300b3945aaf7198b959be

packages/terra-time-input/CHANGELOG.md
avinashg1994: Actually this was an intended change.
As seen [here](https://github.com/cerner/terra-framework/pull/983/files#diff-84e7757ec3b55699f14959cc988cc871R4-R16) the change log entry was incorrectly shown as released when PR #983 was merged.
And since, this change was released on February 4, i have now corrected it here : https://github.com/cerner/terra-framework/pull/1031/commits/0c05483821789068554300b3945aaf7198b959be

emilyrohrbough: Ah nice catch! 👍

packages/terra-theme-provider/src/ThemeProvider.jsx
avinashg1994: Yes, I verified after removing themeName from here and it works as expected.
So, addressed it here : https://github.com/cerner/terra-framework/pull/1031/commits/0c05483821789068554300b3945aaf7198b959be

mjhenkes:
I've discussed this with @ryanthemanuel and @tbiethman. We'd like to make this a major version bump to be safe. As part of this please remove the injected div and just render children. Also make sure you add an upgrade guide and document the major version bump in the change log.

emilyrohrbough:
@mjhenkes Since this is a MVB, should we open up to allow multiple themes to be applied or is Heather's use case pretty unique?

mjhenkes:
@emilyrohrbough, I think it is a rare case to have 2 themes at once and we should not make changes here to allow it.

packages/terra-theme-provider/src/ThemeProvider.jsx
```diff
@@ -1,8 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import className from 'classnames';
-
-const cx = className;
 
 const ThemeProviderThemes = {
```
tbiethman: While we're in MVB mode, do we want to kill these theme constants? Are we using these anywhere?

packages/terra-theme-provider/src/ThemeProvider.jsx
```diff
@@ -18,52 +15,39 @@ const propTypes = {
...
-   * When set to true, applies theme class to HTML element
-   */
-  isGlobalTheme: PropTypes.bool,
-};
-
-const defaultProps = {
-  isGlobalTheme: false,
 };
 
 class ThemeProvider extends React.Component {
```
mjhenkes: Consider turning this into a functional component and using the useEffect hook.

packages/terra-theme-provider/src/ThemeProvider.jsx
avinashg1994: After a quick search in our internal github, I found that there a couple of places where the "**Consumer** theme constant" is being used : https://github.cerner.com/search?q=ThemeProvider.Opts.Themes&type=Code.
As for the "**Mock** Theme constant", that looks like it is only being used in our test pages.

packages/terra-theme-provider/src/ThemeProvider.jsx
```diff
@@ -18,52 +15,27 @@ const propTypes = {
...
-  componentDidMount() {
-    if (this.props.isGlobalTheme === true && this.props.themeName) {
-      document.documentElement.classList.add(this.props.themeName);
+const ThemeProvider = ({
+  children,
+  themeName,
+}) => {
+  useEffect(() => {
+    if (themeName) {
+      document.documentElement.classList.add(themeName);
```
mjhenkes: You should add a ref for the previous theme name and remove it before adding the current theme name (if they are different).

avinashg1994: As per my current implementation, on switching from one theme to another, the previously applied theme will be removed first before the new theme is added.
To implement this i used the following as [reference](https://reactjs.org/docs/hooks-effect.html#explanation-why-effects-run-on-each-update).
Also, this will only execute when the theme name has been changed : [reference](https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects).

The only problem that i am facing with the current approach is that i am getting an exception from PropsTable (only on the main Documentation page). Besides that all the other functionality seems to be the same

**I am not sure if i missed something, but is there a benefit of using the ref approach over this?**

packages/terra-theme-provider/src/ThemeProvider.jsx
mjhenkes: I'd like to remove these constants, they aren't up to date.

packages/terra-theme-provider/src/ThemeProvider.jsx
mjhenkes: @avinashg1994 You are correct, you don't need a ref, which does make things a lot cleaner.

mjhenkes:
@avinashg1994 
Once you:
1. Remove `ThemeProviderThemes` and the exported opts. 
2. Fix the doc site.

You'll have my +1.

packages/terra-theme-provider/src/ThemeProvider.jsx
avinashg1994: Done here : https://github.com/cerner/terra-framework/pull/1031/commits/3747cbde1dbde3f4495849c9f264965539ed05bf

avinashg1994:
@mjhenkes,
All the [requested changes](https://github.com/cerner/terra-framework/pull/1031#issuecomment-593544007) have been made : https://github.com/cerner/terra-framework/pull/1031/commits/e9bc14e3d515a310941ca1d4bb20a2518d22f2de
Also updated the documentation and test pages to reflect the removal of `ThemeProviderThemes`

packages/terra-theme-provider/src/ThemeProvider.jsx
```diff
@@ -32,11 +27,9 @@ const ThemeProvider = ({
     };
   }, [themeName]);
 
-  return children;
+  return <React.Fragment>{children}</React.Fragment>;
```
emilyrohrbough: why can't we directly pass back children?

ryanthemanuel: Short answer is no with our framework:

ERROR in ./packages/terra-theme-provider/src/ThemeProvider.jsx?dev-site-props-table
Module build failed (from ./node_modules/terra-dev-site/config/webpack/loaders/devSitePropsTable.js):
NonErrorEmittedError: (Emitted value instead of an instance of Error) Could not convert file to props table:
/Users/rm012685/releases/terra-framework/packages/terra-theme-provider/src/ThemeProvider.jsx
Error: No suitable component definition found.

PR #1041, opened 2020-02-10
[terra-abstract-modal] Remove State from Abstract Modal

emilyrohrbough:
### Summary
- remove state usage in Abstract Modal
- move dom updates for inhert be triggered on modal content mounting and unmounting


packages/terra-abstract-modal/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ Changelog
 
 Unreleased
 ----------
+### Changes
+* Updated `_ModalContent` to trigger the inhert dom updates when it is mounted and unmouted to remove the updates from the componentDidUpdate hook in the `AbstractModal`
```
mjhenkes:
```suggestion
* Updated `_ModalContent` to trigger the inert dom updates when it is mounted and unmounted to remove the updates from the componentDidUpdate hook in the `AbstractModal`
```

packages/terra-abstract-modal/src/_ModalOverlay.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import classNames from 'classnames/bind';
 import PropTypes from 'prop-types';
-import styles from './ModalOverlay.module.scss';
+import styles from './_ModalOverlay.module.scss';
```
mjhenkes: Why underscore? In js _ means "private" in scss it means that it's a partial.

https://stackoverflow.com/questions/34889962/why-put-in-front-of-the-file-name-or-in-scss-css

emilyrohrbough: Oh really? didn't know that! I'll change back

emilyrohrbough: fixed: https://github.com/cerner/terra-framework/pull/1041/commits/b49155fa7bf4ba19ab4ce4b2b0bc5b27c6459ef8

packages/terra-abstract-modal/src/AbstractModal.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Portal } from 'react-portal';
-import * as KeyCode from 'keycode-js';
+import { KEY_ESCAPE } from 'keycode-js';
```
yuderekyu: nice!

packages/terra-abstract-modal/src/_ModalContent.jsx
```diff
@@ -70,33 +80,46 @@ const ModalContent = React.forwardRef((props, ref) => {
...
-    'is-fullscreen': isFullscreen,
-  }, `layer-${zIndexLayer}`, classNameModal]); // Delete the closePortal prop that comes from react-portal.
+  const modalClassName = cx([
+    'abstract-modal',
+    { 'is-fullscreen': isFullscreen },
+    `layer-${zIndexLayer}`,
+    classNameModal,
+  ]);
 
+  // Delete the closePortal prop that comes from react-portal.
```
yuderekyu: can we also comment on why we remove this prop?

packages/terra-abstract-modal/src/AbstractModal.jsx
```diff
@@ -141,35 +124,34 @@ class AbstractModal extends React.Component {
...
-      if (inert === 1) {
-        mainDocumentElement.removeAttribute('data-overlay-count');
+      if (dataOverlayCount === 1) {
+        mainDocumentElement.removeAttribute('data-abstract-modal-overlay-count');
         mainDocumentElement.removeAttribute('inert');
-        // Ensures aria-hidden is properly cleaned up
-        setTimeout(() => mainDocumentElement.removeAttribute('aria-hidden'), 0);
-      } else if (inert && inert > 1) {
-        mainDocumentElement.setAttribute('data-overlay-count', `${inert - 1}`);
+      } else if (dataOverlayCount && dataOverlayCount > 1) {
```
yuderekyu: nothing major, but i believe we can remove the truthy check because `dataOverlayCount` will always be a number. 

setAttribute will return `null` or `''` if the attribute is not found. casting `null` or `''` to a number will return the number `0`

```suggestion
      } else if (dataOverlayCount > 1) {
```

packages/terra-abstract-modal/src/_ModalContent.jsx
emilyrohrbough: @yuderekyu I'm not sure why this was the case. I could dig in if you want me to, but this is just moving the comment from above to be with the delete actions: https://github.com/cerner/terra-framework/pull/1041/files/13b320052eafd45eb4688fe4a142d0f7202aab06#diff-d613e8d30ee292727e9e04177b5d1807L88

packages/terra-abstract-modal/src/AbstractModal.jsx
emilyrohrbough: Did not know that about getAttribute! Thanks

packages/terra-abstract-modal/src/AbstractModal.jsx
```diff
@@ -141,35 +124,34 @@ class AbstractModal extends React.Component {
...
-        mainDocumentElement.setAttribute('data-overlay-count', `${inert - 1}`);
+      } else if (dataOverlayCount > 1) {
+        mainDocumentElement.setAttribute('data-abstract-modal-overlay-count', `${dataOverlayCount - 1}`);
       }
     }
 
-    setTimeout(() => {
-      if (this.state.modalTrigger && this.state.modalTrigger.focus) {
+    // Allows inert processing to finish or focus will not shift back
+    Promise.resolve().then(() => {
```
tbiethman: Was the setTimeout not working? Seems like browsers could handle this differently based on whether the MutationObserver (or the polyfill) used with `inert` schedules micro- or macrotasks. Some fun reading about the differences in these implementations: https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/

packages/terra-abstract-modal/src/_ModalContent.jsx
```diff
@@ -70,33 +80,46 @@ const ModalContent = React.forwardRef((props, ref) => {
...
     role,
     isFullscreen,
     isScrollable,
     zIndex,
     ...customProps
   } = props;
-  let zIndexLayer = '6000';
 
+  React.useEffect(() => {
+    showModalDomUpdates();
```
tbiethman: Does moving the execution of these functions here buy us anything other than being able to remove the componentDidUpdate implementation in AbstractModal? If not, I kind of prefer how the previous implementation keeps all this logic encapsulated within a single component.

Wanna take a stab at converting AbstractModal to a hooks implementation? :P

packages/terra-abstract-modal/src/AbstractModal.jsx
emilyrohrbough: The timeout was working, I was hoping to remove another timeout from the new-relic logs that get pushed out in our applications. I'll change it back

packages/terra-abstract-modal/src/_ModalContent.jsx
emilyrohrbough: Here's my stab at hooks: https://github.com/cerner/terra-framework/pull/1041/commits/c7ba67ff20f42a88c0615c819b6aeab4f86c23ae

Whatchya feeling?

tbiethman: OoOoOooo I like it.

packages/terra-abstract-modal/src/AbstractModal.jsx
```diff
@@ -71,156 +71,80 @@ const defaultProps = {
...
+    role,
+    rootSelector,
+    onRequestClose,
+    zIndex,
+    ...customProps
+  } = props;
+
+  const modalElementRef = createRef();
+
+  useEffect(() => {
```
tbiethman: Consider using useLayoutEffect for this. It'll execute sooner.

packages/terra-abstract-modal/src/_ModalContent.jsx
```diff
@@ -74,29 +80,45 @@ const ModalContent = React.forwardRef((props, ref) => {
...
+    rootSelector,
     zIndex,
     ...customProps
   } = props;
-  let zIndexLayer = '6000';
 
+  useEffect(() => {
+    // Store element that was last focused prior to modal opening
+    const modalTrigger = document.activeElement;
+    showModalDomUpdates(ref.current, rootSelector);
```
tbiethman: Using the forwarded ref here threw me for a loop. I think this will work, but it's a new one for me.

packages/terra-abstract-modal/src/AbstractModal.jsx
emilyrohrbough: Updated: https://github.com/cerner/terra-framework/pull/1041/commits/2bb1b8369609e498a558bf37c2cb085f56fb03c2#diff-b9d74312c1fdd3afc228794142f831c3R93

packages/terra-abstract-modal/src/AbstractModal.jsx
```diff
@@ -71,156 +71,80 @@ const defaultProps = {
...
-      if ('ontouchstart' in window) {
-        this.modalElement.current.querySelector('[data-terra-abstract-modal-begin]').focus();
-      } else {
-        // Shift focus to modal dialog
-        this.modalElement.current.focus();
+  }, []);
+
+  useEffect(() => {
+    function handleKeydown(e) {
+      if (e.keyCode === KEY_ESCAPE && closeOnEsc && isOpen) {
```
yuderekyu: fun fact - according to the mozilla documentation, `keyCode` is now deprecated in favor of `code`. unfortunately `code` is not supported by edge or ie, yet `keyCode` has full browser support.  just something to keep in mind for the future

https://github.com/cerner/terra-core/issues/2866
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

emilyrohrbough: oh man.... whyyyyyy 😢

PR #1042, opened 2020-02-11
[terra-date-picker] Update to Use Controlled responsive-element 

emilyrohrbough:
### Summary
Update terra-date-picker to use the controlled responsive-element implementation.


packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -394,104 +400,57 @@ class DatePicker extends React.Component {
...
         <ResponsiveElement
+          onChange={this.handleBreakpointChange}
           responsiveTo="window"
-          tiny={portalPicker}
-          medium={popupPicker}
-        />
+        >
+          <ReactDatePicker
+            {...customProps}
+            withPortal={this.state.showPortalPicker}
```
emilyrohrbough: The only difference between the previous portalPicker and popupPicker components was passing in `withPortal`. I updated this to utilize the showPortalPicker state.

StephenEsser:
NIt: We should work on creating consistent commit messages when possible.

There was a previous unofficial discussion here:

https://github.com/cerner/terra-core/pull/2707#issuecomment-545183029

packages/terra-date-picker/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+* Update date-picker to use the the controlled responsive-element
```
StephenEsser: We should drop this into a section

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -137,15 +138,18 @@ class DatePicker extends React.Component {
   constructor(props) {
     super(props);
 
+    const activeBreakpointOnMount = activeBreakpointForSize(window.innerWidth);
```
StephenEsser: What happens if you remove this initial set?

I'm not sure if the constructor is a safe place to analyze the window. The component has not technically mounted yet. The responsive element should immediately re-render the component when mounting. If this default is removed does the component flash?

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -188,6 +192,14 @@ class DatePicker extends React.Component {
     this.isDefaultDateAcceptable = this.validateDefaultDate();
   }
 
+  handleBreakpointChange(activeBreakpoint) {
```
StephenEsser: Can we add some documentation for this function?

packages/terra-date-picker/src/DatePicker.jsx
emilyrohrbough: I added this to prevent the need for a re-render to when it's being mounted and the responsive-element returns the same active breakpoint.

packages/terra-date-picker/src/DatePicker.jsx
emilyrohrbough: As of now, we haven't add much, if any, inline documentation in Terra. To me, this is simple enough, the code is describe what the function is doing, like many other places in the library. I actually like that we don't have inline docs everywhere. Inline docs do add value for helpers, scripts or complex function, but to me it makes the code hard to read when its for simple function like this.

I think it's worth opening up a discussion on creating a standard for inline docs and then backfilling the existing Terra components to have them, but if I add it to just this function, it'd be a one-off in this file and I don't think it should be added unless we intend to backfill all fo the functions, which is out of scope of this PR.

Others thoughts?

StephenEsser: This probably falls into the personal opinion/perspective category, but lack of documentation in other areas of code doesn't mean we shouldn't document any additional code added thereafter.  Adding code to one function doesn't necessitate backfilling every other function in the file.

Inline code documentation is different than documenting the API of a function. This function is rather simple, but it is accepting a parameter and doing something with that parameter.

Private API documentation has always been at the discretion of the developer. Because these functions are private I don't see any harm in either approach. Keeping private documentation at the discretion of the engineer may be the decision that's reached, but I would disagree with discouraging API documentation as a standard for those who choose to do so.

packages/terra-date-picker/src/DatePicker.jsx
StephenEsser: I think the odds are good that this would prevent an extra render most of the time, but it is a gamble. Assigning a breakpoint based on the window width in the constructor prior to the component mounting has no guarantee it will mount with the correct breakpoint.

packages/terra-date-picker/src/DatePicker.jsx
emilyrohrbough: > Adding code to one function doesn't necessitate backfilling every other function in the file.

My opinion on this is probably reflect my OCD tendency for the code/structure to be consistent in file (and project) the code is in. 

> but I would disagree with discouraging documentation as a standard for those who choose to do so.

I didn't mean to indicate we should set a standard to not add document. My thought are, if we want to start pushing to adding in-line documentation, it would be helpful to have an agreement on when/how so it's consistent.

packages/terra-date-picker/src/DatePicker.jsx
emilyrohrbough: > Assigning a breakpoint based on the window width in the constructor prior to the component mounting has no guarantee it will mount with the correct breakpoint.

I'm not sure I understand when the window's width would be different than the width it has on mount. Could you provide some examples or context as to the instances you are thinking of?

My thought it this would be toggling against a true or false boolean so if the breakpoint isn't tiny or small like defined in the window's width when it's mounted, the responsive element would send the correct breakpoint size back.

StephenEsser: Variables initialized in the constructor are created when the component is declared not when the component is mounted. 

Here a demo that will demonstrate this. Load this onto a page and then resize the browser. You'll see that each component was initialized with the the breakpoint of the window, but when they mount they'll still have the constructor breakpoint and not the actual breakpoint.

```jsx
/* eslint-disable max-classes-per-file */
import React, { useContext } from 'react';
import { ActiveBreakpointContext, ActiveBreakpointProvider, activeBreakpointForSize } from 'terra-breakpoints';

class Example extends React.Component {
  constructor() {
    super();

    this.constructorBreakpoint = activeBreakpointForSize(window.innerWidth);
  }

  render() {
    return (
      <div>
        {`Constructor Breakpoint: ${this.constructorBreakpoint}`}
      </div>
    );
  }
}

const BreakpointsExample = (props) => {
  const breakpoint = useContext(ActiveBreakpointContext);

  return (
    <div>
      {`Actual Breakpoint: ${breakpoint}`}
      {props[breakpoint]}
    </div>
  );
};

const Provider = () => {
  const tiny = <Example />;
  const small = <Example />;
  const medium = <Example />;
  const large = <Example />;
  const huge = <Example />;
  const enormous = <Example />;

  return (
    <ActiveBreakpointProvider>
      <BreakpointsExample tiny={tiny} small={small} medium={medium} large={large} huge={huge} enormous={enormous} />
    </ActiveBreakpointProvider>
  );
};

export default Provider;
```

emilyrohrbough: I'm looking through the example you provided, and I see where this could be a concern with a component that is store off it's width when it's initialized, but I added a this comparisons which has similar implementation to the changes I made in this PR and the breakpoints are always matching the `BreakpointsExample` component you linked out.

```js
class Example2 extends React.Component {
  constructor() {
    super();
    this.state = { breakpoint: activeBreakpointForSize(window.innerWidth) };
    this.handleBreakpointChange = this.handleBreakpointChange.bind(this);
  }

  handleBreakpointChange(updatedBreakpoint) {
    this.setState({ breakpoint: updatedBreakpoint });
  }

  render() {
    return (
      <ResponsiveElement responsiveTo="window" onChange={this.handleBreakpointChange}>
        <div>
          {`My Example of Breakpoint in State: ${this.state.breakpoint}`}
        </div>
      </ResponsiveElement>
    );
  }
}
```

I think your concerns should be addressed with the onChange callback passed to the Responsive Element: https://github.com/cerner/terra-framework/pull/1042/files#diff-06f07d3cca7012107347dc63be00c147R409

StephenEsser: I'm hesitant to include performance optimizations that are based on random chance. In this case the end result should be the same with or without the dice roll in the constructor because the responsive element will invoke a re-render before the dom paints. I do think this has a good chance of being correct most of the time, but it's important to understand that it is not guaranteed.

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -394,104 +400,57 @@ class DatePicker extends React.Component {
...
-        onFocus={this.handleOnInputFocus}
-        onSelect={this.handleOnSelect}
-        required={required}
-        customInput={(
-          <DateInput
-            onCalendarButtonClick={this.handleOnCalendarButtonClick}
-            inputAttributes={inputAttributes}
-            isIncomplete={isIncomplete}
-            isInvalid={isInvalid}
-            required={required}
```
benbcai: I noticed that the portalPicker has `required` set here by PR [#923](https://github.com/cerner/terra-framework/pull/923/files#diff-06f07d3cca7012107347dc63be00c147R415) whereas popupPicker below doesn't have it set. After talking to @BenBoersma it was not added for popupPicker most likely by mistake. Consolidating portalPicker and popupPicker should sync them up.

emilyrohrbough: nice catch! update: https://github.com/cerner/terra-framework/pull/1042/commits/0d9fa2162c5263155d82285505180d5af8892939

PR #1044, opened 2020-02-13
[terra-date-time-picker] 'timeZone' prop for adding a specific timezone

### LINKED ISSUE #457: Allow consumer to specify timezone used by the DateTimePicker
# Feature Request

## Description

Currently, the `DateTimePicker` uses the system timezone as seen by the user's browser to control DST related functionality, and to determine the ISO value of the component.  We would like to have the ability to specify a timezone to use instead of the browser's timezone.  This can be an optional property, allowing for fallback to the current behavior if no timezone is provided, which would allow this to be a passive change.

The user could be using a device in one timezone (or have their device configured for a particular timezone), while needing to enter data about something occurring in another timezone.  This is especially apparent when the two timezones have different treatment of DST (the transition might take place on a different date, with a different sized shift, or not at all), but applies for all situations where the timezones have different offsets, as it affects the ISO value of the component.

lokesh-0813:
### Summary
Adds a new prop `initialTimeZone`. This will allow consumers to have date-time-picker in a desired timezone.
terra-date-picker has been adjusted to accept the 'initialTimeZone' prop for calculating "Today's" date accordingly. It also takes care of the DST changes.

closes #457 

### Deployment Link
https://terra-framew-date-time--bkjtzt.herokuapp.com/components/terra-date-time-picker/date-time-picker/date-time-picker

### Tests
Added new Wdio tests.
Updated existing JEST to adhere to changes.


lokesh-0813:
@benbcai I am not sure  of what I have done to set the default timezone is right [here](https://github.com/cerner/terra-framework/pull/1044/files#diff-0a8a568512ba2dc166f3b71e47749752R246) .

I am doubtful and I was thinking you could take a look and provide a better design approach ?

I have tried something like
```
const momentDate = moment(date).tz(timeZone).format();
```

It fails. And I couldn't figure out why exactly the above line was failing.

Also I am thinking `setDefualt` would modify the global object for the browser's moment instance ? I could be wrong here.


Let me know your thoughts.

benbcai:
@lokesh-0813 I tried to start the server using your branch but the page is not loading for me. There are no js errors. Are you able to run it?

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -141,7 +141,7 @@ class DateTimeUtils {
       return false;
     }
 
-    const localizedDateTime = moment.tz(dateTime.format(), moment.tz.guess());
+    const localizedDateTime = moment.tz(dateTime.format(), dateTime.tz());
```
benbcai, line 144: is `dateTime` guaranteed that it is in the correct time zone?

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -238,12 +238,12 @@ class DateTimeUtils {
...
    * @return {object|undefined} - The moment object. Undefined if unable to convert.
    */
-  static createSafeDate(date) {
+  static createSafeDate(date, timeZone) {
     if (!date) {
       return undefined;
     }
 
-    const momentDate = moment(date);
+    const momentDate = (timeZone && moment.tz.zone(timeZone) ? moment.tz(date, timeZone) : moment.tz(date, moment.tz.guess()));
```
benbcai, line 246: When no timezone is passed in, instead of doing`moment.tz(date, moment.tz.guess())` would it also work if we do `moment(date)` just like what it used to do?

benbcai:
Anywhere that creates a new moment object using `moment(...)`, we should replace it with `moment.tz(..., this.props.timeZone)`.  See this [comment](https://github.com/cerner/terra-framework/issues/457#issuecomment-577926828) for the reasoning.

lokesh-0813:
> @lokesh-0813 I tried to start the server using your branch but the page is not loading for me. There are no js errors. Are you able to run it?

I am not sure what the issue was, i have updated the branch. You should be able to run it locally.

packages/terra-date-time-picker/src/DateTimeUtils.js
lokesh-0813: Yes it is.

packages/terra-date-time-picker/src/DateTimeUtils.js
lokesh-0813: Yes we could. updated here 3d9c068396910f04f05e60a58f3c4a0c05532950

benbcai:
@lokesh-0813 After some debugging, the reason why the missing hour `02:00` when changing from `1:00` to `2:00` is subtracting an hour back to `1:00` instead of skipping forward to `3:00` is because `momentDate` in [updateTime](https://github.com/cerner/terra-framework/blob/master/packages/terra-date-time-picker/src/DateTimeUtils.js#L82) has an offset of 360 since it has the timezone set in it. You can check by logging out `momentDate.tz()` and `momentDate._offset`.

I think a better way to update the time and to account for the missing hour is to follow this [example](https://momentjs.com/timezone/docs/#/using-timezones/parsing-ambiguous-inputs/). This is what I changed and it appears to work for me. You can give this a try.

Screen Shot 2020-05-13 at 5 26 51 PM

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -87,12 +87,16 @@ class DateTimeUtils {
...
     const date = moment(time, timeFormat, true);
+    const timeZone = momentDate.tz() || moment.tz.guess();
 
-    if (hasSeconds) {
-      return newDate.hour(date.get('hour')).minute(date.get('minute')).second(date.get('second'));
+    let dateTimeString;
+    if (date.isValid()) {
+      dateTimeString = newDate.format('YYYY-MM-DD').concat(' ').concat(date.format(timeFormat));
+    } else {
+      dateTimeString = newDate.format('YYYY-MM-DD');
```
benbcai, line 96: Since this method is for updating the time in the provided `momentDate`, if the provided `time` is invalid (in this case, `date.isValid()`) would be false), I think we should just exist and return the original `momentDate` instead of proceeding since it would update it with the current time.

packages/terra-date-time-picker/src/DateTimeUtils.js
lokesh-0813: moment(date) fails in a few scenarios and hence is reverted here 574bace57cdc094bdbb3007263af3506f5aa27d7

packages/terra-date-time-picker/src/DateTimeUtils.js
lokesh-0813: Yeah that works too, updated here 52065b6ece74801302990a047817302ab5c59aee

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -87,12 +87,16 @@ class DateTimeUtils {
     const newDate = momentDate.clone();
     const timeFormat = hasSeconds ? 'HH:mm:ss' : 'HH:mm';
     const date = moment(time, timeFormat, true);
+    const timeZone = momentDate.tz() || moment.tz.guess();
 
-    if (hasSeconds) {
```
emilyrohrbough, line 91: why drop the seconds formatting?

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -122,6 +122,11 @@ const propTypes = {
...
    */
   timeVariant: PropTypes.oneOf([DateTimeUtils.FORMAT_12_HOUR, DateTimeUtils.FORMAT_24_HOUR]),
+  /**
+   * Timezone value to indicate in which timezone the date-time component is rendered.
+   * The value provided should be a valid [timezone](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) string, else will default to browser/local timezone.
+   */
+  timeZone: PropTypes.string,
 };
 
 const defaultProps = {
```
emilyrohrbough, line 132: Should this add the default timeZone value? `timeZone: moment.tz.guess()`?

benbcai: I think we should so that we only need to check for the validity of the timezone and not the existence of it in some of the helpers in DateTimieUtil.

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -101,9 +105,10 @@ class DateTimeUtils {
    * @param {boolean} hasSeconds Whether or not seconds should be retrieved
    * @return {string} The time from the date and time string
    */
-  static getTime(time, hasSeconds) {
+  static getTime(time, hasSeconds, timeZone) {
     const timeFormat = hasSeconds ? 'HH:mm:ss' : 'HH:mm';
-    return DateUtil.formatISODate(time, timeFormat);
+    const momentDate = (timeZone && moment.tz.zone(timeZone) ? moment.tz(time, timeZone) : moment(time));
```
benbcai, line 110: If `timezone` is invalid, should the object be created using the guess timezone? Or better yet, just call `createSafeDate`.

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -101,9 +105,10 @@ class DateTimeUtils {
...
+    return DateUtil.formatISODate(momentDate, timeFormat);
```
benbcai, line 111: You should use `formatMomentDate` because `formatISODate` expects a string instead of a moment object.

lokesh-0813: updated here 49f03acf3e9427c81be892413ff0c07bb091048d

packages/terra-date-time-picker/src/DateTimeUtils.js
lokesh-0813: updated here 49f03acf3e9427c81be892413ff0c07bb091048d

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -101,9 +105,10 @@ class DateTimeUtils {
    * @param {boolean} hasSeconds Whether or not seconds should be retrieved
    * @return {string} The time from the date and time string
    */
-  static getTime(time, hasSeconds) {
+  static getTime(time, hasSeconds, timeZone) {
     const timeFormat = hasSeconds ? 'HH:mm:ss' : 'HH:mm';
-    return DateUtil.formatISODate(time, timeFormat);
+    const momentDate = (timeZone && moment.tz.zone(timeZone) ? moment.tz(time, timeZone) : this.createSafeDate(time, timeZone));
```
benbcai, line 110: Are you able to just do `const momentDate = DateTimeUtils.createSafeDate(time, timeZone));` without the checks since createSafeDate already checks all that for you?

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -87,12 +87,16 @@ class DateTimeUtils {
     const newDate = momentDate.clone();
     const timeFormat = hasSeconds ? 'HH:mm:ss' : 'HH:mm';
     const date = moment(time, timeFormat, true);
+    const timeZone = momentDate.tz() || moment.tz.guess();
 
-    if (hasSeconds) {
-      return newDate.hour(date.get('hour')).minute(date.get('minute')).second(date.get('second'));
+    let dateTimeString;
+    if (date.isValid()) {
+      dateTimeString = newDate.format('YYYY-MM-DD').concat(' ').concat(date.format(timeFormat));
```
yuderekyu, line 94: not a big deal, but we could move the return statement inside the if block.
```suggestion
      dateTimeString = newDate.format('YYYY-MM-DD').concat(' ').concat(date.format(timeFormat));
      return moment.tz(dateTimeString, timeZone);
```

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -228,8 +233,8 @@ class DateTimeUtils {
    * @param {boolean} hasSeconds - If true seconds will be converted
    * @return {object} - The moment object representing the given date and time.
    */
-  static convertDateTimeStringToMomentObject(date, time, dateformat, hasSeconds) {
-    return DateTimeUtils.updateTime(DateTimeUtils.createSafeDate(DateUtil.convertToISO8601(date, dateformat)), time, hasSeconds);
+  static convertDateTimeStringToMomentObject(date, time, dateformat, hasSeconds, timeZone) {
+    return DateTimeUtils.updateTime(DateTimeUtils.createSafeDate(DateUtil.convertToISO8601(date, dateformat), timeZone), time, hasSeconds);
```
yuderekyu, line 237: not going to lie, this is a very lengthy return. i wouldn't mind splitting these function calls into separate variables to reduce this.

packages/terra-date-time-picker/src/terra-dev-site/common/DateTimePickerTimezoneExampleTemplate.jsx
```diff
@@ -0,0 +1,66 @@
...
+   */
+  timeZone: PropTypes.string,
+};
+
+const defaultProps = {
+  value: '',
+  timeZone: undefined,
+};
+
+class DateTimePickerTimezoneTemplate extends React.Component {
```
yuderekyu, line 22: use hooks instead? :D

packages/terra-date-time-picker/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ChangeLog
 =========
+### Added
+* Added new prop `timeZone` which takes a valid timezone string. This enables date-time-picker set to a particular timezone. By default is set to local/browser timezone.
```
benbcai, line 4:
```suggestion
* Added new `timeZone` prop which takes a valid timezone string. This enables date-time-picker to set to a particular timezone. The default timezone  is set to the local/browser timezone.
```

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -153,7 +158,7 @@ class DateTimePicker extends React.Component {
     super(props);
 
     this.state = {
-      dateTime: DateTimeUtils.createSafeDate(props.value),
+      dateTime: DateTimeUtils.createSafeDate(props.value, props.timeZone),
```
benbcai, line 161: As discussed offline, the timezone will need to be stored in the state since consumers could set a different timezone after the initial mount. `this.state.dateTime` will be updated to the new timezone should it change.

packages/terra-date-time-picker/src/_TimeClarification.jsx
```diff
@@ -13,7 +13,7 @@ const propTypes = {
   /**
    * An ISO 8601 date time with the ambiguous hour.
    */
-  ambiguousDateTime: PropTypes.string,
+  ambiguousDateTime: PropTypes.object,
```
benbcai, line 16: Is the reason to change the prop type to be a moment object because we can get the timezone from it? Since this is a private class it's not as big of an issue but since this is a first-class prop I feel we shouldn't expect the prop type to be a third party object. Would it be better to keep the type as a string and add a new `timezone` prop instead? If we do decide to change to object, the prop description should be updated.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -196,7 +201,7 @@ class DateTimePicker extends React.Component {
     // If the entered time (this.timeValue) is the missing hour during daylight savings,
     // it needs to be updated to the time in this.state.dateTime to reflect the change and force a render.
     if (this.state.dateTime && DateTimeUtils.isValidTime(this.timeValue, this.props.showSeconds)) {
-      const displayedTime = DateTimeUtils.getTime(this.state.dateTime.format(), this.props.showSeconds);
+      const displayedTime = DateTimeUtils.getTime(this.state.dateTime, this.props.showSeconds, this.props.timeZone);
```
benbcai, line 204: What is `.format()` removed from `this.state.dateTime` since `getTime` is expecting the first parameter to be an ISO string?

packages/terra-date-time-picker/src/DateTimeUtils.js
lokesh-0813: updated here 1e47747132f83358d17d2aaed1f98c6bd4a5cfaf

packages/terra-date-time-picker/src/DateTimeUtils.js
lokesh-0813: updated here 1e47747132f83358d17d2aaed1f98c6bd4a5cfaf

packages/terra-date-time-picker/src/DateTimeUtils.js
lokesh-0813: udpated here 1e47747132f83358d17d2aaed1f98c6bd4a5cfaf

packages/terra-date-time-picker/CHANGELOG.md
lokesh-0813: Updated.

packages/terra-date-time-picker/src/DateTimePicker.jsx
lokesh-0813: Updated here 1e47747132f83358d17d2aaed1f98c6bd4a5cfaf

packages/terra-date-time-picker/src/_TimeClarification.jsx
lokesh-0813: Updated here 1e47747132f83358d17d2aaed1f98c6bd4a5cfaf

packages/terra-date-time-picker/src/DateTimePicker.jsx
lokesh-0813: Reverted to use .format();

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -193,29 +197,29 @@ class DateTimeUtils {
...
   }
 
   /**
    * Gets the long standard time zone offset display. (e.g. America/Chicago CST -6:00)
    * @param {string} ambiguousDateTime - The ISO date time with the ambiguous hour.
    * @return {string} - The long standard time zone offset display.
    */
-  static getStandardExpandedTZDisplay(ambiguousDateTime) {
-    const standardDateTime = moment(ambiguousDateTime);
+  static getStandardExpandedTZDisplay(ambiguousDateTime, timeZone) {
```
emilyrohrbough, line 215: these in-line docs all need to be updated to include the timeZone parameter

packages/terra-date-time-picker/src/_TimeClarification.jsx
```diff
@@ -55,6 +55,10 @@ const propTypes = {
    * Whether the clarification is disabled.
    */
   disabled: PropTypes.bool,
+  /**
+   * Time Zone the date time picker is rendered in.
```
emilyrohrbough, line 59: Could we pull forward the prop description used in the DateTimePicker file? I think the info on the expected format would be helpful: https://github.com/cerner/terra-framework/pull/1044/files#diff-fda357981bfc2d7e9aebab8b200dc00eR123-R126

packages/terra-date-time-picker/src/terra-dev-site/common/DateTimePickerTimezoneExampleTemplate.jsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import DateTimePicker from 'terra-date-time-picker/lib/DateTimePicker';
+import DateTimeUtils from '../../DateTimeUtils';
+
+const propTypes = {
+  /**
+   * The current entered date time. Use for the selected date message.
+   */
+  value: PropTypes.node,
```
emilyrohrbough, line 10: should this be a string value?

packages/terra-date-time-picker/src/terra-dev-site/test/date-time-picker/DateTimePickerWithTimezoneDst.test.jsx
```diff
@@ -0,0 +1,13 @@
...
+import DateTimePickerTimezoneExampleTemplate from '../../common/DateTimePickerTimezoneExampleTemplate';
+
+const DateTimePickerExample = () => (
+  <>
+    <DateTimePickerTimezoneExampleTemplate
+      value="2017-11-05T01:30:00"
+      timeZone="America/Chicago"
+    />
+  </>
+);
```
emilyrohrbough, line 11:
```suggestion
const DateTimePickerExample = () => (
    <DateTimePickerTimezoneExampleTemplate
      value="2017-11-05T01:30:00"
      timeZone="America/Chicago"
    />
);
```

packages/terra-date-time-picker/src/terra-dev-site/test/date-time-picker/DateTimePickerWithTimezoneMissingHour.test.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import DateTimePickerTimezoneExampleTemplate from '../../common/DateTimePickerTimezoneExampleTemplate';
+
+const DateTimePickerExample = () => (
+  <div>
```
emilyrohrbough, line 5: the div wrapper prob isn't needed

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -8,8 +8,8 @@ class DateTimeUtils {
    * @param {string} iSODate - The ISO string
```
emilyrohrbough, line 8: There are alot of changes in these utils files. Are there jest tests to go with them?

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -691,4 +691,34 @@ Terra.describeViewports('DateTimePicker', ['tiny', 'large'], () => {
...
+    before(() => browser.url('/#/raw/tests/terra-date-time-picker/date-time-picker/date-time-picker-with-timezone'));
+
+    Terra.it.matchesScreenshot();
+  });
+
+  describe('With timezone missing hour', () => {
+    before(() => {
+      browser.url('/#/raw/tests/terra-date-time-picker/date-time-picker/date-time-picker-with-timezone-missing-hour');
+      browser.refresh();
+      Terra.hideInputCaret('input[name="terra-time-minute-input"]');
```
emilyrohrbough, line 705: This should happen automatically by terra-toolkit

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -691,4 +691,34 @@ Terra.describeViewports('DateTimePicker', ['tiny', 'large'], () => {
...
+      browser.setValue('input[name="terra-date-input"]', '03/11/2018');
+      browser.setValue('input[name="terra-time-hour-input"]', '02');
+      browser.setValue('input[name="terra-time-minute-input"]', '30');
```
emilyrohrbough, line 708: Can we move this to its own `it` block instead of the before hook?

packages/terra-date-time-picker/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ChangeLog
 =========
+### Added
+* Added new `timeZone` prop which takes a valid timezone string. This enables date-time-picker set to a particular timezone. The default timezone  is set to the local/browser timezone.
```
saket2403, line 4: This should be moved to unreleased section.

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -256,6 +260,14 @@ class DateTimeUtils {
   static isMomentObject(value) {
     return moment.isMoment(value);
   }
+
+  /**
+   * Creates moment object with local timezone.
+   * @return {object|undefined}- The moment object. Undefined if unable to convert.
```
benbcai, line 266: This inline doc is inaccurate. `moment.tz.guess()` returns the local timezone string. For example, `America/Chicago`. It doesn't return a moment object.

packages/terra-date-time-picker/src/DateTimeUtils.js
lokesh-0813: updated here b5f07bc5f04b127f3d75bc46a3651a2e1091b165

packages/terra-date-time-picker/src/_TimeClarification.jsx
lokesh-0813: changed in this commit b5f07bc5f04b127f3d75bc46a3651a2e1091b165

packages/terra-date-time-picker/src/terra-dev-site/common/DateTimePickerTimezoneExampleTemplate.jsx
lokesh-0813: updated here b5f07bc5f04b127f3d75bc46a3651a2e1091b165

packages/terra-date-time-picker/src/terra-dev-site/test/date-time-picker/DateTimePickerWithTimezoneDst.test.jsx
lokesh-0813: Changed here b5f07bc5f04b127f3d75bc46a3651a2e1091b165

packages/terra-date-time-picker/src/terra-dev-site/test/date-time-picker/DateTimePickerWithTimezoneMissingHour.test.jsx
lokesh-0813: Removed here b5f07bc5f04b127f3d75bc46a3651a2e1091b165

packages/terra-date-time-picker/CHANGELOG.md
lokesh-0813: Changed here b5f07bc5f04b127f3d75bc46a3651a2e1091b165

packages/terra-date-time-picker/src/DateTimeUtils.js
lokesh-0813: Changed here b5f07bc5f04b127f3d75bc46a3651a2e1091b165

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -263,7 +272,7 @@ class DateTimeUtils {
 
   /**
    * Creates moment object with local timezone.
-   * @return {object|undefined}- The moment object. Undefined if unable to convert.
+   * @return {string}}- Local/Browser time zone string.
```
emilyrohrbough, line 275:
```suggestion
   * @return {string} - Local/Browser time zone string.
```

packages/terra-date-time-picker/src/DateTimeUtils.js
ryanthemanuel: Yeah let's definitely add some tests.

lokesh-0813: Added here 2b9b95282e579abb629029b54bd70a7c3ec0049a

packages/terra-date-time-picker/src/DateTimeUtils.js
lokesh-0813: changed here 2b9b95282e579abb629029b54bd70a7c3ec0049a

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
lokesh-0813: updated in this commit 2795049548a9f35213a3e290861bea40d3ff2ffe

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
lokesh-0813: removed in this commit 2795049548a9f35213a3e290861bea40d3ff2ffe

packages/terra-date-time-picker/src/terra-dev-site/common/DateTimePickerTimezoneExampleTemplate.jsx
lokesh-0813: changed here b14377f4f34f42d937a36f052d1386c43c50eeb3

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -120,6 +120,11 @@ const propTypes = {
    * If the `variant` prop if set to `12-hour` for one of these supported locales, the variant will be ignored and defaults to `24-hour`.
    */
   timeVariant: PropTypes.oneOf([DateTimeUtils.FORMAT_12_HOUR, DateTimeUtils.FORMAT_24_HOUR]),
+  /**
+   * Timezone value to indicate in which timezone the date-time component is rendered.
+   * The value provided should be a valid [timezone](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) string, else will default to browser/local timezone.
+   */
+  timeZone: PropTypes.string,
```
ShettyAkarsh, line 127: As this is a string. are we validating the timezone prop value ?

lokesh-0813: Yes, we are validating. 

Its done [here](https://github.com/cerner/terra-framework/pull/1044/files/b14377f4f34f42d937a36f052d1386c43c50eeb3#diff-0a8a568512ba2dc166f3b71e47749752R260).

`moment.tz.zone(timeZone)` validates the timeZone string.

packages/terra-date-time-picker/src/DateTimeUtils.js
emilyrohrbough: @lokesh-0813 the tests you added were wdio tests. I think Jest test here to validate these function would be more reliable.

lokesh-0813: @emilyrohrbough  are you talking about tests from this commit 2b9b95282e579abb629029b54bd70a7c3ec0049a ?

emilyrohrbough: Oops! wasn't looking close enough. This is good. 👍

PaulSolDev:

benbcai:
@lokesh-0813 There are several tests covering setting different timezones. However, there is no example on the dev-site to illustrate the behavior when a different timezone is set. Could you add a first class example. I think it would help to add a native select (to avoid adding form-select as a dependency) on the example with a few timezones as options that can choose from.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -203,11 +210,22 @@ class DateTimePicker extends React.Component {
     }
   }
 
+  static getDerivedStateFromProps(nextProps, prevState) {
```
avinashg1994, line 213: Where is this method being used?
Since it is a static method can this be moved into the DatTimeUtils?

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -6,10 +6,11 @@ class DateTimeUtils {
   /**
    * Checks if the ISO string contains the time (hh:mm) part.
    * @param {string} iSODate - The ISO string
+   * @param {string} timeZone- Time zone value for the moment object.
```
avinashg1994, line 9: nit: Space after `timeZone`.

packages/terra-date-time-picker/src/terra-dev-site/test/date-time-picker/Test.test.jsx
```diff
@@ -0,0 +1,78 @@
+import React, { useState } from 'react';
```
avinashg1994, line 1: This file should be renamed to depict its purpose

lokesh-0813: This is temporary to conclude some confusions.

There is a on going offline discussion on the behavior of changing timezone for a date time picker. 
Sure will be deleted once we have a final answer.

packages/terra-date-time-picker/src/DateTimePicker.jsx
lokesh-0813: It's a standard lifecycle method from [react](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops)

packages/terra-date-time-picker/src/terra-dev-site/common/DateTimePickerTimezoneExampleTemplate.jsx
```diff
@@ -0,0 +1,60 @@
...
+   */
+  timeZone: PropTypes.string,
+};
+
+const defaultProps = {
+  value: '',
+  timeZone: DateTimeUtils.getLocalTimeZone(),
+};
+
+function DateTimePickerTimezoneTemplate(props) {
```
avinashg1994, line 22: This can be converted to use the ES6 format.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -211,10 +211,10 @@ class DateTimePicker extends React.Component {
   }
 
   static getDerivedStateFromProps(nextProps, prevState) {
-    if (nextProps.timeZone !== prevState.timeZone) {
+    if (nextProps.initialTimeZone !== prevState.timeZone) {
```
emilyrohrbough, line 214: Do we want to do this still?

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -149,19 +149,19 @@ const defaultProps = {
...
+  initialTimeZone: DateTimeUtils.getLocalTimeZone(),
 };
 
 class DateTimePicker extends React.Component {
   constructor(props) {
     super(props);
 
     this.state = {
-      dateTime: DateTimeUtils.createSafeDate(props.value, props.timeZone),
+      dateTime: DateTimeUtils.createSafeDate(props.value, props.initialTimeZone),
```
emilyrohrbough, line 160: If we don't want this to be able to change, should this be stored on the DateTimePicker and the inital value always be referenced?
```
class DateTimePicker extends React.Component {
  constructor(props) {
    super(props);
    this.timezone = props.initialTimeZone

```

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -144,17 +150,19 @@ const defaultProps = {
   timeInputAttributes: undefined,
   value: undefined,
   timeVariant: DateTimeUtils.FORMAT_24_HOUR,
+  initialTimeZone: DateTimeUtils.getLocalTimeZone(),
```
benbcai, line 153: Do we still want to default this to the local timezone or leave it undefined?

packages/terra-date-time-picker/src/DateTimePicker.jsx
benbcai: No, I don't think we need `getDerivedStateFromProps` anymore since we're going to ignore it.

packages/terra-date-time-picker/src/DateTimePicker.jsx
lokesh-0813: We could get rid of it and adjust the test examples. Setting it to `undefined` will still set it to local timezone [here](https://github.com/cerner/terra-framework/pull/1044/files#diff-0a8a568512ba2dc166f3b71e47749752R260)

packages/terra-date-time-picker/src/DateTimePicker.jsx
lokesh-0813: Changed here ea2630e205a708f19fe9d9c16c9f9cbb247d0434

packages/terra-date-time-picker/src/DateTimePicker.jsx
lokesh-0813: Removed here ea2630e205a708f19fe9d9c16c9f9cbb247d0434

packages/terra-date-time-picker/src/_TimeClarification.jsx
```diff
@@ -55,6 +55,11 @@ const propTypes = {
    * Whether the clarification is disabled.
    */
   disabled: PropTypes.bool,
+  /**
+   * Timezone value to indicate in which timezone the date-time component is rendered.
+   * The value provided should be a valid [timezone](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) string, else will default to browser/local timezone.
+   */
+  timeZone: PropTypes.string,
```
emilyrohrbough, line 62: Should this also be initalTimeZone?

lokesh-0813: Apart from the prop exposed to consumers, I have retained everywhere as "timeZone". If the prop in `_TimeClarification.jsx` confusing, I will go ahead and change it it `initialTimeZone`.

emilyrohrbough: IMO we should keep the naming consistent if the functionality is constant throughout even if it's passes to internal components. If others think it's fine to leave it as it, I'm okay with keeping it.

benbcai: I also vote to make them consistent to minimize any confusion.

packages/terra-date-time-picker/src/DateTimeUtils.js
lokesh-0813: Fixed here 3cadcdb1ff2efc1ba7f4eef489e4bad1f080be6d

packages/terra-date-time-picker/src/_TimeClarification.jsx
lokesh-0813: Changed here 3cadcdb1ff2efc1ba7f4eef489e4bad1f080be6d

emilyrohrbough (approved):

benbcai:
@lokesh-0813 In these [examples](https://terra-framew-date-time--bkjtzt.herokuapp.com/tests/terra-date-time-picker/date-time-picker/date-time-picker-with-timezone) with a different timezone set, when I clicked on the Today button, it is still setting today's date based on the local timezone instead of the timezone that is in context.

lokesh-0813:
> @lokesh-0813 In these [examples](https://terra-framew-date-time--bkjtzt.herokuapp.com/tests/terra-date-time-picker/date-time-picker/date-time-picker-with-timezone) with a different timezone set, when I clicked on the Today button, it is still setting today's date based on the local timezone instead of the timezone that is in context.

Hey Ben, I was not able to reproduce this. Could you walk me through the steps?

benbcai:
> > @lokesh-0813 In these [examples](https://terra-framew-date-time--bkjtzt.herokuapp.com/tests/terra-date-time-picker/date-time-picker/date-time-picker-with-timezone) with a different timezone set, when I clicked on the Today button, it is still setting today's date based on the local timezone instead of the timezone that is in context.
> 
> Hey Ben, I was not able to reproduce this. Could you walk me through the steps?

@lokesh-0813 In the above example, the tz is set to `Asia/Dubai`. My local current date/time is 9/**_22_**/2020 15:05 (Tuesday) and the current date in `Asia/Dubai` is 9/_23_/2020 00:05 (Wednesday). The picker shows `22` as bold for today's date. Selecting the Today button defaults `09/22/2020`. I would expect `23` to be bold and defaults `09/23/2020` as today's date in the context of the `Asia/Dubai` timezone. You would have to choose a timezone and recreate at a time when the current date locally is different from the timezone's current date.

tz

packages/terra-date-time-picker/src/terra-dev-site/doc/example/DateTimePickerTimeZone.jsx
```diff
@@ -0,0 +1,41 @@
+import React, { useState } from 'react';
+import DateTimePicker from 'terra-date-time-picker/lib/DateTimePicker';
+import moment from 'moment';
+import Field from 'terra-form-field';
+import DateTimeUtils from '../../../DateTimeUtils';
+
+function DateTimePickerTimeZone() {
```
pranav300, line 7: This can be converted to use the ES6 format.

pranav300 (approved):

benbcai:
> > > @lokesh-0813 In these [examples](https://terra-framew-date-time--bkjtzt.herokuapp.com/tests/terra-date-time-picker/date-time-picker/date-time-picker-with-timezone) with a different timezone set, when I clicked on the Today button, it is still setting today's date based on the local timezone instead of the timezone that is in context.
> > 
> > 
> > Hey Ben, I was not able to reproduce this. Could you walk me through the steps?
> 
> @lokesh-0813 In the above example, the tz is set to `Asia/Dubai`. My local current date/time is 9/**_22_**/2020 15:05 (Tuesday) and the current date in `Asia/Dubai` is 9/_23_/2020 00:05 (Wednesday). The picker shows `22` as bold for today's date. Selecting the Today button defaults `09/22/2020`. I would expect `23` to be bold and defaults `09/23/2020` as today's date in the context of the `Asia/Dubai` timezone. You would have to choose a timezone and recreate at a time when the current date locally is different from the timezone's current date.
> 

It turns out that each day in the calendar are not updated to the correct utcOffset. They all are derived from the `selected` prop sent down. To address this, what is passed to `selected` first needs to have the utcOffset set correctly.

lokesh-0813:
> Any update on this last change?

Suggestions from Ben is giving vivid results, will have a offline discussion with Ben for the same.

emilyrohrbough:
@lokesh-0813 @benbcai did we determine what was needed to fix/update the examples?

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -691,4 +691,40 @@ Terra.describeViewports('DateTimePicker', ['tiny', 'large'], () => {
 
     Terra.it.validatesElement();
   });
+
+  describe('With timezone', () => {
+    before(() => browser.url('/#/raw/tests/terra-date-time-picker/date-time-picker/date-time-picker-with-timezone'));
+
+    Terra.it.matchesScreenshot();
```
benbcai, line 698: `Terra.it.*` have been deprecated and replaced by `Terra.validates.*` See https://github.com/cerner/terra-toolkit-boneyard/blob/main/docs/Wdio_Utility.md#test-assertion-helpers

benbcai:
The following are issues that were encountered:

1. The day of today that is highlighted in bold in the picker is calculated using the locale timezone instead of the timezone that is provided. It is possible that today’s date in the locale timezone could be yesterday or tomorrow’s date in the context of the provided timezone.
2. Clicking the Today button is calculated using the local timezone’s current date. Clicking the Today button will display the local timezone’s current date when this date could be yesterday or tomorrow’s date in the context of the provided timezone.

For example, the current date/time in America/Chicago is 11/15/2020 11:30 PM while at the same moment in time the current date/time in Australia/Sydney 11/16/2020 04:30 PM. When the `initialTimezone is set to `Australia/Sydney`, the picker show `15` as the current day and clicking the Today button would display 11/15/2020 when it should be 11/16/2020 in the context of the `Australia/Sydney` timezone.


The first attempt to resolve this was to pass down the [utcOffset](https://github.com/cerner/terra-framework/blob/c2f9850f4cba69cab2f64bb7fc0ad2006edc1056/packages/terra-date-picker/src/react-datepicker/index.jsx#L300) so that the moment date object has the correct utc offset. All days in the picker are calculated using this moment date and the utcOffset.

This appears to have resolved the above two issues. However, a new issue this presents is that momentjs does not take daylight savings into account when the utcOffset is set on the date object. For example, when the utcOffset is -5 (for America/Chicago in CST) all CST and CDT date will have -5 as the utcOffset. It should be that CDT dates (dates between November and March) should be converted to an utcOffset of -6. As a result, this throws off one hour when calculating today’s date. That is, if the current date/time is 11/15/2020 11:30 PM, the calculated today’s date ends up to be 11/16/2020 00:30 AM.

The second solution to resolve this is instead of passing down the utcOffset and calculating the date using this utcOffset, we needed to set the timezone on the [selectedDate](https://github.com/cerner/terra-framework/blob/c2f9850f4cba69cab2f64bb7fc0ad2006edc1056/packages/terra-date-picker/src/react-datepicker/index.jsx#L256) that is passed down so that momentjs can handle the daylight savings conversation automatically based on the timezone in the momentjs date object.

lokesh-0813:
@benbcai I have made all the necessary changes to fix the issues that were found.
I have added few tests as well to check "Today's" date getting calculated  correctly.
It would be nice if you took a final look and approve or otherwise need any more changes.

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
lokesh-0813: I think either @saket2403  or @pranav300 will have a separate PR for this.

lokesh-0813:
> @lokesh-0813 @benbcai did we determine what was needed to fix/update the examples?

It took a while but I think all the issues have been addressed.

packages/terra-date-picker/package.json
```diff
@@ -31,6 +31,7 @@
     "focus-trap-react": "^6.0.0",
     "keycode-js": "^2.0.1",
     "moment": "^2.21.0",
+    "moment-timezone": "^0.5.13",
```
benbcai, line 34: moment-timezone is an add-on for moment so we should be able to remove the moment dependency here and [here](https://github.com/cerner/terra-framework/blob/88d3580c37b3965d3b563623e6a8d4843a913d85/packages/terra-date-time-picker/package.json#L32) in terra-date-time-picker. Anywhere we have `import moment from 'moment'` should be replaced with `import moment from 'moment-timezone'`.

Here are some references on why we don't need both:
- https://stackoverflow.com/questions/49729646/using-both-moment-and-moment-tz
- https://thecodebarbarian.com/a-practical-guide-to-moment-timezone.html

packages/terra-date-picker/src/DateUtil.js
```diff
@@ -1,22 +1,33 @@
...
+    if (initialTimeZone && moment.tz.zone(initialTimeZone)) {
+      momentDate = moment.tz(date, DateUtil.ISO_EXTENDED_DATE_FORMAT, true, initialTimeZone);
+    } else {
+      momentDate = moment(date, DateUtil.ISO_EXTENDED_DATE_FORMAT, true);
+    }
+
     if (!momentDate || !momentDate.isValid()) {
       // This should allow DateTime inputs that used to work in the moment.ISO_8601 to still pass but discard the time from the value.
-      momentDate = moment(date.slice(0, 10), DateUtil.ISO_EXTENDED_DATE_FORMAT, true);
+      if (initialTimeZone && moment.tz.zone(initialTimeZone)) {
```
benbcai, line 26: nit: You can store `initialTimeZone && moment.tz.zone(initialTimeZone` in a const and use it in both places. Not sure how expensive it is to make this check.

packages/terra-date-picker/src/react-datepicker/calendar.jsx
```diff
@@ -387,10 +392,10 @@ export default class Calendar extends React.Component {
   }
 
   getDateInView = () => {
-    const { preSelection, selected, openToDate, utcOffset } = this.props
+    const { preSelection, selected, openToDate, initialTimeZone } = this.props
```
benbcai, line 395: Are you now able to remove the `utcOffset` prop and all references to it in react-datepicker since we are now using `initialTimeZone`? I don't think `utcOffset` gets used anymore after these changes. 

The references in these tests [[1](https://github.com/cerner/terra-framework/blob/c2f9850f4cba69cab2f64bb7fc0ad2006edc1056/packages/terra-date-picker/tests/jest/DatePicker.test.jsx#L19)][[2](https://github.com/cerner/terra-framework/blob/4880772b8b92bb05f99ef3f3084a8d1e4c389bdc/packages/terra-date-time-picker/tests/jest/DateTimePicker.test.jsx#L10)] should be updated to pass down `initialTimeZone` to keep a consistent test run.

[setUTCOffset](https://github.com/cerner/terra-framework/blob/e63c8c8ba6484d5122b9fb128973ce894ed733df/packages/terra-date-picker/src/react-datepicker/date_utils.js#L111), [getUTCOffset](https://github.com/cerner/terra-framework/blob/e63c8c8ba6484d5122b9fb128973ce894ed733df/packages/terra-date-picker/src/react-datepicker/date_utils.js#L138-L140) and [isSameUtcOffset](https://github.com/cerner/terra-framework/blob/e63c8c8ba6484d5122b9fb128973ce894ed733df/packages/terra-date-picker/src/react-datepicker/date_utils.js#L235-L241) can also be removed since they are no longer being used.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -124,6 +124,12 @@ const propTypes = {
    * If the `variant` prop if set to `12-hour` for one of these supported locales, the variant will be ignored and defaults to `24-hour`.
    */
   timeVariant: PropTypes.oneOf([DateTimeUtils.FORMAT_12_HOUR, DateTimeUtils.FORMAT_24_HOUR]),
+  /**
+   * Timezone value to indicate in which timezone the date-time component is rendered.
+   * The value provided should be a valid [timezone](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) string, else will default to browser/local timezone.
+   * Note: This value is considered only for the initial render. Changes to an already set time zone will reset to the first set time zone.
+   */
+  initialTimeZone: PropTypes.string,
```
benbcai, line 132: Can you move this and `timeVariant` up so the props are in alphabetical order?

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -77,6 +78,7 @@ class DateTimeUtils {
    * @param {object} momentDate - The moment object to synchronize the date and time.
    * @param {string} time - The time to synchronize with the moment object.
    * @param {boolean} hasSeconds - If true seconds will be processed
+   * @param {string} timeZone - Time zone value for the moment object.
```
benbcai, line 81: Looks like this param is not needed.

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -87,23 +89,28 @@ class DateTimeUtils {
...
     const timeFormat = hasSeconds ? 'HH:mm:ss' : 'HH:mm';
     const date = moment(time, timeFormat, true);
+    const timeZone = momentDate.tz() || moment.tz.guess();
 
-    if (hasSeconds) {
-      return newDate.hour(date.get('hour')).minute(date.get('minute')).second(date.get('second'));
+    let dateTimeString;
+    if (date.isValid()) {
+      dateTimeString = newDate.format('YYYY-MM-DD').concat(' ').concat(date.format(timeFormat));
+      return moment.tz(dateTimeString, timeZone);
```
benbcai, line 97:
```suggestion
    if (date.isValid()) {
      const dateTimeString = newDate.format('YYYY-MM-DD').concat(' ').concat(date.format(timeFormat));
      return moment.tz(dateTimeString, timeZone);
```

packages/terra-date-time-picker/src/DateTimeUtils.js
```diff
@@ -87,23 +89,28 @@ class DateTimeUtils {
     const newDate = momentDate.clone();
     const timeFormat = hasSeconds ? 'HH:mm:ss' : 'HH:mm';
     const date = moment(time, timeFormat, true);
+    const timeZone = momentDate.tz() || moment.tz.guess();
```
benbcai, line 92: Anywhere that does `moment.tz.guess()` can use `getLocalTimeZone`.

packages/terra-date-time-picker/src/terra-dev-site/common/DateTimePickerTimezoneExampleTemplate.jsx
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  return (
+    <>
+      <p>
+        Selected ISO Date Time:
+        <span id="date-time-value">{dateTime}</span>
+      </p>
+      <p>
+        Selected Time Zone:
```
benbcai, line 45: This should say something like `Initial Time Zone` instead of `Selected Time Zone` to avoid thinking that it can be changed on the fly.

packages/terra-date-time-picker/src/terra-dev-site/doc/example/DateTimePickerTimeZone.jsx
```diff
@@ -0,0 +1,41 @@
...
+  };
+
+  return (
+    <>
+      <p>
+        Selected ISO Date Time:
+        {dateTime}
+      </p>
+      <p>
+        Selected Timezone:
```
benbcai, line 25:
```suggestion
        Initial Time Zone:
```

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -691,4 +691,53 @@ Terra.describeViewports('DateTimePicker', ['tiny', 'large'], () => {
...
+
+    Terra.it.matchesScreenshot({ selector: '#root' });
+  });
+
+  describe('With timezone', () => {
+    before(() => {
+      browser.url('/#/raw/tests/terra-date-time-picker/date-time-picker/date-time-picker-with-timezone-missing-hour');
+    });
+
+    it('missing hour', () => {
```
benbcai, line 706: This test step uses a test that has the timezone set so it is more appropriate. We can remove the [test step done here](https://github.com/cerner/terra-framework/blob/ca8fde828d309978e69ab801d59044f5e71626b0/packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js#L70-L82) since it is a duplicate and was using the local timezone, which may not always honor daylight savings. 

Additionally, [this test step](https://github.com/cerner/terra-framework/blob/ca8fde828d309978e69ab801d59044f5e71626b0/packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js#L84-L94) should run against the `date-time-picker-with-timezone-missing-hour` instead of the `date-time-picker-default` to stay consistent.

packages/terra-date-time-picker/src/terra-dev-site/test/date-time-picker/DateTimePickerWithTimezoneMissingHour.test.jsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import DateTimePickerTimezoneExampleTemplate from '../../common/DateTimePickerTimezoneExampleTemplate';
+
+const DateTimePickerExample = () => (
+  <DateTimePickerTimezoneExampleTemplate
+    initialTimeZone="America/New_York"
```
benbcai, line 6: nit: The name of this test file shouldn't be `DateTimePickerWithTimezoneMissingHour` since it has nothing with missing hour. It is a test that can be used to test the missing hour scenario. Are we able to just use one of the other existing tests (e.g.; DateTimePickerWithTimezoneDst) to test this scenario so we don't have to create an extra one? The test steps might just need to clear the default value and then enter the value with the missing hour.

packages/terra-date-picker/package.json
lokesh-0813: Changed here 11b765b40bac98b52e15037a385965c7f9713921

packages/terra-date-picker/src/DateUtil.js
lokesh-0813: Changed in this commit a63c5545ec2835f9c76c1642e14c9902b50c67eb

packages/terra-date-picker/src/react-datepicker/calendar.jsx
lokesh-0813: changed here 9a4bd199149690dd5d681559f8ddc0ad44307779

packages/terra-date-time-picker/src/DateTimeUtils.js
lokesh-0813: removed here e3a49adabb7322a4265611a27d15cd77a0110374

packages/terra-date-time-picker/src/DateTimeUtils.js
lokesh-0813: changed here e3a49adabb7322a4265611a27d15cd77a0110374

packages/terra-date-time-picker/src/DateTimeUtils.js
lokesh-0813: changed here e3a49adabb7322a4265611a27d15cd77a0110374

packages/terra-date-time-picker/src/terra-dev-site/common/DateTimePickerTimezoneExampleTemplate.jsx
lokesh-0813: changed here d09c6b6938c6c9fe12cf082e09fd4ce65bd55da6

packages/terra-date-time-picker/src/terra-dev-site/doc/example/DateTimePickerTimeZone.jsx
lokesh-0813: changed here d09c6b6938c6c9fe12cf082e09fd4ce65bd55da6

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
lokesh-0813: changed here 15688a34fa0eff270ee5c526ff6e65624b2d5fa1

packages/terra-date-time-picker/src/terra-dev-site/test/date-time-picker/DateTimePickerWithTimezoneMissingHour.test.jsx
lokesh-0813: changed here 15688a34fa0eff270ee5c526ff6e65624b2d5fa1

packages/terra-date-time-picker/src/DateTimePicker.jsx
lokesh-0813: 2624eda3c88fd2434edbb3be07736b64b9ab2423 changed here

benbcai (approved):

ryanthemanuel (approved):

chrismichalewicz:
Completed the functional review and updated the labels accordingly.

PR #1045, opened 2020-02-13
[terra-date-time-picker][terra-date-picker] Additional metadata exposed

### LINKED ISSUE #497: DateTime Picker needs more of the props exposed
# Feature Request

## Description
As a consumer, I want a lot of the internal state available so that I dont need to recalculate and duplicate the logic done inside the separate controls.  

This includes:
- When the control is in the ambiguous hour
- When the date control is empty
- When the date control is invalid
- When the time control is empty (potentially to the granularity of both the hour and minute controls)
- When the time control is invalid (potentially to the granularity of both the hour and minute controls)

We are needing a date time component where the control has to be filled out and has to be valid.  While we can default the value, and we are, a user could remove said value and depending on how much information is filled out, we may not be able to delineate between '00:00' and an empty value.  

As for the ambiguity check, our use case is we have 2 timezones that we manage.  If these two timezones diverge, we put a timezone abbreviation of one of the timezones to the right of this control.  The issue arises when this control is ambiguous, it also puts a timezone abbreviation to the right of the control.  We need to be able to suppress our timezone abbreviation if the date time control is showing it's abbreviation.

avinashg1994:
### Summary
The following changes have been made as part of this PR:

**Date Time Picker**

1. Changed `onBlur` callback to now also return **_dateValue, timeValue and isAmbiguousHour_** along with the existing metadata previously provided.
2. Changed `onChange` and `onChangeRaw` callbacks to now return all the metadata being returned from `onBlur` callback, which includes _**iSO, inputValue, dateValue, timeValue, isAmbiguousHour, isCompleteValue and isValidValue**_

**Date Picker**

1. Changed `onChange` and `onChangeRaw` callbacks to now return all the metadata being returned from `onBlur` callback, which includes _**iSO, inputValue, isCompleteValue and isValidValue**_

Resolves #497 

### Additional Details
Updated documentation to reflect the changes and also added tests to check each scenario.


packages/terra-date-time-picker/docs/DateTimePickerEvents.md
```diff
@@ -17,9 +20,25 @@ terra-date-time-picker has several callback events that are slightly different i
  - A date that is filtered by the `filterDate` prop.
  - Not an included date in the `includeDates` prop.
  - Not within the range of the `minDate` and `maxDate` props.
+* There are three parameters in the `onChange` callback prop. The first parameter is the `event` object. The second parameter is the `new date and time` value. The third parameter is the `options` object that contains the following attributes to describe the current state of the input values at the time when the `onChange` callback is triggered.
```
nramamurth: What does ``new date and time`` mean here?

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -203,6 +203,68 @@ class DateTimePicker extends React.Component {
     }
   }
 
+  getMetadata(momentDateTime) {
+    const isCompleteDateTime = DateTimeUtils.isValidDateTime(this.dateValue, this.timeValue, this.state.dateFormat, this.props.showSeconds);
+    let value = '';
```
nramamurth: Let's be explicit here. I'd rename it to something like - `dateTimeInputValue` / `inputValue`.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -203,6 +203,68 @@ class DateTimePicker extends React.Component {
     }
   }
 
+  getMetadata(momentDateTime) {
+    const isCompleteDateTime = DateTimeUtils.isValidDateTime(this.dateValue, this.timeValue, this.state.dateFormat, this.props.showSeconds);
```
nramamurth: nit: I'd move this line closer to where it is first being referenced - right before Line 239, that is.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -203,6 +203,68 @@ class DateTimePicker extends React.Component {
...
+    let value = '';
+    let date = '';
```
nramamurth: Now that this big blob of logic is extracted into a separate function, it could be shortened to get rid of a number of if checks.

```const dateTimeValue = `${dateValue ? dateValue : '' } ${timeValue ? timeValue : '' }`.trim();```

Since `date` and `time` are not being anywhere else but at the end to create `options` object, I'd do - 

```
const options = {
      ...
      dateValue: this.dateValue || '',
      timeValue: this.timeValue || '',
      ...
}
```

File: packages/terra-date-time-picker/src/terra-dev-site/test/date-time-picker/DateTimePickerOnChange.test.jsx [hunk 5]
```diff
@@ -4,20 +4,73 @@ import DateTimePicker from '../../../DateTimePicker';
...
+          {' '}
+          <span id="time-value">{this.state.timeValue}</span>
+          <br />
+          <br />
+          Is Ambiguous?
+          {' '}
+          <span id="ambiguous-date">{this.state.isAmbiguousHour}</span>
+          <br />
+          <br />
+          Is Date Complete?
```
supreethmr: nit: this should be Is-Date-Time Complete since it is validating both :)

supreethmr:
with account to missing hour scenario i was testing [date-time-picker-on-change-raw](https://terra-framework-deploy-pr-1045.herokuapp.com/tests/terra-date-time-picker/date-time-picker/date-time-picker-on-change-raw) and found that time-value was not updated after time change happened due to missing hour(2:30) on time input as shown in below image.

missing-hour-sceanrio

this could be due to below scenario : 
handleChangeRaw() will get triggered before `this.timeValue `gets updated in [componentDidUpdate() ](https://github.com/cerner/terra-framework/blob/c96b4c95077a1037dc072a84933efd466dc87286/packages/terra-date-time-picker/src/DateTimePicker.jsx#L200) during missing hour scenario (I.e. when 2:30 is entered for date 03/13/2016 it will update time to next available hour which 3:30 in this scenario). 

and [getMetadata](https://github.com/cerner/terra-framework/blob/c96b4c95077a1037dc072a84933efd466dc87286/packages/terra-date-time-picker/src/DateTimePicker.jsx#L217) uses `this.timeValue` for setting time value in metaData which will be still the entered time value (2:30). hence in the above test time value is not showing updated time value.

this could be corrected by using momentDateTime object to update date and time value for metaData 
OR 
we can update the timeValue in [handleTimeChange()](https://github.com/cerner/terra-framework/blob/c96b4c95077a1037dc072a84933efd466dc87286/packages/terra-date-time-picker/src/DateTimePicker.jsx#L417) whenever `displayedTimeValue` is updated so that `this.timeValue` remains insync with `displayedTimeValue`. 

```javascript
if (updatedDateTime) {
   displayedTimeValue = DateTimeUtils.getTime(updatedDateTime.format(), this.props.showSeconds);
   if (displayedTimeValue !== this.timeValue)
       this.timeValue = displayedTimeValue;
}
```

cc: @benbcai

packages/terra-date-time-picker/docs/DateTimePickerEvents.md
avinashg1994: `new date and time` was actually to signify the changed date and time which triggered the onChange callback.
But I agree that it can be confusing, so i have now changed it to `date and time` which is sufficient : https://github.com/cerner/terra-framework/pull/1045/commits/1e8dc74b082dc0de2c1002be51b425e1d49f9784

packages/terra-date-time-picker/src/DateTimePicker.jsx
avinashg1994: Addressed here : https://github.com/cerner/terra-framework/pull/1045/commits/1e8dc74b082dc0de2c1002be51b425e1d49f9784

packages/terra-date-time-picker/src/DateTimePicker.jsx
avinashg1994: Addressed here : https://github.com/cerner/terra-framework/pull/1045/commits/1e8dc74b082dc0de2c1002be51b425e1d49f9784

packages/terra-date-time-picker/src/terra-dev-site/test/date-time-picker/DateTimePickerOnChange.test.jsx
avinashg1994: Addressed here : https://github.com/cerner/terra-framework/pull/1045/commits/1e8dc74b082dc0de2c1002be51b425e1d49f9784

packages/terra-date-time-picker/src/DateTimePicker.jsx
avinashg1994: Yes, this suggestion does help shorten and simplify the logic.
Addressed it here : https://github.com/cerner/terra-framework/pull/1045/commits/1e8dc74b082dc0de2c1002be51b425e1d49f9784

benbcai:
@supreethmr @avinashg1994 We may need to call `this.handleTimeChange` after updating `this.timeValue` in [componentDidUpdate](https://github.com/cerner/terra-framework/blob/issue-497-expose-date-time-picker-props/packages/terra-date-time-picker/src/DateTimePicker.jsx#L200) so that the updated time value is notified in onChange/onChangeRaw.

I agree with your first suggestion that we would need to use the passed in `momentDateTime` instead of `this.timeValue` in getMetadata in case it is the missing hour. 

As far as your second suggestion, I'm concerned that this suggestion may re-introduce the issue that was fixed on [#888](https://github.com/cerner/terra-framework/pull/888/files#diff-fda357981bfc2d7e9aebab8b200dc00e).

packages/terra-date-time-picker/CHANGELOG.md
```diff
@@ -3,14 +3,16 @@ ChangeLog
...
+* Update docs to use mdx syntax.
+
 4.27.0 - (February 4, 2020)
 ------------------
 ### Fixed
 * Fixed date-time-picker issue of updating default value programmatically
 
-### Changed
-* Update docs to use mdx syntax.
-
```
benbcai: Is this removed by mistake?

avinashg1994: This is intentional. This was incorrectly merged as released.
So, i have it under Unreleased here : https://github.com/cerner/terra-framework/pull/1045/files?file-filters%5B%5D=.md#diff-26826d473eb00d037e1d708ec3f591bbR9

packages/terra-application-navigation/CHANGELOG.md
```diff
@@ -3,11 +3,12 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+* Fixed Drawer menu not closing in Internet Explorer on clicking outside of drawer menu.
```
ShettyAkarsh: is this part of your change ?

avinashg1994: This change, even-though not a part of this PR, is an intended change, since this changelog entry was incorrectly merged as released.

avinashg1994:
Fixed the missing hour issue, where the metadata was not reflecting the updated time.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -263,49 +315,7 @@ class DateTimePicker extends React.Component {
...
-      }
-
-      const options = {
-        iSO: iSOString,
-        inputValue: value,
-        isCompleteValue: isCompleteDateTime,
-        isValidValue: isValid,
-      };
-
+      const options = this.getMetadata(momentDateTime);
```
nramamurth: Let's rename `options` to `metaData` 🙃

avinashg1994: Addressed here : https://github.com/cerner/terra-framework/pull/1045/commits/a1449c5a19dbeb60ff1189d49f9838ecc5d38e3d

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -192,6 +192,27 @@ class DatePicker extends React.Component {
...
   }
 
+  getMetadata() {
+    const format = DateUtil.getFormatByLocale(this.props.intl.locale);
+    const isCompleteDate = DateUtil.isValidDate(this.dateValue, format);
+    const iSOString = isCompleteDate ? DateUtil.convertToISO8601(this.dateValue, format) : '';
+
+    let isValidDate = false;
+
+    if (this.dateValue === '' || (isCompleteDate && this.isDateWithinRange(DateUtil.createSafeDate(iSOString)))) {
```
supreethmr: this can be simplified to   
```
 if (this.dateValue === '' || this.isDateWithinRange(DateUtil.createSafeDate(iSOString)))
```

avinashg1994: The check for `isCompleteDate` is required because, if the date is incomplete then the `isoString` will be undefined (which may cause an exception).

PR #1066, opened 2020-03-04
[terra-embedded-content-consumer] Updated docs to use MDX syntax

saket2403:
### Summary
Updated docs to use MDX Syntax


packages/terra-embedded-content-consumer/src/terra-dev-site/doc/example/CustomEventsConsumer.jsx
```diff
@@ -1,5 +1,8 @@
 import React from 'react';
 import EmbeddedContentConsumer from 'terra-embedded-content-consumer';
+import { Consumer } from 'xfc';
+
+Consumer.init();
```
emilyrohrbough: why do we need this change?

saket2403: Initially this syntax was in the jsx file which was converted to mdx. Since embedded content consumer needs this line to function properly, moved it to individual examples.

PR #1073, opened 2020-03-13
[terra-time-input] Minute input resizing on modal when cached value selected

### LINKED ISSUE #1050: Terra Time input become two lines when use cached value
# Bug Report
Terra Time input/ DateTimePicker become two lines when using cache value for minutes 

## Description

The Time input/ DateTimePicker will become two lines when I use the cached value. It only happened when I use this component with a Modal manager together. 

## Steps to Reproduce

Code: Put a DateTimePicker or time input in a modal manager
1. Open the modal manager
2. Click minutes field, choose browser cached value

image

## Expected Behavior

Keep the same behavior just like when the user input the value. 

## Environment

* Component Name and Version:  "terra-date-time-picker": "^4.25.0"
* Browser Name and Version: Chrome Version 80.0.3987.116 (Official Build) (64-bit)
* Node/npm Version: Node 10.13.0
* Operating System and version (desktop or mobile): desktop

lokesh-0813:
### Summary
Terra-time-input on modal resizes when given a cached value. This PR addresses the issue with some style changes. Due to new styles most of the terra-time-input wdio screenshots are re generated and few are updated in terra-date-time-picker.

closes #1050


lokesh-0813:
Screenshot 2020-03-13 at 12 15 08 PM

@neilpfeiffer @benbcai 
nit : There is a very small difference I noticed if time-input and time-input with seconds. The difference is significant if they are placed side by side. Especially the space between the dividers and the digits. Let me know if it requires any changes. Thank you.

emilyrohrbough:
Was this bug caused by the cached value having extra whitespace that needed trimmed? Or did we determine why this value was causing the extra width?

benbcai:
@lokesh-0813 We have to make the input values center aligned (instead of left aligned) since we don't want to shrink the time-input width once the minute is entered. I don't think there are that many use cases where a page would have some time-inputs with seconds while others without seconds. If we wanted to keep both consistent, I don't object to also make the one without seconds center aligned. @neilpfeiffer also thought about using a different font for the placeholders (i.e. `mm`) is has a similar width as the numeric digits to the spacing is not as noticeable.

benbcai:
@emilyrohrbough There is no whitespace in the input value when selecting it from the cache. When doing this inside a modal some empty space/object/thing is added to the outside of the minute input. It is visible but there is nothing in the DOM that references that extra space. It is just there. Given what is happening, what is causing this is that the time-input changes/shrinks the minute input once a minute is added. The reason for the width change is because the minute placeholder `mm` requires a slightly bigger width to display the entire placeholder than any two numeric digits. Once the minute is added, it shrinks the width slightly to make it fit better. The resizing seems weird and jumpy so after talking to @neilpfeiffer that we're not doing the resize and that avoid this problem.

ShettyAkarsh:
@lokesh-0813 can you add a test case by rendering the datetimepicker inside the modal ?

lokesh-0813:
> @lokesh-0813 can you add a test case by rendering the datetimepicker inside the modal ?

No @ShettyAkarsh , this PR addresses a issue where input value is selected from browser suggestion/cache.I'm not sure if there is an option to select cache values in WDIO, even if there is a provision, the results will be inconsistent across different devices. The WDIO tests will fail if there is no cached value at the time of running tests.

PaulSolDev:
Functional review completed with @benbcai .  No issues identified.

benbcai:
This issue only happens when the time-input is rendered inside a modal. Since we don't want to make modal-manager as a dev dependency we don't have an existing example that can be used to  validate this. Here if a gif showing the change.

time-input-verification

PR #1085, opened 2020-04-01
[terra-time-input] Updated wdio screenshots

benbcai:
### Summary
Update wdio screenshots for the following components:

- terra-time-input
- terra-modal-manager

### Additional Details
- **terra-time-input**
Previously the css styles of the button from button-group had higher specificity than the styles from time-input. Not sure why. I guess it was like this for some time. Currently, the styles in time-input have higher specificity, which I think is the correct behavior. Not sure what changed since yesterday that corrected this. I know many components got released yesterday. Locking down to previous versions of button and button-group didn't make any difference. The updated screenshots should be the correct styles.

  - Previously:
Screen Shot 2020-04-01 at 2 57 45 PM

  - Now:
Screen Shot 2020-04-01 at 2 58 13 PM

- **terra-modal-manager**
Update wdio screenshots due to the change in https://github.com/cerner/terra-core/pull/2861 to show focus styles when programmatically focused.

@cerner/terra


mjhenkes:
Does this PR need to be combined with https://github.com/cerner/terra-framework/pull/1084 for a successful build?

benbcai:
> Does this PR need to be combined with #1084 for a successful build?

I don't believe so. Screenshots in terra-modal-manager are now also not matching which is what's causing the build failure. I'll check to see what's going on there.

PR #1091, opened 2020-04-08
Focus error during remount.

### LINKED ISSUE #1086: [terra-date-time-picker] Error occurs on focus during a remount
# Bug Report

## Description
There is currently no first-class support for programmatically setting the value of a DateTimePicker. The common workaround of setting a key prop to force a re-mount breaks the blur/focus functionality to switch the user between fields and introduces a console error.

## @ Mentions
@benbcai

benbcai:
### Summary
Fixed error that occurs during remount when setting focus with a null reference to the hour input.

Closes #1086 

### Deployment Link

### Additional Details
The change checks to ensure the reference to `hourInput` for in terra-date-time-picker and `calendarButton` in terra-date-picker still exist before calling focus. It is possible that reference is now nullified after the 100 ms timeout due to a force remount of this component with a new `key` prop value.

It is possible to put focus on the `hourInput`/`calendarButton` other than using the reference handle. However, I don't think we should continue to put the focus on that element when a remount occurs because this essentially refreshes the component with a new state and we should not retain the focus.

@cerner/terra


packages/terra-date-picker/tests/wdio/date-picker-spec.js
```diff
@@ -471,4 +471,17 @@ Terra.describeViewports('Date Picker', ['medium'], () => {
...
   });
+
+  describe('Remounts component', () => {
+    before(() => {
+      browser.url('/#/raw/tests/terra-date-picker/date-picker/date-picker-remount');
+    });
+
+    it('selects a date from date picker', () => {
+      browser.keys('ArrowDown');
+      browser.keys('Enter');
```
supreethmr, line 482: don't we need to click on calendar button to select date here..?

benbcai: Good catch! Updated here: https://github.com/cerner/terra-framework/pull/1091/commits/2898d118261634c3eafec4ddc7b47c0d6ea5a664

PR #1132, opened 2020-05-26
[terra-date-picker] Limit year from 1900 - 2100.

### LINKED ISSUE #1104: Terra Date Time Picker Weird Behavior Of Year Field On Manual Input
# Bug Report

## Description
Date Time Picker shows ambigious year on manually writing the year in the input box provided
https://engineering.cerner.com/terra-ui/components/terra-date-time-picker/date-time-picker/date-time-picker

## Steps to Reproduce
1. Consider any terra date time picker component, for our case we used the  "terra-date-time-picker": "^4.34.0" with Time Variant with 12-hour Time format. Visualisation : YYYY-MM-DD HH:MM AM PM
2. Select a valid date picking from the calendar 05/04/2020.
3. Now Write something gibberish on the month and/or day and/or year fields (make it a invalid date)
4. For our case we changed 05/04/2020 to a complete invalid date 35/25/2465664 in one scenario and a ambigious year scenario but with valid day and month 05/04/2465664.

5. Then again tried selecting a good date by choosing from the datepicker, This time again selected the same date 05/04/2020 but the date picker input box reflects it as 05/04/2465. Next tried with another good date from the picker 05/18 but still the same it becomes 05/18/2465. Seems like It retains the last 4 digit of the year field that was written on the manual input and the year combo box up there shifts to 1900 in both cases. On such scenario, It is expected to either change the date to 05/18/1900 or should come back to the recent year 2020.

This is a bug as user can always type in anything and everything in an input box but needs to be over-riden on select of the date picker to my understanding. Please take a look.

## Environment

* Component Name and Version: DateTime Picker
* Browser Name and Version: Chrome, IE
 npm: '6.9.0',
  ares: '1.10.1-DEV',
  cldr: '32.0',
  http_parser: '2.8.0',
  icu: '60.1',
  modules: '57',
  napi: '4',
  nghttp2: '1.33.0',
  node: '8.16.0',
  openssl: '1.0.2r',
  tz: '2017c',
  unicode: '10.0',
  uv: '1.23.2',
  v8: '6.2.414.77',
  zlib: '1.2.11'
* Webpack Version:
* Operating System and version (desktop or mobile): windows/linux

benbcai:
### Summary

- Limit the supported dates to between 01/01/1900 and 12/31/2100 because the year dropdown only contains years between 1900 and 2100.
- Disable the Today button if today is a disabled day. (i.e.; outside the `minDate`/`maxDate` range, in `excludeDates`, not in `includeDates`, or filtered by `filterDate`)

Closes #1104

### Deployment Link

### Testing
Added new tests.

### Additional Details

@cerner/terra


packages/terra-date-picker/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ Changelog
 
 Unreleased
 ----------
+### Changed
```
emilyrohrbough, line 6: Should this be `###Fixed`?

packages/terra-date-picker/src/DateUtil.js
```diff
@@ -207,8 +207,49 @@ class DateUtil {
     }
     return DateUtil.formatMomentDate(momentDate, format);
   }
+
+  /**
+   * Ensures the maxDate is before DateUtil.MAX_DATE.
+   * @param {String} maxDate The maxDate prop
```
emilyrohrbough, line 213:
```suggestion
   * @param {String} maxDate -vThe maxDate prop.
```

packages/terra-date-picker/src/DateUtil.js
```diff
@@ -207,8 +207,49 @@ class DateUtil {
...
+    if (DateUtil.isBetweenMinMaxDate(maxDate)) {
+      return maxDate;
+    }
+
+    return DateUtil.MAX_DATE;
+  }
+
+  /**
+   * Ensures the minDate is after the DateUtil.MIN_DATE.
+   * @param {String} minDate The minDate prop
```
emilyrohrbough, line 226:
```suggestion
   * @param {String} minDate - The minDate prop.
```

packages/terra-date-picker/src/DateUtil.js
```diff
@@ -207,8 +207,49 @@ class DateUtil {
...
+    if (DateUtil.isBetweenMinMaxDate(minDate)) {
+      return minDate;
+    }
+
+    return DateUtil.MIN_DATE;
+  }
+
+  /**
+   * Check if the date is between DateUtil.MIN_DATE and DateUtil.MAX_DATE.
+   * @param {String} date The date to check. Must be in the 'YYYY-MM-DD' format.
```
emilyrohrbough, line 239:
```suggestion
   * @param {String} date - The date to check. Must be in the 'YYYY-MM-DD' format.
```

packages/terra-date-picker/src/react-datepicker/date_utils.js
```diff
@@ -67,7 +67,13 @@ export function cloneDate (date) {
 
 export function parseDate (value, { dateFormat, locale }) {
   const m = moment(value, dateFormat, locale || moment.locale(), true)
-  return m.isValid() ? m : null
+
+  // Only support dates between 01/01/1900 and 12/31/2010 because the year dropdown only allows 1900 - 2100.
+  if (m.isValid() && m.isAfter('1900-01-01') && m.isBefore('2100-12-31')) {
+    return m;
+  }
```
emilyrohrbough, line 74: Is it possible to use DateUtil.isBetweenMinMaxDate for this check?

packages/terra-date-picker/src/react-datepicker/date_utils.js
```diff
@@ -67,7 +67,13 @@ export function cloneDate (date) {
 
 export function parseDate (value, { dateFormat, locale }) {
   const m = moment(value, dateFormat, locale || moment.locale(), true)
-  return m.isValid() ? m : null
```
emilyrohrbough, line 70: Not suggesting it now, but should we someday merge date_utils.js and DateUtil.js?

packages/terra-date-picker/src/terra-dev-site/doc/date-picker/DatePicker.1.doc.mdx
```diff
@@ -16,7 +16,7 @@ import DatePickerIncomplete from '../example/DatePickerIncomplete?dev-site-examp
...
+DatePicker is an input component that provides users a way to enter or select a date from the date picker. The date format used in the input is localized based on the locale. The display of the month and days of the week in the date picker is also localized based on the locale. DatePicker leverages the [momentjs](http://momentjs.com/docs/) library for date formatting and manipulations. A date can be expressed in many [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) formats. To reduce confusion and complexity, the date props in terra-date-picker must be expressed in a specific format of the ISO 8601 representation of the date. That is, any date prop must conform to the `YYYY-MM-DD` format. Only dates between `01/01/1900` and `12/31/2100` is supported.
```
emilyrohrbough, line 19:
```suggestion
DatePicker is an input component that provides users a way to enter or select a date from the date picker. The date format used in the input is localized based on the locale. The display of the month and days of the week in the date picker is also localized based on the locale. DatePicker leverages the [momentjs](http://momentjs.com/docs/) library for date formatting and manipulations. A date can be expressed in many [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) formats. To reduce confusion and complexity, the date props in terra-date-picker must be expressed in a specific format of the ISO 8601 representation of the date. That is, any date prop must conform to the `YYYY-MM-DD` format. Only dates between `01/01/1900` and `12/31/2100` are supported.
```

packages/terra-date-picker/src/react-datepicker/date_utils.js
benbcai: I agree. Some consumers are using this directly so we should prepend the new file name with an underscore `_` to denote that it's private and not intended for reusable.

packages/terra-date-picker/CHANGELOG.md
benbcai: Maybe. I was thinking that we changed the behavior more so than fixing the underlying problem.
https://github.com/cerner/terra-framework/pull/1132/commits/f37f71c9d677bf16a39b3ee2fb62494cbd0c6c25

packages/terra-date-picker/src/react-datepicker/date_utils.js
benbcai: Definitely. I added this before creating DateUtil.isBetweenMinMaxDate since it is needed in multiple places and forgot to come back here to use it.
https://github.com/cerner/terra-framework/pull/1132/commits/f37f71c9d677bf16a39b3ee2fb62494cbd0c6c25

packages/terra-date-picker/src/react-datepicker/date_utils.js
yuderekyu: based on previous discussions, i believe we decided to use a `private` directory to denote files that are private. we've followed through in application base https://github.com/cerner/terra-application/tree/master/src/application-base/private

ShettyAkarsh:
@benbcai , In the scenario when the user select the date from the calendar dropdown i.e 05/04/2020 and using date input he edits the year to '2345'. so now the date value shows as '05/04/2345'. when the user opens the calendar the year dropdown shows as '2020' itself because of the invalid date format.


Terra-date-time-picker invalid year

so my question was shouldn't the year dropdown show the nearest Max value that the dropdown can accommodate ?  i.e '2100' in this case ? instead of showing the previously selected year ?

ShettyAkarsh:
can we tests for disabling the today button as well ?

benbcai:
> so my question was shouldn't the year dropdown show the nearest Max value that the year dropdown can accommodate ? i.e '2100' in this case ? instead of showing the previously selected year ?

@ShettyAkarsh The logic has always been that if you enter an invalid date (whether it be outside of `01/01/1900 - 12/31/2100`, outside of `minDate` and `maxDate`, is an `excludedDate`, not an `includedDate`, filtered by `filterDate`, or does not conform to the date format), that entry will be ignored and not registered as a valid date. In the scenario that you described, I don't think that we want the component to update to a date (the nearest date in this case) that the user didn't actually enter and without their acknowledgement. I think it would be safer to just leave the date where it was if it is invalid.

@PaulSolDev @neilpfeiffer @jeremyfuksa What are your thoughts from a functional/UX perspective?

benbcai:
> can we have tests added for disabling the today button as well ?

@ShettyAkarsh There are already existing tests that verify the Today button being disabled. Here is one of several examples: https://github.com/cerner/terra-framework/pull/1132/files#diff-a651f87579b2df698d08bd1be80057ed

packages/terra-date-picker/src/react-datepicker/calendar.jsx
```diff
@@ -580,12 +581,15 @@ export default class Calendar extends React.Component {
     if (!this.props.todayButton) {
       return
     }
+
+    const today = getStartOfDate(now(this.props.utcOffset));
```
supreethmr, line 585: Curious to know why we need `getStartOfDate(now(this.props.utcOffset));` when we can pass `this.state.date` as a parameter to `isDayDisabled` method.

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -122,8 +122,8 @@ const defaultProps = {
   inputAttributes: undefined,
   isIncomplete: false,
   isInvalid: false,
-  maxDate: undefined,
-  minDate: undefined,
+  maxDate: '2100-12-31',
+  minDate: '1900-01-01',
```
supreethmr, line 126: **Nit** : since we are adding default date for max and minDate I think we might not require these checks which were there to check if max and minDate are `undefined` : 

1.  [year_dropdown_options.jsx](https://github.com/cerner/terra-framework/blob/8bfdf4bd06564f2b59d3abeb52888b4c05ad5f98/packages/terra-date-picker/src/react-datepicker/year_dropdown_options.jsx#L84-L85) and [year_dropdown.jsx](https://github.com/cerner/terra-framework/blob/8bfdf4bd06564f2b59d3abeb52888b4c05ad5f98/packages/terra-date-picker/src/react-datepicker/year_dropdown.jsx#L79-L80)
```jsx
    const minYear = this.props.minDate ? this.props.minDate.year() : null
    const maxYear = this.props.maxDate ? this.props.maxDate.year() : null
```
2. [Index.jsx](https://github.com/cerner/terra-framework/blob/8bfdf4bd06564f2b59d3abeb52888b4c05ad5f98/packages/terra-date-picker/src/react-datepicker/index.jsx#L547)
```jsx
 const isDateRangePresent = ((typeof this.props.minDate !== 'undefined') && (typeof this.props.maxDate !== 'undefined'))
```

jeremyfuksa:
> I think it would be safer to just leave the date where it was if it is invalid.

Yes. Yes it would.

packages/terra-date-picker/src/react-datepicker/calendar.jsx
benbcai: @supreethmr `getStartOfDate(now(this.props.utcOffset))` always return today's date, which is what is needed to determine if the Today button should be disabled. `this.state.date` is the selected date which can be any date. Furthermore, `getStartOfDate(now(this.props.utcOffset))` is what is currently being used to set as the selected date when the Today button is clicked so I'm just reusing it here.

packages/terra-date-picker/src/DatePicker.jsx
benbcai: @supreethmr Good call! I set the defaults for minDate and maxDate in Index.jsx and removed those checks. https://github.com/cerner/terra-framework/pull/1132/commits/f65579bd2d2d938b6b5c8c9386b763de7dca98f6

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -122,8 +122,8 @@ const defaultProps = {
   inputAttributes: undefined,
   isIncomplete: false,
   isInvalid: false,
-  maxDate: undefined,
-  minDate: undefined,
+  maxDate: DateUtil.MAX_DATE,
```
supreethmr, line 125: Hmm, So is it acceptable to show default value as `DateUtil.MAX_DATE` instead of `2100-12-31` in props table.

Screen Shot 2020-06-01 at 5 00 16 PM

benbcai: I replaced with constants with the actual value. It looks like the actual value doesn't show in the prop table unless the constants are defined in the same file.
https://github.com/cerner/terra-framework/pull/1132/commits/50a1661d838b19bb8058fe44a7bfbc66212ec510

PaulSolDev:

PR #1157, opened 2020-06-15
[terra-menu] Apply theme to menu item

saket2403:
### Summary
Applied theme to menu item to fix clinical theme styles

### Deployment Link

### Additional Details

@cerner/terra

### DIFF
--- a/packages/terra-menu/src/MenuItem.jsx
+++ b/packages/terra-menu/src/MenuItem.jsx
@@ -4,11 +4,13 @@ import Arrange from 'terra-arrange';
 import CheckIcon from 'terra-icon/lib/icon/IconCheckmark';
 import ChevronIcon from 'terra-icon/lib/icon/IconChevronRight';
 import InstructionsForUseIcon from 'terra-icon/lib/icon/IconConsultInstructionsForUse';
-import classNames from 'classnames/bind';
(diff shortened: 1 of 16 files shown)


packages/terra-menu/src/MenuItem.jsx
```diff
@@ -224,9 +226,13 @@ class MenuItem extends React.Component {
     }
 
     return (
-      <li {...attributes} className={itemClassNames} ref={this.setItemNode} role={role} aria-checked={markAsSelected}>
-        {content}
-      </li>
+      <ThemeContext.Consumer>
```
emilyrohrbough, line 229: I'm not familiar with contextTypes, but is this something that could be provided via the contextTypes like `isGroupItem` and `isSelectableMenu`?

mjhenkes: You can only read one context at a time with contextTypes. Since it's taken, we need to wrap.

mjhenkes:
So you're applying the theme context here, but you still need to move the theme variables and reference them in the css files right?

saket2403:
> So you're applying the theme context here, but you still need to move the theme variables and reference them in the css files right?

Its there already, I had moved it while applying theme to menu.

PR #1160, opened 2020-06-22
Implement request animation frame to correct loop limit error.

### LINKED ISSUE #1188: Investigate ResizeObserver Errors
# Bug Report
We're seeing "ResizeObserver loop limit exceeded" errors pretty frequently in our New Relic logs.  It's occurred on the following page https://engineering.cerner.com/terra-ui/components/terra-form-input/form-input/input. We need to see if there's a problem and how to fix it.

pranav300:
### Summary

Updated application navigation, and infinite-list to use animation frames.

Closes #1188 

### Deployment Link

### Additional Details

@cerner/terra


tbiethman:
There is a functional difference in how initial layout happens after this change. The gifs show what happens after a page refresh. Ideally this would stay the same and not show the initial layout.

Before:
old_nav

After:
new_nav

ryanthemanuel:
@pranav300 can you please fix the jest tests?

packages/terra-application-navigation/src/tabs/_Tabs.jsx
```diff
@@ -139,33 +143,16 @@ class Tabs extends React.Component {
...
-      this.resizeObserver.disconnect(this.containerRef.current);
-      this.resizeListenerAdded = false;
-    }
-  }
-
   handleResize(width) {
     // Calculate hide index
     const childrenCount = this.props.navigationItems.length;
-    const moreWidth = width - this.getRollupTabWidth();
+    const moreWidth = this.rollupTabRef.current ? width - this.getRollupTabWidth() : width;
```
dkasper-was-taken, line 155: Instead of inlining it here I might make more sense to put this check in the `getRollupTabWidth` method, where it would return zero if it wasn't present.

pranav300: Updated here https://github.com/cerner/terra-framework/pull/1160/commits/16a7e89d88caf49e801c8cf21fab8dceeaad2f09.

PR #1164, opened 2020-06-25
Fix incompatibility issue between aria-haspopup and link role.

### LINKED ISSUE #1181: [spike] Investigate jsx-a11y/role-supports-aria-props ignore on Tabs
# Bug Report

## Description
In order to quickly get back to a passing CI build we have set the [_TabRollup.jsx](https://github.com/cerner/terra-framework/blob/main/packages/terra-tabs/src/_TabMenu.jsx) file to ignore jsx-a11y/role-supports-aria-props. We need to investigate if this is a problem and how to correct it.

## Steps to Reproduce

1. Remove the jsx-a11y/role-supports-aria-props ignore from _TabRollup.jsx
2. Run eslint
3.
4.

## Expected Behavior
No failures and an accessible tab control

https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html

\- [Dave Kasper](https://github.com/cerner/terra-framework/pull/1166#discussion_r448451433)

## Environment

* Component Name and Version:
* Browser Name and Version:
* Webpack Version:

saket2403:
### Summary
This PR Fixes the jsx-a11y issue between aria-haspopup and role link. As jsx-eslint was upgraded to use wai-aria 1.2  and [here](https://www.w3.org/TR/wai-aria-1.2/#aria-haspopup) is the new features/rule changes wrt aria-haspopup which was changed from global to a restricted attribute.

Fixes #1181

### Deployment Link

### Testing
Updated Jest snapshots.

### Additional Details

@cerner/terra


emilyrohrbough:
@saket2403  was there an issue logged for this or a test that needed updated that disabled this rule?

saket2403:
> @saket2403 was there an issue logged for this or a test that needed updated that disabled this rule?

No issue was logged neither was any test updated that disabled the rule, recent upgrade of jsx-eslint to use wai-ari a1.2 where few rules were changed, introduced this issue. If the fix does not fit accessibility criteria, we can look at disabling the rule.

emilyrohrbough:
@saket2403 Was this a lint warning or you saw the new rules being applied? I think the change looks fine if the screen reader & keyboard interactions remain consistent

saket2403:
Yes it was a lint warning, keyboard interactions are fine , there are slight changes in screen reader behaviour which is why I have added accessibility review label.

emilyrohrbough:
@saket2403 Thanks for the extra info!

supreethmr:
I think removing 'aria-haspopup` will have impact on the menu-buttons of navigation-side-menu. `aria-haspopup` was added on menu-buttons having sub-menu. So that it helps VO to differentiate between normal menu-button and menu-button with sub-menu items (chevron icon). 

Since `aria-haspopup` Indicates the button opens a menu. I think we might need to re-think on removing  of `aria-haspopup`. we can try replacing role `link` with suitable role for 'aria-haspopup` OR might need to add label for menu-buttons with sub-menu to make it accessible for VO.

saket2403:
> I think removing 'aria-haspopup`will have impact on the menu-buttons of navigation-side-menu.`aria-haspopup` was added on menu-buttons having sub-menu. So that it helps VO to differentiate between normal menu-button and menu-button with sub-menu items (chevron icon).
> 
> Since `aria-haspopup` Indicates the button opens a menu. I think we might need to re-think on removing of `aria-haspopup`. we can try replacing role `link` with suitable role for 'aria-haspopup` OR might need to add label for menu-buttons with sub-menu to make it accessible for VO.

I had tried changing role link to menuitem but it requires a parent role which might or might not be there so it was giving accessibility issue in wdio. We can take a look at adding labels

packages/terra-application-navigation/src/tabs/_TabRollup.jsx
```diff
@@ -89,7 +89,6 @@ const TabRollup = ({
       onMouseDown={disableFocusStyles}
       ref={tabRef}
       aria-current={isSelected}
-      aria-haspopup
```
tbiethman, line 92: Would it make more sense to remove the role of link instead? This is the dropdown button that doesn't do any navigation when pressed.

saket2403: Since a role is required , I am guessing `button` should  work instead of link, also it is compatible with aria-haspopup

packages/terra-navigation-side-menu/src/_MenuItem.jsx
```diff
@@ -133,7 +133,6 @@ class MenuItem extends React.Component {
           onKeyDown={this.handleKeyDown}
           onKeyUp={this.handleKeyUp}
           onBlur={this.handleOnBlur}
-          aria-haspopup={hasChevron}
```
dkasper-was-taken, line 136: The fix here would likely not necessarily be to remove aria-haspopup. Only end points should use the 'link' role. Combined with this, it would make more sense to convert the list to use menu/menuitem roles, and the drill-in item use aria-haspopup.

packages/terra-application-navigation/src/tabs/_TabRollup.jsx
dkasper-was-taken: 'button' should be fine here. Though long term I think it might make more sense to use 'menubutton' combined with 'menu/menuitem' in the popup, but that could end up very different as UX is debating the viability and behavior of rollup patterns across cerner.

packages/terra-navigation-side-menu/src/_MenuItem.jsx
dkasper-was-taken: Might do some screen reader testing to see if the elements should be switched to `<a>` tags as well, or if divs are still fine.

saket2403: A combination of menu and menu-item roles works fine. As for the `<div>` and `<a>` tags , no difference in VO.

packages/terra-navigation-side-menu/src/_MenuItem.jsx
```diff
@@ -125,9 +124,10 @@ class MenuItem extends React.Component {
     return (
       <li
         className={cx('list-item')}
+        role="none"
```
yuderekyu, line 127: i'm curious, what happens if we don't explicitly set role to none here?

saket2403: If role is not set explicitly it gives an axe issue `Ensures <li> elements are used semantically`. Also here is [link](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html) for reference

PR #1187, opened 2020-07-13
[terra-application-navigation] Make Tab text style consistent with notifications.

### LINKED ISSUE #1163: [terra-application-navigation] Notification count changes size of nav bar and shifts text
# Feature Request

## Description

The top navigation bar has a different height and the text jumps significantly when going from a state of no notification counts to having counts, or vice versa. As a consumer, I am requesting that the height of the top navigation bar stay the same and the text stay in the same position regardless of the presence of notification counts.

lokesh-0813:
### Summary

Notifications initialised with "0" value do not apply notification styles to tab and results in text having different styles when provided with a non-zero value. This is easily noticeable with providing notification count to a single tab and changing it programmatically.  

Closes #1163 

### Deployment Link

### Testing

Updated Jest. 

### Additional Details

@cerner/terra


ShettyAkarsh:
@lokesh-0813 , can you add the test for this scenario ?

lokesh-0813:
> @lokesh-0813 , can you add the test for this scenario ?

Added here 5e77c28efddb2f17986acd923b307e55aa51ec5c

PR #1207, opened 2020-08-05
Remove babel-eslint dev dep

### LINKED ISSUE #1203: Remove babel-eslint dependency
# Feature Request
Remove babel-eslint dependency from package.json and eslintConfig in order to use the transitive dependency and default parser provided by eslint-config-terra instead.

## Description

eslint-config-terra now contains the dependency on `babel-eslint` and defaults it as our lint parser. we can remove those from this repo now.

### Remove
https://github.com/cerner/terra-framework/blob/main/package.json#L54
https://github.com/cerner/terra-framework/blob/main/package.json#L117

yuderekyu:
### Summary

eslint-config-terra now contains the dependency on babel-eslint and defaults it as our lint parser. we can remove those from this repo now.

Closes #1203 

### Additional Details

@cerner/terra

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,12 +50,6 @@
     ],
     "rules": {
       "react/forbid-prop-types": "off"
-    },
-    "settings": {
-      "polyfills": [
-        "Promise",
-        "Map"
-      ]
     }
   },
   "stylelint": {
--- a/package.json
+++ b/package.json
@@ -51,7 +51,6 @@
     "rules": {
       "react/forbid-prop-types": "off"
     },
-    "parser": "babel-eslint",
     "settings": {
       "polyfills": [
         "Promise",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -51,7 +51,6 @@
     "rules": {
       "react/forbid-prop-types": "off"
     },
-    "parser": "babel-eslint",
     "settings": {
```
emilyrohrbough, line 54: Can we remove these settings as well? These are set in the default config as well: https://github.com/cerner/terra-toolkit/blob/main/packages/eslint-config-terra/eslint.config.js#L43

yuderekyu: https://github.com/cerner/terra-framework/pull/1207/commits/5611ac081c53b9c1764e25aa6a8ce09632bd7838

PR #1216, opened 2020-08-17
[terra-date-picker] Remove dotted border being added on hovering selected date.

### LINKED ISSUE #1215: [date-picker] : dotted border gets added to selected day / active day on mouse enter
# Bug Report
dotted border gets added to selected day / active day on mouse enter when calendar is opened via keyboard.

## Steps to Reproduce
1. navigate to default date picker example : https://engineering.cerner.com/terra-ui/components/terra-date-picker/date-picker/date-picker
2. click on date input and press TAB to access calendar button 📅  of date-picker component.
3. press ENTER to open a calendar and move cursor into days region.
4. dotted border gets added to current day.

date-picker-hover-inconsistency

## Expected Behavior
should not add dotted border on mouse enter. 

## Possible Solution
I suspect the focus styles ( dotted border on day ) would be appearing due to the re-render that caused by state updates on [mouseEnter](https://github.com/cerner/terra-framework/blob/d3d47c6515f352e306f0e69958e06cad787b1b31/packages/terra-date-picker/src/react-datepicker/calendar.jsx#L437). So we might need to evaluate the use case of setting selectingDate on MouseEnter and we can rid of this event if not required to prevent focus styles from getting appeared on mouse enter.

OR

in calendar.jsx value of `keyboardFocus` is set to prop `isCalendarKeyboardFocused` which is part of [condition statement](https://github.com/cerner/terra-framework/blob/d3d47c6515f352e306f0e69958e06cad787b1b31/packages/terra-date-picker/src/react-datepicker/day.jsx#L222) that adds dotted-border around day.

`keyboardFocus` is being set on line [642](
https://github.com/cerner/terra-framework/blob/d3d47c6515f352e306f0e69958e06cad787b1b31/packages/terra-date-picker/src/react-datepicker/calendar.jsx#L642) depending on the props `isCalendarOpenedViaKeyboard` OR `isCalendarKeyboardFocused` value, which will be true when calendar is opened via keyboard. 

here we might need to consider state 'this.state.calendarIsKeyboardFocused` (in this [condition](https://github.com/cerner/terra-framework/blob/d3d47c6515f352e306f0e69958e06cad787b1b31/packages/terra-date-picker/src/react-datepicker/calendar.jsx#L641)) before setting value for `keyboardFoucs` so that focus styles will appear only when calendar is focused keyboard interactions like TAB or Arrow keys.
```diff
+     if((this.props.isCalendarOpenedViaKeyboard || this.props.isCalendarKeyboardFocused) && this.state.calendarIsKeyboardFocused) {
      keyboardFocus = true;
    }

-  if(!keyboardFocus && this.state.calendarIsKeyboardFocused) {
-      keyboardFocus = true;
-   }
```

## Environment

* Component Name and Version: terra-date-picker 

saket2403:
### Summary
Fix to keep dotted border on selected date when calendar is opened via keyboard, instead of it appearing on hover after opening via keyboard.

Closes #1215 

### Deployment Link

https://terra-framew-date-borde-d5q4dv.herokuapp.com/components/terra-date-picker/date-picker/date-picker

### Additional Details

@cerner/terra


packages/terra-date-picker/src/react-datepicker/day.jsx
```diff
@@ -219,7 +219,7 @@ class Day extends React.Component {
     return ['react-datepicker-day', dayClassName, 'react-datepicker-day--' + getDayOfWeekCode(this.props.day), {
       'react-datepicker-day--disabled': this.isDisabled(),
       'react-datepicker-day--selected': this.isSameDay(this.props.selected) && !isDayDisabled(this.props.selected, this.props),
-      'react-datepicker-day--selected-border': this.isSameDay(this.props.preSelection) && !isDayDisabled(this.props.selected, this.props) && (document.activeElement.tagName === 'DIV' || document.activeElement === document.querySelector('[class*="previous"]')) && this.props.isCalendarKeyboardFocused,
+      'react-datepicker-day--selected-border': this.isSameDay(this.props.preSelection) && !isDayDisabled(this.props.selected, this.props) && (document.activeElement === document.querySelector('[aria-label="Open Calendar"]') || document.activeElement.tagName === 'DIV' || document.activeElement === document.querySelector('[class*="previous"]')) && this.props.isCalendarKeyboardFocused,
```
supreethmr, line 222: the new condition will not work for other locales. when aria-label gets translated string.

supreethmr:
Focus border seems to applied (on keyboard open) only on default example. In other examples border still gets added on mouse entering the calendar region.

date_picker_focus

supreethmr:
dotted border is added only when calendar is opened via Enter and seems to be ignored when opened via space key : 
date_picker_focus_space

packages/terra-date-picker/src/react-datepicker/day.jsx
saket2403: using data-attribute instead updated here 92c0e73

saket2403:
> Focus border seems to applied (on keyboard open) only on default example. In other examples border still gets added on mouse entering the calendar region.

> dotted border is added only when calendar is opened via Enter and seems to be ignored when opened via space key :

Fixed 92c0e73

packages/terra-date-picker/src/react-datepicker/day.jsx
```diff
@@ -219,7 +219,7 @@ class Day extends React.Component {
     return ['react-datepicker-day', dayClassName, 'react-datepicker-day--' + getDayOfWeekCode(this.props.day), {
       'react-datepicker-day--disabled': this.isDisabled(),
       'react-datepicker-day--selected': this.isSameDay(this.props.selected) && !isDayDisabled(this.props.selected, this.props),
-      'react-datepicker-day--selected-border': this.isSameDay(this.props.preSelection) && !isDayDisabled(this.props.selected, this.props) && (document.activeElement.tagName === 'DIV' || document.activeElement === document.querySelector('[class*="previous"]')) && this.props.isCalendarKeyboardFocused,
+      'react-datepicker-day--selected-border': this.isSameDay(this.props.preSelection) && !isDayDisabled(this.props.selected, this.props) && (document.activeElement.hasAttribute('data-terra-open-calendar-button') || document.activeElement.tagName === 'DIV' || document.activeElement === document.querySelector('[class*="previous"]')) && this.props.isCalendarKeyboardFocused,
```
benbcai, line 222: In small breakpoints when you open the picker in portal mode with keyboard, the day is correctly styled with the dotted focus style. If you dismiss the portal picker and click on the calendar button to open it again, it continues to have the focus style applied to the day. If you then resize to a larger breakpoint and open the popup picker by clicking the calendar button, again the day continues to have the focus style when it shouldn't.

saket2403: Fixed 111fd4c

benbcai: The logic on line 222 is getting a bit long and hard to understand. Could we possible wrap some of that in a function to make it easier to understand the logic?

saket2403: Done e892b1d

packages/terra-date-picker/src/react-datepicker/index.jsx
```diff
@@ -586,7 +592,7 @@ class DatePicker extends React.Component {
   }
 
   onInputKeyDown = (event) => {
-    if(event.key === 'Enter') {
+    if(event.keyCode === 13 || event.keyCode === 32) {
```
benbcai, line 595: What is the reason for adding the `SPACE` key here? Should this be updated to get the keyCode from `nativeEvent ` and use `KEY_RETURN` and `KEY_SPACE` from keycode-js similar to how [other places](https://github.com/cerner/terra-framework/blob/ca8fde828d309978e69ab801d59044f5e71626b0/packages/terra-application-navigation/src/common/_PopupMenuListItem.jsx#L8-L9) use it?

saket2403: Border was not getting added when opening via space, so added space key also.([comment](https://github.com/cerner/terra-framework/pull/1216#issuecomment-675417899)). Updated to keycode eb41920

benbcai (approved):

packages/terra-date-picker/src/react-datepicker/day.jsx
```diff
@@ -214,12 +214,23 @@ class Day extends React.Component {
       this.props.month !== getMonth(this.props.day)
   }
 
+  isBorderApplied = () => {
+    if(this.isSameDay(this.props.preSelection) && !isDayDisabled(this.props.selected, this.props) 
+    && (document.activeElement.hasAttribute('data-terra-open-calendar-button')
```
emilyrohrbough, line 219: If there are multiple date-pickers on the page, is there any risk that this will apply the border to both?

emilyrohrbough (approved):

mjhenkes (approved):

PR #1229, opened 2020-08-31
[terra-time-input] Fix to allow only 24 hour format for `es` locale

### LINKED ISSUE #1227: [terra-time-input] Only allow 24-hour notation for the `es` language
# Bug Report

## Description
The `es` (spanish) language currently honors the `variant` prop if set to `12-hour` to show the time in the 12-hour notation. The `es` language should only allow the default `24-hour` notation and should ignore the `variant` prop just like other languages (`fr`, `pt`, `sv`) that do not support the 12-hour notation.

## Steps to Reproduce
1.  Go to https://engineering.cerner.com/terra-ui/components/terra-time-input/time-input/time-input
2. Set the locale to `es`
3. Notice that the examples set to the 12-hour variant.

## Possible Solution
A quick solution to this is changing the ["Terra.timeInput.am" and "Terra.timeInput.pm"](https://github.com/cerner/terra-framework/blob/main/packages/terra-time-input/translations/es.json#L2-L3) strings  for `es` to an empty string to allow only the 23-hour notation.

The current logic that relies on this string is a bit confusing and not obvious. We should clearly define in a list which locale can support the 12-hour notation. If a locale is not in this list, then it should only support the 24-hour notation.

## Environment

* Component Name and Version:
* Browser Name and Version:
* Webpack Version:

saket2403:
### Summary
Time Input to allow only 24 hour notation for es locale.

Closes #1227 

### Deployment Link

[Deployment Link](https://terra-framew-time-es-qukm80u2j.herokuapp.com/components/terra-time-input/time-input/time-input)

### Additional Details

@cerner/terra


avinashg1994:
The variant description should also be updated to include `es` in [TimeInput.jsx - L91](https://github.com/cerner/terra-framework/blob/main/packages/terra-time-input/src/TimeInput.jsx#L91) and [DateTimePicker - L119](https://github.com/cerner/terra-framework/blob/main/packages/terra-date-time-picker/src/DateTimePicker.jsx#L119)

saket2403:
> The variant description should also be updated to include `es` in [TimeInput.jsx - L91](https://github.com/cerner/terra-framework/blob/main/packages/terra-time-input/src/TimeInput.jsx#L91) and [DateTimePicker - L119](https://github.com/cerner/terra-framework/blob/main/packages/terra-date-time-picker/src/DateTimePicker.jsx#L119)
> 
> Also, i guess the following jest test should also be updated:
> https://github.com/cerner/terra-framework/blob/main/packages/terra-time-input/tests/jest/TimeInput.test.jsx#L322-L332

Updated 7bf2148

packages/terra-time-input/translations/es.json
```diff
@@ -1,6 +1,6 @@
 {
-  "Terra.timeInput.am": "a. m.",
-  "Terra.timeInput.pm": "p. m.",
+  "Terra.timeInput.am": "",
+  "Terra.timeInput.pm": "",
```
benbcai, line 3: The [current logic](https://github.com/cerner/terra-framework/blob/main/packages/terra-time-input/src/TimeUtil.js#L262-L271) that relies on the existence of the Terra.timeInput.am string to determine the variant is confusing and not obvious. We should clearly define in a list which locale can support the 12-hour notation and check against this list. If a locale is not in this list, then it should only support the 24-hour notation.

These are the locales that have a non-empty string for `Terra.timeInput.am`/`Terra.timeInput.pm` so they are the ones that can support the 12-hour notation.

- en-AU
- en-CA
- en-GB
- en-US
- en
- es-US

saket2403: Updated 4597abe

packages/terra-time-input/tests/jest/TimeInput.test.jsx
```diff
@@ -324,7 +324,7 @@ it('should validate the getVariantFromLocale helper method', () => {
   const wrapper = mountWithIntl(timeInput);
   const testComponent = wrapper.children();
 
-  if (['ar', 'en', 'en-AU', 'en-CA', 'en-GB', 'en-US', 'es', 'es-US'].includes(testComponent.instance().props.intl.locale)) {
+  if (['ar', 'en', 'en-AU', 'en-CA', 'en-GB', 'en-US', 'es-US'].includes(testComponent.instance().props.intl.locale)) {
```
supreethmr, line 327: Can [this](https://github.com/cerner/terra-framework/blob/4597abe4c622d4d9ed83f7d3621e5168206bbd55/packages/terra-time-input/src/TimeInput.jsx#L158) also be updated now..?
```js
 if (TimeUtil.getVariantFromLocale(props) === TimeUtil.FORMAT_12_HOUR) {
    - if (!this.props.intl.messages['Terra.timeInput.am'] || !this.props.intl.messages['Terra.timeInput.pm']) {
       }
    }
```
To

```JS
 if (TimeUtil.getVariantFromLocale(props) !== TimeUtil.FORMAT_12_HOUR) {

 }
```

saket2403: Control was never going inside the condition so, removed it entirely and moved warning to separate condition f7bc706

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -176,6 +166,12 @@ class TimeInput extends React.Component {
         meridiem = this.anteMeridiem;
       }
     }
+    if (this.props.variant === TimeUtil.FORMAT_12_HOUR && TimeUtil.getVariantFromLocale(props) === TimeUtil.FORMAT_24_HOUR) {
+      if (process.env !== 'production') {
+        // eslint-disable-next-line no-console
+        console.warn('This locale only uses 24 hour clock. The ante meridiem and post meridiem will not be displayed');
+      }
```
ShettyAkarsh, line 173: we will need to reset these below variables here. to avoid carry forwarding old meridiem values.

this.anteMeridiem = '';	 
this.postMeridiem = '';

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -155,18 +155,8 @@ class TimeInput extends React.Component {
     let meridiem;
 
     if (TimeUtil.getVariantFromLocale(props) === TimeUtil.FORMAT_12_HOUR) {
```
ShettyAkarsh, line 157: Instead of having two separate if block we can alter this condition like this.

if (this.props.variant === TimeUtil.FORMAT_12_HOUR) {
   if (TimeUtil.getVariantFromLocale(props) === TimeUtil.FORMAT_24_HOUR)) {
    ......................;
   }
  else {
    .....................;
  }
};

packages/terra-time-input/src/TimeInput.jsx
saket2403: Done

packages/terra-time-input/src/TimeUtil.js
```diff
@@ -261,8 +261,9 @@ class TimeUtil {
    */
   static getVariantFromLocale(props) {
     const { intl, variant } = props;
+    const locale12Hr = ['en-AU', 'en-CA', 'en-GB', 'en-US', 'en', 'es-US'];
```
benbcai, line 264: After thinking about this a little more, instead of creating a list of the locales that support both notations it would be better to list the locales that only support the 24-hour notation. The reason being that any new locales we support in the future should by default honor the variant. If we specifically determined that it should only allow 24-hour notation, then we would add it here. Basically, this should be the _denied_ list instead of the _allowed_ list.

saket2403: Changed f773022

packages/terra-time-input/src/TimeUtil.js
```diff
@@ -261,9 +261,9 @@ class TimeUtil {
    */
   static getVariantFromLocale(props) {
     const { intl, variant } = props;
-    const locale12Hr = ['en-AU', 'en-CA', 'en-GB', 'en-US', 'en', 'es-US'];
+    const locale24Hr = ['de', 'es-ES', 'es', 'fi-FI', 'fr-FR', 'fr', 'nl-BE', 'nl', 'pt-BR', 'pt', 'sv-SE', 'sv'];
```
emilyrohrbough, line 264: This will lock the date-time picker to only support terra-defined locales. If someone wants to implement a custom locale, this might  not be correct. is that okay?

saket2403: So should this be reverted back to 12 hr list, or remove list completely and use empty translation logic as used previously. Your thoughts @benbcai ?

benbcai: This logic is saying that the locales defined in this locale are the ones we know that should only use 24-hour notation. All other locales (custom or the other ones we support i.e. en, en-US, en-AU, etc) will continue to honor the variant however it is set.

avinashg1994 (approved):

mjhenkes (approved):

Issue #1244, opened 2020-09-14
[terra-date-input] Cleanup aria-describedby format mask, help, error ids


neilpfeiffer:
## Description

How the aria-describedby is getting populated is creating some unideal markup at render:
#### Source
https://github.com/cerner/terra-framework/blob/0cbfdf5745af49f9e3b4273a34253078e1dd9aed/packages/terra-date-input/src/DateInput.jsx#L500
https://github.com/cerner/terra-framework/blob/0cbfdf5745af49f9e3b4273a34253078e1dd9aed/packages/terra-date-input/src/DateInput.jsx#L554
https://github.com/cerner/terra-framework/blob/0cbfdf5745af49f9e3b4273a34253078e1dd9aed/packages/terra-date-input/src/DateInput.jsx#L593

----
#### Results  

In the **DateInput**, when the additional `monthAttributes`, `dayAttributes` and `yearAttributes` aren't provided they are included as `undefined` :
```
 aria-describedby="terra-date-picker-description-format-e9b925bd-ab19-4345-9c5a-94e924d3dd9f undefined"
```
<img width="782" alt="Screen Shot 2020-09-12 at 8 51 46 PM" src="https://user-images.githubusercontent.com/1355485/93040266-2e4d9480-f60f-11ea-8387-4dfd7f30eded.png">
<img width="767" alt="Screen Shot 2020-09-12 at 8 51 37 PM" src="https://user-images.githubusercontent.com/1355485/93040331-6523aa80-f60f-11ea-932c-032ecc887be1.png">
<img width="777" alt="Screen Shot 2020-09-12 at 8 51 29 PM" src="https://user-images.githubusercontent.com/1355485/93040339-681e9b00-f60f-11ea-831b-f0e108426b0f.png">

\
\
And **DateInputField** results in two spaces before the help ID, and one space after:  
```
   aria-describedby="  terra-date-input-field-description-help-8f589be7-5a7e-4313-8d52-140351fd3213 " 
```
<img width="842" alt="Screen Shot 2020-09-12 at 8 52 42 PM" src="https://user-images.githubusercontent.com/1355485/93040480-bcc21600-f60f-11ea-8dfd-cd645868d07d.png">

Or an extra space before the error ID, a space between the error and help (expected), and extra space after the help ID:
```
[...]

Issue #1193, opened 2020-07-17
Remove placeholders 


benbcai:
# Bug Report

## Description
The contrast of the color of placeholders with their background does not meet the minimum color contrast ratio of 4.5:1. Darkening the placeholder color increases the risk of making it appear as actual entered value. For this reason, placeholders in `terra-date-input`, `terra-date-picker`, `terra-time-input` should be removed and replaced with help text. 

[Comment](https://github.com/cerner/terra-core/issues/2992#issuecomment-651124550) originally from https://github.com/cerner/terra-core/issues/2992 was moved to this issue for guidance: 
See https://github.com/cerner/terra-framework/issues/1193#issuecomment-660363635

## Environment

* Component Name and Version:
* Browser Name and Version:
* Node/npm Version: [e.g. Node 8/npm 5]
* Webpack Version:
* Operating System and version (desktop or mobile):

benbcai:
The alteration direction below is copied from https://github.com/cerner/terra-core/issues/2992#issuecomment-651135365


Alteration directions for:
- terra-date-input
- terra-date-input/lib/DateInputField
- terra-date-picker
- terra-date-picker/lib/DatePickerField
- terra-time-input
- terra-date-time-picker

---
- [ ] The strings (the localized letter abbreviation date/time format masks, e.g. ) that are provided to the text inputs and the `placeholder` prop/html attributes need to be removed. The Date-Input component should remove "Month" as the placeholder text in the select and the text "- Select -" should be used instead to match the form-select components.
- [ ] For each of the base components (non-Field compositions), a new element that matches the styling of the help text prop in terra-form-field needs to be added, and localized letter abbreviation date/time format masks originally put in the placeholder (`MM/DD/YYYY`, `hh:mm:ss`, `Month DD, YYYY`)  is now given to the help text placement 
- [ ] For the component that do not have Field compositions (Time-Input and Date-Time-Picker) need to add Field compositions. All the Field compositions need to remove the help text of the format letter abbreviation masks in the stand-alone/imported inputs, and need to supply the text to the actual `help` prop.  Format notes: If no additional help text is provided, the help text is only the provided format mask e.g. `MM/DD/YYYY`. If additional help text is provided, the format mask should be supplied first at `(MM/DD/YYYY)` followed by a space, followed by the text given to the `help` prop.
- [ ] Remove any css styles and variables for the placeholder in the input elements (exception for Date-Input which will keep the "- Select -" placeholder in the dropdown select).
[...]

PR #1252, opened 2020-09-21
[terra-date-picker] Fix accessibility issues.

pranav300:
### Summary

This PR removes the duplicate id `format` from `DatePickerField`. And cleans up the aria description id's for better accessibility.

Part of #1193

Similar to #1244  

### Deployment Link

### Testing

Added a new test with multiple Datepickerfield.
Added wdio test for the same

### Additional Details

Current behavior:

old

Updated behavior:
new

@cerner/terra


packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -195,7 +195,13 @@ class DatePickerInput extends React.Component {
 
     const label = this.props.ariaLabel ? this.props.ariaLabel : intl.formatMessage({ id: 'Terra.datePicker.date' });
 
-    this.formatDescriptionId = (useExternalFormatMask === false) ? `terra-date-picker-description-format-${this.uuid}` : '';
+    if (useExternalFormatMask === false) {
+      this.formatDescriptionId = `terra-date-picker-description-format-${this.uuid}`;
+    } else if (inputAttributes['aria-describedby']) {
```
supreethmr, line 200: why we need `inputAttributes['aria-describedby']` as an Id here...??? since [format-div ](https://github.com/cerner/terra-framework/blob/ee1e2890c5b5f39b5966708110c07e68a27b05b7/packages/terra-date-picker/src/DateInput.jsx#L246) will only be rendered when  `useExternalFormatMask` is false.

pranav300: It is there for [aria-describedby](https://github.com/cerner/terra-framework/blob/ee1e2890c5b5f39b5966708110c07e68a27b05b7/packages/terra-date-picker/src/DateInput.jsx#L229) in Input to read the help and error message coming from datepickerfield.

If you see the latest version, the voiceover doesn't read the help and error message in datepickerfield. That is due to typo [here](https://github.com/cerner/terra-framework/blob/main/packages/terra-date-picker/src/DatePickerField.jsx#L270) in

packages/terra-date-picker/src/DatePickerField.jsx
```diff
@@ -217,16 +217,10 @@ const DatePickerField = (props) => {
 
   let ariaDescriptionIds;
 
-  if (help && error && isInvalid) {
+  if (error && isInvalid) {
```
emilyrohrbough, line 220: I think the help condition is needed if we are applying `${datePickerId}-help` to `ariaDescriptionIds`.

emilyrohrbough: We could probably break this out to two single checks like: 
```
    if (help) {
      ariaDescriptionIds = `${ariaDescriptionIds} ${datePickerId}-help`;
    }

    if (error && isInvalid) {
      ariaDescriptionIds = `${ariaDescriptionIds} ${datePickerId}-error`;
    }
```
not sure if that is simpler or not

packages/terra-date-picker/src/DateInput.jsx
emilyrohrbough: Do you have a small video clip of the current vs fixed behavior that demonstrates this?

benbcai: Should `useExternalFormatMask` have higher precedence over `aria-describedby` or the other way around? `useExternalFormatMask` will always be false when not using DatePickerField and therefore, it will always use the uuid and ignores `aria-describedby` even if it is provided.

packages/terra-date-picker/src/DatePickerField.jsx
benbcai: I think it should explicitly check for `help` if we're adding `${datePickerId}-help`.

packages/terra-date-picker/src/DatePickerField.jsx
```diff
@@ -267,7 +261,7 @@ const DatePickerField = (props) => {
       <DatePicker
         disabled={disabled}
         id={datePickerId}
-        inputAttribute={mergedInputAttrs}
+        inputAttributes={mergedInputAttrs}
```
benbcai, line 264: 👍

packages/terra-date-picker/src/DatePickerField.jsx
pranav300: What my idea was to always provide `help` for `DatePickerField`, even if the user doesn't provide any help message still format will be there.

packages/terra-date-picker/src/DateInput.jsx
pranav300: Added a check for when both are true here https://github.com/cerner/terra-framework/pull/1252/commits/ef5efe7b86697d7565cee87caa635627f42c26b1.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -195,7 +195,15 @@ class DatePickerInput extends React.Component {
 
     const label = this.props.ariaLabel ? this.props.ariaLabel : intl.formatMessage({ id: 'Terra.datePicker.date' });
 
-    this.formatDescriptionId = (useExternalFormatMask === false) ? `terra-date-picker-description-format-${this.uuid}` : '';
+    if (useExternalFormatMask === false && inputAttributes && inputAttributes['aria-describedby']) {
+      this.formatDescriptionId = `terra-date-picker-description-format-${this.uuid} ${inputAttributes['aria-describedby']}`;
```
benbcai, line 199: Would this add an empty space between the uuid and aria-describedby?

benbcai: I think the issue from [this comment](https://github.com/cerner/terra-framework/pull/1252#discussion_r492108887) still exists with this new change.  `useExternalFormatMask` defaults to false for DatePicker and if the consumer provides `aria-describedby`, we are prepending a uuid to its aria-describedby and pass that down to Input. Consumers probably wouldn't expect their `aria-describedby` to get updated.

Or maybe I'm not understanding what this is trying to do.

yuderekyu: just curious, is the top most if block the most likely scenario? it's ideal if the conditionals follow most likely to least likely :D

pranav300: > I think the issue from this comment still exists with this new change. useExternalFormatMask defaults to false for DatePicker and if the consumer provides aria-describedby, we are prepending a uuid to its aria-describedby and pass that down to Input. Consumers probably wouldn't expect their aria-describedby to get updated.
Or maybe I'm not understanding what this is trying to do.

So, earlier we didn't have any description id's being passed down to the raw component from the field component due to `inputAtrributes`.
Correct me if I am wrong but I believe that we should read the `format, error and help` text, and then any of the user sent `aria-describedby`.


cc / @neilpfeiffer

pranav300: >just curious, is the top most if block the most likely scenario? it's ideal if the conditionals follow most likely to least likely :D

If we want it in if-else order then I think this order of condition is the best. But we can also provide single if blocks, and take care of preference there.

pranav300: >Would this add an empty space between the uuid and aria-describedby?

Yes, it would add spaces between them only when `aria-describedby` is provided.

And for readability I can change the name of `this.formatDescriptionId` to `this.ariaDescriptionId`.

What do you think? @benbcai

benbcai: > > Would this add an empty space between the uuid and aria-describedby?
> 
> Yes, it would add spaces between them only when `aria-describedby` is provided.
> 
> And for readability I can change the name of `this.formatDescriptionId` to `this.ariaDescriptionId`.
> 
> What do you think? @benbcai

I just think it's weird that the `id` as a string has two separate strings.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -195,7 +195,15 @@ class DatePickerInput extends React.Component {
...
+    } else if (useExternalFormatMask === false) {
+      this.formatDescriptionId = `terra-date-picker-description-format-${this.uuid}`;
+    } else if (inputAttributes && inputAttributes['aria-describedby']) {
+      this.formatDescriptionId = inputAttributes['aria-describedby'];
+    } else {
+      this.formatDescriptionId = '';
```
benbcai, line 205: Given the change made in DatePickerField where `aria-describedby` will always be provided, it should never get in here. If it ever gets in here, should this be set to `undefined` instead of an empty string so that aria-describedby is not defined instead of setting to an empty string?

packages/terra-date-picker/src/DateInput.jsx
pranav300: Updated here https://github.com/cerner/terra-framework/pull/1252/commits/b10e52f5589813e424d2e9a198e285884d7eaef4

packages/terra-date-picker/src/DateInput.jsx
pranav300: Updated here https://github.com/cerner/terra-framework/pull/1252/commits/fbcd7ea00aed370b88d57583ddaffc66a755f9f7

packages/terra-date-picker/src/DatePickerField.jsx
```diff
@@ -217,33 +217,29 @@ const DatePickerField = (props) => {
...
-      ariaDescriptionIds = `${datePickerId}-help`;
-    }
-
-    if (error && isInvalid) {
-      ariaDescriptionIds = `${datePickerId}-error`;
-    }
+    ariaDescriptionIds = `${datePickerId}-help`;
   }
 
+  const inputAriaDescriptionId = inputAttributes && inputAttributes['aria-describedby'];
```
ShettyAkarsh, line 226: should we assign only inputAttributes['aria-describedby'] here ?

pranav300: Yes, as we only want to concatenate and use `aria-describedby` for the current scenario.

packages/terra-date-picker/src/DatePickerField.jsx
benbcai: That makes sense. Adding a comment here may help clarify the confusion.

benbcai (approved):

jeremyfuksa (approved):

saket2403 (approved):

packages/terra-date-picker/src/DatePickerField.jsx
ShettyAkarsh: yes, but i was asking like this  'const inputAriaDescriptionId = inputAttributes['aria-describedby'];'

supreethmr (approved):

packages/terra-date-picker/src/DatePickerField.jsx
pranav300: this would throw up if inputAttributes in undefined.

ShettyAkarsh (approved):

packages/terra-date-picker/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Fixed
+  * Fixed description id's in DatePicker for better accessibility.
+
```
avinashg1994, line 7: nit: This can be moved to the existing "Fixed" section.

avinashg1994 (approved):

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -219,8 +229,9 @@ class DatePickerInput extends React.Component {
             onChange={this.handleOnChange}
             onFocus={onFocus}
             onBlur={onBlur}
+            aria-required={required}
```
pranav300, line 232: Added this so that voiceover and jaws read required after the label.

packages/terra-date-picker/CHANGELOG.md
pranav300: updated here https://github.com/cerner/terra-framework/pull/1252/commits/0278a696952f80ca6aca91743ccdc6531926f9cf

PR #1260, opened 2020-09-23
[WIP][terra-pills] New terra pills component

avinashg1994:
### Summary

This draft contains the first iteration and second iteration of the basic, removable and roll-up pill(including container) implementation.

Zeplin(Basic Pill): https://app.zeplin.io/project/59b1bfeae515657fa8268300/screen/5b60abaaa8b621561b120d39
Zeplin(Removable Pill): https://app.zeplin.io/project/59b1bfeae515657fa8268300/screen/5f3d9d37694a9912ac38b2df
Zeplin(Roll-up Pill): https://app.zeplin.io/project/59b1bfeae515657fa8268300/screen/5f3316c0c4dcb05133097579

### Deployment Link

Doc:
https://terra-framew-new-terra--kvc5ti.herokuapp.com/components/terra-pills/pills/pills
Tests:
https://terra-framew-new-terra--kvc5ti.herokuapp.com/tests/terra-pills/pills/pill-list/basic-pill-list


StephenEsser:
@avinashg1994 

There's a few lint errors causing the build the fail, would you mind fixing those to enable a successful deployment?

avinashg1994:
@neilpfeiffer, This PR has been updated to include changes for Roll Up pill as well.

packages/terra-pills/src/PillList.jsx
```diff
@@ -0,0 +1,137 @@
...
+
+const cx = classNamesBind.bind(styles);
+
+const propTypes = {
+  /**
+   * The aria label string value for the pills container.
+   */
+  ariaLabel: PropTypes.string.isRequired,
+  /**
+   * The content to be shown in the pills container. Should only contain the sub-component pills
```
jmsv6d, line 20:
```suggestion
   * The content to be shown in the pills container. Should only contain the sub-component pills.
```

packages/terra-pills/src/PillList.jsx
```diff
@@ -0,0 +1,137 @@
...
+
+  const theme = React.useContext(ThemeContext);
+  const pillListRef = useRef();
+  const rollUpPillRef = useRef();
+  const [rollUpCount, setRollUpCount] = useState(0);
+  const [isRollUpRequired, setIsRollUpRequired] = useState(true);
+
+  const handleCollapse = () => {
+    let pillListHeight = pillListRef.current.getBoundingClientRect().height;
+    const pillHeight = rollUpPillRef.current.getBoundingClientRect().height + 12; // Margin + border width
```
jmsv6d, line 48: Can you obtain margin and border width should be obtained programmatically? Hardcoding it here will make the code brittle if styles change and as themes are applied.

jeremyfuksa: This line also caught my eye. I can see where you got the magic number of `12` from the Zeplin callouts, but this seems like something that needs to be accounted for in CSS.

packages/terra-pills/src/PillList.jsx
```diff
@@ -0,0 +1,137 @@
...
+  } = props;
+
+  const theme = React.useContext(ThemeContext);
+  const pillListRef = useRef();
+  const rollUpPillRef = useRef();
+  const [rollUpCount, setRollUpCount] = useState(0);
+  const [isRollUpRequired, setIsRollUpRequired] = useState(true);
+
+  const handleCollapse = () => {
+    let pillListHeight = pillListRef.current.getBoundingClientRect().height;
```
jmsv6d, line 47: Why is the rollup logic based on height?

packages/terra-pills/src/terra-dev-site/doc/Pills/Pills.1.doc.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+import DefaultPills from '../example/DefaultPills?dev-site-example';
+
+import PropsTable from 'terra-pills/src/Pill?dev-site-props-table';
+
+<Badge />
+
+# Terra Pills
+
+{insert description}
```
jmsv6d, line 11: Missing description

packages/terra-pills/src/Pill.jsx
```diff
@@ -0,0 +1,171 @@
...
+
+  const [open, setOpen] = useState(false);
+  const pillRef = useRef();
+
+  const handleOnClick = () => {
+    setOpen(true);
+  };
+
+  const handleOnRemove = () => {
+    pillRef.current.style.display = 'none';
```
jmsv6d, line 71: This doesn't feel like a very "react" way to handle the remove functionality. Could we update this style based on an `isRemoved` state?

tbiethman: I'd go a step farther and say that there's no need to hide the pill until the consumer stops rendering it, which is what they should be doing in their onRemove handler.

packages/terra-pills/src/Pill.jsx
```diff
@@ -0,0 +1,171 @@
...
+        {onRemove && (
+          <div
+            {...removeButtonProps}
+            className={removeButtonClassNames}
+          >
+            <IconClear height="0.57143rem" width="0.57143rem" />
+          </div>
+        )}
+      </div>
+      {popupConfig && (
```
tbiethman, line 151: I'd recommend not bundling the Popup with the Pill and simply exposing a ref to the Pill so that consumers can_render a Popup themselves if they want to. This decouples the Popup's inherent complexity from the Pill and gives the Pill less responsibility (and thus less surface area for future breaking changes).

I'd boil the props down to:
`label` -> that's what we're using throughout terra-application for props that denote user-visible text
`onSelect` -> Generic selection callback. If not provided, the Pill is not selectable.
`onSelectRemove` -> Remove button selection callback. If not provided, the Pill is not removable.
`refCallback` -> Callback to expose Pill container element for Popup placement.

tbiethman: We talked about this during scrum this morning. @neilpfeiffer and @jeremyfuksa agreed that removing the integrated Popup within the Pill makes sense. However, we still want to be able to create documentation examples that exhibit the desired Popup integration.

mjhenkes: with the dependency on pop up removed should we move this component back to terra-core?

jeremyfuksa: It seems like it becomes a lot more foundational without that dependency. Core feels like the more logical place for this.

jeremyfuksa:
### Keyboard Accessibility/Screen Reader Initial Findings
Tested using the Pill variants in `tests/`

**Basic Pill**
* Cannot get keyboard focus, therefore a screen reader can never read the label contents (VoiceOver and JAWS)

**Removable Pill**
* Spoken Label - Neither screen reader is very descriptive as to what the pill is. While I don't think that "pill" is what the reader needs to say, I don't know that "group" is right either. @scottwilmarth, can you give guidance on this?
  * VoiceOver = "Demo. Group. Currently on Group"
  * JAWS = "Demo Unlabeled 1"
* The screen reader gives no indication that there is a way to remove the pill (VO and Jaws)
* When using the delete key to trigger removal, only JAWS correctly reads the alert that is spawned.

**Popup Pill**
* Same spoken label issue
* VoiceOver gives no indication as to how one accesses the popup content. JAWS is OK.
* Once the popup is shown, VO's instructions on how to access/interact with the popup content doesn't work properly. VO says to use the `Ctrl+Opt+DnArrow` combo to access. This does nothing. Pressing `Ctrl+Opt+UpArrow` at this point shifts focus to the main content container. `Ctrl+Opt+DnArrow` a second time then shifts focus to the popup content. JAWS does not have problems driling into the popup content.

**Removable Popup Pill**
* No unique issues

Did not test Pill List at this time, since all issues should be the same in there.

cc @avinashg1994

packages/terra-pills/src/PillList.jsx
```diff
@@ -0,0 +1,137 @@
...
+      <div
+        className={cx(['roll-up-button'])}
+        onClick={handleRollUp}
+        onKeyDown={handleKeyDown}
+        onMouseDown={handleMouseDown}
+        ref={rollUpPillRef}
+        role="button"
+        tabIndex="0"
+      >
+        {`${rollUpCount} more ...`}
```
ShettyAkarsh, line 111: translations needs to be added for this text.

packages/terra-pills/src/PillList.jsx
```diff
@@ -0,0 +1,137 @@
...
+      handleRollUp();
+    }
+  };
+
+  const handleMouseDown = () => {
+    rollUpPillRef.current.setAttribute('data-terra-pills-show-focus-styles', 'false');
+  };
+
+  let rollUpButton;
+  if (isCollapsible) {
```
ShettyAkarsh, line 100: do we need to check this condition here for rendering the rollup button ?

packages/terra-pills/src/PillList.jsx
avinashg1994: Addressed it here : https://github.com/cerner/terra-framework/pull/1260/commits/65bc8c8eb773f94746a47e8e3f29679e74fef500
Changed the logic to retrieve the margins programatically from the computed styles.

packages/terra-pills/src/Pill.jsx
avinashg1994: This was just temporary for testing purposes for the 1st Iteration.
Have removed it now : https://github.com/cerner/terra-framework/pull/1260/commits/65bc8c8eb773f94746a47e8e3f29679e74fef500 and also updated test page to remove them from there.

jeremyfuksa:
### Accessibility Feedback Update
In order to get proper screen reader recognition, please set the following ARIA roles:

- Basic pill: `role=note`
- Removable and Popup: `role=button`

Also, so @scottwilmarth is able to see and test this component for accessibility, we need to get Travis building this PR.

packages/terra-pills/src/PillList.jsx
avinashg1994: I tried a couple of different approaches but i felt the most appropriate for the scenario where the PillList is rollable and the maximum number of pills including the roll up pill should be incorporated only in the first line.
I am definitely open to suggestion for other alternative approaches.

packages/terra-pills/src/terra-dev-site/doc/Pills/Pills.1.doc.mdx
avinashg1994: Added here : https://github.com/cerner/terra-framework/pull/1260/commits/fdc1985758d02b39d290c92990e2c957e35a28d6

packages/terra-pills/src/Pill.jsx
```diff
@@ -0,0 +1,171 @@
...
+  const handleOnRemove = () => {
+    pillRef.current.style.display = 'none';
+    onRemove();
+  };
+
+  const handleOnKeyDown = (event) => {
+    pillRef.current.setAttribute('data-terra-pills-show-focus-styles', 'true');
+    if (event.keyCode === KEY_RETURN || event.keyCode === KEY_SPACE) {
+      setOpen(true);
+    } else if ((event.keyCode === KEY_DELETE || event.keyCode === KEY_BACK_SPACE) && onRemove) {
```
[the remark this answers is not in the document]
avinashg1994: Yes, the "Space" key will be used for triggering the popup/onSelect callback of the pill.
But, i think in the this case and also from referring other components along with "Delete" the "Back Space" key should also be remove the pill. Especially for windows devices.

packages/terra-pills/src/PillList.jsx
avinashg1994: Yes, So how i have handled roll up is that based on the "isCollapsible" prop i am conditionally rendering the "Roll Up Pill" and the condition on line 130 is to check whether the "Pill List" is collapsed or expanded.

jeremyfuksa:
While @scottwilmarth is doing accessibility testing, please begin preparing for open source styling so that we can keep this PR moving forward.

cc @avinashg1994

PR #1261, opened 2020-09-25
[terra-date-input] Restricted year input to be 1900-2100 and allow users to clear month from select options.

### LINKED ISSUE #1238: [Terra-Date-Input] Restrict year input to be within 1900 and 2100
We are currently restricting the day input to be within 1-31 in Terra-Date-Input. similarly we should do it for year input as well to be within 1900 and 2100.

DateYearInput

### LINKED ISSUE #1241: [terra-date-input] Enhance month select for better usability
# Feature Request
PR #1230 updates the `DateInput` component to remove the placeholders from the inputs ('DD' from day, 'YYYY' from year) and changes the placeholder in the month dropdown select from "Month" to the new UX standard placeholder "_- Select -_" but does not change the behavior of the month select. The previous behavior originally implemented is now not up to the latest UX behavior standards.

## Description

Cerner UX standards dictate that a user should be able to "_clear_" the selected value in a select dropdown after choosing a value by selecting the placeholder option value. The current implementation is unclear to the user, as it shows in the closed/faceup state of the select the placeholder, but when opening the options list, it shows the first option, "January" with a check as if it is selected, even though the user hasn't selected it. Then after a user does select an actual value (e.g. "January", "February", "March", etc.), they are unable to clear it, if it was on accident and not a required field, unable to  return it to an unset value.

| Current behavior  |
| --- |
| date-input-after-PR#1230 |

\
\
The default behavior built-in to the [terra-form-select/.../NativeSelect](https://engineering.cerner.com/terra-ui/components/terra-form-select/form-select/native-select) shows the intended behavior desired for the month selector in `DateInput`, where when the placeholder displays ("_- Select -_"), the user understands nothing has been selected. A user can choose/set an option from the list, then choose the placeholder again to "_clear_" & unset the previously selected/set option.

| Intended behavior - same as Native-Select |
| --- |
| native-select-intended-behavior |

## @ Mentions
Followup for PR #1230
@pranav300 @jeremyfuksa

saket2403:
### Summary
Restricted year input to be within 1900-2100. Allow users to clear selected month by choosing placeholder value from monthe select for better usability.
Closes #1238 
Closes #1241 

### Deployment Link

### Testing
Updated jest test and wdio snapshots.

### Additional Details

@cerner/terra


packages/terra-date-input/src/DateInputUtil.js
```diff
@@ -76,19 +76,18 @@ class DateInputUtil {
    */
   static incrementYear(year) {
     // Handle empty input value
-    if (year === '') {
-      return '1';
+    let numericYear = Number(year);
+    if (numericYear === 0) {
```
benbcai, line 80: What is the reason for allowing to increment to 1900 from year `0` but not year `1`?

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -382,11 +382,12 @@ class DateInput extends React.Component {
...
 
     const inputValue = event.target.value;
-    const maxValue = 9999;
+    const maxValue = 2100;
+    const minValue = 1900;
 
     // When 'Predictive text' is enabled on Android the maxLength attribute on the input is ignored so we have
     // to check the length of inputValue to make sure that it is less then 4.
-    if (inputValue.length > 5 || Number(inputValue) > maxValue) {
+    if (inputValue.length > 5 || Number(inputValue) > maxValue || (inputValue.length === 4 && Number(inputValue) < minValue)) {
```
benbcai, line 390: It allows entering `-` or `+` which it shouldn't. After that it messes up keyboard interaction where I can't delete and doesn't restrict the max length to be 4. Maybe [validNumericInput](https://github.com/cerner/terra-framework/blob/main/packages/terra-date-input/src/DateInputUtil.js#L8-L10) needs to also restrict `-` or `+` to prevent this from happening.

packages/terra-date-input/tests/wdio/date-input-spec.js
```diff
@@ -336,7 +336,7 @@ Terra.describeViewports('Date Input', ['medium'], () => {
       Terra.it.matchesScreenshot();
     });
```
benbcai, line 338: Can we add a test to test clearing the month by selecting the placeholder?

packages/terra-date-input/tests/wdio/date-input-spec.js
```diff
@@ -360,7 +360,7 @@ Terra.describeViewports('Date Input', ['medium'], () => {
         Terra.hideInputCaret('input[name="terra-date-year-date-input"]');
 
         browser.click('input[name="terra-date-year-date-input"]');
-        browser.keys(['9999', 'ArrowUp']);
+        browser.keys(['2100', 'ArrowUp']);
```
benbcai, line 363: We should also add a test to test entering a year outside of 1900 and 2100.

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -382,11 +382,12 @@ class DateInput extends React.Component {
     }
 
     const inputValue = event.target.value;
-    const maxValue = 9999;
+    const maxValue = 2100;
+    const minValue = 1900;
```
supreethmr, line 386: Can we move these to DateInputUtil.js similar to [Date-picker](https://github.com/cerner/terra-framework/blob/6f0bb1db7b670b5d900049295e73d03713ed6059/packages/terra-date-picker/src/DateUtil.js#L252-L253) so that same can be used in DateInputUtils instead of using year values '2100' in multiple places.

packages/terra-date-input/src/DateInputUtil.js
saket2403: This condition is to increment when the field is empty.

packages/terra-date-input/src/DateInput.jsx
saket2403: Added here 6f09098

packages/terra-date-input/tests/wdio/date-input-spec.js
saket2403: Added test 6f09098

packages/terra-date-input/tests/wdio/date-input-spec.js
saket2403: Added test

packages/terra-date-input/src/DateInput.jsx
saket2403: Fixed 6f09098

packages/terra-date-input/src/DateInputUtil.js
```diff
@@ -76,19 +76,18 @@ class DateInputUtil {
...
 
     if (year) {
-      let numericYear = Number(year);
-
-      // Loop back to 1 if incrementing year from "9999"
-      if (numericYear === 9999) {
-        return '1';
+      // Loop back to 1900 if incrementing year from "2100"
+      if (numericYear === DateInputUtil.MaxYearValue) {
+        return '1900';
```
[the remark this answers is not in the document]
saket2403: Done 9d84bf7

packages/terra-date-input/src/DateInputUtil.js
benbcai: If so, shouldn't this check for an empty string like what it did before instead of checking for 0?

saket2403:
> We should prevent `+` and `-` in [validNumericInput](https://github.com/cerner/terra-framework/blob/main/packages/terra-date-input/src/DateInputUtil.js#L8-L10) instead of adding this check in multiple places.

Yes that was the first approach I tried, but handleChange was not getting trigerred on pressing these keys.I think similar issue was there with 'e' and '.' keys. So thatv condition was added like [this](https://github.com/cerner/terra-framework/blob/1ea109bb8eabf4a2d92677cd8348213dafe78743/packages/terra-date-input/src/DateInput.jsx#L312-L316). Thats why I added this condition similar to that.

packages/terra-date-input/src/DateInputUtil.js
saket2403: I had checked for empty string but value is '0' as I found it after putting a breakpoint. It is because of this [line](https://github.com/cerner/terra-framework/blob/1ea109bb8eabf4a2d92677cd8348213dafe78743/packages/terra-date-input/src/DateInput.jsx#L308)

benbcai: Ah, I see. Should that line in [handleDayKeyDown](https://github.com/cerner/terra-framework/blob/0cbfdf5745af49f9e3b4273a34253078e1dd9aed/packages/terra-date-input/src/DateInput.jsx#L308) and [handleDayKeyDown](https://github.com/cerner/terra-framework/blob/0cbfdf5745af49f9e3b4273a34253078e1dd9aed/packages/terra-date-input/src/DateInput.jsx#L272) be updated to default to `''` instead of `'0'` so that it can't increment from 0. It just seem weird that it currently can increment from 0 but not from 1.

benbcai:
> > We should prevent `+` and `-` in [validNumericInput](https://github.com/cerner/terra-framework/blob/main/packages/terra-date-input/src/DateInputUtil.js#L8-L10) instead of adding this check in multiple places.
> 
> Yes that was the first approach I tried, but handleChange was not getting trigerred on pressing these keys. I think similar issue was there with 'e' and '.' keys. So that condition was added like [this](https://github.com/cerner/terra-framework/blob/1ea109bb8eabf4a2d92677cd8348213dafe78743/packages/terra-date-input/src/DateInput.jsx#L312-L316). Thats why I added this condition similar to that.

I see. I think the reason onChange is not getting hit for the plus and minus signs because the input type is `number` set [here](https://github.com/cerner/terra-framework/blob/main/packages/terra-date-input/src/DateInput.jsx#L563-L571). After searching a bit. looks like your change is the suggested solution.

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -279,6 +279,12 @@ class DateInput extends React.Component {
       return;
     }
 
+    // prevent + and - characters from being entered into number input on keyDown
+    if (event.keyCode === 189 || event.keyCode === 187) {
```
yuderekyu, line 283: nitpick - can we stay consistent with how we reference keycodes by using the already imported KeyCode library? this also helps us avoid looking up what each code corresponds to. we could also cleanup the import by only importing the codes we need

packages/terra-date-input/src/DateInputUtil.js
saket2403: Done 3052cfe

packages/terra-date-input/src/DateInput.jsx
saket2403: These 2 keycode values are not there in the Keycode js file

yuderekyu: hmm is KEY_DASH not equivalent to minus? i'm also confused on plus. doing a table lookup tells me 187 is equivalent to equals. 

this logic could also be extracted into a named helper

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -499,7 +510,7 @@ class DateInput extends React.Component {
           disabled={this.props.disabled}
           aria-describedby={`${this.formatDescriptionId} ${this.props.monthAttributes['aria-describedby']}`}
         >
-          <option value="" hidden>{this.props.intl.formatMessage({ id: 'Terra.date.input.monthPlaceholder' })}</option>
+          <option value="">{this.props.intl.formatMessage({ id: 'Terra.date.input.monthPlaceholder' })}</option>
```
benbcai, line 513: In Firefox when no month is initially selected in the month select, all the months appear as disabled and italicized similar to this `- Select -` placeholder but they are actually active.

Screen Shot 2020-10-13 at 1 33 04 PM

saket2403: Looks like this [line](https://github.com/cerner/terra-framework/blob/4fa0427ed5c424247a12cc66c4b711d8943b09b5/packages/terra-date-input/src/DateInput.jsx#L478) is causing this behaviour in firefox. As a workaround thought of adding this style for all options except the placeholder one, which is similar to this [block](https://github.com/cerner/terra-framework/blob/main/packages/terra-date-input/src/DateInput.module.scss#L232-L237).
```  
    option:not([value='']) {
      color: transparent;
      font-style: normal;
      text-shadow: 0 0 0 #1c1f21;
    }
```

packages/terra-date-input/src/DateInput.jsx
saket2403: Added keycode in utils file. 9f1f78c

packages/terra-date-input/src/DateInput.jsx
saket2403: Done 9f1f78c

benbcai (approved):

packages/terra-date-input/src/DateInputUtil.js
```diff
@@ -245,5 +243,11 @@ DateInputUtil.inputType = {
   MONTH: 1,
   DAY: 2,
 };
+DateInputUtil.MinYearValue = 1900;
+DateInputUtil.MaxYearValue = 2100;
+DateInputUtil.keyCode = {
+  KEY_PLUS: 187,
```
yuderekyu, line 249: just so i understand, these keycodes aren't in the version of keycode js that we pull in?

saket2403: Yes

PaulSolDev:

packages/terra-date-input/src/DateInput.jsx
emilyrohrbough, line 513: I thought placeholders were in the process of being removed? Should this still be hidden?

packages/terra-date-input/src/DateInputUtil.js
```diff
@@ -77,18 +77,17 @@ class DateInputUtil {
   static incrementYear(year) {
     // Handle empty input value
     if (year === '') {
-      return '1';
+      return DateInputUtil.MinYearValue.toString();
```
emilyrohrbough, line 80: If year is not an empty string or is not defined, we return null. Should it always return the min year value if undefined/null?

packages/terra-date-input/tests/wdio/date-input-spec.js
```diff
@@ -405,7 +441,7 @@ Terra.describeViewports('Date Input', ['medium'], () => {
         Terra.hideInputCaret('input[name="terra-date-year-date-input"]');
 
         browser.click('input[name="terra-date-year-date-input"]');
-        browser.keys(['1', 'ArrowDown']);
+        browser.keys(['7654']);
```
emilyrohrbough, line 444: This visuals here shows 765, should an warning message show or if the first number is larger than the first digit in the min/max year should it allow you to enter it at all?

packages/terra-date-input/src/DateInputUtil.js
emilyrohrbough: Can we add a comment on here that they arn't included, seems a little odd if you didn't know that.

packages/terra-date-input/src/DateInput.jsx
saket2403: This placeholder was not removed, and this is a fix for #1241

packages/terra-date-input/src/DateInputUtil.js
saket2403: On incrementing a null/undefined value it starts from 1 so changed it to start from 1900.

packages/terra-date-input/tests/wdio/date-input-spec.js
saket2403: cc @benbcai , your thoughts?

benbcai: I think it would be difficult to handle "if the first number is larger than the first digit in the min/max year should it allow you to enter it at all?". The reason being the first digit entered may not always represent the first digit in the intended. For example, the user may enter `8` first and use the left arrow key to then enter `201` with the intention to enter the year 2018. It would seem annoying to prevent entering 8 as the first digit in this case.

This test case resembles when the user is pasting `7654` in the input. I think it is more ideal and less confusing to prevent the entire paste event instead of allowing it but just paste the first 3 digits `765` and cutting off the last digit because it would exceed the max year. Pasting may not be a very common use case here let alone pasting an invalid value. So I don't have a strong opinion here.

packages/terra-date-input/src/DateInputUtil.js
emilyrohrbough: Is this null return necessary then? https://github.com/cerner/terra-framework/pull/1261/files/9f1f78c7cefb5afde6ab25888703d2490c19a16d#diff-b5cce12f1b45fdee1f1ddd3837cbc55c5088424ddf957de7a28046fe64ecf1b4R98

packages/terra-date-input/tests/wdio/date-input-spec.js
```diff
@@ -396,7 +417,22 @@ Terra.describeViewports('Date Input', ['medium'], () => {
...
+        Terra.hideInputCaret('input[name="terra-date-day-date-input"]');
+        Terra.hideInputCaret('input[name="terra-date-year-date-input"]');
+
+        browser.click('input[name="terra-date-year-date-input"]');
+        browser.keys(['1900', 'ArrowDown']);
+      });
+
+      Terra.it.matchesScreenshot();
+    });
+
```
emilyrohrbough, line 434: The Terra.it helpers are being deprecated. Can we update?
```suggestion
    it('Year Input DOWN_ARROW cycles to 2100 when the year has reached 1900', () => {
      browser.url('/#/raw/tests/terra-date-input/date-input/default-date-input');
      browser.refresh();
        
      // Removes the blinking cursor to prevent screenshot mismatches.
      Terra.hideInputCaret('input[name="terra-date-day-date-input"]');
      Terra.hideInputCaret('input[name="terra-date-year-date-input"]');

      browser.click('input[name="terra-date-year-date-input"]');
      browser.keys(['1900', 'ArrowDown']);

      Terra.validates.screenshot();
    });

```

packages/terra-date-input/tests/wdio/date-input-spec.js
```diff
@@ -101,6 +101,27 @@ Terra.describeViewports('Date Input', ['medium'], () => {
...
+    it('Selects January in month select', () => {
+      browser.url('/#/raw/tests/terra-date-input/date-input/day-month-year-date-input');
+      Terra.hideInputCaret('input[name="terra-date-day-date-input"]');
+      Terra.hideInputCaret('input[name="terra-date-year-date-input"]');
+      browser.click('select[name="terra-date-month-date-input"]');
+      browser.keys(['ArrowDown']);
+      browser.keys(['Enter']);
+    });
+
+    Terra.it.validatesElement('First Month Selected');
```
emilyrohrbough, line 114:
```suggestion
      browser.keys(['Enter']);
      Terra.validates.element('First Month Selected');
    });

    
```

packages/terra-date-input/tests/wdio/date-input-spec.js
```diff
@@ -101,6 +101,27 @@ Terra.describeViewports('Date Input', ['medium'], () => {
...
+
+    it('Selects Placeholder option to clear selected value', () => {
+      browser.click('select[name="terra-date-month-date-input"]');
+      browser.keys(['ArrowUp']);
+      browser.keys(['Enter']);
+    });
+
+    Terra.it.validatesElement('Value Cleared');
+  });
```
emilyrohrbough, line 123:
```suggestion
      browser.keys(['Enter']);
      Terra.validates.element('Value Cleared');
    });
  });
```

packages/terra-date-input/src/DateInputUtil.js
saket2403: Added comment c212e4e

emilyrohrbough (approved):

jeremyfuksa (approved):

PR #1285, opened 2020-10-22
[terra-modal-manager] Automatically hide expand button at certain breakpoints

### LINKED ISSUE #1177: [terra-disclosure-manager] Evaluate hiding the expand button automatically on certain breakpoints
# Feature Request

## Description

Evaluate hiding the expand button automatically on certain breakpoints

saket2403:
### Summary
Hide expand button for certain breakpoints where user is unable to maximize although expand button is present.
Closes #1177 

### Deployment Link

### Testing
Updated jest and wdio tests

### Additional Details

@cerner/terra


packages/terra-modal-manager/src/ModalManager.jsx
```diff
@@ -49,6 +49,16 @@ const widthFromSize = {
 };
 
 class ModalManager extends React.Component {
+  static checkIsExpandable() {
+    if (document.querySelector(['[data-terra-abstract-modal]'])) {
+      const { height, width } = document.querySelector(['[data-terra-abstract-modal]']).getBoundingClientRect();
+      if (window.innerHeight === height + 20 && window.innerWidth === width + 20) {
```
emilyrohrbough, line 55: What breakpoint is this? Should this be on certain terra breakpoints?

tbiethman: The AbstractModal has a maximum size of 100% - 20px for its height and width. This is checking to see if the modal has reached that size. 

But I'm not quite sure how this would work, as at the time this check occurs, I don't believe the AbstractModal has even mounted yet? Maybe we're getting subsequent renders that make this work appropriately. Ideally we'd be able to cross-reference the current active breakpoint with the current modal size and determine whether or not maximize is possible even before rendering, but that logic would get pretty complicated.

supreethmr: I think we can use [dimensions / size values that are passed to modal in `renderModal()`](https://github.com/cerner/terra-framework/blob/580b034bbe0f225b58e20c33619844af2e3cff3e/packages/terra-modal-manager/src/ModalManager.jsx#L84-L86) instead of using `data-terra-abstract-modal` height and width for calculation to hide maximize on header.
```jsx
For Dimensions :
 shouldHideMaximize = (windowHeight <= manager.disclosure.dimensions.height + 20 && windowWidth <= manager.disclosure.dimensions.width + 20);

For Size : 
 shouldHideMaximize = (windowHeight <= heightFromSize[manager.disclosure.size] + 20 && windowWidth <= widthFromSize[manager.disclosure.size] + 20);
```

supreethmr:
We might need to handle screen orientation change also as shown in below image. maximize button still remains on modal even when there is no space for maximize on changing screen orientation. 

menudemo
with current implementation change in orientation has no effects on modal manager hence maximize button stays on header and gets removed on re-render(clicking the button).

we might need  listeners for orientation change / resize-observer to handle this scenario.

packages/terra-modal-manager/src/ModalManager.jsx
saket2403: Updated 7f5a84c

packages/terra-modal-manager/src/ModalManager.jsx
```diff
@@ -68,12 +68,17 @@ class ModalManager extends React.Component {
...
 
     const classArray = ['modal-manager'];
+    const windowHeight = window.innerHeight;
+    const windowWidth = window.innerWidth;
+    let hideMaximize = false;
     const isFullscreen = manager.disclosure.isMaximized || manager.disclosure.size === availableDisclosureSizes.FULLSCREEN;
     if (!isFullscreen) {
       if (manager.disclosure.dimensions) {
         classArray.push(`height-${manager.disclosure.dimensions.height}`, `width-${manager.disclosure.dimensions.width}`);
+        hideMaximize = (windowHeight <= manager.disclosure.dimensions.height + 20 && windowWidth <= manager.disclosure.dimensions.width + 20);
```
ShettyAkarsh, line 78: we could rename this hideMaximize variable as hideResizeButton.

ryanthemanuel:
Spoke with @tbiethman @neilpfeiffer and @mjhenkes. We think this one is going to get too messy to fix right now. It will be fixed in application v2 with our new modal implementation. Closing this PR.

PR #1292, opened 2020-10-28
react-intl upgrade

### LINKED ISSUE #1290: Update components to work with react-intl v2-v5
# Feature Request
In an endeavor to minimize the number of major version bumps we want to make the updates for our components to work with react-intl v5 passive wherever possible.

## Description
For this update the peer dependencies need to be updated. `intlShape` Is no longer exported since v3, so the imports for that must be removed and PropType for intl object updated. An undocumented non-passive change was made in react-intl for the `FormattedMessage` component, when a function is passed as a child of the component the function is now given the param of an array of strings instead of the joined string. This causes many of the components that we are rendering in these functions to have prop type warnings as they are expecting a string. There are a few approaches that we could take to make these components continue to work as expected for all the react-intl version. I believe for most of these cases we can switch from using the `FormattedMessage` component to `injectIntl` + `intl.formatMessage()` or the function provided to `FormattedMessage` can do a check if the param is an array and join the strings, if so. 

I have begun this work at https://github.com/cerner/terra-framework/tree/react-intl-upgrade
This work should be able to be completed and released prior to the framework portions of the upgrade being completed.

In order to test these changes against react-intl v5 the package.json can be updated to pull in:
`    "terra-aggregate-translations": "git+https://github.com/cerner/terra-aggregate-translations.git#upgrade-intl",
    "terra-dev-site": "git+https://github.com/cerner/terra-dev-site.git#react-intl-upgrade",
`

jmsv6d:
### Summary
Made updates to components so they will be compatible with react-intl v2-v5.  

Closes #1290 

### Deployment Link

### Testing

Ran locally with react-intl v2 and v5, and tests pass

### Additional Details

Updates required:

- Peer dependency updated to range
- Remove `intlShape` imports as react-intl stopped exporting it in V3
- `FormattedMessage` no longer passes a string as the param for the child function, but instead an array of strings. Changes were made to either use `injectIntl` + `intl.formatMessage()` or to check if the param is an array and join the strings, to prevent propType warning for the components rendered.
- `injectIntl` wraps the component with a function component in newer versions, this caused some issues with `refs`, particularly `DatePicker` required an update to use a refCallback internally. Note: The api for `injectIntl` to forwardRef is different between v2 and v3 

@cerner/terra


packages/terra-abstract-modal/src/_ModalContent.jsx
```diff
@@ -138,15 +138,29 @@ const ModalContent = forwardRef((props, ref) => {
         ref={ref}
       >
         <FormattedMessage id="Terra.AbstractModal.BeginModalDialog">
```
emilyrohrbough, line 140: Should we take advantage of the new fallback message logic ? Where if a translation is missing, it will fallback to the value defined here?
https://formatjs.io/docs/react-intl/components#usage

It might end up with mis-matched locales...but might be better than the translation key?

packages/terra-application-links/src/tabs/_TabMenu.jsx
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import Popup from 'terra-popup';
 import { matchPath } from 'react-router-dom';
 import * as KeyCode from 'keycode-js';
-import { FormattedMessage } from 'react-intl';
+import { injectIntl } from 'react-intl';
```
emilyrohrbough, line 6: Would it be worth making this change throughout instead of concatenating the chunks array passed down by FormattedMessage?

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
```diff
@@ -30,6 +30,12 @@ const propTypes = {
    * Bounding container for the menu, will use window if no value provided.
    */
   boundingRef: PropTypes.func,
+
```
emilyrohrbough, line 33:
```suggestion
```

packages/terra-date-picker/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Updated to support a peer dependency of react-intl v2-v5\
```
emilyrohrbough, line 6:
```suggestion
  * Updated to support a peer dependency of react-intl v2-v5
```

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -480,6 +480,7 @@ class DatePicker extends React.Component {
                 useExternalFormatMask={useExternalFormatMask}
               />
             )}
+            customInputRef="inputRefCallback"
```
emilyrohrbough, line 483: why is this needed?

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -40,7 +40,7 @@ const propTypes = {
   /**
    * intl object programmatically imported through injectIntl from react-intl.
    * */
-  intl: intlShape.isRequired,
+  intl: PropTypes.shape({ locale: PropTypes.string }).isRequired,
```
emilyrohrbough, line 43: I know it won't happen now, but do you think we should utilize react-intl's timeZone config option in the intlProvider?

packages/terra-abstract-modal/src/_ModalContent.jsx
mjhenkes: Is the fallback logic available for all versions we support? IMO i'm ok if we don't fall back, if we're falling back there's something wrong with their app setup and we should make it obvious.

packages/terra-date-picker/src/react-datepicker/calendar.jsx
```diff
@@ -112,7 +112,7 @@ export default class Calendar extends React.Component {
      * @private
      * Internationalization object with translation APIs. Provided by `injectIntl`.
      */
-    intl: intlShape,
+    intl: PropTypes.shape({}),
```
mjhenkes, line 115: Why are these shapes not as defined as the others?

packages/terra-abstract-modal/src/_ModalContent.jsx
jmsv6d: We'll probably want to update to use that once we do next major version bumps on components, but in order to continue supporting the components to be used with the older versions of react-intl we can't.

packages/terra-application-links/src/tabs/_TabMenu.jsx
jmsv6d: I tried to mostly switch to `injectIntl`, the problem is that the wrapper that `injectIntl` creates is a functional component, so when I made that change in a few of the components it caused issues if the `ref` for the component was being used in a test or example. `react-intl` v2 has a different api for forwarding refs than later versions.

packages/terra-date-picker/src/DatePicker.jsx
jmsv6d: In later versions of `react-intl` the wrapper created by `injectIntl` is a function component. The `react-date-picker` code utilized the `ref` of the date input for focus manipulation, but the DateInput component being wrapped in a function component means it doesn't have a `ref`. There was already logic in the `react-date-picker` to pass the name of an alternative prop to pass the `ref` callback to, so I added an `inputRefCallback` prop to the DateInput that would pass it down to the `ref` of the input element and this prop indicates that's what the `react-date-picker` should use.

packages/terra-date-picker/src/react-datepicker/calendar.jsx
jmsv6d: It's because this component doesn't actually use the `intl` object, it just passes it through as a prop. I had only defined the shape with what each component used from the `intl` object in lieu of defining the whole shape everywhere . If I dig down to where it's actually used, it looks like it's being used for the `formatMessage` function,  I can go ahead and define that here.

packages/terra-date-time-picker/src/DateTimePicker.jsx
jmsv6d: I don't see why not. We could easily add that as a prop to ApplicationBase to allow that to be set.

mjhenkes (approved):

packages/terra-application-header-layout/src/ApplicationHeaderLayout.jsx
```diff
@@ -20,7 +20,7 @@ const propTypes = {
   /**
    * intl object programmatically imported through injectIntl from react-intl.
    * */
-  intl: intlShape.isRequired,
+  intl: PropTypes.shape({ formatMessage: PropTypes.func }),
```
ryanthemanuel, line 23: Should these be required?

ryanthemanuel (approved):

packages/terra-application-header-layout/src/ApplicationHeaderLayout.jsx
jmsv6d: Oops, yeah.

packages/terra-application-navigation/package.json
```diff
@@ -28,7 +28,7 @@
   "peerDependencies": {
     "react": "^16.8.5",
     "react-dom": "^16.8.5",
-    "react-intl": "^2.4.0",
+    "react-intl": ">=2.4.0, <6.0.0",
```
benbcai, line 31: Should we make this `">=2.8.0, <6.0.0"` to be consistent with the other packages?

benbcai (approved):

PR #1324, opened 2020-11-20
[terra-layout] Remove Terra.it helpers from wdio

saket2403:
### Summary
Remove Terra.it helpers from wdio as they are deprecated.

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-layout/tests/wdio/layout-slide-panel-spec.js
```diff
@@ -1,33 +1,32 @@
...
+    browser.url('/raw/tests/terra-layout/layout/layout-slide-panel-example');
     browser.click('#test-toggle-1');
     browser.waitForExist('[class*=_panel_][aria-hidden="false"]');
+    Terra.validates.element('toggle click', { selector: '#site' });
   });
-
-  Terra.it.matchesScreenshot({ selector: '#site' });
 });
 
 describe('Toggle the slide panel double click', () => {
```
yuderekyu, line 21: all three it blocks toggle the same example. we could remove the inner describe blocks and have the it blocks directly under `Terra.describeViewports`. is that out of scope for this pr? what are your thoughts?

saket2403: First test is for medium viewport and the other 2 are for huge, so I think thats the reason why they are outside the `Terra.describeViewports`.

packages/terra-layout/tests/wdio/layout-slide-panel-spec.js
```diff
@@ -1,33 +1,32 @@
...
   before(() => {
-    browser.url('/#/raw/tests/terra-layout/layout/layout-slide-panel-example');
+    browser.url('/raw/tests/terra-layout/layout/layout-slide-panel-example');
     browser.waitForExist('[class*=_panel_][aria-hidden="true"]');
   });
-
-  Terra.it.matchesScreenshot({ selector: '#site' });
+  it('Toggles the slide panel and hidden styles', () => {
+    Terra.validates.element('default', { selector: '#site' });
+  });
```
supreethmr, line 9: we can have it() just like below without before() since here we have only one it() 
```suggestion
  it('Toggles the slide panel and hidden styles', () => {
     browser.url('/raw/tests/terra-layout/layout/layout-slide-panel-example');
     browser.waitForExist('[class*=_panel_][aria-hidden="true"]');
     Terra.validates.element('default', { selector: '#site' });
  });
```

PR #1378, opened 2021-02-25
[terra-application-navigation] Add Ids to Navigation Items, Settings, Logout, and Help

Andrew-Shoemake:
### Summary

Adds the ability to add an id to navigation items, the settings button, logout button, and help button. This will be used for pendo integrations. 

### Testing

Tested via updated unit tests/snapshots

### Additional Details

Internal Cerner Jira: https://jira2.cerner.com/browse/CONNECT-29174 

@cerner/terra


packages/terra-application-navigation/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
```
supreethmr, line 5:
```suggestion
* Added
```
Since we are adding new props and not making changes to existing props.

packages/terra-application-navigation/src/ApplicationNavigation.jsx
```diff
@@ -63,6 +63,18 @@ const propTypes = {
    * `navigationItems` array.
    */
   activeNavigationItemKey: PropTypes.string,
+  /**
+    * The id for the logout utility item.
+    */
```
supreethmr, line 68: nit : formatting seems to off here compare to other props.

supreethmr:
Can you link the issue reference  requesting the changes done in this PR...??

packages/terra-application-navigation/src/common/_PopupMenu.jsx
```diff
@@ -59,6 +63,10 @@ const propTypes = {
      * Key used as react key, and returned in the onSelect
      */
     key: PropTypes.string.isRequired,
+    /**
+     * The id for the menuItem.
+     */
+    id: PropTypes.string,
```
supreethmr, line 69: I am not able to find where this prop is being assigned..? do we need this prop in popMenu...???

packages/terra-application-navigation/src/common/_PopupMenu.jsx
```diff
@@ -165,6 +173,7 @@ const PopupMenu = ({
   if (onSelectFooterItem) {
     endContent = (
       <Button
+        id={logoutId || undefined}
```
supreethmr, line 176:
```suggestion
        id={logoutId}
```
i feel both are same.

packages/terra-application-navigation/src/common/_PopupMenu.jsx
```diff
@@ -193,6 +202,7 @@ const PopupMenu = ({
         <ul className={cx('utility-list')} aria-label={title} ref={listRef} role={role} tabIndex="0" onKeyDown={handleKeyDown}>
           {menuItems.map(item => (
             <PopupMenuListItem
+              id={item.id || undefined}
```
supreethmr, line 205:
```suggestion
              id={item.id}
```

mjhenkes (changes requested):
@Andrew-Shoemake 
Thanks for putting out this PR. 

I would like to see some changes to the approach. Instead of adding individual id's to extensionsItems, navigationItems, utilityItems, logout, settings, and help, I would like you to add a single Id to the application navigation component and generate the other items id's based on the application navigation id. 

You can use the following template for creating the item ids:
`<ApplicationNavigation id>-<item type>-<item key>` 

Logout, settings and help may not fall directly into this pattern but we can still make them unique. 

We will also have to update documentation for this approach to help consumers understand how the id's are generated.

Changing to generate id's based of a single id will help prevent our API from exploding and keeps id's within the control of the component.

Andrew-Shoemake:
@mjhenkes, if I'm understanding correctly you're saying something like.

The ApplicationNavigation element has an id of 'main-navigation' and then when build the individual navigation/extension items there we would get something like 

'id of the navigation element' + 'they key of the navigation item' 

for example

there's a navigation item with the key of 'home' the code would then generate the id for that nav item as 'main-navigation-home'.

Then for stuff like settings/help/etc we would just generate stuff like 'main-navigation-help-btn'.

mjhenkes:
Pretty close.

Given:
ApplicationNavigation element has an id of `main-navigation`
It has one Navigation Item with a key of `item-1`.
it has one utility Item with a key of `item-1`. (keys only need to be unique within the group)
It has one extension Item with a key of `item-2`.

the navigation item would have a generated id of: `main-navigation-navigationItem-item-1`
the utility item would have a generated id of: `main-navigation-utilityItem-item-1`
the extension item would have a generated id of" `main-navigation-extensionItem-item-2`
help would have a generated id of : `main-navigation-utilityItemHelp`
logout  would have a generated id of : `main-navigation-utilityItemLogout`
settings item would have a generated id of : `main-navigation-utilityItemSettings`

Andrew-Shoemake:
I have updated per the comments. I'm still working on updating tests, but wanted to get the updates out so I can see if I'm going in the right direction.

mjhenkes:
@Andrew-Shoemake You're on the right track with these changes.

packages/terra-application-navigation/src/drawer-menu/_DrawerMenu.jsx
```diff
@@ -168,6 +178,7 @@ const DrawerMenu = ({
         ) : null}
         {onSelectHelp ? (
           <DrawerMenuListItem
+            id={utilityItemId(id, "Help")}
```
mjhenkes, line 181: this would generate the same id as if someone used the key of 'help'. Though that is unlikely I'd like to avoid it by giving this a slightly different id:
instead of `${appNavId}-${itemType}-${itemKey}`
something like `${appNavId}-${itemType}${reservedItemKey}` (there might be a better name than `reservedItemKey`)

Same comment for settings and logout.

packages/terra-application-navigation/CHANGELOG.md
Andrew-Shoemake: Switched wording to "added"

packages/terra-application-navigation/src/utility-menu/_UtilityMenu.jsx
```diff
@@ -78,6 +90,7 @@ const UtilityMenu = ({
 
   if (onSelectHelp) {
     menuItems.push({
+      id: utilityItemId(id, 'TerraHelp'),
```
Andrew-Shoemake, line 93: Would we want to make a special function that generates this rather than inputting 'TerraHelp'?

packages/terra-application-navigation/src/utils/helpers.js
```diff
@@ -61,6 +61,22 @@ const useAnimatedCount = (countRef, countValue) => {
   }, [countRef, countValue]);
 };
 
+function itemId(appNavId, itemType, itemKey) {
+  return appNavId ? `${appNavId}-${itemType}-${itemKey}` : null;
```
Andrew-Shoemake, line 65: Not sure if returning null is the right thing to do here when there is no appNavId. I assume we don't want to enforce ids if the consumer doesn't want them.

mjhenkes: if the user hasn't entered an app nav id we shouldn't add ids at all.

packages/terra-application-navigation/src/utility-menu/_UtilityMenu.jsx
mjhenkes: I lean towards making a special function, but it's probably fine either way.

packages/terra-application-navigation/src/ApplicationNavigation.jsx
Andrew-Shoemake: This prop has been removed. Marking as resolved.

packages/terra-application-navigation/src/common/_PopupMenu.jsx
Andrew-Shoemake: To follow up on this, it's used down on line 206 when iterating through the menu items.

packages/terra-application-navigation/src/common/_PopupMenu.jsx
Andrew-Shoemake: From what I understand the ``` || undefined ``` prevents id from being put in the dom if item.id is null. However, if there is a more standard way to do that I'll change it.

packages/terra-application-navigation/src/utility-menu/_UtilityMenu.jsx
Andrew-Shoemake: Added special function to generate the help, settings, and logout ids

packages/terra-application-navigation/src/drawer-menu/_DrawerMenu.jsx
Andrew-Shoemake: This has been addressed

mjhenkes:
@Andrew-Shoemake can you add a changelog entry for terra-tabs?

packages/terra-application-navigation/src/ApplicationNavigation.jsx
```diff
@@ -210,6 +215,7 @@ const ApplicationNavigation = ({
           returnFocusOnDeactivate: false,
         }}
         className={cx('drawer-menu-focus-trap-container')}
+        id={id}
```
tbiethman, line 218: I don't think I'd expect the default id to show up on the FocusTrap element. This would likely be better placed on the outer container on line 447.

Andrew-Shoemake: Yup, you're right. I tested with just removing that completely and it's actually not needed at all.

mjhenkes (approved):

tbiethman (approved):

mjhenkes:
Thanks for the contribution @Andrew-Shoemake!
This will get released next Tuesday.

PR #1381, opened 2021-03-03
[terra-application-navigation] Prevent error when using focus trap with react intl v5

mjhenkes:
### Summary

Focus trap will throw a "cannot read property 'ownerDocument' of null" when the child component isn't an html element. See:  https://github.com/focus-trap/focus-trap-react/issues/44#issuecomment-543770159

The drawer menu is wrapped in injectIntl. In our current version it is an html element, but in react intl v5 it nolonger is an html element.

I believe this could be fixed by upgrading focus trap, but the easiest safest fix is to wrap the drawer menu in a div... which is what I have done here.

Closes #

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-application-navigation/src/ApplicationNavigation.jsx
```diff
@@ -211,20 +211,29 @@ const ApplicationNavigation = ({
...
-          onSelectHelp={onSelectHelp ? generateMenuClosingCallback(onSelectHelp) : undefined}
-          onSelectLogout={onSelectLogout ? generateMenuClosingCallback(onSelectLogout) : undefined}
-          utilityItems={utilityItems}
-          onSelectUtilityItem={onSelectUtilityItem ? generateMenuClosingCallback(onSelectUtilityItem) : undefined}
-          notifications={notifications}
-        />
+        {/*
+          The following div has been added to prevent a "Cannot read property 'ownerDocument' of null" error when using react-intl v5.
+          See link for more info: https://github.com/focus-trap/focus-trap-react/issues/44#issuecomment-543770159
+          This div could be removed with a newer version of FocusTrap
```
yuderekyu, line 217: want to log an issue to remove the div, once we upgrade focus trap?

yuderekyu (approved):

packages/terra-application-navigation/src/ApplicationNavigation.jsx
mjhenkes: I will log issues to update react-focus trap

mjhenkes: https://github.com/cerner/terra-application/issues/145
https://github.com/cerner/terra-core/issues/3388
https://github.com/cerner/terra-application/issues/145

benbcai (approved):

tbiethman (approved):

PR #1387, opened 2021-03-18
Consume jest-config-terra

mjhenkes:
### Summary

This is an example upgrade to the proposed jest-config-terra configuration package.

Notes on the added config:
The only additional config here is setting up enzyme. 

Enzyme setup is not included in jest-config-terra as I'm trying to be agnostic to the enzyme/testing-library:react debate and the jest config should be able to be used in no react modules without extraneous dependencies.

### Deployment Link

### Additional Details

@cerner/terra


tbiethman (approved):
👍 👍

ryanthemanuel (approved):

PR #1403, opened 2021-04-23
[terra-date-picker] Fix focus trap issue for mobile view.

### LINKED ISSUE #1390: Focus Trap Error when the portal DatePicker is open and disabled
# Bug Report

## Description

If the date picker is in a state when it is open AND disabled, focus trap will throw `You can't have a focus-trap without at least one focusable element`. This only happens with the non-inline portaled version of the date picker (i.e. mobile viewports). [Focus trap error](https://github.com/focus-trap/focus-trap/blob/4.0.2/index.js#L199)

This seems to happen because the portal and focus trap are getting rendered because of this [check](https://github.com/cerner/terra-framework/blob/terra-date-picker%404.58.0/packages/terra-date-picker/src/react-datepicker/index.jsx#L865) but not the calendar because of this [check](https://github.com/cerner/terra-framework/blob/terra-date-picker%404.58.0/packages/terra-date-picker/src/react-datepicker/index.jsx), so when the focus trap is mounted and activates, it cannot find any tabbable nodes [here](https://github.com/focus-trap/focus-trap/blob/4.0.2/index.js#L194), so it throws.

## Steps to Reproduce

1. Have focus on an element other than the date picker.
2. Click on the date picker button AND disable the date picker field.

We have workflows where when we blur fields like inputs or textareas, we will make a network call and disable fields on the page. We had one specific case where if the blur was caused by clicking on a date picker button, we would get this focus trap error.

From some investigation, the check for [opening the portal and focus trap](https://github.com/cerner/terra-framework/blob/main/packages/terra-date-picker/src/react-datepicker/index.jsx#L865) seems suspicious compared to other checks within the component for opening the calendar [1](https://github.com/cerner/terra-framework/blob/main/packages/terra-date-picker/src/react-datepicker/index.jsx#L690) and [2]( https://github.com/cerner/terra-framework/blob/main/packages/terra-date-picker/src/react-datepicker/index.jsx#L894) 

## Possible Solution

Updating the check [here](https://github.com/cerner/terra-framework/blob/terra-date-picker%404.58.0/packages/terra-date-picker/src/react-datepicker/index.jsx#L865) to match other checks for opening the date picker.
```js
(this.state.open && !this.props.disabled) || this.props.inline
```
Making this change locally seemed to fix our issue.

## Environment

* Component Name and Version: terra-date-picker 4.58.0
* Browser Name and Version: Chrome v89
* Node/npm Version: [e.g. Node 8/npm 5] Node 10.18.0, npm 6.14.0
* Webpack Version: 4.46.0
* Operating System and version (desktop or mobile): macOS 10.15.7 mobile

## @ Mentions

@tbiethman

saket2403:
### Summary
Fix focus trap issue for mobile view.

Closes #1390 

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-date-picker/src/react-datepicker/index.jsx
```diff
@@ -862,7 +862,7 @@ class DatePicker extends React.Component {
               : null
           }
           {
-            this.state.open || this.props.inline
+            (this.state.open && !this.props.disabled) || this.props.inline
```
benbcai, line 865: Can we add a test to test this when it is open and then disable it?

saket2403: Added

benbcai (approved):

lokesh-0813 (approved):

dkasper-was-taken (approved):

pranav300 (approved):

PR #1416, opened 2021-06-02
Upgrade keycode-js to v3.1.0 

### LINKED ISSUE #1404: Update keycode-js version to the latest major version
# Story

## Description

As a consumer of terra, I'd like to be on the latest version of keycode-js

## Motivation

So that I am sure to get the latest version that has the most recent functionality and bug fixes

## Definition of Done

We'll know we're done when each package's version of keycode-js has it's version updated, the code is tested, and merged to main.

saket2403:
### Summary
Upgrade keycode-js to v3.1.0 
Closes #1404 

### Deployment Link

### Additional Details

@cerner/terra


mjhenkes (approved):

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -280,7 +280,7 @@ class DateInput extends React.Component {
     }
 
     // prevent + and - characters from being entered into number input on keyDown
-    if (event.keyCode === DateInputUtil.keyCode.KEY_MINUS || event.keyCode === DateInputUtil.keyCode.KEY_PLUS) {
+    if (event.keyCode === KeyCode.KEY_EQUALS || event.keyCode === KeyCode.KEY_DASH) {
```
yuderekyu, line 283: these keys don't seem to match the previous ones 👀

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -322,7 +322,7 @@ class DateInput extends React.Component {
...
```
yuderekyu, line 325: same question here

packages/terra-date-input/src/DateInput.jsx
mjhenkes: At first i thought there might be a bug here but I think this is ok. Dash and minus are equivalent (keycode 189) and Equals is equivalent to plus as well when you factor in shift (keycode 187). We don't really have to take shift into account here because it doesn't hurt to prevent = as well as +.

benbcai: https://github.com/kabirbaidhya/keycode-js/blob/master/dist/keycode.cjs.js#L112-L114

benbcai (approved):

packages/terra-date-input/src/DateInput.jsx
yuderekyu: ah, thanks for clearing that up! 😅

yuderekyu (approved):

PR #1425, opened 2021-06-30
[terra-collapsible-menu-view] Fix to allow users to add custom click event for menu buttons.

### LINKED ISSUE #1412: terra-collapsible-menu-view overrides consumer onClick when expanded
# Bug Report

## Description
The menu item behaves differently between if the menu item with a submenu is expanded vs. collapsed. The terra-collapsible-menu-view has its own onClick handler for when the menu button is expanded so it can open and close the menu https://github.com/cerner/terra-framework/blob/main/packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItem.jsx#L146

In the process it is overwriting any onClick handler passed by the consumer. Whereas the terra-menu item in the collapsed menu will call the onClick handler provided by the consumer after doing its own onClick logic https://github.com/cerner/terra-framework/blob/main/packages/terra-menu/src/MenuItem.jsx#L134.

## Steps to Reproduce
1. 

## Expected Behavior
The menu item behaves the same whether the menu item with a submenu is expanded or collapsed.

## Possible Solution
The onClick seems is a first class prop for Menu.Item so removing that would be non-passive. An update to collapsible menu view to also call the onClick handler of the consumer when expanded might be a possible solution

## Environment

* Component Name and Version:
* Browser Name and Version:
* Webpack Version:

## @ Mentions
@jmsv6d

saket2403:
### Summary
 Allow users to add custom click event for menu buttons.

Closes #1412 

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-collapsible-menu-view/src/terra-dev-site/test/collapsible-menu-view/DefaultCollapsibleMenuView.test.jsx
```diff
@@ -71,6 +71,7 @@ class CollapsibleMenuViewDemo extends React.Component {
           text="Menu Button 1"
           key="MenuButton1"
           className="MenuButton1"
+          onClick={this.handleToggleOneOnChange}
```
ShettyAkarsh, line 74: can we have different event created for this example ? like handleMenuButtonOnClick ? to be more clear on what we are trying to do.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItem.jsx
```diff
@@ -83,6 +83,13 @@ class CollapsibleMenuViewItem extends React.Component {
     super(props);
     this.setButtonNode = this.setButtonNode.bind(this);
     this.getButtonNode = this.getButtonNode.bind(this);
+    this.handleButtonClick = this.handleButtonClick.bind(this);
+  }
+
+  handleButtonClick(event) {
```
benbcai, line 89: Is it correct that onClick didn't do anything before because [this.handleButtonClick](https://github.com/cerner/terra-framework/pull/1425/files#diff-73b8965c3503d92ed6bbad70f722b2db46bf3e9d58183ff7e5417ad4cc419bdfR153) was undefined?

benbcai (approved):

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItem.jsx
```diff
@@ -83,6 +83,13 @@ class CollapsibleMenuViewItem extends React.Component {
...
+    if (this.props.onClick) {
```
jmsv6d, line 90: Since this function is actually doing anything, why don't we just remove this line: https://github.com/cerner/terra-framework/pull/1425/files#diff-73b8965c3503d92ed6bbad70f722b2db46bf3e9d58183ff7e5417ad4cc419bdfR153
the onClick prop would just get passed in through the attributes. No need to wrap it.

saket2403: Done

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItem.jsx
saket2403: Yes

packages/terra-collapsible-menu-view/src/terra-dev-site/test/collapsible-menu-view/DefaultCollapsibleMenuView.test.jsx
saket2403: Done b04c426

jmsv6d (approved):

ryanthemanuel (approved):

PR #1442, opened 2021-07-30
[WIP][terra-pills] New terra-pills component

### LINKED ISSUE #1441: New terra-pill component
# Feature Request
Create the Pill component based on the OCS designs posted in zeplin:
Basic Pill:
https://app.zeplin.io/project/59b1bfeae515657fa8268300/screen/5b60abaaa8b621561b120d39

Removable Pill: 
https://app.zeplin.io/project/59b1bfeae515657fa8268300/screen/5f3d9d37694a9912ac38b2df

Roll Up Pill: 
https://app.zeplin.io/project/59b1bfeae515657fa8268300/screen/5f3316c0c4dcb05133097579

lokesh-0813:
### Summary
Creating draft PR for initial review.

Closes #1441

### Deployment Link

https://terra-framew-terra-basi-dx0eli.herokuapp.com/tests/cerner-terra-framework-docs/pills/pills/single-pill
### Testing

Included few test examples.
WDIO - In Progress
JEST - In Progress

### Additional Details

Modified the wdio-config to support scoped packages.

Things that is in progress/don't work:

1)Accessibility testing not complete.

@cerner/terra


mjhenkes:
It seem like whether or not a pill is truncated is calculated only at first render. On screen resize the pop is either not shown on a truncated pill or still available on a non truncated pill.
pills resize small to large
pills resize large to small

packages/terra-pills/LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2021-present Cerner Innovation, Inc.
```
mjhenkes, line 190:
```suggestion
   Copyright 2021 - present Cerner Innovation, Inc.
```

packages/terra-pills/README.md
```diff
@@ -0,0 +1,24 @@
...
+- [Documentation](https://engineering.cerner.com/terra-ui/components/terra-pills/pills/pills)
+- [LICENSE](#license)
+
+## Getting Started
+
+- Install from [npmjs](https://www.npmjs.com): `npm install terra-pills`
+
+## LICENSE
+
+Copyright 2021 Cerner Innovation, Inc.
```
mjhenkes, line 18:
```suggestion
Copyright 2021 - present Cerner Innovation, Inc.
```

packages/terra-pills/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "files": [
+    "*.log",
+    "target",
+    "reports",
+    "tests/**/__snapshots__/",
+    "aggregated-translations",
+    "dev-site-config",
+    "screenshots",
+    "errorScreenshots",
+    "generated-themes"
```
mjhenkes, line 34: files is the list of files to include in the package, vs .npmignore which is the list of files to exclude. You'll want to change this to something like: https://github.com/cerner/terra-framework/blob/main/packages/terra-framework-docs/package.json#L28-L30

there are a couple of files included by default so you don't have to worry about those: https://docs.npmjs.com/cli/v6/configuring-npm/package-json#files

packages/terra-pills/src/Pills.jsx
```diff
@@ -0,0 +1,313 @@
...
+  children: PropTypes.node,
+  /**
+   * @private
+   * The intl object to be injected for translations.
+   */
+  intl: PropTypes.shape({ formatMessage: PropTypes.func }).isRequired,
+  /**
+   * Determines if the Pills group should be shown as rolled up or not, limited to a single line of display.
+   */
+  isCollapsed: PropTypes.bool,
```
mjhenkes, line 45: I'm not sure the isCollapsed and onSelectRollUp api's are going to be useful to consumers. I'm not sure when and how consumers would know when to force a tabs to collapse, once they're un collapsed theres no common way to collapse them and they will expand the area where you expand them so they can't easily fit in short containers.

Optionally maybe we could offer two states, single line or 'normal' non collapsed. The normal variant would never collapse and the single line variant would always be a single line but offer a disclosure mechanism to show and modify the collapsed pills.

Or maybe we could scope out collapsed for the first release?

ShettyAkarsh: @neilpfeiffer , @scottwilmarth  can you guys provide your thoughts on the above comment ?

neilpfeiffer: We'll be discussing the API and tech design further today and will post direction of how to address.

cc: @scottwilmarth

mjhenkes: I talked with @neilpfeiffer, on this topic, I think there was some confusion on what `isCollapsed` meant, so we'd like to change the `isCollapsed` prop to `isSingleLine`. A single line pill container may be collapsed or not depending on how many pills it contains.

packages/terra-framework-docs/src/terra-dev-site/test/pills/SelectablePills/SelectableAndRemovablePill.test.jsx
```diff
@@ -0,0 +1,91 @@
...
+      id: 'terra-pills-example-disclosure-removable-pill-bronchitis',
+    },
+    {
+      label: 'fibro',
+      ref: useRef(),
+      id: 'terra-pills-example-disclosure-removable-pill-fibro',
+    },
+  ];
+  const [pillsState, setPillsState] = useState(pillsData);
+  const [openPillIndex, setOpenPillIndex] = useState(undefined);
```
supreethmr, line 25:
```suggestion
  const [openPillIndex, setOpenPillIndex] = useState(-1);
```
Nit : Can we use -1 instead of `undefined`

packages/terra-framework-docs/src/terra-dev-site/test/pills/SelectablePills/SelectablePill.test.jsx
```diff
@@ -0,0 +1,83 @@
...
+      id: 'terra-pills-example-disclosure-removable-pill-bronchitis',
+    },
+    {
+      label: 'fibro',
+      ref: useRef(),
+      id: 'terra-pills-example-disclosure-removable-pill-fibro',
+    },
+  ];
+  const [pillsState] = useState(pillsData);
+  const [openPillIndex, setOpenPillIndex] = useState(undefined);
```
supreethmr, line 25: Nit : 
```suggestion
  const [openPillIndex, setOpenPillIndex] = useState(-1);
```

packages/terra-framework-docs/src/terra-dev-site/test/pills/Pills/CollapsiblePills.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+import Pills from 'terra-pills/lib/index';
+import styles from '../PillListTestCommon.module.scss';
+
+const cx = classNames.bind(styles);
+
+const CollapsibleBasicPills = () => {
+  const pillsData = [
+    {
+      label: 'asthma',
+      ref: useRef(),
```
tbiethman, line 12: There are two similarly named APIs in React for creating ref objects, `createRef` and `useRef`.

[createRef](https://reactjs.org/docs/refs-and-the-dom.html#creating-refs) will do exactly as it says, create a new instance of a ref object and return it. Every time you call createRef, you get a new ref object instance.

[useRef](https://reactjs.org/docs/hooks-reference.html#useref), on the other hand, is a hook that will create a single instance of a ref object and return that instance every time it is called, which should be with every update of component it's used in.

```jsx
const myRef = useRef();

// is functionally equivalent to

const myRef = useState(createRef())[0];
```

That said, your usage of useRef here should likely be replaced with createRef. Though your usage doesn't technically violate the [rules of hooks](https://reactjs.org/docs/hooks-rules.html), it's only because you're creating this pillsData object with every update, unnecessarily. The value is only used the first render to initialize state and is ignored thereafter. If you used createRef here, you could initialize your state like this:
```js

```jsx
const [pillsState] = useState(() => ([{
  label: 'asthma',
  ref: createRef(),
  id: 'terra-pills-example-disclosure-removable-pill-asthma',
}, ...]));
```

packages/terra-pills/src/SelectablePills.jsx
```diff
@@ -0,0 +1,337 @@
...
+
+    // needs to be verified - lp052179
+    if (isCollapsed && prevPillsCount.current === pillCount) {
+      focusNode.current = 0;
+    }
+
+    // focus the first pill( be it pill or a roll up pill)
+    if (pills.length > 0 && pills.length > focusNode.current) {
+      currentPill.current = pills[focusNode.current].id;
+      setTabIndex('0');
```
ShettyAkarsh, line 152: Can we declare the these values as a constant ?

const tabbingOn  = 0
const tabbingOff = -1

lokesh-0813:
> It seem like whether or not a pill is truncated is calculated only at first render. On screen resize the pop is either not shown on a truncated pill or still available on a non truncated pill.
> pills resize small to large
> pills resize large to small

fixed in this commit 91a9d8766efc2524f8114e0d1f9332318e52b26f

packages/terra-pills/LICENSE
lokesh-0813: fixed in this commit 91a9d8766efc2524f8114e0d1f9332318e52b26f

packages/terra-pills/README.md
lokesh-0813: fixed in this commit 91a9d8766efc2524f8114e0d1f9332318e52b26f

packages/terra-pills/src/Pills.jsx
lokesh-0813: renamed in this commit 91a9d8766efc2524f8114e0d1f9332318e52b26f

packages/terra-pills/src/SelectablePills.jsx
lokesh-0813: `setTabIndex('0')` is actually a function.  It sets "-1" for the pill that is about to lose focus and "0" for the pill that is about to receive focus. 

[Here](https://github.com/cerner/terra-framework/pull/1442/files#diff-30d1befc1f5f254f8705585f759e7da122677fbeb6b8f3ec84ff694b83b53bd6R107) is a reference to it .

ShettyAkarsh: Yup, what i was mentioning above is instead of passing the hardcoded value (0,-1 in this case) everytime. can we declare it as a constant and pass ?

packages/terra-pills/src/Private/Pill.module.scss
```diff
@@ -0,0 +1,197 @@
+@import '~terra-mixins/lib/Mixins';
+
+// Themes
+@import './../clinical-lowlight-theme/Pill.module';
+@import './../orion-fusion-theme/Pill.module';
+
+:local {
```
pranav300, line 7: Can we separate the styles into different files as per the jsx component?

packages/terra-pills/src/Private/PillsUtils.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  return startIndex;
+};
+
+const getFirstFocusableNode = (pills) => {
+  let firstIndex = -1;
+  if (pills.length > 0) {
+    for (let i = 0; i < pills.length; i += 1) {
+      if ((pills[i].hasAttribute('data-terra-pill-removable')) || (pills[i].hasAttribute('data-terra-rollup-pill')) || (pills[i].hasAttribute('aria-haspopup'))) {
```
ShettyAkarsh, line 40: do we need this check for roll up pill here ? as we are only trying to loop through the pills having attribute 'data-terra-pill'

packages/terra-pills/src/Pills.jsx
```diff
@@ -0,0 +1,383 @@
...
+  useEffect(() => {
+    const pills = [...pillsRef.current.querySelectorAll('[data-terra-pill]')];
+    const rollUpPill = pillsRef.current.querySelector('[data-terra-rollup-pill]');
+    const nextFocusableNode = PillsUtils.getNextFocusableNode(pills, focusNode.current);
+    const prevFocusableNode = PillsUtils.getPreviousFocusableNode(pills, focusNode.current);
+    const startIndex = PillsUtils.handleRollUp(pillsRef);
+
+    if (isPillDeleted.current && React.Children.count(children) > 0) {
+      if (nextFocusableNode === -1 && prevFocusableNode === -1 && rollUpPill === null) {
+        focusPillsContainer();
```
ShettyAkarsh, line 163: So are we still focusing the pills container  when all the pills have been removed ?

packages/terra-pills/src/Pills.jsx
```diff
@@ -0,0 +1,383 @@
...
+        setTabIndex('0');
+        focusCurrentNode();
+        isRollUpRemoved.current = false;
+      }
+    }
+  }, [children, updatedCount]);
+
+  const focusNextNode = (pills, rollUpPill) => {
+    const allPills = React.Children.toArray(children);
+    const nextFocusableNode = PillsUtils.getNextFocusableNode(pills, focusNode.current + 1);
```
ShettyAkarsh, line 210:
```suggestion
    const nextFocusableNode = PillsUtils.getNextFocusableNode(pills, focusNode.current);
```

packages/terra-pills/src/Pills.jsx
```diff
@@ -0,0 +1,383 @@
...
+    if (focusNode.current + 1 === pills.length && rollUpPill) {
+      setTabIndex('-1');
+      currentPill.current = rollUpPill.getAttribute('id');
+      setTabIndex('0');
+      focusCurrentNode();
+      return;
+    }
+
+    // focus the next pill
+    if (focusNode.current + 1 < pills.length) {
```
ShettyAkarsh, line 222:
```suggestion
    if (focusNode.current + 1 < pills.length || focusNode.current + 1 === pills.length) {
```
we could remove the if block between L212-L219.  along with the above changes mentioned for L210. add the extra condition here. so that we can remove the extra if block.

packages/terra-pills/src/Private/Pill.module.scss
lokesh-0813: Yes we can, there are more changes to existing styles soon. So will sort it out then.

packages/terra-pills/src/Pills.jsx
```diff
@@ -0,0 +1,383 @@
...
+      }
+
+      if (rollUpPill && startIndex >= 1) {
+        currentPill.current = pills[focusNode.current].id;
+      } else if (rollUpPill) {
+        currentPill.current = rollUpPill.getAttribute('id');
+      } else if (prevFocusableNode >= 0) {
+        focusNode.current = prevFocusableNode;
+        currentPill.current = pills[focusNode.current].id;
+      } else {
```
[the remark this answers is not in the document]
lokesh-0813: Thats the only condition that's left out, so i thought it was safe to skip it. I will add it if its necessary, let me know.

packages/terra-pills/src/Pills.jsx
lokesh-0813: No we cannot remove that,  the block between L212 - L219 is specific to rollup pill if it exists and only applies if the last focused pill is just before the roll up pill.

 Where as the block from L222 - 237 is for pills surrounded by other focusable pills.

packages/terra-pills/src/Private/PillsUtils.js
lokesh-0813: nice catch, I forgot to remove it earlier.

Removed in this commit 29ffe143cdcdef2efffe4c60df4142ad768a15d5

packages/terra-pills/src/SelectablePills.jsx
```diff
@@ -0,0 +1,371 @@
...
+  }, [children]);
+
+  const setTabIndex = (val) => {
+    const currentNode = currentPill.current ? selectablePillsRef.current.querySelector(`[id=${currentPill.current}]`) : null;
+    if (currentNode) {
+      currentNode.setAttribute('tabIndex', val);
+    }
+  };
+
+  const focusCurrentNode = () => {
```
pranav300, line 114: Similarly, we can apply optional chaining over here:
```jsx
const focusCurrentNode = () => currentPill.current && selectablePillsRef.current.querySelector(`[id=${currentPill.current}]`)?.focus();
```

packages/terra-pills/src/Pills.jsx
```diff
@@ -0,0 +1,383 @@
...
+  useLayoutEffect(() => {
+    if (isSingleLine) {
+      generateRollUp();
+    } else {
+      generateExpansion();
+    }
+  }, [children, generateExpansion, generateRollUp, isSingleLine]);
+
+  const setTabIndex = (val) => {
+    const currentNode = currentPill.current ? pillsRef.current.querySelector(`[id=${currentPill.current}]`) : null;
```
pranav300, line 107: Similarly, we can apply optional chaining over here:
```jsx
const setTabIndex = (val) => currentPill.current && selectablePillsRef.current.querySelector(`[id=${currentPill.current}]`)?.setAttribute('tabIndex', val);
```

packages/terra-pills/src/Pills.jsx
ShettyAkarsh: But in case when there is one selectable pill and a rollup pill. if the focus is on first selectable pill. then current 'getNextFocusableNode' returns  '-1 ' if we pass 'focusNode.current + 1' here. it should return 0 if we are using pills index or 1 if we are going with the pills count.

packages/terra-pills/src/Pills.jsx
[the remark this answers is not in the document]
ShettyAkarsh: if there is a way 'nextFocusableNode' might set to -1. then it would be better to set the condition here.

packages/terra-pills/src/Pills.jsx
ShettyAkarsh: yes that correct. by adding the condition here and removing the above if block. will that not work for both the scenarios ?

packages/terra-pills/src/Pills.jsx
lokesh-0813: If there is just one focusable pill and a rollup pill, clicking on "right-arrow" should focus the rollup pill.
In this case you are right,  "getNextFocusableNode' returns '-1'.  But since there is just one pill, the block between L213 - L 218 is hit and is returned. The negative value of '-1' is neglected.

packages/terra-pills/package.json
mjhenkes: This still needs fixed or the package won't be useable when released.

packages/terra-pills/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "terra-pills",
+  "main": "lib/index.js",
```
mjhenkes, line 3: Make the package private until first release.
```suggestion
  "main": "lib/index.js",
  "private": true,
```

mjhenkes:
SelectablePills and Pills are nearly identical and contain a lot of repeated code. I believe that you could use selectable pills as base class for pills where you restrict the API. This would reduce memory footprint and make the package easier to maintain.

mjhenkes:
Has the documentation for Pills not been added yet?

packages/terra-pills/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "terra-pills",
```
mjhenkes, line 2: Lets make this a scoped package
```suggestion
  "name": "@cerner/terra-pills",
```

packages/terra-framework-docs/src/terra-dev-site/test/pills/Pills/CollapsiblePills.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+        {pillsState.map((pill, index) => (
+          <Pills.Pill
+            label={pill.label}
+            id={pill.id}
+            key={pill.id}
+            pillKey={pill.label}
+            isRemovable
+            metaData={{ index }}
+            // eslint-disable-next-line no-param-reassign
+            refCallback={(node) => { pill.ref.current = node; }}
```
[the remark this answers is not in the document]
tbiethman: To be completely clear, we should be providing a ref along with the onSelect callback to simplify our consumers' implementations. We should leave a refCallback in place on the pill to give consumers an option, but providing a ref in the callback should be sufficient for most use cases.

packages/terra-pills/package.json
lokesh-0813: changed here ee1622133ac54f2deacf7efbaa829413121f73e6

packages/terra-pills/src/Pills.jsx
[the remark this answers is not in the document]
lokesh-0813: added here 6bb6d29c876a914ff4bac5cc4d42707b47563011

lokesh-0813:
> Has the documentation for Pills not been added yet?

cc : @neilpfeiffer

lokesh-0813:
> SelectablePills and Pills are nearly identical and contain a lot of repeated code. I believe that you could use selectable pills as base class for pills where you restrict the API. This would reduce memory footprint and make the package easier to maintain.

Although the container markup of both "SelectablePills" and "Pills" look identical, the key changes are 

1) when we render the pills, we modify and add certain props both in Pills and Selectable pills. Adding a condition might look a bit shabby. 
2) The pills in "SelectablePills" are all homogenous, while "Pills"(Basic Pills) have both selectable and non-selectable pills. It's straight forward to restrict the "onSelect" API, but keyboard navigation becomes a lot messier with various conditions. Also the initial mount and after deletion of a pill have different approaches for setting the first/next focusable node. I thought it would be cleaner to maintain separate files considering the nature of pills rendered within the containers.

I have extracted most of the repetitive code to "PillsUtils.js".

Let me know your thoughts.

packages/terra-pills/package.json
lokesh-0813: changed here 5158250e39187534d43a615c7ee12bd447c7ab6b

packages/terra-pills/package.json
lokesh-0813: changed here 5158250e39187534d43a615c7ee12bd447c7ab6b

packages/terra-pills/package.json
lokesh-0813: changed here 5158250e39187534d43a615c7ee12bd447c7ab6b

packages/terra-framework-docs/src/terra-dev-site/test/pills/Pills/CollapsiblePills.test.jsx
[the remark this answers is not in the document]
lokesh-0813: updated "onSelect" callback to return "pillRef". Adjusted the test examples.

Changed in this commit c862d829afb03ebc25181ec8648fd069bce01aa7

packages/terra-pills/src/Pills.jsx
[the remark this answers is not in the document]
lokesh-0813: changed here 6bb6d29c876a914ff4bac5cc4d42707b47563011

packages/terra-pills/src/Pills.jsx
pranav300: How about we update the function to something like this:
```jsx
  const focusNextNode = (pills, rollUpPill) => {
    const allPills = React.Children.toArray(children);
    const nextFocusableNode = PillsUtils.getNextFocusableNode(pills, focusNode.current + 1);

    // focus the next pill
    if (focusNode.current + 1 < pills.length || focusNode.current + 1 === pills.length) {
      setTabIndex('-1');
      // if there no focusable pills or the next pill is rollUpPill, focus the rollUpPill
      if ((nextFocusableNode === focusNode.current || focusNode.current + 1 === pills.length) && rollUpPill) {
        currentPill.current = rollUpPill.getAttribute('id');
      } else if (nextFocusableNode !== -1) {
        focusNode.current = nextFocusableNode;
        currentPill.current = allPills[focusNode.current].props.id;
      }
      setTabIndex('0');
      focusCurrentNode();
    }
  };
```

packages/terra-pills/src/Pills.jsx
```diff
@@ -0,0 +1,385 @@
...
+    if (isRemovable) {
+      // If there is only one pill and is removable
+      if (pills.length === 1 && rollUpPill === null) {
+        focusPillsContainer();
+        return;
+      }
+      if (pills.length - 1 < React.Children.count(children)) {
+        setTabIndex('-1');
+        if (focusNode.current === 0) { // If the first pill is deleted, focus the next pill
+          focusNode.current = 0;
```
pranav300, line 267: Is this condition necessary, can't we have something like this:
```jsx
if (focusNode.current > 0) focusNode.current -= 1;
```

packages/terra-pills/src/private/_Pill.jsx
```diff
@@ -0,0 +1,290 @@
...
+    { 'is-selectable': !!onSelect || (isBasicPill && isTruncated) },
+    { 'is-removable': isRemovable },
+    { 'is-selectable-and-removable': (!!onSelect && isRemovable) || (isRemovable && isTruncated) },
+  ]);
+
+  const removeButtonClassNames = cx([
+    'pill-remove-button',
+  ]);
+
+  const popupConent = () => {
```
pranav300, line 225: Typo

packages/terra-framework-docs/src/terra-dev-site/test/pills/Pills/RemovablePill.test.jsx
```diff
@@ -0,0 +1,64 @@
...
+      label: 'bronchitis',
+      id: 'terra-removable-pill-bronchitis1',
+      isRemovable: true,
+    },
+    {
+      label: 'fibro',
+      id: 'terra-removable-pill-fibro1',
+      isRemovable: false,
+    },
+  ];
```
supreethmr, line 36: can we move this pills data to separate file and loop it through `React.Children.map` in the required test files.

packages/terra-framework-docs/src/terra-dev-site/test/pills/SelectablePills/CollapsibleSelectablePills.test.jsx
```diff
@@ -0,0 +1,108 @@
...
+    {
+      label: 'cardiac',
+      id: 'terra-pills-example-disclosure-removable-pill-cardiac',
+    },
+    {
+      label: 'allergies',
+      id: 'terra-pills-example-disclosure-removable-pill-allergies',
+    },
+  ];
+  const [pillsState, setPillsState] = useState(pillsData);
```
supreethmr, line 37: nit : 
```suggestion
  const [pills, setPills] = useState(pillsData);
```

packages/terra-pills/src/Pills.jsx
pranav300: How about we update the function to something like this:
```jsx
  const focusNextNode = (pills, rollUpPill) => {
    const allPills = React.Children.toArray(children);
    const nextFocusableNode = PillsUtils.getNextFocusableNode(pills, focusNode.current + 1);

    // focus the next pill
    if (focusNode.current + 1 < pills.length || focusNode.current + 1 === pills.length) {
      setTabIndex('-1');
      // if there no focusable pills or the next pill is rollUpPill, focus the rollUpPill
      if ((nextFocusableNode === focusNode.current || focusNode.current + 1 === pills.length) && rollUpPill) {
        currentPill.current = rollUpPill.getAttribute('id');
      } else if (nextFocusableNode !== -1) {
        focusNode.current = nextFocusableNode;
        currentPill.current = allPills[focusNode.current].props.id;
      }
      setTabIndex('0');
      focusCurrentNode();
    }
  };
```

packages/terra-pills/src/Pills.jsx
```diff
@@ -0,0 +1,385 @@
...
+      return undefined;
+    });
+
+    const reducedArray = pills.splice(0, updatedCount);
+
+    return reducedArray;
+  };
+
+  return (
+    <ResponsiveElement responsiveTo="window" onResize={handleWidthChange}>
```
tbiethman, line 356: You likely want to be responsive to "element" here. It's more expensive, but as-is, changes to the container size that aren't the result of the window size changing will not be handled.

packages/terra-framework-docs/src/terra-dev-site/test/pills/SelectablePills/SelectableAndRemovablePill.test.jsx
lokesh-0813: Undefined works similar.

packages/terra-framework-docs/src/terra-dev-site/test/pills/Pills/RemovablePill.test.jsx
lokesh-0813: The pill data is not identical in most scenarios. If there arises such a situation, I shall move it to a single file.

packages/terra-pills/src/Pills.jsx
lokesh-0813: changed in this commit f2c63babe26a06cb3b0ee3ba1deb77728bd5a0ef

PR #1445, opened 2021-08-09
[terra-date-input] Fix to prevent non-numeric input in safari

### LINKED ISSUE #1436: Terra Date Input doesn't work as expected in Safari browser
# Bug Report

## Description
If you use terra date input in Safari browser, you can enter letters to the fields despite that the type of input is "number". After you enter these letters, you can't delete them using the 'Backspace' or 'Delete' button. The issue happens only in the Safari browser.

## Steps to Reproduce

1. Go to https://engineering.cerner.com/terra-ui/components/terra-date-input/date-input/date-input using Safari browser
2. Try to enter letters in the 'Year' and 'Day' fields.
3. Then, try to delete them.

Screenshot 2021-07-21 at 12 35 56

## Expected Behavior
If you use other browsers, you can't write letters to these inputs.

## Possible Solution
I suggest moving this line to the top so users can override the type of input or any other solution that you can offer.
https://github.com/cerner/terra-framework/blob/6e6145e5138de8d56598975cf35316f0f794a920/packages/terra-date-input/src/DateInput.jsx#L552

saket2403:
### Summary
Safari allows non numeric input for input type=number by default, so added a fix to prevent non-numeric input.

Closes #1436 

### Deployment Link

### Additional Details

@cerner/terra


ShettyAkarsh (approved):

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -264,6 +266,17 @@ class DateInput extends React.Component {
...
+  /**
+   * Takes a key press from the day and year input, and processes it based on the value of the keycode
+   * Prevents non-numeric characters from being entered in Safari browser.
+   * @param {Object} event Event object generated from the event delegation.
+   */
+  handleKeyPress = event => {
+    const keyCode = (typeof event.which === 'undefined') ? event.keyCode : event.which;
+    const input = String.fromCharCode(keyCode);
+    if (!input.match(/^[0-9]+$/) && !event.metaKey) event.preventDefault();
+  }
```
supreethmr, line 278: both event.keyCode and event.which are deprecated : 
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode and https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which 

So instead of `keyCode` and `which` can we use [key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) property of keyboard event object. which returns the actual value of the key pressed instead of unicode value. This eliminates need of converting unicode value to string again for comparsion with regex.


```suggestion
  handleKeyPress = event => {
    const input = event.key;
    if (!input.match(/^[0-9]+$/) && !event.metaKey) event.preventDefault();
  }
```

pranav300 (approved):

supreethmr (approved):

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -264,6 +266,16 @@ class DateInput extends React.Component {
...
   }
 
+  /**
+   * Takes a key press from the day and year input, and processes it based on the value of the keycode
+   * Prevents non-numeric characters from being entered in Safari browser.
+   * @param {Object} event Event object generated from the event delegation.
+   */
+  handleKeyPress = event => {
+    const input = event.key;
+    if (!input.match(/^[0-9]+$/) && !event.metaKey) event.preventDefault();
```
benbcai, line 276: Does this change work if a non-numeric key is copied/pasted in the input using mouse click menu or the Command-C/V keys instead of entering the non-numeric key directly using the keyboard?  If copy/paste doesn't work, something to try is to move this logic to `handleValueChange` similar to [this](https://github.com/cerner/terra-framework/blob/terra-date-picker%404.65.0/packages/terra-date-picker/src/DateInput.jsx#L149-L151).

benbcai: Talked to Saket to keep the flexing of the input types and how handle preventing character key for Safari. Also need to handle copy and paste characters.

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -264,6 +265,26 @@ class DateInput extends React.Component {
...
   }
 
+  /**
+   * Checks Paste event in the day and year input, and processes it based on the value of the keycode
+   * Prevents non-numeric characters from being entered in Safari browser.
+   * @param {Object} event Event object generated from the event delegation.
+   */
+  handlePaste = event => {
+    const input = (event.clipboardData || window.clipboardData).getData('text');
+    if (!input.match(/^[0-9]+$/)) event.preventDefault();
```
benbcai, line 275: Would it work if we validate the key using `DateInputUtil.validNumericInput(input)` so that we are consistent with the other places in how we validate key inputs? Same for `handleKeyPress`.

Something like `if (!DateInputUtil.validNumericInput(input)) event.preventDefault();`

saket2403: Done 78e90ec

benbcai (approved):

ryanthemanuel (approved):

PR #1456, opened 2021-09-17
Remove auto focus from date-time-picker

BenBoersma:
### Summary

Removing the autoadvance focus that happens when you enter a valid date. Also turning back on all tests and regenerating terra-date-time-picker wdio screenshots (these probably need to get checked later to make sure they are correct)

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -311,21 +311,6 @@ class DateTimePicker extends React.Component {
     } else {
       this.setState({ dateTime: updatedDateTime });
     }
-
-    if (isDateValid && (event.currentTarget === this.dateInputComponent || this.isDateSelected)) {
```
benbcai, line 315: `this.isDateSelected` and `this.dateInputComponent` and  are needed only for advancing the focus to the hour input. Since we're no longer doing this, all references to `this.isDateSelected` and `this.dateInputComponent` leading up to where `lastInputRefCallback` is set in [DateInput.jsx](https://github.com/cerner/terra-framework/blob/split-date-picker-inputs/packages/terra-date-picker/src/DateInput.jsx#L199-L219) can also be removed.

benbcai: https://github.com/cerner/terra-framework/pull/1456/commits/e29ff7ad8be31f0a85b93ae550f3fc0c556f7415

benbcai (approved):

mjhenkes (approved):

PR #1464, opened 2021-10-05
Main merge and regenerate screenshots

BenBoersma:
### Summary
Merging in main to `split-date-picker-inputs` branch and regenerating some screenshots.

Also fix where terra-form-input was overwriting some css in terra-date-picker. Fixed by increasing specificity on css applied to terra-date-picker inputs.

Closes #

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-date-picker/src/DatePicker.module.scss
```diff
@@ -81,9 +81,9 @@
     }
   }
 
-  .date-input-month,
-  .date-input-day,
-  .date-input-year {
+  .date-input .date-input-month,
+  .date-input .date-input-day,
+  .date-input .date-input-year {
```
BenBoersma, line 86: This change was to deal with an issue where terra-form-input css would overwrite this css due to specificity rules. This seems to work but I'd happy with second opinions.
@tbiethman @benbcai

benbcai:
@BenBoersma Would it be easier to merge main on the `split-date-picker-inputs` branch, then issue a PR with only the new changes you add on top of it. The way this PR is now is it has all the other changes that are already in main and that makes it harder to identify the actual changes to review.

BenBoersma:
> @BenBoersma Would it be easier to merge main on the `split-date-picker-inputs` branch, then issue a PR with only the new changes you add on top of it? The way this PR is now is it has all the other changes that are already in main and that makes it harder to identify the actual changes to review.

@benbcai Merged main into `split-date-picker-inputs` branch and then remerged `split-date-picker-inputs` into this one. only differences should now be my changes and the regenerating of screenshots.

packages/terra-date-picker/src/DatePicker.module.scss
benbcai: Does it work if we just move these inside the .date-input class on line 144 so that we don't need to add it to each one of these?

PR #1469, opened 2021-10-08
Time input a11y 2

xenoworf:
### Summary
UXPLATFORM-5728 TimeInput screen reader and voice input support

This is a starting point for screen reader support. Preliminary support for voice input. We will solicit feedback from the A11Y community via Slack before releasing this change.

**Enhancements and fixes**
1. Use i18n messages with placeholders rather than concatenating them together.
2. The field has an accessible name, but it's not visible. That is the wrapper's responsibility.
3. Each sub-field (e.g. hour, minute) has its own format description (e.g. 'Format: hh').
4. Each sub-field has an accessible name <field name> <subfield name>. It is not visible though.
5. The meridian buttons are semantically grouped together and have an accessible name.
6. The field has an invisible region that reads the full human-readable localized value, e.g. "09:22 pm".
7. When the user changes the field to another valid time (e.g. 09:23pm) the invisible region becomes a status region and will read off the updated times.

The next step is to add more A11Y to the consumers of this component. For example, the visible label (E.g. 'Birth Time') of a web form consumer should be updated to pass that label into this component via props, so that users using assistive technologies will perceive the accessible name as the visible name, and vice-a-versa.

**TODOs (work not done in this change, but identified for later)**
1. Give the meridiem buttons accessible names that include the context of which time input they're grouped with, for cases where multiple time inputs are presented at once.
2. i18n the Time Input's default label.
3. Jest and WDIO tests to prevent regressions.
4. Update A11Y to address feedback from the accessible community.
5. Test against IE 10 and 11

Passed all TimeInput tests manually using VoiceOver + Edge on macOS

### Testing
I've manually tested Voice Over, macOS's native screen reader, with Edge against all of the raw Time Input scenarios linked to form the dev site.

### DIFF
--- a/packages/terra-time-input/package.json
+++ b/packages/terra-time-input/package.json
@@ -33,6 +33,7 @@
     "terra-button-group": "^3.0.0",
     "terra-form-input": "^4.4.0",
     "terra-theme-context": "^1.7.0",
+    "terra-visually-hidden-text": "^2.0.0",
     "uuid": "^3.3.2"
   },
   "scripts": {
--- a/packages/terra-time-input/src/TimeInput.jsx
+++ b/packages/terra-time-input/src/TimeInput.jsx
@@ -9,8 +9,11 @@ import { injectIntl } from 'react-intl';
 import uuidv4 from 'uuid/v4';
 
 import * as KeyCode from 'keycode-js';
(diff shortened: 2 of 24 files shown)


elliott-hoffman-cerner:
Hmmm, my git-fu has failed as this is showing way too many commits. Should be one commit. I'll fix this.

PR #1480, opened 2021-11-02
Bug? TimeInputs are sharing hourAttributes

elliott-hoffman-cerner:
### Summary
I'm seeing strange behavior: when I have many TimeInputs on one page, they are all sharing the same hourAttributes! Perhaps I'm doing something wrong, but I can't figure out what.

### How to reproduce
Visit theDeployment/raw/tests/terra-time-input/time-input/twelve-hour/a-11-y-name-description and inspect the _Accessibility > ARIA Attributes_ of the 1st and 2nd TimeInputs on the page. They both have the same aria-label property, even though that value is only assigned on the 1st TimeInput. In my code the aria label is being set by conditionally adding it to the hours attribute of the TimeInput input. 

### Expected behavior
The aria-label property is only assigned to the 1st TimeInput's hour input.

### Code lines of note
The first TimeInput gets its aria-label set here:
https://github.com/cerner/terra-framework/pull/1480/files#diff-732b6688487eadd7f36d6f81eaa310986eaaedb98eb4faa11bfca2e34eac6afbR11

The second TimeInput does not get an assigned aria-label: https://github.com/cerner/terra-framework/pull/1480/files#diff-732b6688487eadd7f36d6f81eaa310986eaaedb98eb4faa11bfca2e34eac6afbR16

This is my code where I noticed the problem. I'm adding aria-label to the Hour attribute of the TimeInput: https://github.com/cerner/terra-framework/pull/1480/files#diff-37c08a2b0d74d4e14dd7fc4e7e160bb6ac190dd983aebaaf30397c71621b735dR755

### Deployment Link

https://terra-framework-deployed-pr-1480.herokuapp.com/


elliott-hoffman-cerner:
@tbiethman figured it out: I'm mutating the default value of the hourAttributes prop, which is causing all subsequent TimeInputs to get the same value. Closing this PR. Thanks, TIL not to do that!

Issue #996, opened 2019-12-11
Enable hot keys for Terra Date/Time field


JC014157:
# Feature Request

## Description
Allow for end users to quickly enter Today and Now with keyboard shortcuts `T` and `N`.  
Allow users to quickly increment and decrement date and time with keyboard shortcuts `+` and `-`.

## Additional Context / Screenshots
These are common short cut keys that are used for date/time controls in PowerChart as well as other Cerner applications. Adding them into Terra will maintain consistency with other existing Win64 and Java applications.

---
## Future Enhancements
Additional keyboard shortcuts to allow for end users to quickly enter keyboard shortcuts for:
first and last day of the year (`Y` and `R`), first and last day of the month (`M` and `H`), first and last day of the week (`W` and `K`), etc.

pranav300:
I have some doubts regarding this issue -
1. Should we also enable hot keys even when the calendar is open?
2. And should I also enable hot keys for the new component terra-date-input?

cc / @ryanthemanuel, @BenBoersma

benbcai:
@pranav300 We will have a meeting today to review and discuss how to handle this request. We will provide more information later.

benbcai:
@pranav300 Our meeting got rescheduled to later this week. I will update you when we have more information.

ShettyAkarsh:
@benbcai, Any updates on this issue ?

benbcai:
@ShettyAkarsh There are still some ongoing conversations and engaging the human factor team around how this should work in Terra. UX will provide updates once we know more.

scottwilmarth:
Providing shortcuts is a great way to provide for disabled users, especially mobility challenged keyboard-only users. The use of "t" for today, "n" for now, and others are standard keyboard shortcuts that are already existing within both clinical and revenue cycle products. Some clinical products use the "Shift +" in addition to the letter, but the best practice would be to not require "Shift" in addition to a letter. While single letter shortcuts should not be used usually, they are perfectly acceptable in this instance because the date-time input would be in focus for the key commands to work. Here are the basic rules for implementing the commands: 1. Must be expected behavior (consistent and predictable - go with what users already use for the letter, but lose the "Shift +"). 2. Must only occur when the input has focus. 3. Must be documented in solution documentation. 
More investigation would be needed to document the current date-time shortcuts across clinical and revenue cycle solutions (possibly consumer as well) and reconciled for final commands.

neilpfeiffer:
[...]

PR #1483, opened 2021-11-04
[terra-date-input] Adding initial hot keys

BenBoersma:
### Summary
 Adding 'T', '-', and '+' hot keys to set date on an empty date input or to increment or decrement the date by a day.

Closes #

### Deployment Link

- [date-input](https://terra-framew-terra-date-5asksy.herokuapp.com/components/terra-date-input/date-input/date-input)
- [date-input-field](https://terra-framew-terra-date-5asksy.herokuapp.com/components/terra-date-input/date-input/date-input-field)
- [Hotkeys Doc](https://terra-framew-terra-date-5asksy.herokuapp.com/components/terra-date-input/date-input/date-input-hotkeys)

### Additional Details

@cerner/terra


packages/terra-date-input/src/DateInput.jsx
```diff
@@ -442,6 +470,35 @@ class DateInput extends React.Component {
     }
   }
 
+  /**
+   * Sets state to new value dependent on existing date
+   * @param {Object} event Event object generated from the event delegation.
+   * @param {Number} addDays Adds days to current date or today's date.
+   */
+  setHotKeyDate(event, addDays) {
+    if (!this.props.value || event.keyCode === KeyCode.KEY_T) {
```
benbcai, line 479: We should use the current state of the date instead of using this.props.value. For example, if this.props.value was 01/01/2021 but the user changed it to 02/02/2020 and pressed `T`, `-`, or `+`, we would want to calculate the new date using 02/02/202 instead of the value passed in.

BenBoersma: Fixed here: https://github.com/cerner/terra-framework/pull/1483/commits/520855d1e17fd222b3585c138de09be411922f8a#diff-ea35ec190c52db171ef5f774b071e80db0b038abff0e19580b310959fc582329R481

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -442,6 +468,30 @@ class DateInput extends React.Component {
...
+   * Sets state to new value dependent on existing date
+   * @param {Object} event Event object generated from the event delegation.
+   * @param {Number} addDays Adds days to current date or today's date.
+   */
+  setHotKeyDate(event, addDays) {
+    let dateObj;
+    if (!this.state.month || !this.state.year || !this.state.day || addDays === 0) {
+      dateObj = new Date();
+    } else {
+      dateObj = new Date(`${this.state.year}-${this.state.month}-${this.state.day}`);
```
benbcai, line 481: What would happen if a month is selected (this.state.month is valid) but both `this.state.year` and `this.state.day` contain a value but they are invalid? For example: day is `6` and year is `202`. In this case it would get in here and create an invalid Date object to add `addDays`.

The logic might be easier if we check to ensure we have a full and complete date (month, day, and year are valid), then create the date based on these values and then add `addDays` to it. Otherwise, we don't have a complete date value and should fall back to use today's date.

BenBoersma: Added some logic for what constitutes a valid day and a valid year: https://github.com/cerner/terra-framework/pull/1483/commits/9d7f89c03340c4863ac691af55912ac4bd0c7e81

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -442,6 +468,32 @@ class DateInput extends React.Component {
...
+  /**
+   * Sets state to new value dependent on existing date
+   * @param {Object} event Event object generated from the event delegation.
+   * @param {Number} addDays Adds days to current date or today's date.
+   */
+  setHotKeyDate(event, addDays) {
+    const validYear = (this.state.year.length === 4 && Number(this.state.year) < DateInputUtil.MaxYearValue && Number(this.state.year) > DateInputUtil.MinYearValue);
+    const validDay = (Number(this.state.day) <= 31 && Number(this.state.day) > 0);
+    let dateObj;
+    if (!this.state.month || addDays === 0 || !validYear || !validDay) {
```
benbcai, line 480: If any of the year, day or month is invalid but not all three are empty (basically we don't have a complete date), and I press either `-` or `+`, it will get in here and create today's day. According to the requirements, we should ignore the -/+ key press and do nothing instead of continuing to add/subtract a day from today's date.

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -442,6 +468,32 @@ class DateInput extends React.Component {
...
   }
 
+  /**
+   * Sets state to new value dependent on existing date
+   * @param {Object} event Event object generated from the event delegation.
+   * @param {Number} addDays Adds days to current date or today's date.
+   */
+  setHotKeyDate(event, addDays) {
+    const validYear = (this.state.year.length === 4 && Number(this.state.year) < DateInputUtil.MaxYearValue && Number(this.state.year) > DateInputUtil.MinYearValue);
+    const validDay = (Number(this.state.day) <= 31 && Number(this.state.day) > 0);
```
benbcai, line 478: The date-input as it is today allows to use the Up arrow key to increase the day to 31 regardless how many days are in a given month. For example, February has only 28 days on non-leap years but it allows to enter up to 31.  When this occur and the user presses the `+` key, instead of updating the date `March 01 2021` it would actually update it to `March 04 2021`. So it is actually 3 days (31 - 28) ahead of what it should be. We should discuss how to best handle this type of scenario.


In this workflow, I pressed the UP key several times to increase the day to 31 and then press the `+` just once. It advanced the date to `March, 04 2021`

date-input-hotkeys

packages/terra-date-input/src/DateInput.jsx
benbcai: After further discussion and clarification, pressing the `-` or `+` key when the date is incomplete should continue to increment/decrement from today's date. The above comment can be ignored.

For example, if month is `February` and day is `01` but the year is `20` which is incomplete, pressing `+` should set to tomorrow's date (today + 1).

packages/terra-date-input/src/DateInput.jsx
BenBoersma: The changes here https://github.com/cerner/terra-framework/pull/1483/commits/64ce6f898cee6c84c6a041722c74baa14a2d84c7#diff-ea35ec190c52db171ef5f774b071e80db0b038abff0e19580b310959fc582329R479 should deal with this. We now consider a date that has more days in it then the maximum an "invalid" date.

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -442,6 +468,32 @@ class DateInput extends React.Component {
     }
   }
 
+  /**
+   * Sets state to new value dependent on existing date
+   * @param {Object} event Event object generated from the event delegation.
+   * @param {Number} addDays Adds days to current date or today's date.
+   */
+  setHotKeyDate(event, addDays) {
+    const validYear = (this.state.year.length === 4 && Number(this.state.year) < DateInputUtil.MaxYearValue && Number(this.state.year) > DateInputUtil.MinYearValue);
```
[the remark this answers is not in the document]
BenBoersma: Destructed here https://github.com/cerner/terra-framework/pull/1483/commits/b2a9136417f7946cf943d10bef8aab26a9fc5ca6

packages/terra-date-input/src/DateInput.jsx
benbcai: I think making days that exceed the total number of days in the month as invalid makes sense. However, from an end user's perspective the new change doesn't seem like what they would expect.

Using the same workflow as above, instead of advancing the date to `March, 04 2021` like what it did before it now changes the date from the invalid `February, 31 2021` date to tomorrow's date (December, 09, 2021) when pressing the `+` key. I think the user would expect it to advance the date to `March, 01 2021`.

This is what I'm thinking but would like to get @neilpfeiffer and @chrismichalewicz 's thoughts.
- When we have an incomplete date (missing either a month, day, or year), pressing `-`/`+` should change to yesterday/tomorrow respectively.

- When we have an invalid date (have month, day and year but the date is not a valid date; e.g,; outside the range of a valid date), pressing `-`/`+` should change to the previous/next valid date respectively.

If the current behavior is not what we want, we could check if we have a 2 digit day that exceeds the total number of days in a month and if so advance to the first day of the next month.

date_input

elliott-hoffman-cerner:
What about spec tests for pressing the keys in a Month input? I see day and year but I didn't see month tests.

elliott-hoffman-cerner (commented):
What about spec tests for pressing the keys in a Month input? Seems like those would be good to have.

BenBoersma:
> What about spec tests for pressing the keys in a Month input? Seems like those would be good to have.
@elliott-hoffman-cerner 
Due to innate browser handling of key presses in <Select> tags, we held off implementing the hot keys for the month select option. The specfic issue was our select has a default option that is "- Select -" so when we would press the "-" key it would automatically move the month select to that option before doing our hotkey logic. When that happens we never get a valid date so we would always use today. Similar issue with the "T" key. There was general concerns about later hotkeys depending on what they were.

BenBoersma:
> > Will the dev site doc to describe the hotkey behavior be part of this PR?
> 
> I think we should have some documentation describing what hotkeys are supported, their behaviors, and call out that hotkeys is not supported for the month select. We can update the documentation as we add more key sequences in the future.

Added an initial doc site here: https://github.com/cerner/terra-framework/pull/1483/commits/dd3e6dc41a56878639934809718021976e808f6e Can change it if we need to.

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -246,9 +248,33 @@ class DateInput extends React.Component {
   }
 
   /**
- * Takes a key input from the month select, and processes it based on the value of the keycode.
- * @param {Object} event Event object generated from the event delegation.
- */
+   * Takes a key input and processes it based on if it is a hot key otherwise passes it to the apprioate input type key down handler.
```
benbcai, line 251:
```suggestion
   * Takes a key input and processes it based on if it is a hot key otherwise passes it to the appropriate input type key down handler.
```

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -246,9 +248,33 @@ class DateInput extends React.Component {
...
+   * @param {Object} event Event object generated from the event delegation.
+   * @param {DateInputUtil.inputType} inputType Type definition of the input that recieved the keydown event.
```
benbcai, line 253:
```suggestion
   * @param {DateInputUtil.inputType} inputType Type definition of the input that received the keydown event.
```

packages/terra-date-input/src/DateInput.jsx
benbcai: We discussed this scenario with the team and decided that the `-`/`+` keys will always change to yesterday/tomorrow respectively regardless if the value is incomplete or invalid.

benbcai (approved):

elliott-hoffman-cerner (approved):

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -442,6 +468,36 @@ class DateInput extends React.Component {
...
 
+  /**
+   * Sets state to new value dependent on existing date
+   * @param {Object} event Event object generated from the event delegation.
+   * @param {Number} addDays Adds days to current date or today's date.
+   */
+  setHotKeyDate(event, addDays) {
+    const { year, month, day } = this.state;
+    // The Date constructor uses 0-based index for month, using our set month value will give the next month but by also using 0 for day
+    // we move the day back by one giving the set month's final day.
```
elliott-hoffman-cerner, line 479: This is too clever to be maintained I think. It's very tricky and seems likely to be broken in the future when another person has to fix, reason-about, etc. Even if it works... I'm not comfortable with this.

elliott-hoffman-cerner (changes requested):
I am terrified of the logic on lines

packages/terra-date-input/src/DateInput.jsx
elliott-hoffman-cerner: Use Momentjs, and you'll have a one-liner that is dead easy to understand and far less risky to maintain. https://momentjs.com/docs/#/parsing/is-valid/

```
moment().isValid();
```

elliott-hoffman-cerner: I mean specifically the momentjs w/o the timezones: ~ 22kb

elliott-hoffman-cerner: @BenBoersma showed me offline that it's not as simple as using momentjs.isValid() because that doesn't really check if the user entered month 35.

Here is another tactic: Since a valid date should parse to itself, but an invalid date will parse to something else, try comparing the date generated by the inputs back against the original input values like so: https://codepen.io/elliott-hoffman-cerner/pen/jOGBYVb

BenBoersma: Okay reworked the logic here: https://github.com/cerner/terra-framework/pull/1483/commits/fe464dfb64ddf3f72184bfea172d374d81b17473

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -442,6 +468,38 @@ class DateInput extends React.Component {
...
+   */
+  setHotKeyDate(event, addDays) {
+    let { year, month, day } = this.state;
+    year = Number(year);
+    month = Number(month) - 1;
+    day = Number(day);
+    const dateFromInput = new Date(year, month, day);
+    const validYear = (year < DateInputUtil.MaxYearValue && year > DateInputUtil.MinYearValue && dateFromInput.getFullYear() === year);
+    const validMonth = (dateFromInput.getMonth() === month);
+    const validDay = (day > 0 && day <= 31 && dateFromInput.getDate() === day);
```
elliott-hoffman-cerner, line 484: You don't need to check if the day is within 1 and 31 anymore, because `dateFromInput.getDate() === day` will be false if day falls outside of the valid days for that month. Same if day is "cat", -1, etc. 

Try boundary testing on my codepen and you'll see these cases are correctly invalid dates w/o those extra checks:
* day= '0'
* day= '-1'
* day= 'cat'
* day=(one day over the last day of the month)

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -442,6 +468,38 @@ class DateInput extends React.Component {
...
+    let dateObj;
+    if (!month || addDays === 0 || !validYear || !validDay || !validMonth) {
```
elliott-hoffman-cerner, line 486: Likewise here you don't need to be checking that month is truthy, because validMonth will be false if month is false-y. Try some boundary tests w/o this check and it should work fine.

elliott-hoffman-cerner (changes requested):
There are some superfluous checks in the latest date validation code - but once those are removed I think it's good to go.

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -442,6 +468,38 @@ class DateInput extends React.Component {
     }
   }
 
+  /**
+   * Sets state to new value dependent on existing date
+   * @param {Object} event Event object generated from the event delegation.
+   * @param {Number} addDays Adds days to current date or today's date.
```
elliott-hoffman-cerner, line 474: I see that setHotKeyDay will set the date to today if addDays is zero. You should mention that in this param's doc since it's special, and surprising (though valid), behavior.

elliott-hoffman-cerner (changes requested):

elliott-hoffman-cerner:
General comment: it would be even better if you could keep state from becoming invalid. As written, you're waiting until the key press handler to check state validity - but what if the state is invalid yet the user never presses these keys? Would the state keep being invalid and isn't that dangerous? Assuming the answer is yes - the refactor would be to place the validation before the state is updated.

(state) Garbage in -> Garbage out, you know?

packages/terra-date-input/src/DateInput.jsx
BenBoersma: Removed here: https://github.com/cerner/terra-framework/pull/1483/commits/2a6582f14d8b118be226db1e1816aa5418005e26#diff-ea35ec190c52db171ef5f774b071e80db0b038abff0e19580b310959fc582329R484

packages/terra-date-input/src/DateInput.jsx
BenBoersma: Removed month check here: https://github.com/cerner/terra-framework/pull/1483/commits/2a6582f14d8b118be226db1e1816aa5418005e26#diff-ea35ec190c52db171ef5f774b071e80db0b038abff0e19580b310959fc582329R486

packages/terra-date-input/src/DateInput.jsx
BenBoersma: Added sentence caveat for case 0: https://github.com/cerner/terra-framework/pull/1483/commits/2a6582f14d8b118be226db1e1816aa5418005e26#diff-ea35ec190c52db171ef5f774b071e80db0b038abff0e19580b310959fc582329R474

BenBoersma:
> General comment: it would be even better if you could keep state from becoming invalid. As written, you're waiting until the key press handler to check state validity - but what if the state is invalid yet the user never presses these keys? Would the state keep being invalid and isn't that dangerous? Assuming the answer is yes - the refactor would be to place the validation before the state is updated.
> 
> (state) Garbage in -> Garbage out, you know?
> 
> Now, if the answer is "no it's fine to let the state be garbage" then I argue we don't need this state at all: instead just pull the values from the input within your event hander.

I think for the most part we already have code that does this. I know you can't type in 1800 as a date for instance cuz when you try to put in the last 0 it won't let you, same with trying to type 33 for a day. The validation is for "unfinished" dates that have been put in.

elliott-hoffman-cerner (approved):

packages/terra-date-input/src/terra-dev-site/doc/DateInput/DateInputHotkeys.3.doc.mdx
```diff
@@ -0,0 +1,11 @@
+import { Badge } from 'terra-date-input/package.json?dev-site-package';
+
+<Badge />
+
+# Terra Date Input Hotkeys
```
neilpfeiffer, line 5: Updated wording for the doc page to have a bit more descriptions and change exposed terminology from "hotkeys" to "keyboard shortcuts". Also would recommend changing the filename to `KeyboardShortcuts.3.doc.mdx`.

```suggestion
import { Badge } from 'terra-date-input/package.json?dev-site-package';

<Badge />

# Keyboard Shortcuts for Terra Date Input

The Terra Date Input offers additional functionality with specific keyboard shortcuts to save time when 
interacting with date inputs. 

Pressing any of the supported keys while the input is in focus will populate the inputs with the listed 
shortcut response.

| Key / Sequence | Description |
|-|-|
| T | Sets the date value to the current date (Today). |
| - | Decrements the date value by 1 day. If blank or not a valid date, input will set value to Yesterday. |
| + | Increments the date value by 1 day. If blank or not a valid date, input will set value to Tomorrow. |

_Note: For any key listed, pressing the designated keyboard character key is not specific to uppercase or 
lowercase and works for both cases (e.g. if capslock or shift key is also being applied). Uppercase-only usage
is not required._

## Release notes for specific versions

### 1.32.0

* Initial release. This adds the key listeners to the day number input and the year number input, but does not 
add the listeners or shortcut responses to the month dropdown select input, since pressing keys with the dropdown 
in focus will contradict with the select's existing behavior of attempting to match the localized month name strings. 
This will continue to be explored with how to best add the keyboard shortcut behavior for the month dropdown potentially 
in a future release.
```

packages/terra-date-input/src/DateInput.jsx
neilpfeiffer: Doing any kind of partial interpretation may become extremely difficult and challenging with any additional key shortcut keys we add (start/end of month, start/end of week, stand/end of year, etc), trying to make assumptions for what a user intended may create some very complex log and never be perfectly correct - therefore it will be the simplest and most consistent to function with _*incomplete*_ and _*invalid*_ as the same response and ignore any entered information and use today as the reference value. 

Only a complete and valid/accurate date should result in using that date as the reference value instead of today/now.

neilpfeiffer (approved):
Tested: PR #1483
Partially Resolves: Issue #996

\
**Reviewed:**
- Adding the initial implementation for keyboard shortcuts to set 'today', adding one day, and subtracting one day to the Date-Input. 

\
**Fix with Followup Issues:**
- Resolving how to incorporate the keyboard shortcut listeners with the Month select dropdown, such that it does not conflict with default behavior of the native select when entering key letters attempting to match the localized Month string name values.
- This does not add the t/n shortcut keys for any other inputs like the `terra-date-picker`, the `terra-time-input`, or the `terra-date-time-picker`, which will be resolved with additional future PRs.
\
\
Verified by @neilpfeiffer, ready to merge after updating documentation references from 'hotkeys' to 'keyboard shortcuts': https://github.com/cerner/terra-framework/pull/1483#discussion_r772438963 and https://github.com/cerner/terra-framework/pull/1483#discussion_r772440161

 **UX Reviewed** |
-----------|

packages/terra-date-input/src/terra-dev-site/doc/DateInput/DateInputHotkeys.3.doc.mdx
BenBoersma: Addressed all 3 in this commit: https://github.com/cerner/terra-framework/pull/1483/commits/394ad584572a44d9b47afadf6e86f11556995fdc

packages/terra-date-input/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added
+  * Added hotkeys 'T', '-', and '+' for auto-setting dates
```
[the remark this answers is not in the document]
BenBoersma: Fixed here https://github.com/cerner/terra-framework/pull/1483/commits/394ad584572a44d9b47afadf6e86f11556995fdc#diff-069b7fc951aacd3270743dba8d041e42d4da0beda10e7a14065e20a994c9d722R6

PR #1488, opened 2021-11-11
[date-picker] Ensure onChangeRaw gets called when date is selected from calendar.

jmsv6d:
### Summary

onChangeRaw was not getting called when a date was selected from the calendar. Updated to be consistent.

### Deployment Link

### Testing

Tested locally

### Additional Details

@cerner/terra


packages/terra-date-picker/src/react-datepicker/index.jsx
```diff
@@ -531,6 +531,10 @@ class DatePicker extends React.Component {
   }
 
   handleSelect = (date, event) => {
+    const value =  DateUtil.formatISODate(date, DateUtil.getFormatByLocale(this.props.intl.locale));
+    if (this.props.onChangeRaw) {
+      this.props.onChangeRaw(event, value)
```
sdadn, line 536: nit
```suggestion
      this.props.onChangeRaw(event, value);
```

packages/terra-date-picker/src/react-datepicker/index.jsx
```diff
@@ -531,6 +531,10 @@ class DatePicker extends React.Component {
   }
 
   handleSelect = (date, event) => {
+    const value =  DateUtil.formatISODate(date, DateUtil.getFormatByLocale(this.props.intl.locale));
+    if (this.props.onChangeRaw) {
+      this.props.onChangeRaw(event, value);
```
benbcai, line 536: If I open the calendar and select a date, this would trigger onChangeRaw as expected. If I go back to the calendar and select the same date that I previously selected, would onChangeRaw get triggered again even when the actual date on the input didn't get changed?

The [onChange](https://github.com/cerner/terra-framework/blob/terra-date-picker%404.69.1/packages/terra-date-picker/src/react-datepicker/index.jsx#L570-L572) here checks if the value really got changed before calling onChange. I wonder if we need to do something similar for onChangeRaw so it doesn't get called when the value didn't get change.

sdadn (approved):

packages/terra-date-picker/src/react-datepicker/index.jsx
jmsv6d: I've added the check, but we actually have the [`allowSameDate`](https://github.com/cerner/terra-framework/blob/main/packages/terra-date-picker/src/DatePicker.jsx#L506) prop set to true, so `onChange` and `onChangeRaw` are both getting called if the same date is selected again regardless.

benbcai (approved):

BenBoersma (approved):

PR #1496, opened 2021-12-23
snaps changed from recent terra-core changes

elliott-hoffman-cerner:
### Summary
While checking on main, I noticed that a couple snaps are different after consuming recent Terra Core changes around buttons. They seem legit to me.

### Deployment Link
TBD

### Additional Details
If I'm correct these update snaps are the result of consuming changes to NOT set tab indexen to -1 as often, and to add the aria-pressed attr to button groups.

### DIFF
--- a/packages/terra-date-picker/CHANGELOG.md
+++ b/packages/terra-date-picker/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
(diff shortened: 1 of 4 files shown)


jmsv6d (approved):

sdadn (approved):

elliott-hoffman-cerner:
This was fixed by #1483

PR #1497, opened 2021-12-30
Time input a11y

elliott-hoffman-cerner:
### Summary
This PR makes the Time Input as accessible as possible while remaining passive to all the little quirks and features (e.g. automatically advancing the focus). There are a couple known issues I can't crack - see below.

The approach here is to wrap the existing inputs of the component in higher order components that provide accessibility. The idea being that these can eventually be folded into the Inputs themselves.

The main new features are:
1. There is a new, (sadly, for passivity) optional `label` prop. When supplied, it is used as the label of the Time Input by the screen reader. When it's omitted, there is a default value "Time". I'm open to a different default - it's easy to change.
2. Each component (e.g. minute, hour, am button) is described by the screen reader upon focus. Here is an example: If I place focus on the hour field I might hear "Hour, required invalid text input, a two-digit twenty-four hour value". That phrase is a combination of description text we must localize ("a two-digit twenty-four hour value") and parts that the screen reader computes based on our use of HTML attrs, including some aria-attrs ("required", "invalid", "text input"). 
3. Items that don't make sense when read by a screen reader are no longer read. This includes the ":" time seperators and the mask text, e.g. "hh:mm:ss". Item 2 above takes the place of these phrases. 
4. The complete and localized time phrase, e.g. "09:22 pm", will be read if present. The phrase will also be read when the time is changed to a different and valid value, e.g. "09:23 pm". The time is not read if the time is complete or equivalent to the previous time value, so "09:22" and "9:22" are equivalent. Note: we are not localizing the time value, only the phrasing of the complete time. There is no conversion of timezones, etc.

### Known issues
1. For reasons I can't figure out, the doc tests about **mobile incomplete and mobile invalid are not reading the fields correctly**. The screen reader does not read them as incomplete/invalid despite the attributes looking correct... The non-mobile versions of these doc-tests are read correctly. I need to figure out what's different.
2. **One DateTime jest test is broken by this change**, and I don't know why yet - I'm not familiar with that component.
3. **The screen reader user will not be read any indication that the toggle button state is invalid, despite the toggle button visually indicating (red box) that the button is invalid**. There is no way for the screen reader to recognize that the consumer has marked a am/pm toggle button as invalid because buttons don't have such a concept. Please tell me if there is some way to express "invalid button state" using HTML attrs. I found none.

### Deployment Link
TBD

### Testing
I went through every test scenario in the doc-site and manually tried macOS VoiceOver on each. I also inspected the [Accessibility Pane](https://developer.chrome.com/docs/devtools/accessibility/reference/#pane) to make sure the attributes were correct. 

I added a new Accessibility doc site test so that our consumers can be shown the features. That is just a bare minimum. We really should write a small guide on accessibility for this component some day.

I did not run this through exhaustive A11Y auditing or testing against Windows screen readers or iOS's voiceOver implementation. 

### Next steps
* I'm going to work w/ the rest of the team post holidays to fix the broken test(s).
* The next big PR will introduce a Time Input Field component, which does the tricky wrapping of the Time Input in such a way as to marry the A11Y of this PR with the needs of a field.


elliott-hoffman-cerner:
I spoke w/ @benbcai  and @jmsv6d just now, and I'm going to push a few changes into this PR to address their feedback:
- I'm going to measure the render time and memory changes due to my change, and report back.
- I'm going to move any of my new `...propSpreads` as high as possible in the render functions.
- I will switch to using the translation files for my intl messages rather than my current in-line messages. The fear is that our ancient react-intl lib + our CI won't be able to extract those inline messages. Pity.

elliott-hoffman-cerner:
I got tired of, and made to mistakes while, moving my inline react-intl messages (see below) to our translation files. [So I wrote a bridge that allows us to use the inline messages](https://github.com/cerner/terra-framework/pull/1497/commits/23ec62615e90aafcb0f03ef17be7455e22c3da46) WITHOUT upgrading react-intl (though we should!) and/or modifying terra-aggregate-translations.

Inline messages offer several advantages over key-lookups:
* Easy for the developer and translator to see how the message will be used - because the message is WHERE it will be used. Also, the inline message format includes metadata like a description for translators.
* Less errors will happen when we don't have to tediously copy/paste the message IDs back and forth.

Example inline message:
```javascript
function hourDescription() {
  if (is12Hour) {
    return intl.formatMessage({
      id: 'Terra.timeInput.hourDescriptionTwelve',
      defaultMessage: 'A two-digit 12-hour value',
      description: `Explains to screen reader users that the hour field needs a two digit hour. This will be read
      only when a screen reader is enabled. It is never displayed. It will be read when the user has focused on
      the hour input or when the screen reader is reading the page to the user.`,
    });
``` 

With my adapter, it will be placed in our aggregated translations files like normal:
```javascript
var messages = {
  "Terra.timeInput.hourDescriptionTwelve": "A two-digit 12-hour value",
```

In the future, once we upgrade to react-intl 6, we should also update terra-aggregated translations to extract these messages and retain all of the metatdata (e.g. the description string). 

Hopefully we can start using inline message anywhere in terra-framework. We can do the same thing in our other mono-repos if we wish.

@jmsv6d, @benbcai, @neilpfeiffer

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -52,6 +53,14 @@ const propTypes = {
...
    */
   name: PropTypes.string.isRequired,
+  /**
+   * An identifier used by assistive technologies like screen readers to briefly describe this time input to users.
+   * The label is not rendered visually.
+   *
+   * **BEST PRACTICE FOR ACCESSIBILITY**: you _SHOULD_ set this to match whatever visible label you give in your UI.
+   * Currently this is optional for passivity reasons, but it will become mandatory in a future major version.
+   */
+  label: PropTypes.string,
```
jmsv6d, line 63: This is described in the comment, but in the interest of hand holding our consumers, maybe the name of this prop should be a little more explicit. `a11yLabel` or `visuallyHiddenLabel`, something like that. I could see us ending up with support questions of people trying to add a visual label using this prop thinking it behaved like a form field.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -724,30 +720,168 @@ class TimeInput extends React.Component {
...
+     * NOTE: the description does not take the place of validation or error messages.
+     */
+    function hourDescription() {
+      if (is12Hour) {
+        return intl.formatMessage({
+          id: 'Terra.timeInput.hourDescriptionTwelve',
+          defaultMessage: 'A two-digit 12-hour value',
+          description: `Explains to screen reader users that the hour field needs a two digit hour. This will be read
+          only when a screen reader is enabled. It is never displayed. It will be read when the user has focused on
+          the hour input or when the screen reader is reading the page to the user.`,
```
sdadn, line 755: Is the use of backticks (`) intentional? this will also preserve the whitespaces so the the second two lines will be tabbed when printed, regardless of indentation of the first line.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -169,15 +183,8 @@ class TimeInput extends React.Component {
         meridiem = this.anteMeridiem;
       }
     }
-    if (this.props.variant === TimeUtil.FORMAT_12_HOUR && TimeUtil.getVariantFromLocale(props) === TimeUtil.FORMAT_24_HOUR) {
-      if (process.env.NODE_ENV !== 'production') {
-        // eslint-disable-next-line no-console
-        console.warn('This locale only uses 24 hour clock. The ante meridiem and post meridiem will not be displayed');
-      }
-      this.anteMeridiem = '';
-      this.postMeridiem = '';
```
benbcai, line 178: This previously would clear anteMeridiem and postMeridiem if the locale only supports the 24-HOUR variant. It seems `this.postMeridiem` and `this.postMeridiem` will now always return `AM` and `PM` respectively. Would that have any impact?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -724,30 +720,168 @@ class TimeInput extends React.Component {
...
 
-    const formatDescriptionId = `terra-time-input-description-format-${this.uuid}`;
-
-    const format = showSeconds
-      ? `(${intl.formatMessage({ id: 'Terra.timeInput.hh' })}:${intl.formatMessage({ id: 'Terra.timeInput.mm' })}:${intl.formatMessage({ id: 'Terra.timeInput.ss' })})`
-      : `(${intl.formatMessage({ id: 'Terra.timeInput.hh' })}:${intl.formatMessage({ id: 'Terra.timeInput.mm' })})`;
+    function mask() {
+      if (showSeconds) {
+        return intl.formatMessage({
+          id: 'Terra.timeInput.maskHourMinute',
```
benbcai, line 730: Should the id be `Terra.timeInput.maskHourMinuteSecond`?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -724,30 +720,168 @@ class TimeInput extends React.Component {
...
+          defaultMessage: '(hh:mm:ss)',
+          description: 'A visual hint that the Time Input expects two-digit hour, minute and second.',
+        });
+      }
+      return intl.formatMessage({
+        id: 'Terra.timeInput.maskHourMinuteSecond',
```
benbcai, line 736: Should the id be `Terra.timeInput.maskHourMinute`?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -724,30 +720,168 @@ class TimeInput extends React.Component {
...
+        defaultMessage: '(hh:mm)',
+        description: 'Like Terra.timeInput.maskHourMinute but no second.',
```
benbcai, line 738: Should we update the description to be similar to the one above like this instead of referencing Terra.timeInput.maskHourMinute?

`A visual hint that the Time Input expects two-digit hour and minute.`

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -724,30 +720,195 @@ class TimeInput extends React.Component {
...
+    // Fan out some component-level props into input-level a11y attributes. See the big comment below for more info.
+    inputAttributes.isInvalid = isInvalid;
+    inputAttributes.disabled = disabled;
+    inputAttributes.required = required;
+
+    const a11yString = TimeUtil.getA11YTimeValue(this.props, this.state, this.postMeridiem);
+
+    function groupLabel() {
+      if (label) {
+        return intl.formatMessage({ // fix
```
jmsv6d, line 774: Fix comment still here, is there a remaining issue?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -724,30 +720,168 @@ class TimeInput extends React.Component {
...
+    }
+
+    // Fan out some component-level props into input-level a11y attributes. See the big comment below for more info.
+    inputAttributes.isInvalid = isInvalid;
+    inputAttributes.disabled = disabled;
+    inputAttributes.required = required;
+
+    const a11yString = TimeUtil.getA11YTimeValue(this.props, this.state, this.postMeridiem);
+
+    const groupLabel = intl.formatMessage({ // fix
```
benbcai: Is there still something that needs fixed here?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -724,30 +720,195 @@ class TimeInput extends React.Component {
...
+          description: `The label that will only be read to screen readers. It is prefixed with the time input's name,
+          e.g. 'Time of Birth', so that screen reader users can pick this specific hour field out of a list of many
+          hour fields on the same page. The minute and second screen reader labels won't contain the name because they
+          will always follow their labeld hour field. We didn't want to say the label too many times.`
+        }, {label});
+      }
+      return intl.formatMessage({
+        id: 'Terra.timeInput.hourLabelDefault',
+        defaultMessage: 'Hour',
+        description: `Like Terra.timeInput.hourLabel but used in situations where the consumer of the time input did not
```
jmsv6d, line 804: I feel like referencing the id of another message in the description could get confusing if the messages are ever not right next to each other.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -724,30 +720,195 @@ class TimeInput extends React.Component {
...
+      return intl.formatMessage({
+        id: 'Terra.timeInput.inputGroupValueDefault',
+        defaultMessage: 'Time',
+        description: `Same meaning as groupLabel, only for situations where the consumer has not provided a label for us
+      to plug in.`,
+      });
+    }
+
+    function hoursLabel() {
+      if(label){
```
jmsv6d, line 791: I would have thought this would have caused a lint error.
```suggestion
      if (label) {
```

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -724,30 +720,195 @@ class TimeInput extends React.Component {
...
-          <Input
+          <AccessibleInput
             {...inputAttributes}
             {...hourAttributes}
-            aria-label={intl.formatMessage({ id: 'Terra.timeInput.hours' })}
+            /**
+             * If the entire Time Input isInvalid, then pass isInvalid into each wrapped input, so that the screen
+             * reader users will get an indication that something is invalid. It's not perfect because both inputs will
+             * be marked invalid even though it's the combination of both that is really the problem. For example,
+             * 09:88 is a valid hour and an invalid minute, but both hour and minute will */
```
jmsv6d, line 910:
```suggestion
             * 09:88 is a valid hour and an invalid minute, but both hour and minute will be marked invalid.*/
```

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -724,30 +720,168 @@ class TimeInput extends React.Component {
...
+
+    const a11yString = TimeUtil.getA11YTimeValue(this.props, this.state, this.postMeridiem);
+
+    const groupLabel = intl.formatMessage({ // fix
+      id: 'Terra.timeInput.inputGroupValue',
+      defaultMessage: '{label}',
+      description: `This will be read by screen readers as the reader moves into the group of inputs. It is intended to
+      help the user understand "you are about to enter a section of the page where many different inputs all work
+      together for this one concept of time.`,
+    }, { label });
```
benbcai: The PR description mentions that the `label` prop gets defaulted to "Time" if it is not set. I see it gets defaulted to an empty string in `defaultProps` but don't see where it gets set to "Time".

packages/terra-time-input/package.json
```diff
@@ -30,9 +30,11 @@
     "classnames": "^2.2.5",
     "keycode-js": "^3.1.0",
     "prop-types": "^15.5.8",
-    "terra-button-group": "^3.0.0",
+    "terra-button-group": "^3.60.1",
+    "terra-form-field": "^4.20.0",
```
jmsv6d, line 34: Where is this dependency being used?

packages/terra-time-input/src/TimeUtil.js
```diff
@@ -269,6 +288,141 @@ class TimeUtil {
 
     return variant;
   }
+
+  /**
+   * Is the Time Input operating as a 12-hour variant?
```
jmsv6d, line 293:
```suggestion
   * Indicates if the Time Input is operating as a 12-hour variant.
```

packages/terra-time-input/src/TimeUtil.js
```diff
@@ -269,6 +288,141 @@ class TimeUtil {
...
+   * @param {Object} The TimeInput props.
+   * @returns {Boolean} True iff the locale allows a 12-hour time and the variant is 12-hour time.
```
jmsv6d, line 295:
```suggestion
   * @returns {Boolean} True if the locale allows a 12-hour time and the variant is 12-hour time.
```

packages/terra-time-input/src/TimeUtil.js
```diff
@@ -269,6 +288,141 @@ class TimeUtil {
...
+   */
+  static is12Hour(props) {
+    return TimeUtil.getVariantFromLocale(props) === TimeUtil.FORMAT_12_HOUR;
+  }
+
+  /**
+   * Is the Time Input's state representing a 12-hour time in the PM?
```
jmsv6d, line 302:
```suggestion
   * Indicates if the Time Input's state is representing a 12-hour time in the PM.
```

packages/terra-time-input/src/TimeInput.jsx
benbcai: Should we also call out that the default label is "Time" if not set?

packages/terra-time-input/src/TimeUtil.js
```diff
@@ -269,6 +288,141 @@ class TimeUtil {
...
+   * @param {Object} props The TimeInput props.
+   * @param {Object} state The TimeInput state.
+   * @param {String} postMeridiem The PM string to compare against.
+   * @returns {Boolean} True iff the time input is in 12-hour mode and the state is a PM time.
```
jmsv6d, line 306:
```suggestion
   * @returns {Boolean} True if the time input is in 12-hour mode and the state is a PM time.
```

packages/terra-time-input/src/TimeUtil.js
```diff
@@ -269,6 +288,141 @@ class TimeUtil {
...
+  static isPM(props, state, postMeridiem) {
+    return TimeUtil.is12Hour(props) && state.meridiem === postMeridiem;
+  }
+
+  /**
+   * Build a UTC time string in kind-of-but-not-really ISO 8601 format.
+   *
+   * WARNING: Read the name. This has very peculiar behavior you probably don' want ot use for new features. I'm
+   * keeping this around as it was, because it's got strange behavior I'm not sure how to break. It has a bug where a
+   * person could type leave some fields blank and see a value containing 'NaN', like 'TNaN:22'. Not sure if that is
```
jmsv6d, line 317:
```suggestion
   * person could leave some fields blank and see a value containing 'NaN', like 'TNaN:22'. Not sure if that is
```

packages/terra-time-input/src/TimeInput.jsx
benbcai: Should this just use the provided "label" as is and not need to do any translation?

packages/terra-time-input/src/_AccessibleInput.jsx
```diff
@@ -0,0 +1,93 @@
+import VisuallyHiddenText from 'terra-visually-hidden-text';
+import PropTypes from 'prop-types';
+import React from 'react';
+import Input from 'terra-form-input';
+import uuidv4 from 'uuid/v4';
+
+const propTypes = {
+  value: PropTypes.string,
```
jmsv6d, line 8: This prop is missing a description

packages/terra-time-input/src/_AccessibleInput.jsx
```diff
@@ -0,0 +1,93 @@
...
+  disabled: PropTypes.bool,
+  /** Set to true to mark as invalid for screen readers. */
+  isInvalid: PropTypes.bool,
+};
+
+const defaultProps = {
+  value: undefined,
+  label: undefined,
+  isInvalid: false,
+
```
jmsv6d, line 28:
```suggestion
```

packages/terra-time-input/src/_AccessibleValue.jsx
```diff
@@ -0,0 +1,62 @@
+import React, { useState } from 'react';
+import VisuallyHiddenText from 'terra-visually-hidden-text';
+import PropTypes from 'prop-types';
+
+const propTypes = {
+  // The value you want read only when changed. An undefined value is never read.
```
jmsv6d, line 6:
```suggestion
  /**
   * The value you want read only when changed. An undefined value is never read.
   */
```

packages/terra-time-input/src/_AccessibleValue.jsx
```diff
@@ -0,0 +1,62 @@
...
+  value: PropTypes.string,
+  readThis: PropTypes.string,
```
jmsv6d, line 8: Prop missing description

packages/terra-time-input/src/_AccessibleValue.jsx
```diff
@@ -0,0 +1,62 @@
...
+};
+
+const defaultProps = {
+  value: undefined,
```
jmsv6d, line 12: Default props of `undefined` are unnecessary. If a prop is not defined it will have an undefined value.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -724,30 +720,168 @@ class TimeInput extends React.Component {
...
+              readThis={intl.formatMessage({ // fix
+                id: 'Terra.timeInput.textValue',
+                // eslint-disable-next-line formatjs/no-camel-case
+                defaultMessage: 'Time {a11yString}',
+                description: `Similar to Terra.timeInput.labeledTextValue, but we want the screen reader to say "time"
+                before reading the value because no label was provided. It would be confusing to hear "09 22" right
+                after you typed "22" or "09". So instead we can say "Time 09 22".`,
+              }, { a11yString })}
+            />
+          )}
```
benbcai, line 895: It seems like this is creating the same AccessibleValue but with a different "readThis" depending on whether or not a label is provided. Is it possible to consolidate these into one AccessibleValue by creating the valid label before?

`const a11yLabel = label || intl.formatMessage({ id: 'Terra.timeInput.textValue' })`

After we have a11yLabel that we know is not empty, then create AccessibleValue without the duplication?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -724,30 +720,195 @@ class TimeInput extends React.Component {
...
+
+    function groupLabel() {
+      if (label) {
+        return intl.formatMessage({ // fix
+          id: 'Terra.timeInput.inputGroupValue',
+          defaultMessage: '{label}',
+          description: `This will be read by screen readers as the reader moves into the group of inputs. It is intended to
+      help the user understand "you are about to enter a section of the page where many different inputs all work
+      together for this one concept of time.`,
+        }, { label });
```
benbcai, line 780: I would think we should just use the provided label as is without having to do any kind of formatting/translation.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -103,6 +112,7 @@ const defaultProps = {
   isInvalidMeridiem: false,
   minuteAttributes: {},
   hourAttributes: {},
+  label: '',
```
benbcai, line 115: Should we just let this be undefined instead of an empty string?

elliott-hoffman-cerner:
@benbcai I measured size and speed and found no difference between main branch and my feature branch. 
Transfer was ~14.4MB, DOMContentLoaded ~ 833ms and Load was ~ 1.1s.

packages/terra-time-input/src/TimeInput.jsx
elliott-hoffman-cerner: @jmsv6d OK, I will change it to a11yLabel. 
@benbcai  yes, but it must be localized so I am defaulting it to Time further down in the code using formatMessage(). I'll see if I can set the prop default to be the output of a formatMessage(). If not, I will keep it as is.

packages/terra-time-input/src/TimeInput.jsx
elliott-hoffman-cerner: Yes, it's intentional so I can have this multi-line phrase easily wrapped to 120 chars - my personal max width. Do we have a stance on max width like 80 chars or 120 chars?

I'll see about removing the extra indentation.

elliott-hoffman-cerner: These descriptions will only appear in our code and (one day, when we upgrade react-intl and update our CI/CD) maybe in our translation team's tools. End users will never see these descriptions.

packages/terra-time-input/src/TimeInput.jsx
elliott-hoffman-cerner: I saw no impact to that change. When I made this change I looked around for any logic that would flex based on AM/PM being set and found none.

packages/terra-time-input/src/TimeInput.jsx
elliott-hoffman-cerner: Good catch, I'll change it. Because I'm using inline messages rather than the translation files, this has no real impact but it is a typo.

packages/terra-time-input/src/TimeInput.jsx
elliott-hoffman-cerner: Yep will fix.

packages/terra-time-input/src/TimeInput.jsx
elliott-hoffman-cerner: Will do. I must have been tired when I coded this 😄 .

packages/terra-time-input/src/TimeInput.jsx
elliott-hoffman-cerner: No, I will remove it.

packages/terra-time-input/src/TimeInput.jsx
elliott-hoffman-cerner: CTRL+F is always there? What would you suggest as an alternative.

packages/terra-time-input/src/TimeInput.jsx
elliott-hoffman-cerner: You're right. I'll double check my linter settings and try again.

packages/terra-time-input/package.json
elliott-hoffman-cerner: Ah this was left in by mistake. It will eventually be needed when I wrap Time Input in Field to create the TimeInputField component but that is for later. I will remove it.

packages/terra-time-input/src/TimeUtil.js
elliott-hoffman-cerner: Not a typo, _iff_ meaning _If, and only if_.

packages/terra-time-input/src/TimeUtil.js
elliott-hoffman-cerner: Like above, _iff_ meaning _if and only if_.

packages/terra-time-input/src/_AccessibleInput.jsx
elliott-hoffman-cerner: I will add one.

packages/terra-time-input/src/_AccessibleValue.jsx
elliott-hoffman-cerner: Will add.

packages/terra-time-input/src/_AccessibleValue.jsx
elliott-hoffman-cerner: I see. I'll change it.

packages/terra-time-input/src/TimeInput.jsx
elliott-hoffman-cerner: Could be, but I'm not sure if in another language there would be other words given?

packages/terra-time-input/src/TimeInput.jsx
elliott-hoffman-cerner: I'll try it.

packages/terra-time-input/src/TimeInput.jsx
elliott-hoffman-cerner: I'll try it.

packages/terra-time-input/src/TimeInput.jsx
elliott-hoffman-cerner: Thinking about this more, I've convinced myself it's OK. Will do.

packages/terra-time-input/src/TimeInput.jsx
sdadn: In that case, you can try this. Though if it's only on the dev side then ig this is more of an optional nit.
```suggestion
          description: 'Explains to screen reader users that the hour field needs a two digit hour. This will be read' +
                       'only when a screen reader is enabled. It is never displayed. It will be read when the user has focused on' +
                       'the hour input or when the screen reader is reading the page to the user.',
```

sdadn: Here's a more accurate visualization of the result:
CleanShot 2022-01-07 at 10 54 44@2x

elliott-hoffman-cerner:
@benbcai  and @jmsv6d  I addressed most of your feedback on 1352af2d28c1ce02722eaa96e0c01f0dab8e5f0d. Take a look plz.

packages/terra-time-input/src/TimeInput.jsx
elliott-hoffman-cerner: It is not possible to default the a11yLabel (aka label) to a localized string because the locale of the runtime context is not known when that default is assigned to the class. I think what I can do is add something to the constructor taht computes a default and use that to simplify the code base. I'll try it next.

elliott-hoffman-cerner: It worked! See 89815ff1f9e218cf28a06c70822292773a2cf1ec.

packages/terra-time-input/src/TimeInput.jsx
elliott-hoffman-cerner: I'm going to keep with my back ticks for now, because it's the easiest way to reflow the text if it is edited - no need to chop the text into new strings.

elliott-hoffman-cerner:
@benbcai @sdadn @jmsv6d I took the liberty of resolving any comments that either I fixed or I'm choosing not to make changes for. Let me know if there is anything you feel strongly different about and we can discuss.

packages/terra-time-input/src/TimeInput.jsx
benbcai: The expectation should be that the higher order component would already have translated the label that is passed to time-input. It's similar to other components (e.g. [terra-form-field](https://github.com/cerner/terra-core/blob/terra-form-fieldset%402.63.2/packages/terra-form-field/src/Field.jsx#L182)) that would just use the passed in label as is.

.gitignore
```diff
@@ -36,6 +36,7 @@
 /spec/reports/
 /test/tmp/
 /test/version_tmp/
+/translations/*.*
```
jmsv6d, line 39: Why are the translations files being added to the gitignore?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -571,6 +574,26 @@ class TimeInput extends React.Component {
     this.handleValueChange(event, TimeUtil.inputType.HOUR, this.state.hour.toString(), selectedKey);
   }
 
+  get a11yLabel() {
+    const { a11yLabel, intl } = this.props;
+    return a11yLabel || intl.formatMessage({
```
jmsv6d, line 579: Has a translations jira been logged for the new translation strings?

.gitignore
elliott-hoffman-cerner: I don't remember why, so I'll remove this.

packages/terra-time-input/src/TimeInput.jsx
elliott-hoffman-cerner: Just now, yes.

.gitignore
elliott-hoffman-cerner: I remember now! OK, so the translation directory here is only created/used by my new technique of gathering up inline messages from the source JSX files. This dir is populated on demand by the message generator  - so like if I run the local server. The messages in this dir's translation files will be placed in the normal aggregated translation dir - so in my opinion we should not version the translation files in this directory. The reason I want to version the directory is so that I don't have to (re)create it as part of populating it.

elliott-hoffman-cerner: I remember now: this new root-level translation dir is just a container to hold the intermediate form of messages generated from inline messages in our JSX (like the ones I introduced in this PR). All of the message placed in this dir will be folded into the normal aggregated_translations directory, so there is no need to keep the intermediate step (the msg files in this dir). 

For the inline messages, they are defined in the JSX and only exist in the root trans dir as a way to be compatible w/ our ancient react-intl translation aggregation process.

jmsv6d: My concern is around the translation strings getting released with the component. They need to exist in the node module or they won't be found by aggregate translations in the projects that consume it. If the extracted messages are going to the root level instead of the package level then that needs to be changed.

packages/terra-date-time-picker/tests/jest/DateTimePicker.test.jsx
```diff
@@ -88,9 +88,9 @@ it('Should not throw any errors while date value is outside of the Min, Max rang
   const dateYearInput = datePicker.find({ name: 'terra-date-year-input', type: 'text' }).at(0);
   dateYearInput.simulate('change', { target: { value: '2017' } });
   expect(() => {
-    const hourInput = datePicker.find({ name: 'terra-time-hour-input' }).at(0);
+    const hourInput = datePicker.find({ name: 'terra-time-hour-input' }).at(2);
```
elliott-hoffman-cerner, line 91: @benbcai I couldn't use DIVs because then each input would be on a new line, or I would need new styles. Instead, I discovered that the 3rd node in this find() object is the input.  The first and 2nd nodes are a _function_ and a _host_. I think the function is the intl wrapper and the host is the frag? IDK. It works though.

benbcai: That makes sense and this change looks good to me.

.gitignore
elliott-hoffman-cerner: Couldn't figure it out and so I moved the translations to the traditional place - in the message files. PR is updated.

elliott-hoffman-cerner:
I updated the changelog just now, in theory that will fix the travic CI fail.

jmsv6d (commented):
I think this looks good if we can get a passing build and once we've got the new translated strings included.

elliott-hoffman-cerner:
I'm in merge/rebase hell because this feature branch is so far out of date - so I'm going to close this PR and open a new one to make it easier. The new PR will not have any functional change, merely testing changes.

PR #1498, opened 2022-01-03
#PR to fix main branch build

ShettyAkarsh:
Updated jest snapshots to fix main branch build


elliott-hoffman-cerner (commented):
I think this was resolved already by #1483. Please try building against based off the latest main and you should see it working fine.

PR #1503, opened 2022-01-20
[terra-time-input] Adding initial keyboard shortcuts

BenBoersma:
### Summary

Adding initial keyboard shortcuts of "N" (set time to now), "-" (decrement minute by 1), "+" (increment minute by 1), "A" (set to Ante meridiem), and "P" (set to post meridiem).

Closes #

### Deployment Link

https://terra-framework-deployed-pr-#.herokuapp.com/

### Additional Details

Thank you for contributing to Terra.
@cerner/terra

PR #1507, opened 2022-01-28
[terra-date-picker] Add keyboard shortcuts

BenBoersma:
### Summary
Adding initial keyboard shortcuts of "T" (set date to today), "-" (decrement day by 1, set to today then decrement if no valid date),  and "+" (increment day by 1, set to today then increment if no valid date).

Closes #

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -521,6 +521,67 @@ const DatePickerInput = (props) => {
     }
   };
 
+
+
+  const handleInputKeydown = (event, inputType) => {
+    const { day, month, year } = date;
```
elliott-hoffman-cerner, line 527: Where is date defined?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -521,6 +521,67 @@ const DatePickerInput = (props) => {
...
+    let inputDate;
+    let formattedDate;
+    if (day.length === 2 && month.length === 2 && year.length === 4) {
+      inputDate = DateUtil.convertToISO8601(`${year}-${month}-${day}`, DateUtil.ISO_EXTENDED_DATE_FORMAT);
+      formattedDate = DateUtil.strictFormatISODate(inputDate, momentDateFormat);
```
elliott-hoffman-cerner, line 532: What is the difference between this step and the preceding line?

packages/terra-date-picker/tests/wdio/date-picker-spec.js
```diff
@@ -672,6 +672,335 @@ Terra.describeViewports('Date Picker', ['medium'], () => {
...
+          browser.url('/#/raw/tests/terra-date-picker/date-picker/date-picker-default');
+        });
+
+        it('sets Date to yesterday on month input', () => {
+          Terra.hideInputCaret('input[name="terra-date-month-date-input"]');
+          $('input[name="terra-date-month-date-input"]').click();
+          browser.keys('-');
+
+          const currentDate = new Date();
+          currentDate.setDate(currentDate.getDate() - 1);
```
elliott-hoffman-cerner, line 795: I find this variable naming confusing.

packages/terra-date-picker/tests/wdio/date-picker-spec.js
```diff
@@ -672,6 +672,335 @@ Terra.describeViewports('Date Picker', ['medium'], () => {
...
+        beforeEach(() => {
+          browser.url('/#/raw/tests/terra-date-picker/date-picker/date-picker-start-date');
+        });
+
+        it('sets Date to valid date minus 1 day if pressed in month', () => {
+          Terra.hideInputCaret('input[name="terra-date-month-date-input"]');
+          $('input[name="terra-date-month-date-input"]').click();
+          browser.keys('-');
+
+          expect($('input')).toHaveValue('2017-03-31');
```
elliott-hoffman-cerner, line 872: How are you setting the initial condition of 2017-03-30?

packages/terra-date-picker/src/DateInput.jsx
BenBoersma: https://github.com/cerner/terra-framework/blob/5a4b1cf7be14410bd8173b971135be0edbba2ff9/packages/terra-date-picker/src/DateInput.jsx#L171

packages/terra-date-picker/tests/wdio/date-picker-spec.js
BenBoersma: the URL has a date input with a starting value of `2017-04-01`
https://github.com/cerner/terra-framework/blob/5a4b1cf7be14410bd8173b971135be0edbba2ff9/packages/terra-date-picker/src/terra-dev-site/test/date-picker/DatePickerStartDate.test.jsx

packages/terra-date-picker/tests/wdio/date-picker-spec.js
BenBoersma: yea i can easily go through and on the `-` and `+` shortcut tests rename to yesterday and tomorrow for the variable names

packages/terra-date-picker/src/DateInput.jsx
BenBoersma: format, 
`inputDate` = YYYY-MM-DD
`formatted` = locale format that is being displayed (so in my case, for example MM/DD/YYYY)

elliott-hoffman-cerner (approved):

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -521,6 +521,67 @@ const DatePickerInput = (props) => {
...
+      dateDispatch({ day: nextDayValues.day, month: nextDayValues.month, year: nextDayValues.year });
+      return;
+    }
+    if (event.key === '=' || event.key === '+') {
+      if (validDate) {
+        inputDate = DateUtil.incrementDateByDay(inputDate, DateUtil.ISO_EXTENDED_DATE_FORMAT);
+      } else {
+        inputDate = DateUtil.incrementDateByDay(DateUtil.getCurrentDate(), DateUtil.ISO_EXTENDED_DATE_FORMAT);
+      }
+      formattedDate = DateUtil.strictFormatISODate(inputDate, momentDateFormat);
```
sdadn, line 567: Can you add a comment before each `if` statement about what each block does? Something like 
```js
// jump to current date
if (event.key === 't' || event.key === 'T') {
```

packages/terra-date-picker/tests/wdio/date-picker-spec.js
BenBoersma: Changed here https://github.com/cerner/terra-framework/pull/1507/commits/88f3dab1071c7e5c7e94082247ddafb878b2631d

elliott-hoffman-cerner (approved):

jmsv6d (approved):

packages/terra-date-picker/src/DateInput.jsx
BenBoersma: Added here https://github.com/cerner/terra-framework/pull/1507/commits/69edd85edb2fc48fdfa70c5799fe8c7284906461

chrismichalewicz:
This has been functionally reviewed.

sdadn (approved):

neilpfeiffer (approved):
Tested: PR #1507
Partially Resolves: Issue #996
Followup to: PR #1483 and PR #1503

\
**Reviewed:**
- Adding the initial implementation for keyboard shortcuts to set 'today', adding one day, and subtracting one day to the Date-Picker component.
- Verified usability was functioning as expected on desktop device with keyboard.
- Does not function on mobile since is locked to the numeric keyboard only (not able to press alphabetic keys).

\
**Fix with Followup Issues:**
- Usability Research: consider moving cursor to the day input when '+' and '-' are pressed, or re-examine input cursor/selection behavior (perhaps highlighting full value). Is unexpected that +/- changes minutes while cursor remains in the month or year, but up/down changes the month or year.
- Does not apply to when the Calendar dropdown/popup is open, which will be handled in an additional enhancement PR.
- Accessibility interaction hints and feedback will be added with an additional enhancement PR.
- This does not add the t/n shortcut keys for any other inputs, like the `terra-date-time-picker` yet, which will be resolved with additional future PRs.
\
\
Verified by @neilpfeiffer, ready to merge.

 **UX Reviewed** |
-----------|

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -521,6 +521,69 @@ const DatePickerInput = (props) => {
...
+    const { day, month, year } = date;
+    let inputDate;
+    let formattedDate;
+    if (day.length === 2 && month.length === 2 && year.length === 4) {
+      inputDate = DateUtil.convertToISO8601(`${year}-${month}-${day}`, DateUtil.ISO_EXTENDED_DATE_FORMAT);
+      formattedDate = DateUtil.strictFormatISODate(inputDate, momentDateFormat);
+    }
+    const validDate = DateUtil.isValidDate(formattedDate, momentDateFormat);
+
+    // set date to today
```
supreethmr, line 536: should we restrict this shortcuts to work only for input type day 
`inputType === DateUtil.inputType.DAY` . Since we can only increment or decrement or set only a day field with this shortcuts. should we prevent increment and decrement of day when the focus is on month / year input when`T` / `t` or `-` or `+` keys are pressed.

other inputs on date-picker does already support arrow keys for increment and decrement of year, month and day when the focus is on respective input.

neilpfeiffer: Thanks @supreethmr, the current code is functioning as intended, mimicking how the legacy date and time inputs work in Cerner Win32/Win64 products, where the keyboard shortcut commands for `T`  `-` and `+`  work anywhere within the date field input (it visually looks like a single input as well), and will be what users expect.

Although I noticed it felt slightly odd during testing as well, that it appears much more visually noticeable in Terra than it does in the Win32/Win64, so I will be investigating further to see what micro-interaction differences there are, and what we can do to improve the visual continuity to new key listeners working equally in any node/input (month/day/year/etc) without calling specific noticeability to just the day (and minute) inputs.

Also for awareness, there will be more future stories where we add even more existing Cerner key shortcuts that will have additional behaviors that apply to the full date value ('W' for start of the week, 'K' for the end of the week, 'M' for the start of the month, 'H' for the end of the month, 'Y' for start of the year, 'R' for end of the year, and a few others), so they will have more effect than simply the Day and will also work in any of the inputs.

PR #1511, opened 2022-02-04
Test padding above and below TimeInput

elliott-hoffman-cerner:
### Summary
This change makes it easy to see if a change to TimeInput is causing padding changes. I was in a situation where it was tricky to know whether a change I made would perturb the layouts consuming TimeInput - so this change adds a test snapshot that clearly shows it:

```
<p> some text above
<timeinput>
<p> some text below
```

If the padding is changed we should see it when the diff shows spacing changes in between timeinput and the top or bottom p.


benbcai (approved):

jmsv6d (approved):

packages/terra-time-input/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Added
+  * Added a test of time-input padding.
```
sdadn, line 6: nit:
```suggestion
  * Added a test for time-input padding. 
```

PR #1513, opened 2022-02-07
Time a11y omega

xenoworf:
### Summary
**This is a redo of #1497. It's the same functionality, but I was in merging hell so I created this PR. Any comments left in that PR were already handled prior to this PR.**

This PR makes the Time Input as accessible as possible while remaining passive to all the little quirks and features (e.g. automatically advancing the focus). There are a couple known issues I can't crack - see below.

The approach here is to wrap the existing inputs of the component in higher order components that provide accessibility. The idea being that these can eventually be folded into the Inputs themselves.

The main new features are:
1. There is a new, (sadly, for passivity) optional `a11yLabel` prop. When supplied, it is used as the label of the Time Input by the screen reader. When it's omitted, there is a default value "Time". I'm open to a different default - it's easy to change.
2. Each component (e.g. minute, hour, am button) is described by the screen reader upon focus. Here is an example: If I place focus on the hour field I might hear "Hour, required invalid text input, a two-digit twenty-four hour value". That phrase is a combination of description text we must localize ("a two-digit twenty-four hour value") and parts that the screen reader computes based on our use of HTML attrs, including some aria-attrs ("required", "invalid", "text input"). 
3. Items that don't make sense when read by a screen reader are no longer read. This includes the ":" time seperators and the mask text, e.g. "hh:mm:ss". Item 2 above takes the place of these phrases. 
4. The complete and localized time phrase, e.g. "09:22 pm", will be read if present. The phrase will also be read when the time is changed to a different and valid value, e.g. "09:23 pm". The time is not read if the time is complete or equivalent to the previous time value, so "09:22" and "9:22" are equivalent. Note: we are not localizing the time value, only the phrasing of the complete time. There is no conversion of timezones, etc.

### Known issues
1. For reasons I can't figure out, the doc tests about **mobile incomplete and mobile invalid are not reading the fields correctly**. The screen reader does not read them as incomplete/invalid despite the attributes looking correct... The non-mobile versions of these doc-tests are read correctly. I need to figure out what's different.
2. **The screen reader user will not be read any indication that the toggle button state is invalid, despite the toggle button visually indicating (red box) that the button is invalid**. There is no way for the screen reader to recognize that the consumer has marked a am/pm toggle button as invalid because buttons don't have such a concept. Please tell me if there is some way to express "invalid button state" using HTML attrs. I found none.

### Deployment Link
https://terra-framew-time-a11y--tjwx9m.herokuapp.com/home/terra-framework/home

### Testing
I went through every test scenario in the doc-site and manually tried macOS VoiceOver on each. I also inspected the [Accessibility Pane](https://developer.chrome.com/docs/devtools/accessibility/reference/#pane) to make sure the attributes were correct. 

I added a new Accessibility doc site test so that our consumers can be shown the features. That is just a bare minimum. We really should write a small guide on accessibility for this component some day.

I did not run this through exhaustive A11Y auditing or testing against Windows screen readers or iOS's voiceOver implementation. 

### Next steps
* In a future phase, we should introduce a Time Input Field component, which does the tricky wrapping of the Time Input in such a way as to marry the A11Y of this PR with the needs of a field.


packages/terra-date-input/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Fixed
+  * Tests.
```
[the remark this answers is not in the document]
xenoworf: Updated in latest commit.

xenoworf:
> Fails
> 🚫	Please include a CHANGELOG entry for each changed package this PR. Looks like a CHANGELOG is missing for:
> * terra-date-time-picker
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [dabc4ce](https://github.com/cerner/terra-framework/commit/dabc4ce7a94fc3a876ea3b6bdc9972967779a653)

Divide by zero, bot: the change _was_ the changlog.

jmsv6d (approved):

packages/terra-date-input/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Fixed
```
jmsv6d, line 5: This was added to the wrong packages changelog, I believe this was ment for terra-date-time-picker, that's why the danger script is failing.

packages/terra-time-input/translations/en-GB.json
```diff
@@ -1,11 +0,0 @@
-{
```
benbcai, line 1: What is the reason for deleting some of these translations files (nl-BR; pt-BR, fr_FR, etc)?

packages/terra-time-input/src/TimeUtil.js
```diff
@@ -240,6 +259,130 @@ class TimeUtil {
...
+  static isPM(props, state, postMeridiem) {
+    return TimeUtil.is12Hour(props) && state.meridiem === postMeridiem;
+  }
+
+  /**
+ * Build a UTC time string in kind-of-but-not-really ISO 8601 format.
+ *
+ * WARNING: Read the name. This has very peculiar behavior you probably don' want ot use for new features. I'm
+ * keeping this around as it was, because it's got strange behavior I'm not sure how to break. It has a bug where a
+ * person could leave some fields blank and see a value containing 'NaN', like 'TNaN:22'. Not sure if that is
```
sdadn, line 287: Should we log a jira to investigate this bug?

packages/terra-time-input/src/_Minute.jsx
```diff
@@ -0,0 +1,67 @@
...
+  static STRINGS = [...Minute.NUMBERS.map((num) => String(num).padStart(2, 0)), ...Minute.NUMBERS.map((num) => String(num))];
+
+  /**
+   * Factory method to create a Minute from a string. Use this to take form input and normalize it for later operations.
+   * @param {String} string A string expressing an minute. Can be be like '1' or '01'.
+   * @returns a Minute object or undefined if the minute doesn't make sense.
+   */
+  static FromString(string) {
+    if (!Minute.STRINGS.includes(string)) {
+      // throw new Error(`Expected a string in ${Minute.STRINGS}`);
```
sdadn, line 13: is there a reason this was commented out?

sdadn:
+10 for all the documentation in the code. You are an inspiration.

packages/terra-time-input/translations/en-GB.json
xenoworf: There were no differences in those files from their parent locale - only dual maintenance. For example _fr_FR_ (French, France) and _fr_ (French) were identical. End-users in fr_FR locale will receive the fr localization at runtime. We should add these back only if one requires different L10n, such as some of the Spanish locales have.

packages/terra-time-input/src/TimeUtil.js
```diff
@@ -177,6 +181,11 @@ class TimeUtil {
     return '00';
   }
 
+  /**
+   * Returns a string representation of the time's hour, or ''.
+   * @param {*} time String time formatted like hh:mm or hh:mm:ss.
+   * @returns A zero-padded two digit hour, 00 <= mm < 24.
+   */
```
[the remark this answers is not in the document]
xenoworf: 🏆

packages/terra-date-input/CHANGELOG.md
xenoworf: Oh thanks!

packages/terra-time-input/src/_Minute.jsx
xenoworf: I will delete it. It's left over from an earlier approach where I thought raising an error was a good idea. I changed to returning `undefined` instead.

packages/terra-time-input/src/_Hour.jsx
```diff
@@ -0,0 +1,134 @@
...
+
+  /**
+   * Factory method to create an Hour from a string. Use this to take form input and normalize it for later operations.
+   * @param {String} string A string expressing an hour found within the mode(see below). Can be be like '1' or '01'.
+   * @param {Hours.MODES} mode Tells the factory how to interpret the string literal. Is it a 24-hour hour? A
+   * 12-hour AM value, or a 12-hour PM value? Knowing the mode lets the constructor normalize the value to be
+   * a 24-hour clock hour value.
+   * @returns an Hour object or undefined if the hour doesn't make sense.
+   */
+  static FromString(string, mode = Hour.TWENTY_FOUR_HOUR) {
```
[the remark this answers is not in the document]
xenoworf: Will fix

packages/terra-time-input/src/TimeUtil.js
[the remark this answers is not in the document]
xenoworf: Thanks to this emoji convo I caught a couple doc typos. Will fix.

packages/terra-time-input/translations/en-GB.json
benbcai: The pattern we have been using has always been including the translations files for all supported locales (even if the strings for the locale are identical to the language. I think our [aggregate-translations package](https://github.com/cerner/terra-toolkit/blob/main/packages/terra-aggregate-translations/src/aggregate-messages.js#L11) will warn if we have missing locale files. 

This would also introduces inconsistency with other existing packages. 'm not totally against removing translations files that have no difference with the base language file. If we do go this route and to minimize inconsistencies, I think we should make an effort to do this across all packages, potentially update our generators, update the logic in aggregate-translations, potentially update workflow-server has it currently has a validation script that would fail if there are missing strings that are not translated for a given locale.

So there is quite a bit of work that is needed to get done in harmony if we want to get this done right.

Given this, I personally think we should keep these files for now until we have a better plan in place to remove the matching locales.

jmsv6d: @benbcai actually there was already a change made in aggregate translations to not warn about that any more: https://github.com/cerner/terra-toolkit/commit/36de8d91d5905f5aa25659749299113168233376
And workflow-framework only checks that there are translations for the base locales, not regional translations:
https://github.cerner.com/orion/workflow_server/blob/master/client/tests/jest/validate-translations.test.js#L224-L231

sdadn (approved):

packages/terra-time-input/translations/en-GB.json
jmsv6d: I don't think we need to necessarily make intentional changes to remove the extraneous translations in all our components, but when the translations team tells us that there's no difference between a regional locale and the base locale, there's not any need to actually include them, since aggregate-translations will automatically fallback, it reduces the chance of copy/paste errors. I don't see any harm in cleaning them up if we happen to be updating translations anyway.

xenoworf: I verified that our L10n falls back correctly. I visited the deployment, flipped the locale to fr_FR and saw the correct fr fallbacks. Given that and the comments above, I think this is fine. I agree no need to rip out translations right away. It can be done over time as we touch projects.

image

benbcai (approved):

chrismichalewicz:
Functionally reviewed

PR #1512, opened 2022-02-07
[terra-date-input] Keyboard shortcut detection fix for firefox

BenBoersma:
### Summary

Swaps to using `event.key` to detect keyboard shortcuts instead of `event.keyCode` since KeyCodes can differ between browsers. 

Closes #

### Deployment Link

https://terra-framework-deployed-pr-#.herokuapp.com/

### Additional Details

Thank you for contributing to Terra.
@cerner/terra

PR #1520, opened 2022-02-23
[terra-date-time-picker] Add support for date-picker and time-input keyboard shortcuts

### LINKED ISSUE #996: Enable hot keys for Terra Date/Time field

BenBoersma:
### Summary
Adding initial support of time-input keyboard shortcuts of "N" (set time to now, if date is blank or invalid set date to today as well), "-" (decrement minute by 1, set date to today if no valid date and time to now if no valid time, then decrement), and "+" (increment minute by 1, set date to today if no valid date and time to now if no valid time, then increment).

Also includes some edge case fixing in time-input for increment and decrementing time in roll-over cases.

Closes #996

### Deployment Link

https://terra-framew-terra-date-ij3eje.herokuapp.com/components/cerner-terra-framework-docs/date-time-picker/date-time-picker

### Additional Details

@cerner/terra


packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -647,3 +647,791 @@ Terra.describeViewports('DateTimePicker', ['tiny', 'large'], () => {
...
+          expect($('input[name="terra-date-day-input"]')).toHaveValue((`0${today.date()}`).slice(-2));
+          expect($('input[name="terra-time-hour-input"]')).toHaveValue('12');
+          expect($('input[name="terra-time-minute-input"]')).toHaveValue('00');
+          expect($('input[name="terra-time-second-input"]')).toHaveValue('00');
+          expect($('button[class*="meridiem"]')).toHaveAttrContaining('class', 'is-selected');
+        });
+      });
+
+      describe('if valid date', () => {
+        it('in the hour input should not change date and should leave it as AM', () => {
```
xenoworf, line 1229: These docs strings are a bit confusing. Seems like we're testing that `A` does nothing? I believe the intent is to test that if the date is a valid AM date then nothing happens? If so please change line 1228 and that will make it clear.

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -647,3 +647,791 @@ Terra.describeViewports('DateTimePicker', ['tiny', 'large'], () => {
...
+          expect($('input[name="terra-date-year-input"]')).toHaveValue('2017');
+          expect($('input[name="terra-date-month-input"]')).toHaveValue('08');
+          expect($('input[name="terra-date-day-input"]')).toHaveValue('08');
+          expect($('input[name="terra-time-hour-input"]')).toHaveValue('10');
+          expect($('input[name="terra-time-minute-input"]')).toHaveValue('30');
+          expect($('input[name="terra-time-second-input"]')).toHaveValue('00');
+          expect($('button[class*="meridiem"]')).toHaveAttrContaining('class', 'is-selected');
+        });
+
+        it('in the second input should not change date and should leave it as AM', () => {
```
xenoworf, line 1259: A couple test scenarios I'm not seeing, but think are important:
* 12-hour PM time, press A and see the time switch to PM
* 24-hour clock, press A and see nothing changes.

xenoworf (changes requested):

packages/terra-date-time-picker/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added
+  * Added keyboard shortcuts `N`, `-`, `+`, `A`, and `P` for auto-setting date and time values.
```
benbcai, line 6: Do we want to call out `T` also?

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -318,9 +318,57 @@ class DateTimePicker extends React.Component {
...
+    if (event.type === 'keydown') {
+      const timeArray = time.split(':');
+      if (event.key === 'n' || event.key === 'N') {
+        previousDateTime = DateUtil.createSafeDate(DateUtil.getCurrentDate(), this.initialTimeZone);
+        validDate = true;
+        dateChanged = true;
+      }
+
+      if (event.key === '-' || event.key === '_') {
+        if (!validDate) {
```
benbcai, line 335: There appears to be an issue when calculating the missing hour when daylight saving begins in the Spring. When the time is `3:00` and using the minus sign `-` to subtract the minute it is expected to go to `1:59` but instead it cycles back to `3:59`.
Feb-24-2022 18-19-40

sdadn (approved):

packages/terra-date-time-picker/src/terra-dev-site/doc/date-time-picker/KeyboardShortcuts.3.doc.mdx
```diff
@@ -0,0 +1,28 @@
...
+| T              | Date  | Sets the date value to the current date (Today).                                                                                                                                             |
+| -              | Date  | Decrements the day value by 1 day. If blank or not a valid date, input will set value to current date minus 1 day.                                                                           |
+| +              | Date  | Increments the day value by 1 day. If blank or not a valid time, input will set value to current date plus 1 day.                                                                            |
+| N              | Time  | Sets the time value to the current time (Now). Also sets date to the current date if there is no date or invalid date                                                                        |
+| -              | Time  | Decrements the minute value by 1 minute. If blank or not a valid time or date, input will set date to current date and value to current time minus 1 minute.                                 |
```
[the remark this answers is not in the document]
neilpfeiffer: @xenoworf I'll remind you that our requirements for these specific hotkey interactions are based on matching existing learned behaviors from internal win32 components. The T/N suite of hotkeys/keyboard shortcuts are not per-input node focused, but per date/time object focused, as such they only will ever affect a single node per concept (day for date, hour for time).

 E.G. pressing T sets today, we will add pressing M sets it to the first of the month, Y sets it to the first of the year. Plus (+) and minus (-) in this case are not synonymous with the up and down arrow in each input note unit, but work with the same consistency as the previous T, M, Y does, since the +/- are used as modifiers to those keys.

packages/terra-date-time-picker/CHANGELOG.md
BenBoersma: So I wasn't sure how accurate that would be, because since we pull in date-picker, as soon as date-picker consumed its changes `T` was supported? so these changes mostly deal with the "extra" that has to happen with time shortcuts and dates.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -318,9 +318,57 @@ class DateTimePicker extends React.Component {
...
+
+    if (event.type === 'keydown') {
+      const timeArray = time.split(':');
+      if (event.key === 'n' || event.key === 'N') {
+        previousDateTime = DateUtil.createSafeDate(DateUtil.getCurrentDate(), this.initialTimeZone);
+        validDate = true;
+        dateChanged = true;
+      }
+
+      if (event.key === '-' || event.key === '_') {
```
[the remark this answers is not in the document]
xenoworf: I talked offline w/ @neilpfeiffer and L10N is a future enhancement.

packages/terra-date-time-picker/CHANGELOG.md
xenoworf: From a consumer or end-users POV, they will not see or care about the internal distinction between this component and others it consumes. As a simple solution: how about adding a link in the Date Time Picker doc to reference the additional time commands?

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
BenBoersma: Updated the doc strings to hopefully be clearer in this commit: https://github.com/cerner/terra-framework/pull/1520/commits/cf7f6e84415a91dbdd1f7d0a327de6c14e65293b

packages/terra-date-time-picker/src/DateTimePicker.jsx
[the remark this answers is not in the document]
xenoworf: @chrismichalewicz FYI - we should enhance the DateTimePicker's keyboard commands to support the different keyboard layouts used in the locales that we support.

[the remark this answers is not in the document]
BenBoersma: I believe this is out of scope for now, part of the requirements are (sanitized for public github)

> Hotkey translations
> Localization of hot key sequences with i18n will be handled as future task after another JIRA is complete.

packages/terra-date-time-picker/CHANGELOG.md
xenoworf: Oh sorry, I see you indeed have the `T` referenced in the doc!! My bad 🤦🏼

packages/terra-date-time-picker/src/terra-dev-site/doc/date-time-picker/KeyboardShortcuts.3.doc.mdx
```diff
@@ -0,0 +1,28 @@
+import { Badge } from 'terra-date-time-picker/package.json?dev-site-package';
+
+<Badge />
+
+# Keyboard Shortcuts for Terra Date Time Picker
+
+The Terra Date Time Picker offers additional functionality with specific keyboard shortcuts to save time when interacting with date and time inputs in terra-date-time-picker.
```
xenoworf, line 7:
```suggestion
The Terra Date Time Picker offers additional functionality with specific keyboard shortcuts to save time when interacting with date and time inputs in terra-date-time-picker.

_Note: these commands assume a QWERTY keyboard layout. Other layouts will probably still work, but they have not yet been tested._
```

xenoworf (changes requested):
Just a minor doc change I'd like you to add. Other than that it looks good.

xenoworf (approved):

packages/terra-date-time-picker/src/DateTimePicker.jsx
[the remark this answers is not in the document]
chrismichalewicz: @xenoworf  Late to this party but +1 to @neilpfeiffer  and @BenBoersma  in tht L10N is outside the scope of this project.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -318,9 +318,57 @@ class DateTimePicker extends React.Component {
...
+          dateChanged = true;
+        }
+
+        if (timeArray[0] === '23' && timeArray[1] === '59') {
+          previousDateTime = previousDateTime.subtract(1, 'd');
+          dateChanged = true;
+        }
+      }
+
+      if (event.key === '=' || event.key === '+') {
```
benbcai, line 347: These `event.key` checks should really be an `elseif`.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -318,9 +318,57 @@ class DateTimePicker extends React.Component {
...
+        if (!validDate) {
+          previousDateTime = DateUtil.createSafeDate(DateUtil.getCurrentDate(), this.initialTimeZone);
+          validDate = true;
+          dateChanged = true;
+        }
+      }
+    }
+
+    if (dateChanged) {
+      this.handleDateChange(event, previousDateTime.format('YYYY-MM-DD'));
```
benbcai, line 370: I don't think we can call handleDateChange here because we already call `this.handleChange` and/or `this.handleChangeRaw` at the end of this function, which will eventually call the `onChange` callback. Call `this.handleDateChange` here again will also eventually call `onChange` callback. As a result, consumers would receive the `onChange` callback twice for every single change.

packages/terra-time-input/CHANGELOG.md
```diff
@@ -19,6 +19,9 @@
 * Added
   * Added keyboard shortcuts `N`, `-`, `+`, `A`, and `P` for auto-setting time values.
 
+* Fixed
+  * Fixed some edge case scenarios for keyboard shortcuts
```
benbcai, line 23: This should be moved up to the Unreleased section.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -460,7 +460,18 @@ class TimeInput extends React.Component {
...
+          } else {
+            if (variant === TimeUtil.FORMAT_12_HOUR && hour === '12') {
+              if (meridiem === this.anteMeridiem) {
+                meridiem = this.postMeridiem;
+              } else {
+                meridiem = this.anteMeridiem;
+              }
+            }
+            hour = TimeUtil.decrementHour(hour);
+          }
```
benbcai, line 474: We should add some tests for the changes in this file.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -318,9 +318,57 @@ class DateTimePicker extends React.Component {
 
   handleTimeChange(event, time) {
```
benbcai, line 319: This function is getting so big and complicated that it is hard to follow what it is doing. We should move some of this new logic to a separate function.

benbcai: We should probably just keep it this way for this iteration but it would help to add some code comments for each set of logic to make it easier to understand.

packages/terra-date-time-picker/src/DateTimePicker.jsx
BenBoersma: Changed here https://github.com/cerner/terra-framework/pull/1520/commits/e31d2250dc60b5bc379214c65e7312ff889cf584#diff-063daa4349be2bbad9c58209d280bfaee7a04b7e429163ebdb7784768f501a5cR332

packages/terra-date-time-picker/src/DateTimePicker.jsx
BenBoersma: You were correct, and it left over from problem with partial dates, removed here https://github.com/cerner/terra-framework/pull/1520/commits/e31d2250dc60b5bc379214c65e7312ff889cf584#diff-063daa4349be2bbad9c58209d280bfaee7a04b7e429163ebdb7784768f501a5cL369

packages/terra-time-input/CHANGELOG.md
BenBoersma: Fixed here. https://github.com/cerner/terra-framework/pull/1520/commits/e31d2250dc60b5bc379214c65e7312ff889cf584#diff-72cbeeaa7b7acb5b5ff08c157dd2db817a1cb487850786e696399735fb4898e1R5

packages/terra-date-time-picker/src/DateTimePicker.jsx
BenBoersma: Added some code comments for the keyboard shortcut section https://github.com/cerner/terra-framework/pull/1520/commits/e31d2250dc60b5bc379214c65e7312ff889cf584#diff-063daa4349be2bbad9c58209d280bfaee7a04b7e429163ebdb7784768f501a5cR326

packages/terra-date-time-picker/src/DateTimePicker.jsx
BenBoersma: This bit of logic should solve it for `-`, without effecting Fall DST hour issues. Feel free to test it https://github.com/cerner/terra-framework/pull/1520/commits/e31d2250dc60b5bc379214c65e7312ff889cf584#diff-063daa4349be2bbad9c58209d280bfaee7a04b7e429163ebdb7784768f501a5cR345-R352

`+` didn't need to be fixed because the auto skipping of the spring hour worked in its favor to just auto bump up.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -318,14 +318,64 @@ class DateTimePicker extends React.Component {
...
+        }
+
+        if (timeArray[0] === '23' && timeArray[1] === '59') {
+          // Time is 23:59 which means previous datetime should be on the wrong day, lower it by 1
+          previousDateTime = previousDateTime.subtract(1, 'd');
+        }
+
+        const changedDateTime = DateTimeUtils.updateTime(previousDateTime, this.timeValue, this.props.showSeconds);
+        const checkHour = previousDateTime.subtract(1, 'minute').hours();
+        previousDateTime.add(1, 'minute');
```
benbcai, line 347: Because the `add` and `subtract` APIs mutates the original moment and we have no intention to change the value of `previousDateTime` here, instead of subtracting a minute and then adding a minute back, it seems to make more sense and cleaner to just make a clone of `previousDateTime` to use for `checkHour`.

Something like this to prevent having to change the value of previousDateTime.

> const checkHour = previousDateTime.**clone()**.subtract(1, 'minute').hours();

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -318,14 +318,64 @@ class DateTimePicker extends React.Component {
...
+        // `+` add 1 minute`
+        if (!validDate) {
+          // Date is not valid, set date to today making it valid
+          previousDateTime = DateUtil.createSafeDate(DateUtil.getCurrentDate(), this.initialTimeZone);
+          validDate = true;
+        }
+
+        if (timeArray[0] === '00' && timeArray[1] === '00') {
+          // Time is 00:00 which means previous datetime should be on the wrong day, increase it by 1
+          previousDateTime = previousDateTime.add(1, 'd');
```
benbcai, line 363: We need to check to ensure that by adding 1 day does not go beyond the MAX supported date of 12/31/2100. For example, if the date is 12/31/2022 and the `+` key is pressed, it would update the date to `01/01/2101` which was previously not possible to enter because it exceed the [max date](https://github.com/cerner/terra-framework/blob/terra-date-picker%404.72.0/packages/terra-date-picker/src/DateUtil.js#L535-L536). 

This workflow should also be handled in terra-date-picker with the hot key changes but it currently does not. It would be best if this logic can be handled only in terra-date-picker so that terra-date-time-picker doesn't need to care about it.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -318,14 +318,64 @@ class DateTimePicker extends React.Component {
...
+        // `-` subtract 1 minute`
+        if (!validDate) {
+          // Date is not valid, set date to today making it valid
+          previousDateTime = DateUtil.createSafeDate(DateUtil.getCurrentDate(), this.initialTimeZone);
+          validDate = true;
+        }
+
+        if (timeArray[0] === '23' && timeArray[1] === '59') {
+          // Time is 23:59 which means previous datetime should be on the wrong day, lower it by 1
+          previousDateTime = previousDateTime.subtract(1, 'd');
```
benbcai, line 342: See similar [comment](https://github.com/cerner/terra-framework/pull/1520/files#r820890748) below regarding subtracting a day that goes beyond the supported MIN date of 01/01/1900.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -318,14 +318,64 @@ class DateTimePicker extends React.Component {
...
+      if (event.key === 'n' || event.key === 'N') {
+        // `N` set Date to today
+        previousDateTime = DateUtil.createSafeDate(DateUtil.getCurrentDate(), this.initialTimeZone);
+        validDate = true;
+      } else if (event.key === '-' || event.key === '_') {
+        // `-` subtract 1 minute`
+        if (!validDate) {
+          // Date is not valid, set date to today making it valid
+          previousDateTime = DateUtil.createSafeDate(DateUtil.getCurrentDate(), this.initialTimeZone);
+          validDate = true;
```
benbcai, line 337: It is possible for consumers to pass in a date time value that is outside the range of the `minDate` and `maxDate` as illustrated in this [example](https://engineering.cerner.com/terra-framework/tests/terra-date-time-picker/date-time-picker/date-time-picker-default-date-out-of-range) with this [test](https://github.com/cerner/terra-framework/blob/terra-date-picker%404.72.0/packages/terra-date-time-picker/src/terra-dev-site/test/date-time-picker/DateTimePickerDefaultDateOutOfRange.test.jsx#L6-L8).

When this scenario occurs (i.e. [isDefaultDateAcceptable](https://github.com/cerner/terra-framework/blob/terra-date-picker%404.72.0/packages/terra-date-picker/src/DatePicker.jsx#L214) would be false), we have logic [here](https://github.com/cerner/terra-framework/blob/terra-date-picker%404.72.0/packages/terra-date-picker/src/DatePicker.jsx#L319-L323) to automatically clear the invalid initial date when the date input receives focus.

This logic appears to be broken when pressing the `N` key to populate both the date and time. The logic works on the initial input focus where both the date and time are cleared. After pressing the `N` key to set the current date and time, putting focus in the date input should no longer clear the date again for the second time. It should be cleared only once for the entire lifecycle of the component. 

Whenever we set the date and time to a valid value using any of these shortcut keys, date-time-picker needs to communicate with date-picker that the initial date is now no longer considered out of range and isDefaultDateAcceptable should no longer be false to avoid the date from being cleared a second time.

out-of-range-date

packages/terra-time-input/src/TimeInput.jsx
BenBoersma: Added here https://github.com/cerner/terra-framework/pull/1520/commits/23db2ef02289816b2cbf9619172dc497a918286f#diff-2bc93a8867c8555b5abe36e5b499d356c82131d59cb0390dcc5668fb928a8ba0R126

packages/terra-date-time-picker/src/DateTimePicker.jsx
BenBoersma: Changed here https://github.com/cerner/terra-framework/pull/1520/commits/d798c38e2975d58fd2abecc55bc87db4aed4ac68#diff-063daa4349be2bbad9c58209d280bfaee7a04b7e429163ebdb7784768f501a5cR358

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -318,14 +318,64 @@ class DateTimePicker extends React.Component {
...
-    const validDate = DateUtil.isValidDate(this.dateValue, this.state.dateFormat) && this.isDateTimeAcceptable(DateTimeUtils.convertDateTimeStringToMomentObject(this.dateValue, this.timeValue, this.state.dateFormat, this.props.showSeconds, this.initialTimeZone));
+    let validDate = DateUtil.isValidDate(this.dateValue, this.state.dateFormat) && this.isDateTimeAcceptable(DateTimeUtils.convertDateTimeStringToMomentObject(this.dateValue, this.timeValue, this.state.dateFormat, this.props.showSeconds, this.initialTimeZone));
     const validTime = DateTimeUtils.isValidTime(this.timeValue, this.props.showSeconds);
-    const previousDateTime = this.state.dateTime ? this.state.dateTime.clone() : null;
+    let previousDateTime = this.state.dateTime ? this.state.dateTime.clone() : null;
+
+    if (event.type === 'keydown') {
+      // Keyboard shortcut is pressed, split the time value into hour and minute inputs
+      const timeArray = this.timeValue.split(':');
+      if (event.key === 'n' || event.key === 'N') {
```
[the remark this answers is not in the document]
BenBoersma: Added the `handleDateChange` function call back in here https://github.com/cerner/terra-framework/pull/1520/commits/d798c38e2975d58fd2abecc55bc87db4aed4ac68#diff-063daa4349be2bbad9c58209d280bfaee7a04b7e429163ebdb7784768f501a5cR388
but modified with a new argument that when skips calling `handleChange` to avoid double calls

packages/terra-date-time-picker/src/DateTimePicker.jsx
BenBoersma: Fixed in Date-picker here https://github.com/cerner/terra-framework/pull/1520/commits/ea6e6e82248964185d5d81ae2f63f8e50f6603ed#diff-dafaabc64f600b92fbfa865adda17d41223fea5f885e15310c2e88582c58e0dcR502 
along with adding tests for it

Currently still 100% correctly implemented in date-time-picker for when increasing or decreasing the minute causes us to go over. but tests for that have been added. https://github.com/cerner/terra-framework/pull/1520/commits/d798c38e2975d58fd2abecc55bc87db4aed4ac68#diff-8f8488b16a2c0f0e4089aa01407d27d8ba3116e1f11e277b8d3554ad88972292R977

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -303,29 +303,96 @@ class DateTimePicker extends React.Component {
...
+        if (!validDate) {
+          // Date is not valid, set date to today making it valid
+          previousDateTime = DateUtil.createSafeDate(DateUtil.getCurrentDate(), this.initialTimeZone);
+          validDate = true;
+          dateChanged = true;
+        }
+      }
+    }
+    if (dateChanged) {
+      this.handleDateChange(event, previousDateTime.format('YYYY-MM-DD'), null, true);
```
benbcai, line 389: I thought this was removed based on this [comment](https://github.com/cerner/terra-framework/pull/1520#discussion_r815213080). What is the reason for adding this back as this will trigger the onChange callback twice?

benbcai: After a closer look, I see what is being done here. It feels hacky and anti-pattern to do things this way by calling handeDateChange. If it needs any state or class variables need to be updated, it should be more direct to just update what is needed directly here instead of relying and cherry picking the logic in handleDateChange.

packages/terra-date-picker/src/DateUtil.js
```diff
@@ -499,7 +499,14 @@ class DateUtil {
    * @return {moment} The incremented moment date
    */
   static incrementDateByDay(date, format) {
-    return moment(date, format, true).add(1, 'd').format('YYYY-MM-DD');
+    const momentDate = moment(date, format, true).add(1, 'd');
+    if (momentDate.year() < Number(DateUtil.MIN_YEAR)) {
+      return DateUtil.MIN_DATE;
```
benbcai, line 504: It's not a big deal but I don't think it will ever get in this scenario because the year should never be 1899 to being with.

packages/terra-date-picker/src/DateUtil.js
```diff
@@ -508,7 +515,14 @@ class DateUtil {
    * @return {moment} The decremented moment date
    */
   static decrementDateByDay(date, format) {
-    return moment(date, format, true).subtract(1, 'd').format('YYYY-MM-DD');
+    const momentDate = moment(date, format, true).subtract(1, 'd');
+    if (momentDate.year() < Number(DateUtil.MIN_YEAR)) {
+      return DateUtil.MIN_DATE;
+    } else if (momentDate.year() > Number(DateUtil.MAX_YEAR)) {
+      return DateUtil.MAX_DATE;
```
benbcai, line 522: Same as the above comment where it should ever get in this scenario because the year should never be greater than 2100 to being with.

packages/terra-date-time-picker/src/DateTimePicker.jsx
BenBoersma: Fixed Here https://github.com/cerner/terra-framework/pull/1520/commits/3ad7a15a91e5339a735c4c70ca93065c8d227b64#diff-063daa4349be2bbad9c58209d280bfaee7a04b7e429163ebdb7784768f501a5cL388

packages/terra-date-picker/src/DateUtil.js
BenBoersma: We shouldn't get into the scenario right now, but I would prefer leaving it in for protecting future uses of method just in case.

packages/terra-date-picker/src/DateUtil.js
BenBoersma: We shouldn't get into the scenario right now, but I would prefer leaving it in for protecting future uses of method just in case.

packages/terra-date-time-picker/src/DateTimePicker.jsx
BenBoersma: Added in a private prop that skips this logic in date-picker since date-time-picker also has its own version of the logic
https://github.com/cerner/terra-framework/pull/1520/commits/72af1e82737101eb86da6557fe50f9b3f1798f82#diff-2da0e35de52b21bfa57a60f6cd02cbde9957d81df2f3856f0ff48c61d68c5974R46

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -447,6 +452,10 @@ class DatePicker extends React.Component {
       selectedDateInPicker = this.state.selectedDate;
     }
 
+    if (this.isDefaultDateAcceptable && !this.isDateWithinRange(selectedDateInPicker)) {
+      selectedDateInPicker = undefined;
```
benbcai, line 456: Should we do this only for uncontrolled component? I would think if consumers are using a controlled component, we should just let them manage the value instead of clearing it.

BenBoersma: I'm not sure, basically the date-time-picker is using a controlled version of date-picker, and the issue we were having was that when we assigned a value to date-picker in date-time and it was out of range, `selectedDateinPicker` would still be the set value. If that value was wildly inaccurate and not within the same year of the accepted range it cause issues with the calender dropdown where the only way for the user to get to the accepted range was to click the left arrow to manually go through months until you got there.

benbcai (approved):

packages/terra-date-picker/src/DatePicker.jsx
benbcai: We can keep it this way. I just think that as a controlled DatePicker and from a DatePicker perspective, it shouldn't update the value passed in since it should allow the higher order component manage all its state and value.

chrismichalewicz:
Functionally Reviewed

packages/terra-date-time-picker/src/DateTimePicker.jsx
neilpfeiffer: Resolved in this comment https://github.com/cerner/terra-framework/pull/1520/files#r824176148

> **[@BenBoersma](https://github.com/BenBoersma)** --  [(Mar 10, 2022, 3:33 PM CST)](https://github.com/cerner/terra-framework/pull/1520/files#r824176148):
>
> Fixed in Date-picker here [ea6e6e8#diff-dafaabc64f600b92fbfa865adda17d41223fea5f885e15310c2e88582c58e0dcR502](https://github.com/cerner/terra-framework/commit/ea6e6e82248964185d5d81ae2f63f8e50f6603ed#diff-dafaabc64f600b92fbfa865adda17d41223fea5f885e15310c2e88582c58e0dcR502) along with adding tests for it
> 
> Currently still 100% correctly implemented in date-time-picker for when increasing or decreasing the minute causes us to go over. but tests for that have been added. [d798c38#diff-8f8488b16a2c0f0e4089aa01407d27d8ba3116e1f11e277b8d3554ad88972292R977](https://github.com/cerner/terra-framework/commit/d798c38e2975d58fd2abecc55bc87db4aed4ac68#diff-8f8488b16a2c0f0e4089aa01407d27d8ba3116e1f11e277b8d3554ad88972292R977)

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -313,19 +313,103 @@ class DateTimePicker extends React.Component {
...
+        }
+
+        if (timeArray[0] === '23' && timeArray[1] === '59') {
+          // Time is 23:59 which means previous datetime should be on the wrong day, lower it by 1
+          previousDateTime = previousDateTime.subtract(1, 'd');
+          dateChanged = true;
+        }
+
+        const changedDateTime = DateTimeUtils.updateTime(previousDateTime, this.timeValue, this.props.showSeconds);
+        const checkHour = previousDateTime.clone().subtract(1, 'minute').hours();
```
neilpfeiffer, line 354: For Daylight Savings handling when DST begins, the keyboard shortcuts for `+` & `-` are performing as expected when focus is in the time input, correctly skipping the missing hour when incrementing and decrementing time (`01:59`+`+`=`03:00`, `03:00`+`-`=`01:59`), but when focus is in the date input, the time is not being handled correctly.

The current behavior when the date is any day before or after when DST begins, (e.g. `March 06, 2020`) and the time is between 02:00 and 03:00, valid on March 6th (e.g. `03/06/2020 02:30`), with focus in the date input, pressing `+` these number counts, changes the value to:
1. `03/07/2020 02:30` (correct)
2. `03/08/2020 03:30` (correct)
3. `03/09/2020 03:30` (incorrect)
4. `03/10/2020 03:30` (incorrect)

And same is true in reverse starting with `03/12/2020 02:30` with pressing `-`, the time is correct on `03/08/2020` at `03:30`, but should go back to `02:30` starting on `03/07/2020`:

incorrect-DST-start-missing-hour-and-hotkeys-PR#1520

---

The correct expected behavior is when starting with the time between 02:00 and 03:00 on the days before/after the start of DST (e.g. `03/06/2020 02:30`), for time to shift up an hour on the day of the start of DST when the missing hour occurs, but then go back to the original time for days before or after when the missing hour does not occur. Using the same example above when pressing `+` these number counts, the correct values would be:
1. `03/07/2020 02:30` (correct)
2. `03/08/2020 03:30` (correct)
3. **`03/09/2020 02:30`** (correct)
4. `03/10/2020 02:30` (correct)
5. `03/11/2020 02:30` (correct)

---

neilpfeiffer (approved):
Tested: PR #1530
Resolves: Issue #996 (this is the last PR to complete the initial request)

**Reviewed:**
- Combines the initial implementation for keyboard shortcuts to set 'today', adding one day, and subtracting one day in the Date-Picker component, adding into the Date-Time-Picker with special handling for DST and coordinated usage with Times.
- Combines the initial implementation for keyboard shortcuts to set 'now', adding one minute, subtracting one minute, plus 12 hour 'A' and 'P' quick setting midnight/noon in the Time-Input, adding into the Date-Time-Picker with special handling for DST and coordinated usage with Dates.
- Verified usability was functioning as expected on desktop device with keyboard.
- Does not function on mobile since is locked to the numeric keyboard only (not able to press alphabetic keys).

\
\
**Related PRs already complete:**
- [[terra-date-input] Adding initial hot keys #1483](https://github.com/cerner/terra-framework/pull/1483#pullrequestreview-836580668)
- [[terra-time-input] Adding initial keyboard shortcuts #1503](https://github.com/cerner/terra-framework/pull/1503#pullrequestreview-866725797) 
- [[terra-date-picker] Add keyboard shortcuts #1507](https://github.com/cerner/terra-framework/pull/1507#pullrequestreview-868516062)
- [[terra-date-input] Keyboard shortcut detection fix for firefox #1512](https://github.com/cerner/terra-framework/pull/1512#pullrequestreview-881681826)

\
\
**Fix with Followup Issue:**
- Usability Research: consider moving cursor to the day input when '+' and '-' are pressed, or re-examine input cursor/selection behavior (perhaps highlighting full value). Is unexpected that +/- changes days while cursor can remain in the month or year, (up/down would change the month or year but +/- does not).
- Does not apply to when the Calendar dropdown/popup is open, which will be handled in an additional enhancement PR.
- Accessibility interaction hints and feedback will be added with an additional enhancement PR.
- **Discovered Bug:** Issue # _(adding new issue)_ -- when the date or time 
- **Introduces a new Bug:** Issue # _(adding new issue)_ -- For Daylight Savings, the `+` and `-` keyboard shortcut behaviors need a minor adjustment to when time is between 02:00 and 03:00, and focus is in the date input, to correctly pass over the DST start day when the missing hour occurs to adjust the time forward ("spring forward"), but then to return the time back on the days before or after. See comment: https://github.com/cerner/terra-framework/pull/1520#pullrequestreview-923848063

\
\
Verified by @neilpfeiffer, ready to merge.

 **UX Reviewed** |
-----------|

packages/terra-date-time-picker/src/DateTimePicker.jsx
neilpfeiffer: Approving current state and leaving this bug fix for a future enhancement to resolve. https://github.com/cerner/terra-framework/pull/1520#pullrequestreview-923876182

PR #1522, opened 2022-02-24
Date input a11y

xenoworf:
### Summary
A11y improvements for DateInput and DateInputField!

- Proper labels and descriptions for each control.
- Proper grouping of controls.
- Aria-live reading of the date value.
- Updated translations for better L10N and to support the A11Y improvements.
- Re-using the `AccessibleValue` and `AccessibleInput` private components from `time-input`
- Interactive doc tests added: play with the settings to see the components in action!

# TODOs (I'm working on these right now)
- Translations, in progress.
- WDIO tests.

### Deployment Link
TBD

### Testing
I wrote two test harnesses and manually tried all the prop combos I could think of. I will add additional tests soon.


packages/terra-date-input/src/DateInput.jsx
```diff
@@ -5,15 +5,24 @@ import classNamesBind from 'classnames/bind';
 import ThemeContext from 'terra-theme-context';
 import uuidv4 from 'uuid/v4';
 import { injectIntl } from 'react-intl';
-import Input from 'terra-form-input';
 import * as KeyCode from 'keycode-js';
 
+import AccessibleValue from 'terra-time-input/lib/_AccessibleValue';
+import AccessibleInput from 'terra-time-input/lib/_AccessibleInput';
```
benbcai, line 11: Both `_AccessibleValue` and `_AccessibleInput` are meant to be private modules in terra-input. Having to add the terra-time-input dependency to be able to use these private modules seems anti-pattern. I don't know of a good and easy option other than to duplicate these for terra-date-input. A long term strategy would be to create a common reusable accessibility toolkit package that can provides some of these reusable modules.

xenoworf: I agree it is an anti-pattern. Would it be OK if I left it as-is for now, and then submitted another PR to expose these as two new reusables in terra-core? I feel it is OK to "get away with it" for a few days or a week or so while I make this change. AFIAK end users won't notice.

benbcai: @xenoworf  Yes, we should leave it this way for now until when we see more use cases where we need more of these common modules and we can then plan to create a common place to store them.

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -5,15 +5,24 @@ import classNamesBind from 'classnames/bind';
 import ThemeContext from 'terra-theme-context';
 import uuidv4 from 'uuid/v4';
 import { injectIntl } from 'react-intl';
-import Input from 'terra-form-input';
```
benbcai, line 8: Since it no longer needs terra-form-input and this appears to be the only place that imports it, can it now be removed as a dependency?

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -565,7 +597,8 @@ class DateInput extends React.Component {
           onFocus={this.handleMonthFocus}
           onBlur={this.handleMonthBlur}
           disabled={this.props.disabled}
-          aria-describedby={ariaDescriptionId}
+          aria-invalid={this.props.isInvalid}
+          aria-required={this.props.required}
```
benbcai, line 601: Not sure if it is out of scope for this change. Should we also add `aria-disabled` if `this.props.disabled` is true?

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -500,6 +512,11 @@ class DateInput extends React.Component {
     }
   }
 
+  // Returns the display format, considering the locale and the prop.
+  computedDisplayFormat() {
+    return DateInputUtil.computedDisplayFormat(this.props.displayFormat, this.props.intl.locale);
```
benbcai, line 517: 👍

packages/terra-time-input/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * Fixed
   * Accessibility improvements.
+  * Supply isInvalid as well as aria-invalid to AccessibleInput
```
benbcai, line 9: This should be moved up to the Unreleased section.

packages/terra-date-input/src/DateInput.jsx
xenoworf: I think yes. I'll try it.

packages/terra-date-input/src/DateInput.jsx
xenoworf: Yes, it should. I will add that.

packages/terra-date-input/src/DateInput.jsx
xenoworf: Actually, I must leave it in because my test harness (doc tests) require this input.

packages/terra-date-input/src/DateInput.jsx
xenoworf: Logged #1523 to clean up later.

packages/terra-date-input/src/DateInput.jsx
xenoworf: Fixed with 0c9deda2a90b10bb13dcd44a55b11f8d203f86ef

packages/terra-date-input/src/DateInput.jsx
xenoworf: 🙇🏻‍♂️

packages/terra-time-input/CHANGELOG.md
xenoworf: Fixed with 998f1f0f978b7efd32b26ea09259745a1c3b9583

packages/terra-date-input/src/DateInputField.jsx
```diff
@@ -206,17 +206,25 @@ const DateInputField = (props) => {
...
+        {/* The text of the legend, besides any decorations. */}
         {legend}
-        {required && !isInvalid && hideRequired && <span className={cx('required-hidden')}>*</span>}
-        {showOptional && !required
-          && (<span className={cx('optional')}>{intl.formatMessage({ id: 'Terra.date.input.optional' })}</span>)}
+
+        {/* We only include the following span so that elements do not pop around in the layout when isInvalid changes. */}
+        {required && !isInvalid && hideRequired && (<span className={cx('required-hidden')}>*</span>)}
+
+        {/* This bit controls whether the (Optional suffix displays after the legend text */}
```
xenoworf, line 226:
```suggestion
        {/* This bit controls whether the (Optional) suffix displays after the legend text */}
```

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -5,15 +5,24 @@ import classNamesBind from 'classnames/bind';
...
 import styles from './DateInput.module.scss';
 
 const cx = classNamesBind.bind(styles);
 
 const propTypes = {
+  /**
+   * An identifier used by assistive technologies like screen readers to briefly describe this time input to users.
+   * The label is not rendered visually.
+   *
+   * **BEST PRACTICE FOR ACCESSIBILITY**: you _SHOULD_ set this to match whatever visible label you give in your UI.
```
supreethmr, line 22: Can we add a UX Accessibility Badge for this prop like below :

Screen Shot 2022-02-28 at 8 49 24 PM


```suggestion
   * ![IMPORTANT](https://badgen.net/badge/UX/Accessibility/blue)  **BEST PRACTICE FOR ACCESSIBILITY**: you _SHOULD_ set this to match whatever visible label you give in your UI.
```

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -738,21 +787,28 @@ class DateInput extends React.Component {
...
     // Using the state of month, day, and year to create a formatted date value
     let dateValue = '';
+    let completeDateValue;
 
     if (month.length > 0 || day.length > 0 || year.length > 0) {
       dateValue = `${year}-${month}-${day}`;
     }
 
-    this.formatDescriptionId = `terra-date-input-description-format-${this.uuid}`;
+    if (dateValue.length === 'xx-xx-xxxx'.length) {
```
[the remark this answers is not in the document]
xenoworf: 🙇🏻

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -738,21 +787,28 @@ class DateInput extends React.Component {
...
 
     const format = DateInputUtil.getDateFormat(this.props);
+    const label = a11yLabel || this.props.intl.formatMessage({ id: 'Terra.date.input.labelDefault' });
 
     return (
       <div
         {...customProps}
         className={dateInputClassNames}
         ref={(element) => { this.dateInputContainer.current = element; if (refCallback) { refCallback(element); } }}
+        role={isA11yControlled ? undefined : 'group'}
```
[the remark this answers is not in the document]
xenoworf: I used `undefined` because that is the default value of role in HTML.
```
> let div = document.createElement('div')
undefined
> div.role
undefined
```

xenoworf:
For the record, that last commit was also feedback from @supreethmr! My brain shorted out when I added the commit msg.

packages/terra-date-input/src/DateInput.jsx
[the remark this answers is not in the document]
eawww: image
The default value of any attribute that doesn't have a value is `null`. This mixup illustrates the point pretty well: with `null` you know you're getting a value; it's just a non-value value. `undefined` just means there's no defined value for the thing you're trying to access. 
In React, even if you pass `undefined` as the value of an element's attribute, it will still render an element with a `null` value for that attribute.

packages/terra-date-input/src/terra-dev-site/test/date-input/common/_RadioFieldGenerator.jsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import Radio from 'terra-form-radio';
+import RadioField from 'terra-form-radio/lib/RadioField';
+import PropTypes from 'prop-types';
+
+const propTypes = {
+  // The text of the legend tag.
+  fieldLegend: PropTypes.string,
+  // The name of every Radio
```
sdadn, line 9: Period for consistency
```suggestion
  // The name of every Radio.
```

packages/terra-date-input/src/DateInput.jsx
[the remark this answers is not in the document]
xenoworf: TIL!

sdadn (approved):

packages/terra-date-input/translations/sv.json
```diff
@@ -1,24 +1,29 @@
...
+  "Terra.date.input.april": "April",
+  "Terra.date.input.august": "Augusti",
+  "Terra.date.input.dayDescription": "Ett tvåsiffrigt tal",
+  "Terra.date.input.dayInputLabel": "Dag",
+  "Terra.date.input.dayInputLabelWithDefaultName": "Dag i datum",
+  "Terra.date.input.dayInputLabelWithName": "Dag i {a11yLabel}",
+  "Terra.date.input.december": "December",
+  "Terra.date.input.february": "Februari",
+  "Terra.date.input.formatDayMonthYear": "(DD månad ÅÅÅÅ)",
+  "Terra.date.input.formatMonthDayYear": "(månad DD ÅÅÅÅ)",
```
benbcai, line 11: The month is capitalized for all other locals and the M in Månad for SV was capitalized before. It is made lower case intentionally?

xenoworf: I asked this question to the Swedish translator.

xenoworf: My guess is that Month used to be capitalized only because of old line 4 it was the only word and so got a Title Case capitalization. Now that the translation strings are phrases perhaps the new capitalization is more normal to a native speaker. I'll update the PR if they say it's wrong.

benbcai (approved):

benbcai (approved):

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -738,21 +789,28 @@ class DateInput extends React.Component {
...
     const format = DateInputUtil.getDateFormat(this.props);
+    const label = a11yLabel || this.props.intl.formatMessage({ id: 'Terra.date.input.labelDefault' });
 
     return (
       <div
         {...customProps}
         className={dateInputClassNames}
         ref={(element) => { this.dateInputContainer.current = element; if (refCallback) { refCallback(element); } }}
+        role={isA11yControlled ? null : 'group'}
+        aria-label={isA11yControlled ? undefined : label}
```
eawww, line 811:
```suggestion
        aria-label={isA11yControlled ? null : label}
```

eawww (approved):
Pointing out one more spot where `undefined` is passed as an attribute in the changes. 
Looks good though! 👍

PR #1208, opened 2020-08-05
[WIP][terra-date-picker][terra-date-time-picker] Split Date Picker Input

avinashg1994:
### Summary

Split Date Picker Input into 3 separate inputs: day, month and year.

Closes #468 

### Deployment Link

https://terra-framework-deployed-pr-#.herokuapp.com/

### Testing

Jests and WDIOs

PR #1563, opened 2022-07-21
Fixes Memory Leak issue of Date-Picker

supreethmr:
### Summary

In version date-picker@V4.69.0 date-picker input field was splitted into three individual inputs to make date-picker more accessible for keyboard and screen-reader users. ( PR: https://github.com/cerner/terra-framework/pull/1208)

As part of PR #1208 [DateInput.jsx was re-written](https://github.com/cerner/terra-framework/pull/1208/files?file-filters%5B%5D=.jsx&show-deleted-files=true&show-viewed-files=true#diff-ed641a9a9e86097e537f02f796210b67af89a5b2eefac5add0745b656086de82) with react hooks, All the class component lifecycle methods and other methods are converted into hooks and arrow functions.

This code changes introduced usage of `useEffect` to state value for date : https://github.com/cerner/terra-framework/blob/76126d5b85d376f9cbc6a956899f18addd5b6ed3/packages/terra-date-picker/src/DateInput.jsx#L231 
```jsx
// Sets the date state based on the passed in value prop, or if it changes via a calendar click.
  useEffect(() => {
    if (DateUtil.isValidDate(value, momentDateFormat)) {
      const parseDate = DateUtil.getDateInputValues(DateUtil.dateOrder.YMD, dateValue, '-');
      dateDispatch(parseDate);
    } else if (dateValue === '') {
      dateDispatch({ day: '', month: '', year: '' });
    }
  }, [momentDateFormat, dateValue, value]);
```
`dateDispatch` is reducer hook of react added to perform updates to multiple state objects as a single state object. This process of updating multiple state within `useEffect` was causing memory leak by leaving few object references uncleared on heap memory. This was leading into `memory out of bound exception` in rev-cycle-scheduling app.

Memory leak was happening only when date-picker was used with SWT window. With modern browsers  date-picker seems to work fine without crashing the application but we could see slight increase in heap memory consumption by date-picker version greater than 4.68.0 in modern browsers as well.

In general state updates are not preferable actions to perform within `useEffect()` as it could make component to get into infinite update state if not handled properly.

With this PR the `useEffect` that was added just to update the state depending on changes to value prop have been removed. The functionality of updating date value / date selection can be achieved without state value, similar to how it was handled on class component prior to version 4.69.0 : https://github.com/cerner/terra-framework/blob/terra-date-picker%404.68.0/packages/terra-date-picker/src/DateInput.jsx#L241  

Similar approach has been used in this PR to update the date values. Along with changes to state update use of `useRef` and `useCallBack` hooks have been removed to reduce the side effects caused by extensive / unwanted usage of hooks.

`Map()` Data structure has been replaced with `WeakMap()` to reduce risk of memory leaks that tends to happen with `Maps()` as mentioned here  :  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap
> The second inconvenience is a memory leak because the arrays ensure that references to each key and each value are maintained indefinitely. 

Closes #

### Deployment Link

### Testing

Code changes are tested with existing WDIO and Jest Test to ensure code changes remains passive.
Code changes were also tested on the environment were the memory leak issue was identified.
Below are the screenshots showing heap memory consumption of application having terra-date-picker

**Before Fix with date-picker version >= 4.69.0**
memoryleak

_**As we can see in the above graph the section colored with yellow represents heap memory seems to be growing whenever the window having date-picker is opened. memory seems to be not restored after component is loaded**_

**After Fix**
memoryleakfix

_**We can notice in the above graph memory seems to be released indicated by dip in graph after component is loaded. memory seems to be restored after component is loaded**_

### Additional Details

 

@cerner/terra


packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -1,9 +1,7 @@
+/* eslint-disable no-unused-vars */
```
saket2403, line 1: This eslint-disable should be removed.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -198,23 +186,21 @@ const DatePickerInput = (props) => {
     delete inputAttributes.id;
   }
 
+  let dayInputRef = document.getElementById(dayInputId);
+  let monthInputRef = document.getElementById(monthInputId);
+  let yearInputRef = document.getElementById(yearInputId);
```
benbcai, line 191: We would get null input refs if no inputAttributes.id was passed in. If so, wouldn't it throw an error later on when it tries to access these refs?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -198,23 +186,21 @@ const DatePickerInput = (props) => {
...
-    monthInputRef.current = node;
-  }, []);
-
-  const setYearRef = useCallback(node => {
-    yearInputRef.current = node;
-  }, []);
+  let date = useMemo(() => ({ day: '', month: '', year: '' }), []);
+  // Sets the date state based on the passed in value prop, or if it changes via a calendar click.
+  if (DateUtil.isValidDate(value, momentDateFormat)) {
+    date = DateUtil.getDateInputValues(DateUtil.dateOrder.YMD, dateValue, '-');
```
benbcai, line 202: Wouldn't the separator based on the locale instead of using `-`?

benbcai: Never mind, please ignore my comment. `dateValue` is guaranteed to be in the ISO8601 format.

packages/terra-date-picker/src/react-datepicker/index.jsx
```diff
@@ -136,7 +136,7 @@ class DatePicker extends React.Component {
     /**
      * Highlight range of dates with custom classes.
      */
-    highlightDates: PropTypes.array,
+    highlightDates: PropTypes.instanceOf(WeakMap),
```
benbcai, line 139: 👍

packages/terra-date-picker/src/react-datepicker/calendar.jsx
```diff
@@ -99,7 +99,7 @@ export default class Calendar extends React.Component {
     /**
      * Highlight range of dates with custom classes.
      */
-    highlightDates: PropTypes.instanceOf(Map),
+    highlightDates: PropTypes.instanceOf(WeakMap),
```
benbcai, line 102: Did this play a role in contributing to the memory leak? Or is it just better in general to not keep a strong reference to the map keys?

sdadn (approved):

packages/terra-date-picker/src/DateInput.jsx
supreethmr: [Removed](https://github.com/cerner/terra-framework/pull/1563/commits/bafb9129abf5f0b7aa459d3567ccb02523fbdd58)

packages/terra-date-picker/src/DateInput.jsx
supreethmr: That's correct !! this would be null when `id` is not provided. Actually in DateInput ref's gets assigned on render by the node provided by terra-form-input' refCallback. hence no need of `document.getElementById()` to get element reference. 
Updated [here](https://github.com/cerner/terra-framework/pull/1563/commits/51a58975f288f50212a95e64c66f2e76766e1a92#diff-ed641a9a9e86097e537f02f796210b67af89a5b2eefac5add0745b656086de82)

packages/terra-date-picker/src/react-datepicker/calendar.jsx
supreethmr: No, but This change was done more as precaution of preventing possible leaks that could happen in some edge cases with key objects.

udaymattam (approved):

benbcai (approved):

BenBoersma (approved):

saket2403 (approved):

ShettyAkarsh (approved):

PR #1567, opened 2022-09-07
[terra-date-time-picker] Improve keyboard navigation when moving over the start date of DST.

### LINKED ISSUE #1554: [terra-date-time-picker] DST missing hour keyboard shortcut improvement
# Bug Report

## Description

PR #1520 **[terra-date-time-picker] Add support for date-picker and time-input keyboard shortcuts** -- when completed contained a bug in the handling of keyboard shortcuts when incrementing/decrementing days if the time was set between 02:00 and 03:00 (AM) and the user moved the date over the start day of DST when there is a missing hour.

Link to the originating comment detailing the problem: [../terra-framework/pull/1520#pullrequestreview-923848063](https://github.com/cerner/terra-framework/pull/1520#pullrequestreview-923848063)

## Steps to Reproduce

>For Daylight Savings handling when DST begins, the keyboard shortcuts for `+` & `-` are performing as expected when focus is in the time input, correctly skipping the missing hour when incrementing and decrementing time (`01:59`+`+`=`03:00`, `03:00`+`-`=`01:59`), but when focus is in the date input, the time is not being handled correctly.
>
>The current behavior when the date is any day before or after when DST begins, (e.g. `March 06, 2020`) and the time is between 02:00 and 03:00, valid on March 6th (e.g. `03/06/2020 02:30`), with focus in the date input, pressing `+` these number counts, changes the value to:
>1. `03/07/2020 02:30` (correct)
>2. `03/08/2020 03:30` (correct)
>3. `03/09/2020 03:30` (incorrect)
>4. `03/10/2020 03:30` (incorrect)
>
>And same is true in reverse starting with `03/12/2020 02:30` with pressing `-`, the time is correct on `03/08/2020` at `03:30`, but should go back to `02:30` starting on `03/07/2020`:
>
> incorrect-DST-start-missing-hour-and-hotkeys-PR#1520
>

---

## Expected Behavior

The correct expected behavior is when starting with the time between 02:00 and 03:00 on the days before/after the start of DST (e.g. `03/06/2020 02:30`), for time to shift up an hour on the day of the start of DST when the missing hour occurs, but then go back to the original time for days before or after when the missing hour does not occur. Using the same example above when pressing `+` these number counts, the correct values would be:
1. `03/07/2020 02:30` (correct)
2. `03/08/2020 03:30` (correct)
3. **`03/09/2020 02:30`** (correct)
4. `03/10/2020 02:30` (correct)
5. `03/11/2020 02:30` (correct)

---

saket2403:
### Summary

Closes #1554 

### Deployment Link

### Additional Details

@cerner/terra

### DIFF
--- a/packages/terra-date-time-picker/src/DateTimePicker.jsx
+++ b/packages/terra-date-time-picker/src/DateTimePicker.jsx
@@ -297,6 +297,16 @@ class DateTimePicker extends React.Component {
       const previousDateTime = this.state.dateTime ? this.state.dateTime.clone() : DateTimeUtils.createSafeDate(formattedDate, this.initialTimeZone);
       updatedDateTime = DateTimeUtils.syncDateTime(previousDateTime, date, this.timeValue, this.props.showSeconds);
 
+      if (previousDateTime.isDST() && previousDateTime.hours() === 3) {
+        if (!updatedDateTime.isDST() && (event.key === '-' || event.key === '_')) {
+          updatedDateTime.subtract(1, 'hours');
+        }
+
+        if (!previousDateTime.subtract(1, 'days').isDST() && updatedDateTime.isDST() && (event.key === '+' || event.key === '=')) {
+          updatedDateTime.subtract(1, 'hours');
+        }
+      }
+
       if (isTimeValid) {
         // Update the timeValue in case the updatedDateTime falls in the missing hour and needs to bump the hour up.
         this.timeValue = DateTimeUtils.getTime(updatedDateTime.format(), this.props.showSeconds, this.initialTimeZone);
--- a/packages/terra-date-time-picker/src/DateTimePicker.jsx
+++ b/packages/terra-date-time-picker/src/DateTimePicker.jsx
@@ -298,25 +298,28 @@ class DateTimePicker extends React.Component {
       const previousDateTime = this.state.dateTime ? this.state.dateTime.clone() : DateTimeUtils.createSafeDate(formattedDate, this.initialTimeZone);
       updatedDateTime = DateTimeUtils.syncDateTime(previousDateTime, date, this.timeValue, this.props.showSeconds);
 
+      // Conditions to subtract an hour when date is moved past the start of DST when navigating using keyboard shortcut keys.
       if (previousDateTime.isDST() && previousDateTime.hours() === 3) {
-        let flag = false;
+        let isHourSubtracted = false;
         if (!updatedDateTime.isDST()) {
           updatedDateTime.subtract(1, 'hours');
-          flag = true;
+          isHourSubtracted = true;
         }
 
         if (!previousDateTime.subtract(1, 'days').isDST() && updatedDateTime.isDST()) {
           updatedDateTime.subtract(1, 'hours');
-          flag = true;
+          isHourSubtracted = true;
         }
 
+        // Condition to add back an hour if initial time was 3AM and date is moved past start of DST.
         // eslint-disable-next-line no-underscore-dangle
-        if (previousDateTime && previousDateTime._a[3] === 3 && flag) {
+        if (previousDateTime && previousDateTime._a[3] === 3 && isHourSubtracted) {
           updatedDateTime.add(1, 'hours');
-          flag = false;
+          isHourSubtracted = false;
         }
       }
 
+      // Conditions to subtract an hour when date is moved past the start of DST when using keyboard to input (Numeric Keys and Delete).
       // eslint-disable-next-line no-underscore-dangle
       if (this.prevDateTime && this.prevDateTime._a[3] === 2 && this.prevDateTime.hours() === 3) {
         updatedDateTime.subtract(1, 'hours');
--- a/packages/terra-date-time-picker/src/DateTimePicker.jsx
+++ b/packages/terra-date-time-picker/src/DateTimePicker.jsx
@@ -311,14 +311,14 @@ class DateTimePicker extends React.Component {
         }
 
         // eslint-disable-next-line no-underscore-dangle
-        if (previousDateTime._a[3] === 3 && flag) {
+        if (previousDateTime && previousDateTime._a[3] === 3 && flag) {
           updatedDateTime.add(1, 'hours');
           flag = false;
         }
       }
 
       // eslint-disable-next-line no-underscore-dangle
-      if (this.prevDateTime._a[3] === 2 && this.prevDateTime.hours() === 3) {
+      if (this.prevDateTime && this.prevDateTime._a[3] === 2 && this.prevDateTime.hours() === 3) {
         updatedDateTime.subtract(1, 'hours');
       }
 
--- a/packages/terra-date-time-picker/src/DateTimePicker.jsx
+++ b/packages/terra-date-time-picker/src/DateTimePicker.jsx
@@ -197,6 +197,7 @@ class DateTimePicker extends React.Component {
     this.handleOnRequestClose = this.handleOnRequestClose.bind(this);
     this.dateTimePickerContainer = React.createRef();
     this.containerHasFocus = false;
+    this.prevDateTime = DateTimeUtils.createSafeDate(props.value, this.initialTimeZone);
   }
 
   componentDidMount() {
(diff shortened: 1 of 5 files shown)


packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -1442,6 +1442,30 @@ Terra.describeViewports('DateTimePicker', ['large'], () => {
             expect($('input[name="terra-time-hour-input"]')).toHaveValue('02');
             expect($('input[name="terra-time-minute-input"]')).toHaveValue('00');
           });
+
+          it('should should show correct time when moved over the start date of  DST', () => {
```
sdadn, line 1446:
```suggestion
          it('should show the correct time when the day moves past the start date of  DST', () => {
```

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -297,6 +297,16 @@ class DateTimePicker extends React.Component {
       const previousDateTime = this.state.dateTime ? this.state.dateTime.clone() : DateTimeUtils.createSafeDate(formattedDate, this.initialTimeZone);
       updatedDateTime = DateTimeUtils.syncDateTime(previousDateTime, date, this.timeValue, this.props.showSeconds);
 
+      if (previousDateTime.isDST() && previousDateTime.hours() === 3) {
+        if (!updatedDateTime.isDST() && (event.key === '-' || event.key === '_')) {
```
benbcai, line 301: 1. Do we need to also handle using the UP and DOWN arrow keys to increment and decrement the day?
2. How about the workflow where the user manually changes the date from 03/08/2020 to 03/09/20200? Should the hour also change back from `03` back to `02`

For example:
Step 1: date/time is `03/09/2020 02:30`
Step 2: Manually change only the day from `03/09/2020` to `03/08/2020`. The hour is auto updated from `02` to `03`.
Step 3: Manually change only the day back to `03/09/2020`. Shouldn't the hour also get reverted back to `02`?

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -1442,6 +1442,30 @@ Terra.describeViewports('DateTimePicker', ['large'], () => {
...
+            $('input[name="terra-date-day-input"]').click();
+            browser.keys('+');
+            browser.keys('-');
+
+            expect($('input[name="terra-date-year-input"]')).toHaveValue('2020');
+            expect($('input[name="terra-date-month-input"]')).toHaveValue('03');
+            expect($('input[name="terra-date-day-input"]')).toHaveValue('07');
+            expect($('input[name="terra-time-hour-input"]')).toHaveValue('02');
+            expect($('input[name="terra-time-minute-input"]')).toHaveValue('30');
+          });
```
supreethmr, line 1468: can we also have screenshot captured for this scenario to show up the updated time values.

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -1442,6 +1442,30 @@ Terra.describeViewports('DateTimePicker', ['large'], () => {
...
+            browser.url('/raw/tests/cerner-terra-framework-docs/date-time-picker/date-time-picker-default-dst');
+            browser.refresh();
+            $('input[name="terra-date-year-input"]').click();
+            browser.keys('2020');
+            $('input[name="terra-date-month-input"]').click();
+            browser.keys('03');
+            $('input[name="terra-date-day-input"]').click();
+            browser.keys('07');
+            $('input[name="terra-time-hour-input"]').click();
+            browser.keys('02');
```
supreethmr, line 1456:
```suggestion
            browser.keys('03');
```
Can we keep time as 3:30 and validate upon `+` key press validate the missing hour.

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -1442,6 +1442,30 @@ Terra.describeViewports('DateTimePicker', ['large'], () => {
...
+            $('input[name="terra-time-minute-input"]').click();
+            browser.keys('30');
+            $('input[name="terra-date-day-input"]').click();
+            browser.keys('+');
```
supreethmr, line 1460: can we validate the time values for `+` key press and then for `-` key press to ensure the same behavior ( missing hour is updated ) for both key interactions.

if there are no test written for `up` and `down` arrow key we would need test case for  ⬆️ and ⬇️ key.

packages/terra-date-time-picker/src/DateTimePicker.jsx
supreethmr: > How about the workflow where the user manually changes the date from 03/08/2020 to 03/09/20200? Should the hour also change back from 03 back to 02

Since it works ( resets time to default value 02:30 on changing date using shortcut keys ) when provided the initial date that as missing hour for example : `const initialValue = moment().tz('America/Chicago').format('2020-03-08T02:30:00');` I think it should behave same when user manually enters a date that as missing hour.

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
saket2403: Missing Hour is 2AM-3AM so any time with hour as 2 will change to 3 and then back to 2 for all other days.

supreethmr: Sorry I meant if we can test like this lets input `03/07/2020 T02:30` and  do `+` on day input and validate if time has changed from `02:30` to `03:30` . Same for up and down arrow.

supreethmr (approved):

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -297,6 +297,16 @@ class DateTimePicker extends React.Component {
       const previousDateTime = this.state.dateTime ? this.state.dateTime.clone() : DateTimeUtils.createSafeDate(formattedDate, this.initialTimeZone);
       updatedDateTime = DateTimeUtils.syncDateTime(previousDateTime, date, this.timeValue, this.props.showSeconds);
 
+      if (previousDateTime.isDST() && previousDateTime.hours() === 3) {
+        if (!updatedDateTime.isDST()) {
+          updatedDateTime.subtract(1, 'hours');
```
benbcai, line 302: It appears that one hour is subtracted from the time when it should not if the hour is at `03` and changing the date to date light savings. For example.

1. Default date/time is `03/09/2020 03:57`.
2. Change the date to `03/08/2020` The time correctly remains unchanged at `03:57` because it is not ambiguous.
3. Increment or decrement the date again to `03/07/2020` or `03/09/2020`. Notice that the hour is incorrectly adjusted to `02:57` when it is expected to remain `03:57`.


Sep-12-2022 12-05-35

neilpfeiffer: Good catch @benbcai - the component should not auto-adjust for values that did not qualify for missing hour, like being set to any 3am value.

saket2403: Fixed [24a4450](https://github.com/cerner/terra-framework/pull/1567/commits/24a4450b2b1795629270c352acdb666c572a7897)

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -297,6 +298,30 @@ class DateTimePicker extends React.Component {
...
+
+        // eslint-disable-next-line no-underscore-dangle
+        if (previousDateTime && previousDateTime._a[3] === 3 && flag) {
+          updatedDateTime.add(1, 'hours');
+          flag = false;
+        }
+      }
+
+      // eslint-disable-next-line no-underscore-dangle
+      if (this.prevDateTime && this.prevDateTime._a[3] === 2 && this.prevDateTime.hours() === 3) {
```
benbcai, line 321: Could we use `previousDateTime` instead of needing to declare `this.prevDateTime`?

This new logic seems confusing to follow where we potentially subtract an hour from updatedDateTime a couple times and then potentially add the hour back. Adding some code comments might help for readability.

saket2403: When clearing the date value using Backspace the time value gets cleared out entirely from the `previousDateTime` object. So `previousDateTime,hours` is 0. Hence added the new variable. I'll add comments

neilpfeiffer:
Tested: PR #1567
Resolves: Issue #1554 and UXPLATFORM-6643

**Reviewed:**
- Resolves DST bug created in PR #1520 when shortcut keys were added to date-time-picker, as detailed on issue #1554 and https://github.com/cerner/terra-framework/pull/1520#pullrequestreview-923848063 and includes suggested interactions for how to handle.

\
**Fix with Followup Issue:**
- With this resolution, it is clear after testing the interactions that consumers would benefit from a followup enhancement to bring additional visibility and awareness to users by alerting them the situation is occurring, and how the date-time-picker has auto-handled the situation for them (and for users to double-check if that automatic handling is correct or fix if incorrect). The exact UI behaviors and details have not been designed, so suggestion is to work with UX to determine best approach. Enhancement request UXPLATFORM-7557 has been logged to track this work.



 **UX Reviewed** |
-----------|

PR #1621, opened 2023-03-29
[terra-slide-panel-manager] Allow consuming teams ability to define heading levels

artpark:
### Summary
Added a `level` prop to `terra-slide-panel-manager` as a passthrough to the underlying `terra-action-header` component. Screen readers should now pick up the correct heading level according to what level is passed. Level changes are not visual. 

### Testing
Jest tests. WDIO tests are not necessary because there are no visual changes.


packages/terra-slide-panel-manager/src/SlidePanelManager.jsx
```diff
@@ -15,6 +15,11 @@ const propTypes = {
    * disclosure capabilities through the DisclosureManger's context API.
    */
   children: PropTypes.node,
+  /**
+   * Sets the heading level of the disclosed panel. One of `1`, `2`, `3`, `4`, `5`, `6`. This helps screen readers to announce appropriate heading levels.
+   * Changing 'level' will not visually change the style of the content.
+   */
+  level: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
```
eawww, line 22: Since `level` is set on the SlidePanelManager, if a consumer wanted to disclose a slide panel of level 2 and then disclose a second slide panel of level 3 from within the level 2 one, would they be able to do that?

packages/terra-slide-panel-manager/src/SlidePanelManager.jsx
```diff
@@ -31,6 +36,7 @@ const propTypes = {
 };
 
 const defaultProps = {
+  level: 1,
```
eawww, line 39: Default should be 2 or 3. Since it's within the same context as the main page, it should be a subheading of the main page, which should only have one h1.

sycombs (approved):

packages/terra-slide-panel-manager/src/SlidePanelManager.jsx
artpark: The header is shared across multiple slides, so a consumer would not be able to do that, only one level is allowed

packages/terra-slide-panel-manager/src/SlidePanelManager.jsx
artpark: Changed to default level 2 [ca3f033](https://github.com/cerner/terra-framework/pull/1621/commits/ca3f0331cd8ced41a275a64b33cc4a1fc91ec6ee)

packages/terra-slide-panel-manager/src/SlidePanelManager.jsx
eawww: That might be an issue. BRB bugging some people.

eawww: Talked to Scott and he said it's cool!

eawww (approved):
🤘

PR #1638, opened 2023-04-17
[terra-slide-panel] Contain focus in panel content when fullscreen

smason0:
### Summary

Changes how SlidePanel handles tab order when the panel content is open and fullscreen. Prevents main content (under panel content) from being focused when panel content is fullscreen either via `isFullscreen` prop or a narrow viewport size.

### Testing

Tested manually in Chrome/Firefox/Safari on MacOS, and in Edge on Windows 10.

### Additional Details

Preview: https://engineering.cerner.com/terra-framework/pull/1638

@cerner/terra

### DIFF
--- a/packages/terra-slide-panel/CHANGELOG.md
+++ b/packages/terra-slide-panel/CHANGELOG.md
@@ -5,6 +5,9 @@
 * Added
   * Added roles and visually hidden text to slide panel.
 
+* Changed
+  * Updated main content to not be focus-able when panel content is fullscreen.
+
 ## 3.38.1 - (April 12, 2023)
 
 * Fixed
--- a/packages/terra-slide-panel/CHANGELOG.md
+++ b/packages/terra-slide-panel/CHANGELOG.md
@@ -6,7 +6,7 @@
   * Added roles and visually hidden text to slide panel.
 
 * Changed
-  * Updated main content to not be focus-able when panel content is fullscreen.
+  * Updated main content to not be focusable when panel content is fullscreen.
 
 ## 3.38.1 - (April 12, 2023)
 
--- a/packages/terra-slide-panel/src/SlidePanel.module.scss
+++ b/packages/terra-slide-panel/src/SlidePanel.module.scss
@@ -77,9 +77,8 @@
     }
   }
 
-  // Focus trap
+  // Remove main content from tab order when it's covered by slide panel on narrow screen
   .is-open {
-    // Removes main content from tab access when it's covered by slide panel on narrow screen
     // max-width is set to 1 pixel less than "min-width: 768px" breakpoint to avoid the main content being invisible at width 768px when the panel has already moved to the side, as both max-width and max-width include that width
     @media (max-width: 767px), print {
       > .main {
--- a/packages/terra-slide-panel/src/SlidePanel.module.scss
+++ b/packages/terra-slide-panel/src/SlidePanel.module.scss
@@ -77,6 +77,23 @@
     }
   }
 
+  // Focus trap 
+  .is-open {
+    // Removes main content elements from tab order as slide panel lays over them on narrow screen
+    @media (min-width: 0px), print {
+      > .main {
+        visibility: hidden;
+      } 
+    }
+
+    // Returns main content elements to tab order
+    @media (min-width: 768px), print {
+      > .main {
+        visibility: visible;
+      }
+    }
+  }
+
   // Overlay
   /* stylelint-disable max-nesting-depth */
   .is-open[data-slide-panel-panel-behavior='overlay'] {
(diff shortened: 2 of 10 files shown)


packages/terra-slide-panel/src/SlidePanel.module.scss
```diff
@@ -77,6 +77,17 @@
     }
   }
 
+  // Focus trap
```
trandrew1023, line 80: Since the user isn't necessarily trapped within the slide panel, maybe this could  say something like "Hide main"

adoroshk: Replaced "Focus trap" with a description: "Remove main content from tab order when it's covered by slide panel on narrow screen"

packages/terra-slide-panel/tests/jest/SlidePanel.test.jsx
artpark, line 1: How come the tests have switched to using shallow render instead?

packages/terra-slide-panel/tests/wdio/slide-panel-spec.js
```diff
@@ -53,39 +53,39 @@ Terra.describeViewports('Slide panel', ['large'], () => {
...
-      $('#test-slide [aria-hidden="false"]').waitForExist();
+      $('[aria-label="Panel content area"]').waitForExist();
       browser.pause(150);
 
       browser.keys(['Tab']);
-      $('#focus-button').isFocused();
+      expect($('#focus-button').isFocused()).toBeTruthy();
       browser.keys(['Enter']);
-      $('#test-slide [aria-hidden="true"]').waitForExist();
+      $('[aria-label="Main content area"]').waitForExist();
```
trandrew1023, line 86: I wonder if this one should still be "#test-slide [aria-hidden="true"]" since the main content should already exist (or is this waitForExist even needed since the main content exists?)

smason0: Good point. I think I need to test that the panel is closed a different way: `[aria-label="Panel content area"][aria-hidden="true"]` I don't want to just rely on the aria-hidden, because the changes in this story allow the main content to also be given aria-hidden="true", which could get ambiguous with how many checks we do throughout the test. Will make that change where needed

trandrew1023: Gotcha that makes sense, I just realized "test-slide" is the entire main/slide panel component is not just the panel div itself. Checking the panel is closed using aria-hidden = true looks good to me 👍

sycombs:
Can we also add a WDIO test for the new behavior to see if it does not tab out of the fullscreen slide panel? Maybe something like checking if an element outside the slide panel does not receive focus.

smason0:
> Can we also add a WDIO test for the new behavior to see if it does not tab out of the fullscreen slide panel? Maybe something like checking if an element outside the slide panel does not receive focus.

Sure, I can add a test to make sure the main content doesn't receive focus when tabbing from the fullscreen panel

packages/terra-slide-panel/tests/jest/SlidePanel.test.jsx
sdadn: I think that's a good call and makes the tests more robust.

packages/terra-slide-panel/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 * Added
   * Added roles and visually hidden text to slide panel.
 
+* Changed
+  * Updated main content to not be focus-able when panel content is fullscreen.
```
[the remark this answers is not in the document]
smason0: Fixed: https://github.com/cerner/terra-framework/pull/1638/commits/b8bc458138a6a53820cf661e479db32fe3c68529

smason0:
> > Can we also add a WDIO test for the new behavior to see if it does not tab out of the fullscreen slide panel? Maybe something like checking if an element outside the slide panel does not receive focus.
> 
> Sure, I can add a test to make sure the main content doesn't receive focus when tabbing from the fullscreen panel

Done: https://github.com/cerner/terra-framework/pull/1638/commits/0a9cf583c04f020462da26ed7240bbb00389ef0a

sycombs (approved):

sdadn (approved):

eawww (approved):

mjpalazzo:
Steven and Alla walked me through the enhancements to Slide Panel in Full Screen mode. I could not use Shift-Tab to navigate to main content when Slide Panel is full screen. I also tested with the browser window size decreased so other slide panel sizes are full screen. When I was on an interactive control in the Slide Panel and pressed Shift-Tab, focus went to the prior region outside of the Slide Panel and out of the Main Content. I compared this to current behavior where Shift-Tab brought focus to controls in the Main Content. Great job! This change makes keyboard navigation compliant with accessibility guidelines for this control.

PR #1646, opened 2023-04-20
[terra-modal-manager] Changed modal name to be the same as h1 title of the modal dialog

KoushikKalli:
### Summary
When entering a modal, the h1 title of the modal should be read by the screen reader. 

**What was changed:**
The required behavior was achieved by introducing an additional non-hidden empty span with tabIndex=-1, data-terra-abstract-modal-begin attribute and assign focus on dialog opening.

### Testing

This change was tested using:
- [x] Jest
- [x] WDIO

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

**This PR resolves:**

UXPLATFORM-8733

---

@cerner/terra

### DIFF
--- a/packages/terra-abstract-modal/CHANGELOG.md
+++ b/packages/terra-abstract-modal/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## Unreleased
 
-  * Changed
-    * Updated to read the title of the modal by the screen reader when a modal is opened.
+  * Added
+    * Added screen reader support to read the title of the modal when it is open.
 
 ## 3.39.0 - (February 14, 2023)
 
--- a/packages/terra-abstract-modal/CHANGELOG.md
+++ b/packages/terra-abstract-modal/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+  * Changed
+    * Updated to read the title of the modal by the screen reader when a modal is opened.
+
 ## 3.39.0 - (February 14, 2023)
 
 * Changed
--- a/packages/terra-abstract-modal/src/_ModalContent.jsx
+++ b/packages/terra-abstract-modal/src/_ModalContent.jsx
@@ -138,9 +138,8 @@ const ModalContent = forwardRef((props, ref) => {
         ref={ref}
       >
         {
-          /* UXPLATFORM-8733: Adding an empty span with data-terra-abstract-modal-begin attribute
-             to receive focus when the dialog is opened.
-          */
+          // This empty span with data-terra-abstract-modal-begin attribute
+          //  receives focus when the dialog is opened.
         }
         <span data-terra-abstract-modal-begin tabIndex="-1" />
         <FormattedMessage id="Terra.AbstractModal.BeginModalDialog">
--- a/packages/terra-abstract-modal/src/_ModalContent.jsx
+++ b/packages/terra-abstract-modal/src/_ModalContent.jsx
@@ -137,6 +137,12 @@ const ModalContent = forwardRef((props, ref) => {
         role={role}
         ref={ref}
       >
+        {
+          /* UXPLATFORM-8733: Adding an empty span with data-terra-abstract-modal-begin attribute
+             to receive focus when the dialog is opened.
+          */
+        }
+        <span data-terra-abstract-modal-begin tabIndex="-1" />
         <FormattedMessage id="Terra.AbstractModal.BeginModalDialog">
           {text => {
             // In the latest version of react-intl this param is an array, when previous versions it was a string.
--- a/packages/terra-abstract-modal/src/inertHelpers.js
+++ b/packages/terra-abstract-modal/src/inertHelpers.js
@@ -17,8 +17,7 @@ function showModalDomUpdates(modalElement, rootSelector) {
 
     mainDocumentElement.setAttribute('data-abstract-modal-overlay-count', `${dataOverlayCount + 1}`);
 
-    // Handle focus shift for VoiceOver on iOS
-    if ('ontouchstart' in window) {
+    if (modalElement.querySelector('[data-terra-abstract-modal-begin]')) {
       modalElement.querySelector('[data-terra-abstract-modal-begin]').focus();
     } else {
       // Shift focus to modal dialog
--- a/packages/terra-modal-manager/tests/wdio/modal-manager-spec.js
+++ b/packages/terra-modal-manager/tests/wdio/modal-manager-spec.js
@@ -316,13 +316,13 @@ Terra.describeViewports('ModalManager - Behaviors', ['large'], () => {
       it(' is on the first element when modal when opened', () => {
         $('#root-component .disclose-small').click();
         $('[class*="slide-group"] #DemoContainer-1 .maximize').waitForDisplayed({ timeout: 1000 });
-        
-        const firstElementWithModalBeginAttribute =  $('[aria-modal="true"][role="dialog"] [data-terra-abstract-modal-begin="true"]');
+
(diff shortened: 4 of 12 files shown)


packages/terra-abstract-modal/src/_ModalContent.jsx
```diff
@@ -137,6 +137,12 @@ const ModalContent = forwardRef((props, ref) => {
         role={role}
         ref={ref}
       >
+        {
+          /* UXPLATFORM-8733: Adding an empty span with data-terra-abstract-modal-begin attribute
```
trandrew1023, line 141: I think we can remove the internal identifier here

KoushikKalli: Since we have not concretely landed on not including JIRA's, we decided to include it as it could be useful for someone internal in triaging especially because we are introducing a new element in the component DOM.

packages/terra-modal-manager/tests/wdio/modal-manager-spec.js
```diff
@@ -313,10 +313,16 @@ Terra.describeViewports('ModalManager - Behaviors', ['large'], () => {
     });
 
     describe('Modal Focus', () => {
-      it('focuses on the modal when opened', () => {
+      it(' is on the first element when modal when opened', () => {
```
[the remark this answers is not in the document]
KoushikKalli: Removed.

packages/terra-abstract-modal/src/inertHelpers.js
```diff
@@ -17,8 +17,7 @@ function showModalDomUpdates(modalElement, rootSelector) {
 
     mainDocumentElement.setAttribute('data-abstract-modal-overlay-count', `${dataOverlayCount + 1}`);
 
-    // Handle focus shift for VoiceOver on iOS
-    if ('ontouchstart' in window) {
```
sycombs, line 21: What was the reason this was removed?

packages/terra-modal-manager/tests/wdio/modal-manager-spec.js
```diff
@@ -313,10 +313,16 @@ Terra.describeViewports('ModalManager - Behaviors', ['large'], () => {
...
 
     describe('Modal Focus', () => {
-      it('focuses on the modal when opened', () => {
+      it('is on the first element when modal when opened', () => {
         $('#root-component .disclose-small').click();
         $('[class*="slide-group"] #DemoContainer-1 .maximize').waitForDisplayed({ timeout: 1000 });
-        expect($('[aria-modal="true"][role="dialog"]').isFocused()).toEqual(true);
+
+        const firstElementWithModalBeginAttribute = $('[aria-modal="true"][role="dialog"] [data-terra-abstract-modal-begin="true"]');
+        expect(firstElementWithModalBeginAttribute.getText() === '').toEqual(true);
```
sycombs, line 321: Would something like this work?
```suggestion
        expect(firstElementWithModalBeginAttribute.getText()).toBe('');
```

packages/terra-abstract-modal/src/_ModalContent.jsx
trandrew1023: Gotcha, I'm fine with that if others agree thats ok since it could be helpful. The reason I was thinking we shouldn't add this is because it wouldn't follow existing Terra documentation given I haven't found Jira numbers in the code anywhere. If anything, the comment is very descriptive so the reason is clear enough and git blame/history should be able to link anyone needing to audit/review the reason can trace it back to this PR which can include the Jira number. @artpark/@sdadn would having the Jira number be ok or is this something that is avoided in the source code?

sycombs: Given that the Jira number is specified in the PR description, would that not give sufficient context as to when/why the change was made?

sdadn: The JIRA is already included in the PR, which can easily be found using the git blame history. I don't think there's much value in adding the JIRA number here. A comment is more than sufficient to convey this information.

packages/terra-abstract-modal/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+  * Changed
+    * Updated to read the title of the modal by the screen reader when a modal is opened.
```
sdadn, line 6:
```suggestion
  * Added
    * Added screen reader support to read the title of the modal when it is open.
```

sdadn (changes requested):

packages/terra-modal-manager/tests/wdio/modal-manager-spec.js
```diff
@@ -313,10 +313,16 @@ Terra.describeViewports('ModalManager - Behaviors', ['large'], () => {
     });
 
     describe('Modal Focus', () => {
-      it('focuses on the modal when opened', () => {
+      it(' is on the first element when modal when opened', () => {
         $('#root-component .disclose-small').click();
         $('[class*="slide-group"] #DemoContainer-1 .maximize').waitForDisplayed({ timeout: 1000 });
-        expect($('[aria-modal="true"][role="dialog"]').isFocused()).toEqual(true);
+
+        const firstElementWithModalBeginAttribute = $('[aria-modal="true"][role="dialog"] [data-terra-abstract-modal-begin="true"]');
```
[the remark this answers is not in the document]
KoushikKalli: The space between dialog & data-terra-abstract-modal-begin is intentional. 

For example, $('[a][b]') returns the element with both _a_ & _b_ whereas, $('[a] [b]') first finds element matching _a_ and then looks for a child element of _a_ matching _b_.

packages/terra-abstract-modal/src/_ModalContent.jsx
KoushikKalli: I was thinking more about when all these changes are squashed and merged into main and if we need to review a year from now. That said, we are good with removing the JIRA reference.

trandrew1023 (approved):

packages/terra-modal-manager/tests/wdio/modal-manager-spec.js
```diff
@@ -313,10 +313,16 @@ Terra.describeViewports('ModalManager - Behaviors', ['large'], () => {
...
+      it('is on the first element when modal when opened', () => {
         $('#root-component .disclose-small').click();
         $('[class*="slide-group"] #DemoContainer-1 .maximize').waitForDisplayed({ timeout: 1000 });
-        expect($('[aria-modal="true"][role="dialog"]').isFocused()).toEqual(true);
+
+        const firstElementWithModalBeginAttribute = $('[aria-modal="true"][role="dialog"] [data-terra-abstract-modal-begin="true"]');
+        expect(firstElementWithModalBeginAttribute.getText()).toBe('');
+
+        const expectedParentElement = $('[aria-modal="true"][role="dialog"]');
+        expect(firstElementWithModalBeginAttribute.parentElement().getHTML() === expectedParentElement.getHTML()).toEqual(true);
```
sdadn, line 324: Can this also be updated to:
```suggestion
        expect(firstElementWithModalBeginAttribute.parentElement().getHTML).toBe(expectedParentElement.getHTML());
```

KoushikKalli: @sycombs & @sdadn - Good suggestion with the toBe() matcher. A new thing that we learned. 

This suggestion is missing parentheses. So, I made a manual commit.

sdadn: Oops, I missed a few in my suggestion. Thanks for doing the manual commit!

PR #1649, opened 2023-04-26
[terra-menu] Added showHeader prop

sugan2416:
### Summary

**What was changed:**

- Added `showHeader` prop to toggle header display on/off
- Fixed header context for JAWS

**Why it was changed:**

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-XXXX 

---

@cerner/terra


supreethmr:
> Always display header for sub-menu 

Does this mean the existing menu does not displays header for sub-menu items now ??

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -17,6 +17,7 @@ import styles from './Menu.module.scss';
 
 const cx = classNames.bind(styles);
 const menuHeaderId = `terra-menu-headertitle-${uuidv4()}`;
+const menuTopHeaderId = `terra-menu-headertitle-${uuidv4()}`;
```
supreethmr, line 20:
```suggestion
const menuTopHeaderId = `terra-menu-headertitle-${uuidv4()}`;
```
since only one header is displayed at any given moment of time `menuHeaderId` can be used for both header !!

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -380,6 +381,7 @@ class MenuContent extends React.Component {
               totalItems,
               itemIndex,
               intl: this.props.intl,
+              'aria-describedby': !MenuUtils.isMac() && !this.props.index && this.props.showHeader && index === 0 ? menuTopHeaderId : undefined,
```
supreethmr, line 384: why do we need aria-described by only for Mac ?

sugan2416:
> > Always display header for sub-menu
> 
> Does this mean the existing menu does not displays header for sub-menu items now ??

Existing menu also always displays header for sub-menu, this was just mentioned as a functional behaviour (no change was done for this). Will remove this line to avoid confusion

packages/terra-menu/src/_MenuContent.jsx
sugan2416: This was added (for JAWS) to fix the issue where the header was not announced when main-menu is opened. In case of VO, it does get announced but for jaws, when the focus is on first item, it fails to provide header context.
 Hence have added this for only main-level menu and for the first focusable item only.

packages/terra-menu/src/_MenuContent.jsx
sugan2416: https://github.com/cerner/terra-framework/pull/1649/commits/f3c198c07d05ed1fe309e66c569eac8d214ed205

ShettyAkarsh (approved):

supreethmr (approved):

PR #1656, opened 2023-05-15
UXPLATFORM-8912: Introduction of WorklistDataGrid component

KoushikKalli:
### Summary
Introduced a new component - Worklist data grid. The Terra Worklist Data Grid is a container that enables users to navigate the grid information using directional navigation keys.

### Testing
This change was tested using:

- [x] WDIO

### Reviews

- [x] UX review
- [x] Accessibility review
- [x] Functional review

**This PR resolves:**

UXPLATFORM-8912

---

@cerner/terra


packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+  columnHeaders: WorklistDataGridPropTypes.columnHeaderShape,
+  /**
+   * Data for content in the body of the Grid. Rows will be rendered in the order given.
+   */
+  rows: PropTypes.arrayOf(WorklistDataGridPropTypes.rowShape),
+  /**
+   * String that specifies the default column width. Any valid CSS width value is accepted.
+   * This value will be used if no overriding width value is provided on a per-column basis.
+   */
+  columnHeaderDefaultWidth: PropTypes.string,
```
cm9361, line 37: Wouldn't we expect the columns to specify their width?  If a width is not provided, I would think that it would be a constant value and not an accepted prop.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+  /**
+   * String that specifies the default row height. Any valid CSS width value is accepted.
+   * This value will be used if no overriding height value is provided on a per-row basis.
+   */
+  rowDefaultHeight: PropTypes.string,
```
cm9361, line 42: This should probably just be rowHeight.  The framework would determine this value and pass the appropriate height.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+  /**
+   * String that specifies the column height. Any valid CSS height value is accepted.
+   */
+  columnHeaderHeight: PropTypes.string,
```
cm9361, line 46: It looks like the columnHeaderShape also specified the column header height.  Do we need to remove the value from one of the props?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+   * Number indicating the index of the column that represents row header. Index is 0 based and cannot exceed one less than the number of columns in the grid.
+   */
+  rowHeaderIndex: PropTypes.number,
+};
+
+const defaultProps = {
+  rowHeaderIndex: 0,
+  columnHeaderDefaultWidth: '200px',
+  columnHeaderHeight: '2.5rem',
+  rowDefaultHeight: '2.5rem',
```
cm9361, line 57: How did we identify this value for the default heights for the column header and row height?  How many lines does this value allow?  Also, does the default column header height account for the potential error icon or sort indicator?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+
+const propTypes = {
+  /**
+   * String representing an attribute which identifies the element (or elements) that labels the grid.
+   */
+  ariaLabelledby: PropTypes.string,
+  /**
+   * String that labels the grid for accessibility.
+   */
+  ariaLabel: PropTypes.string,
```
cm9361, line 19: Do we need both of these attributes?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+    super(props);
+
+    this.handleClick = this.handleClick.bind(this);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+    this.focussedRow = 0;
+    this.focussedCol = 0;
+  }
+
+  componentDidMount() {
+    this.grid = document.querySelector(`#${this.props.id}`);
```
cm9361, line 71: We should use document.getElementById if we are using an id value.  We would not use querySelector in this scenario.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+    const focussedCell = this.grid.rows[this.focussedRow].cells[this.focussedCol];
+    focussedCell.setAttribute('tabIndex', '0');
+
+    // add event listeners to handle user key events in the grid.
+    this.grid.addEventListener('keydown', this.handleKeyDown);
+    this.grid.addEventListener('click', this.handleClick);
+  }
+
+  handleKeyDown(event) {
+    if (event.target?.parentNode?.parentNode?.parentNode?.id !== this.props.id) {
```
cm9361, line 82: Looking at this code, it almost seems like the event listener is at the wrong level?  Do we need to allow event bubbling to simplify this code?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+    const height = columnData.height || this.props.columnHeaderHeight;
+    const columnHeaderClassName = classNames(cx('worklist-data-grid-column-header'));
+    return (
+      /* eslint-disable react/forbid-dom-props */
+      <th key={columnData.id} id={columnData.id} role="columnheader" className={columnHeaderClassName} tabIndex="-1" style={{ width, height }}>{columnData.displayName}</th>
+    );
+  }
+
+  buildSingleRow(row, rowIndex) {
+    const backgroundClassName = classNames(cx({ striped: !!(rowIndex % 2) }));
```
cm9361, line 190: Can't we use pseudo selectors to make this logic much easier?  We could use even or odd (whichever is correct)

https://www.w3.org/Style/Examples/007/evenodd.en.html

packages/terra-worklist-data-grid/src/WorklistDataGrid.module.scss
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  .worklist-data-grid-cell-data {
+    // Style for the table cell data
+    border: var(--terra-worklist-data-grid-cell-border, 1px solid #dedfe0);
+    height: 100%;
+    position: relative;
+  }
+
+  [role='grid'] .striped {
```
cm9361, line 43: Wouldn't we want this to be a descendant of the worklist class instead of using the grid role?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+      columnHeaderDefaultWidth,
+      ...customProps
+    } = this.props;
+
+    const theme = this.context;
+
+    const gridClassNames = classNames(cx('worklist-data-grid', theme.className), customProps.className);
+
+    return (
+      <div className={cx(['container'])}>
```
cm9361, line 222: Why do we need a div container element around the table element?  We could apply the necessary styles to the table directly, correct?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+    const width = columnData.width || this.props.columnHeaderDefaultWidth;
+    const height = columnData.height || this.props.columnHeaderHeight;
+    const columnHeaderClassName = classNames(cx('worklist-data-grid-column-header'));
+    return (
+      /* eslint-disable react/forbid-dom-props */
+      <th key={columnData.id} id={columnData.id} role="columnheader" className={columnHeaderClassName} tabIndex="-1" style={{ width, height }}>{columnData.displayName}</th>
+    );
+  }
+
+  buildSingleRow(row, rowIndex) {
```
cm9361, line 189: I would rename this method to simply "buildRow".  Also, is the row index important to this method?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+    const backgroundClassName = classNames(cx({ striped: !!(rowIndex % 2) }));
+    return (
+      <tr key={rowIndex} className={backgroundClassName} height={row.height || this.props.rowDefaultHeight}>
```
cm9361, line 192: We should only have to calculate the row height once, correct?  We should not have to determine the value for each table row as it should not change.

packages/terra-worklist-data-grid/src/WorklistDataGrid.module.scss
```diff
@@ -0,0 +1,47 @@
+// Themes
+@import './orion-fusion-theme/WorklistDataGrid.module';
+@import './clinical-lowlight-theme/WorklistDataGrid.module';
+
+:local {
+  .container {
+    height: 100%;
+  }
+
+  [role='grid'] [tabIndex='0']:focus {
```
cm9361, line 10: Wouldn't we want to use the worklist data grid class for the selector?

packages/terra-worklist-data-grid/src/proptypes/columnHeaderCellShape.js
```diff
@@ -0,0 +1,18 @@
...
+   */
+  id: PropTypes.string.isRequired,
+  /**
+   * String of text to render within the column header cell.
+   */
+  displayName: PropTypes.string,
+  /**
+   * Boolean value indicating whether or not the cell can accept focus.
+   */
+  acceptsFocus: PropTypes.bool,
```
cm9361, line 15: Is there a reason that a column headers would not accept focus?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+        break;
+      case KeyCode.KEY_DOWN:
+        nextRow += 1;
+        break;
+      case KeyCode.KEY_LEFT:
+        nextCol -= 1;
+        break;
+      case KeyCode.KEY_RIGHT:
+        nextCol += 1;
+        break;
```
cm9361, line 105: It appears that some of these cases can be combined because they perform the same action.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+    this.focussedRow = 0;
+    this.focussedCol = 0;
+  }
+
+  componentDidMount() {
+    this.grid = document.querySelector(`#${this.props.id}`);
+
+    const focussedCell = this.grid.rows[this.focussedRow].cells[this.focussedCol];
+    focussedCell.setAttribute('tabIndex', '0');
```
cm9361, line 74: Can't you use .tabIndex directly instead of using a setAttribute call?  Same comment for other areas of the code.  The tabIndex property is a global property.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+      return;
+    }
+    if (nextCol < 0 || nextRow < 0) {
+      event.preventDefault();
+      return;
+    }
+
+    // Remove the ability to receive visible focus as only one cell should be able to receive tab focus.
+    // This is also how tabbing back to the grid ensures the grid last cell with the focus will be re-entered.
+    this.grid.rows[currentFocussedRow].cells[currentFocussedCol].setAttribute('tabIndex', '-1');
```
cm9361, line 132: I believe you should be able to set the tabIndex property directly instead of using the setAttribute method.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+    this.focussedRow = event.target.parentElement.rowIndex;
+    this.focussedCol = event.target.cellIndex;
+    const clickedCell = event.target;
+    clickedCell.setAttribute('tabIndex', '0');
+    clickedCell.focus();
+    event.preventDefault();
+  }
+
+  getNumberOfHeaderRows() {
+    return this.props.columnHeaders?.pinnedColumns?.length > 0 ? 1 : 0;
```
cm9361, line 159: Why would pinned columns be a property of the column headers?  Whether or not a column is pinned should be part of the columns themselves.

packages/terra-worklist-data-grid/src/proptypes/rowShape.js
```diff
@@ -0,0 +1,15 @@
+import PropTypes from 'prop-types';
+import cellShape from './cellShape';
+
+const rowShape = PropTypes.shape({
+  /**
+   * String that specifies the row height. Any valid CSS height value is accepted.
```
cm9361, line 6: Should each row have an id as exists in the current API?

packages/terra-worklist-data-grid/src/proptypes/columnHeaderShape.js
```diff
@@ -0,0 +1,15 @@
+import PropTypes from 'prop-types';
+import columnHeaderCellShape from './columnHeaderCellShape';
+
+const columnHeaderShape = PropTypes.shape({
```
cm9361, line 4: This should really be a column shape.

packages/terra-worklist-data-grid/src/proptypes/rowShape.js
```diff
@@ -0,0 +1,15 @@
...
+   */
+  height: PropTypes.string,
```
cm9361, line 8: The row height should be part of the main worklist.  It will not vary per row.  It would not be part of the row shape.

packages/terra-worklist-data-grid/src/proptypes/columnHeaderShape.js
```diff
@@ -0,0 +1,15 @@
...
+  /**
+   * String that specifies the column header height. Any valid CSS height value is accepted.
+   */
+  height: PropTypes.string,
```
cm9361, line 8: This property should be part of the main data grid.  It would not be part of the column header shape.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+};
+
+const defaultProps = {
+  rowHeaderIndex: 0,
+  columnHeaderDefaultWidth: '200px',
+  columnHeaderHeight: '2.5rem',
+  rowDefaultHeight: '2.5rem',
+};
+
+class WorklistDataGrid extends React.Component {
```
cm9361, line 60: If we are creating a new component, it should be a functional component instead of a class component.

cm9361:
Did UX specify that we need to add a row header capability to the worklist data grid?  To my knowledge, this is not a capability of either the MPage or Orion Worklist.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+    // Remove the ability to receive visible focus as only one cell should be able to receive tab focus.
+    // This is also how tabbing back to the grid ensures the grid last cell with the focus will be re-entered.
+    this.grid.rows[currentFocussedRow].cells[currentFocussedCol].setAttribute('tabIndex', '-1');
+
+    this.focussedRow = nextRow;
+    this.focussedCol = nextCol;
+
+    const elementWithTabStop = this.grid.rows[nextRow].cells[nextCol];
+    elementWithTabStop.setAttribute('tabIndex', '0');
+    elementWithTabStop.focus();
```
cm9361, line 139: Why do we have to manually handle the navigation.  Can't this be done using built-in browser capabilities and the tabIndex property?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+    event.preventDefault();
+  }
+
+  handleClick(event) {
+    // Is this event for something that is on the grid
+    if (event.target?.parentNode?.parentNode?.parentNode?.id !== this.props.id) {
```
cm9361, line 145: This logic makes me think that the event listener is at the wrong level.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+      return (
+        <tr height={this.props.columnHeaderHeight}>
+          {colHeader.pinnedColumns.map(columnData => (this.buildSingleColumnHeading(columnData)))}
+        </tr>
+      );
+    }
+    return undefined;
+  }
+
+  buildSingleColumnHeading(columnData) {
```
cm9361, line 179: It doesn't seem like we should need a separate method for building a column header.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+    const width = columnData.width || this.props.columnHeaderDefaultWidth;
+    const height = columnData.height || this.props.columnHeaderHeight;
+    const columnHeaderClassName = classNames(cx('worklist-data-grid-column-header'));
+    return (
+      /* eslint-disable react/forbid-dom-props */
```
cm9361, line 184: Why are we disabling the linter here?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+            {this.buildRows(rows)}
+          </tbody>
+        </table>
+      </div>
+    );
+  }
+}
+
+WorklistDataGrid.propTypes = propTypes;
+WorklistDataGrid.contextType = ThemeContext;
```
cm9361, line 241: We would want to use a functional component so that we are not locked in to a single context for this component.

packages/terra-worklist-data-grid/src/proptypes/columnHeaderCellShape.js
```diff
@@ -0,0 +1,18 @@
+import PropTypes from 'prop-types';
+
+const columnHeaderCellShape = PropTypes.shape({
+  /**
+   * String representing a unique identifier for the column header cell.
+   */
+  id: PropTypes.string.isRequired,
```
cm9361, line 7: Do the headers need identifiers?  What would be their purpose?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  getNumberOfHeaderRows() {
+    return this.props.columnHeaders?.pinnedColumns?.length > 0 ? 1 : 0;
+  }
+
+  getCellData(cell, cellColumnIndex) {
+    const cellClassName = { className: classNames(cx(this.props.rowHeaderIndex === cellColumnIndex ? 'worklist-data-grid-row-header' : 'worklist-data-grid-cell-data')) };
+    const tabIndex = { tabIndex: '-1' };
+    return this.props.rowHeaderIndex === cellColumnIndex ? (<th key={cellColumnIndex} {...tabIndex} {...cellClassName} role="rowheader">{cell.cellContent}</th>) : (<td key={cellColumnIndex} {...tabIndex} {...cellClassName}>{cell.cellContent}</td>);
```
cm9361, line 165: You might need to use something like react-html-parser (https://www.npmjs.com/package/react-html-parser) to render your cell content.  That way you can handle HTML fragments and plain text.  In the future there might need to be a separate strategy for handling cell content that is a React fragment.  I am not sure if there is something that can handle both.

packages/terra-worklist-data-grid/src/proptypes/cellShape.js
```diff
@@ -0,0 +1,10 @@
+import PropTypes from 'prop-types';
+
+const cellShape = PropTypes.shape({
+  /**
+   * String indicating content to be displayed for the cell.
+   */
+  text: PropTypes.string,
```
cm9361, line 7: Your shape does not match your example JSON data.  In the example JSON, this is called cellContent.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+   */
+  rowDefaultHeight: PropTypes.string,
+  /**
+   * String that specifies the column height. Any valid CSS height value is accepted.
+   */
+  columnHeaderHeight: PropTypes.string,
+  /**
+   * Number indicating the index of the column that represents row header. Index is 0 based and cannot exceed one less than the number of columns in the grid.
+   */
+  rowHeaderIndex: PropTypes.number,
```
cm9361, line 50: The Worklist Framework for MPages and Orion do not use the row header concept.  Talking with Andy Plumb, it appears that this concept is for compatibility with the flowsheet.  I totally understand that need.  However, how do we remove the row header for the Worklist in this implementation?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+    return undefined;
+  }
+
+  buildSingleColumnHeading(columnData) {
+    const width = columnData.width || this.props.columnHeaderDefaultWidth;
+    const height = columnData.height || this.props.columnHeaderHeight;
+    const columnHeaderClassName = classNames(cx('worklist-data-grid-column-header'));
+    return (
+      /* eslint-disable react/forbid-dom-props */
+      <th key={columnData.id} id={columnData.id} role="columnheader" className={columnHeaderClassName} tabIndex="-1" style={{ width, height }}>{columnData.displayName}</th>
```
cm9361, line 185: The id attribute must be unique for the entire webpage and is a global attribute. Do we need an id for the header element?  If so, we need to ensure that it is unique in all cases.  If you have two worklist data grids in a page for some reason, you DOM would be invalid.  You might need to also prepend the data grid id at a minimum to ensure uniqueness.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  getCellData(cell, cellColumnIndex) {
+    const cellClassName = { className: classNames(cx(this.props.rowHeaderIndex === cellColumnIndex ? 'worklist-data-grid-row-header' : 'worklist-data-grid-cell-data')) };
+    const tabIndex = { tabIndex: '-1' };
+    return this.props.rowHeaderIndex === cellColumnIndex ? (<th key={cellColumnIndex} {...tabIndex} {...cellClassName} role="rowheader">{cell.cellContent}</th>) : (<td key={cellColumnIndex} {...tabIndex} {...cellClassName}>{cell.cellContent}</td>);
+  }
+
+  buildColumnHeaders(colHeader) {
+    if (colHeader?.pinnedColumns?.length > 0) {
```
cm9361, line 169: Why do we need a check for pinned columns to render the column headers.  The column headers would not change based on being pinned, correct?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+
+  render() {
+    const {
+      id,
+      ariaLabelledby,
+      ariaLabel,
+      columnHeaders,
+      rows,
+      columnHeaderDefaultWidth,
+      ...customProps
```
cm9361, line 214: If we want to allow the class name to be set, we should make it a first class prop instead of leveraging custom props.  However, do consumers need the ability to set the class name on this component?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
cm9361, line 145: I get it now.  The logic is incorrect.  You should be using event.currentTarget instead of event.target.  You can read the documentation for context.  However, this logic will not work as the project gets more complex.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+
+  componentDidMount() {
+    this.grid = document.querySelector(`#${this.props.id}`);
+
+    const focussedCell = this.grid.rows[this.focussedRow].cells[this.focussedCol];
+    focussedCell.setAttribute('tabIndex', '0');
+
+    // add event listeners to handle user key events in the grid.
+    this.grid.addEventListener('keydown', this.handleKeyDown);
+    this.grid.addEventListener('click', this.handleClick);
```
cm9361, line 78: This is not the proper way to add events in React.  You would use the onClick attribute in the JSX for the click event and onKeyDown for the keydown event.  We should not need to add JavaScript event listeners and be able to leverage the React synthetic events.  It might change the implementation of the event handlers.

mjpalazzo:
Great job on this initial version of the Worklist Data Grid! I validated basic keyboard navigation with Tab and arrow keys. Visual focus moved as expected. 
I also validated with JAWS. JAWS announced the headers and cell contents as expected. I was able to navigate the cells and JAWS announced the cell contents. When I changed rows. JAWS announced the row header and then the value in the cell. Navigating horizontally, JAWS announced only the values in the cells only as expected.
I did not try to test any of the other JAWS keyboard shortcuts for grids. I believe these short cuts are in another story.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
cm9361, line 46: The header height is not really associated to each column.  I would suggest that we just have a headerHeight prop.

mjpalazzo:
I did additional testing with VoiceOver. VoiceOver works as expected. I could navigate the table with up down left and right arrow keys. VO announces the cell position in addition to the cell value. Control Option C announces the column header of the column of the cell in focus. Control Option R announces the header of the row header of the row of the cell in focus. Tabbing out of the table and then tabbing back in returns focus to the last cell in focus.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Renamed the method. RowIndex was used to provide a key for the row. But, as we added id for the row, we changed the key to use id and removed rowIndex.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: We believe the column will eventually need a unique id. However, in its current state, one is not necessary. So, we have removed it.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: We had to disable it because the linter expects us not to pass props to DOM nodes. But we needed to provide the style. Also, we had noticed that this particular rule has been disabled several times in our code base. So, we followed existing precedent.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: A separate method is not absolutely needed but it was added for better readability.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Pinned/overflow columns are not yet implemented. We have restructured the data to only consider columns at this time.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Agreed but we are only considering content type text in this first iteration.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+    clickedCell.focus();
+    event.preventDefault();
+  }
+
+  getNumberOfHeaderRows() {
+    return this.props.columnHeaders?.pinnedColumns?.length > 0 ? 1 : 0;
+  }
+
+  getCellData(cell, cellColumnIndex) {
+    const cellClassName = { className: classNames(cx(this.props.rowHeaderIndex === cellColumnIndex ? 'worklist-data-grid-row-header' : 'worklist-data-grid-cell-data')) };
```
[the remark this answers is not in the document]
KoushikKalli: Changed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Pinned/overflow columns are not yet implemented. We have restructured the data to only consider columns at this time.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+    elementWithTabStop.focus();
+    event.preventDefault();
+  }
+
+  handleClick(event) {
+    // Is this event for something that is on the grid
+    if (event.target?.parentNode?.parentNode?.parentNode?.id !== this.props.id) {
+      return;
+    }
+    // Remove Tab stop from previous cell in table that has focus and set it to the cell that was clicked.
```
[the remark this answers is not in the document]
KoushikKalli: Using the tabIndex, the browser’s built-in capabilities handle tabbing in and out of the grid. The other keys such as arrows, home and end, needed to be handled manually.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: 1. The grid does not handle many things in its current state and we expect this to evolve with future development.
2. The current logic was created to handle only when the cell was clicked and not an element in the cell. As only the entire cell is interactable at this time, we removed the check.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Using the tabIndex, the browser’s built-in capabilities handle tabbing in and out of the grid. The other keys such as arrows, home and end, needed to be handled manually.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Changed.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: We believe the cases are different.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: As we are dealing with only text in the cells, removed this logic now.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Modified to use OnClick & OnKeyDown

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Changed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Changed to use getElementById

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+};
+
+class WorklistDataGrid extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.handleClick = this.handleClick.bind(this);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+    this.focussedRow = 0;
+    this.focussedCol = 0;
```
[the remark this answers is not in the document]
KoushikKalli: We will make the change but long live our gracious king.... ;)

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: The current default values were placeholders we used during development. We removed them.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,244 @@
...
+  columnHeaderHeight: PropTypes.string,
+  /**
+   * Number indicating the index of the column that represents row header. Index is 0 based and cannot exceed one less than the number of columns in the grid.
+   */
+  rowHeaderIndex: PropTypes.number,
+};
+
+const defaultProps = {
+  rowHeaderIndex: 0,
+  columnHeaderDefaultWidth: '200px',
```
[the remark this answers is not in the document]
KoushikKalli: The current default values were placeholders we used during development. We removed them.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: @chrismichalewicz & @eawww - please provide your feedback on the following questions:
Do we support row header concept in the worklist data grid?
1. If yes, should the consumer be able to build a grid without the row header? Based on current story requirements, if a row header position is not provided, the worklist data grid treats the 1st column as the row header.
2. If yes, is there a use case for the row header to be anything other than the 1st column?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: As the grid has the concepts of column header and row header, we felt columnHeaderHeight makes it clearer for the consumer. But we changed it to columnHeight.

KoushikKalli: Agreed, removed height from columnHeaderShape.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Yes, columns can specify their widths. However, rather than forcing the consumer to specify a width for each column, we provide the concept of default column width. For example, if all columns have the same width, consumer could just specify the default rather than providing same width for each column. When any column needs to a different width compared to default, only that column will provide a value to override. That said, we changed the prop name to be ‘columnWidth’.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Yes, the consumer will have option to choose whichever one they prefer for their use case.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Since each row could potentially have a different height, we provided this prop to make it easier for the consumer to specify only 1 height and override it for row(s) requiring a different height. Similar to columnWidth, we changed the prop name to ‘rowHeight’.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Please see response in the question below.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Changed.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: @chrismichalewicz & @eawww will ultimately have to make this determination. For now, we think it's more flexible to give the consumer the ability to have different row heights.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: There is no need at this time. We will remove it.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Correct, removed the div container.

packages/terra-worklist-data-grid/src/WorklistDataGrid.module.scss
KoushikKalli: Changed to use the class instead of the role selector for the grid.

packages/terra-worklist-data-grid/src/WorklistDataGrid.module.scss
KoushikKalli: As we changed the implementation to even-odd selectors, this section the code became moot and is removed.

packages/terra-worklist-data-grid/src/proptypes/cellShape.js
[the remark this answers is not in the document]
KoushikKalli: That is correct and we share the same expectation. As this is first iteration of the grid, we kept it simple by using string. As we get more requirements from UX, we expect a separate story will be created to handle complex cell content.

KoushikKalli: Great catch, modified.

packages/terra-worklist-data-grid/src/proptypes/columnHeaderCellShape.js
KoushikKalli: For this 1st iteration, no. But we added them as they could be helpful in event handling when advanced features such as column sorting, resizing etc. are developed.

packages/terra-worklist-data-grid/src/proptypes/columnHeaderCellShape.js
KoushikKalli: In the event that headers provide no functionality, we do not see the need to give them focus. As we discussed, the determination to accept focus can be computed based on specific props. For example, a column that does not allow sorting or resizing need not accept focus. @chrismichalewicz & @eawww, please weigh in on this suggestion and include it for the story that deals with this functionality.

packages/terra-worklist-data-grid/src/proptypes/columnHeaderShape.js
KoushikKalli: These were initially separated as we expected different properties for a _th_ vs _td_. But because we don’t have that requirement right now, we will refactor & simplify.

packages/terra-worklist-data-grid/src/proptypes/columnHeaderShape.js
KoushikKalli: Agreed and removed.

packages/terra-worklist-data-grid/src/proptypes/rowShape.js
KoushikKalli: At this point, we don’t have the need for it, but we will introduce it to accommodate future features such as row selection etc.

packages/terra-worklist-data-grid/src/proptypes/rowShape.js
KoushikKalli: @chrismichalewicz  & @eawww, could you clarify if all rows of the grid will have same height or can row height vary? We assumed that they could vary based on what we saw in Clinical data grid. We can refactor if this assumption is incorrect.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
cm9361: Once you update to a functional component, you can use a ref callback to get the reference to the element.

eawww (commented):
Sorry it's taken me so long to get to this. 

First off, keyboard  works perfectly in JAWS! Nice job.

### Found Issues
- Row header column should be styled the same as normal body cells
- The example should incorporate Worklist style data instead of Flowsheet style data (see attached screenshot). We can fill it out later with all the complex stuff, for now it should at least have text values representative of the type of data it will ultimately display.
image

- Please make the aria-label on the grid something like "Worklist" instead of "aria-label"
- As keyboard shortcuts are added, they should be listed on the Worklist Data Grid doc page.


### Weird VoiceOver Problem
- Arrow key takes focus out of grid while using VO
- In VoiceOver, Home/End behavior jumps VFI out of grid in FF and Chrome but focus can still be moved around with arrow keys in the grid itself. In Safari, moves the entire focus out of the grid.
- In VoiceOver, arrow keys only read contents of newly focused cell. Must use VO key + Arrows instead.

Best as we can tell, this is an issue with VoiceOver. It does it on the APG example as well. We should probably add a future task to investigate/remediate/log an issue with Apple and/or browser manufacturers.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,211 @@
...
+  rows: PropTypes.arrayOf(WorklistDataGridPropTypes.rowShape),
+  /**
+   * String that specifies the default width for columns in the grid. Any valid CSS width value is accepted.
+   * To override the default value, provide width for the column that needs to be overridden.
+   */
+  columnWidth: PropTypes.string,
+  /**
+   * String that specifies the column height. Any valid CSS height value is accepted.
+   */
+  columnHeight: PropTypes.string,
```
cm9361, line 41: Is this really the headerHeight?  Based of the discussion in the UX meeting yesterday, it looks like we will only have the column headers with a header style treatment.  In that case, this would end up just being the header height for the worklist, right?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
cm9361: Based on the meeting yesterday, this column index will not actually look like a header but will be used for the voice over capabilities.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,211 @@
...
+    elementWithTabStop.tabIndex = 0;
+    elementWithTabStop.focus();
+    event.preventDefault();
+  };
+
+  const handleClick = (event) => {
+    // Remove Tab stop from previous cell in table that has focus and set it to the cell that was clicked.
+    grid.current.rows[focusedRow.current].cells[focusedCol.current].tabIndex = -1;
+    focusedRow.current = event.target.parentElement.rowIndex;
+    focusedCol.current = event.target.cellIndex;
```
cm9361, line 139: Since you are using the target.  You will need to use something similar to the JavaScript closest API to get the proper td and tr elements.  This will make you logic future proof as well.  The closest method is not support in IE 11.  However, there are polyfills that can be used or you can create your own implementation.  You want to find the closest td and get the cellIndex.  Then use closest on the td element to find the parent tr element to get the rowIndex.  This logic would work in any scenario.

https://stackoverflow.com/questions/45656949/how-to-return-the-row-and-column-index-of-a-table-cell-by-clicking

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
brithom: As discussed, we renamed this to columnHeaderHeight.

packages/terra-worklist-data-grid/README.md
```diff
@@ -0,0 +1,37 @@
...
+|-|-|
+| react | ^16.8.5 |
+| react-dom | ^16.8.5 |
+| react-intl | ^2.8.0 |
+
+<!-- AUTO-GENERATED-CONTENT:END -->
+
+## LICENSE
+
+Copyright 2019 - 2020 Cerner Innovation, Inc.
```
sycombs, line 31:
```suggestion
Copyright 2023 - present Cerner Innovation, Inc.
```

packages/terra-worklist-data-grid/NOTICE
```diff
@@ -0,0 +1,13 @@
+Copyright 2019 - 2020 Cerner Innovation, Inc.
```
sycombs, line 1:
```suggestion
Copyright 2023 - present Cerner Innovation, Inc.
```

packages/terra-worklist-data-grid/tests/wdio/worklist-data-grid-spec.js
```diff
@@ -0,0 +1,109 @@
+const selector = '#default-terra-worklist-data-grid';
+
+Terra.describeViewports('WorklistDataGrid', ['tiny', 'medium', 'large'], () => {
+  describe('with keyboard navigation', () => {
+    beforeEach(() => {
+      browser.url('/raw/tests/cerner-terra-framework-docs/worklist-data-grid/default-worklist-data-grid');
+    });
+
+    afterEach(async () => {
+      await expect(browser.$$('[role="grid"] [tabIndex="0"]')).toBeElementsArrayOfSize(1);
```
sycombs, line 10: Why do we need to have this expect statement after every test?

sycombs (changes requested):

packages/terra-worklist-data-grid/tests/wdio/worklist-data-grid-spec.js
```diff
@@ -0,0 +1,109 @@
...
+
+      Terra.validates.element('tab-selection-off-grid', { selector });
+    });
+
+    it('arrow keys', () => {
+      browser.keys(['Tab']
+        .concat(new Array(5).fill('ArrowDown'))
+        .concat(new Array(6).fill('ArrowRight'))
+        .concat(new Array(3).fill('ArrowUp'))
+        .concat(new Array(2).fill('ArrowLeft')));
```
sycombs, line 35: +1 to this approach as they use a lot of keystrokes.

packages/terra-worklist-data-grid/tests/wdio/worklist-data-grid-spec.js
brithom: Only one cell on the table should have tabIndex=0. If there are multiple cells with tabIndex=0 (e.g if we forgot to remove the tabIndex from the previous cell that has focus), then it would be possible to tab to multiple cells on the table and this would violate the requirement of one tab stop.

sycombs (approved):

sycombs (approved):

mjpalazzo:
I validated keyboard navigation with Edge and Safari. All is working as expected. When I tabbed out of the table and then Shift tabbed back in, focus returned to the cell I was on before tabbing out. I did not try Home and End keys because my Mac does not have them and mapped key options I tried did not work. 
I validated with JAWS and row headers and columns headers were announced as expected as I navigated with the arrow keys.
VoiceOver did not announce row and column headers as required when I changed columns or changed rows. At this moment navigation with VoiceOver is failing. Also VoiceOver announces navigation instructions with each cell change which is unnecessary and annoying.

mjpalazzo:
I will be online for a brief time on Wednesday morning, so please contact me if you would like to discuss.

mjpalazzo (approved):
Works as expected with Edge and JAWS. I found some issues with VoiceOver that I will document in a Jira for future remediation.

mjpalazzo (approved):
Works as expected with Edge and JAWS. I found some issues with VoiceOver that I will document in a Jira for future remediation.

cm9361 (approved):

packages/terra-worklist-data-grid/tests/wdio/worklist-data-grid-spec.js
sdadn: `before` and `after` functions are intended to be only used for setup and teardown of tests. Not only is it bad practice to put assertions/expects in them but any failures will also prevent execution of subsequent tests. These should be moved into the `it` block of the test itself. Since these are tests, we don't need to worry too much about repetition

brithom: @sdadn, I understand your point but I do not agree completely. Even tests should be clear and succinct where possible as they too must be maintained, The reason for the setup is to ensure that you know the state of the environment before the test begins. We think that asserting that the actual state is consistent with the expected state can be done in the before/setup. Similarly, teardown is meant to get the state back to some known state so asserting that you have gotten back to that state can be part of the after/teardown.
The actual test then become more succinct as it focuses only on what is actually essential to that test. In this example, the before and after are each one line but consider the case where more elaborate setup/teardown is required. Repeating the code in all the tests would not be desirable.

sdadn: I am only referring to the `expect` function being called in the `afterEach` function. You aren't preforming any operation here, you're only asserting that the state hadn't changed. Assertions are part of the test itself and therefore should be included in the `it` block.

sdadn: In the case of a failure, this also prevents execution of subsequent tests which isn't behavior that we want:

CleanShot 2023-05-31 at 11 27 01

brithom: We do want the test to fail for any test where the expectation is violated. I will reach out to you and discuss.

sdadn (approved):

packages/terra-worklist-data-grid/tests/wdio/worklist-data-grid-spec.js
brithom: After discussion with @Saad and @sycombs, we made the change.

eawww (commented):
Noticed a very weird behavior in JAWS. When traversing the Cardiac Index row, Oxygen Flow Rate row, and some other rows horizontally, it's reading the dates at the top correctly, but then reading the number before the decimal as the ordinal. (ex: 2.8 being read as "second point eight"). This seems like a JAWS error but maybe there's something we could do about it. Gonna bring this up when I meet with Brooks this afternoon.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,211 @@
...
+      id={id}
+      role="grid"
+      aria-labelledby={ariaLabelledby}
+      aria-label={ariaLabel}
+      className={gridClassNames}
+      onClick={handleClick}
+      onKeyDown={handleKeyDown}
+    >
+      <tbody>
+        {buildColumns(columns)}
```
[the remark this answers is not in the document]
cm9361: @eawww - I would not expect us to have a header that spans multiple rows.  I don't see this as a plausible use case for the worklist data grid at least.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,217 @@
...
+    focusedCol.current = nextCol;
+
+    const elementWithTabStop = grid.current.rows[nextRow].cells[nextCol];
+    elementWithTabStop.tabIndex = 0;
+    elementWithTabStop.focus();
+    event.preventDefault();
+  };
+
+  const handleClick = (event) => {
+    const clickedCell = event.target.closest('td,th') || document.activeElement.closest('td,th');
```
cm9361, line 136: It would be really good to have a comment here.  The code looks correct.  However, a comment might be useful since the logic is not straightforward.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,217 @@
...
+    if (!clickedCell) {
+      event.preventDefault();
```
cm9361, line 138: Why are we preventing the default behavior in this scenario?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,217 @@
...
+    focusedCol.current = clickedCell.cellIndex;
+
+    clickedCell.tabIndex = 0;
+    clickedCell.focus();
+    event.preventDefault();
+  };
+
+  const getCellData = (cell, cellColumnIndex) => {
+    const tabIndex = { tabIndex: '-1' };
+    return props.rowHeaderIndex === cellColumnIndex ? (<th key={cellColumnIndex} {...tabIndex} className={cx('worklist-data-grid-row-header')}>{cell.content}</th>) : (<td key={cellColumnIndex} {...tabIndex} className={cx('worklist-data-grid-cell-data')}>{cell.content}</td>);
```
cm9361, line 153: Maybe this is for a separate JIRA, but I thought we landed on the worklist does not have row headers.  You can style the cells with a classname without needing to worry about creating a distinction between td and th elements.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
cm9361: I think if we handle the onMouseDown event, the behavior would mimic MS Excel and it would simplify this logic as the user would not be allowed to drag before the event triggers.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -0,0 +1,211 @@
...
+  const buildColumn = (columnData) => {
+    const width = columnData.width || props.columnWidth;
+    const height = props.columnHeaderHeight;
+    return (
+      /* eslint-disable react/forbid-dom-props */
+      <th key={columnData.id} className={cx('worklist-data-grid-column-header')} tabIndex="-1" style={{ width, height }}>{columnData.displayName}</th>
+    );
+  };
+
+  const buildColumns = (allColumns) => {
```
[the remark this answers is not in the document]
cm9361: The buildColumn method is actually the method building the header row not this one.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
brithom: The mouseDown would also work but I do not think that it will be ideal when it comes to handling selection and notifying the consumer. We can always refactor later.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
brithom: I think it can be removed. We will test without it.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Our understanding is that the grid will have row headers for screen reader purposes but they will be styled similar to regular cells.

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/ChangeLog.1.doc.mdx
```diff
@@ -0,0 +1,7 @@
+import { Badge } from 'terra-worklist-data-grid/package.json?dev-site-package';
+
```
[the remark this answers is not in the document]
KoushikKalli: Updated.

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/WorklistDataGrid.1.doc.mdx
```diff
@@ -0,0 +1,45 @@
+import { Badge } from 'terra-grid/package.json?dev-site-package';
+
```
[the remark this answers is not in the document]
KoushikKalli: Updated.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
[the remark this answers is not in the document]
KoushikKalli: The name of this method has gone through multiple iterations in this review starting with buildColumnHeaders. We are going to leave it as is for now.

cm9361 (approved):

brooksienoodlesoup (approved):
Tab moves into and out of grid as expected.
Keyboard shortcuts produce expected screen reader results.
Code markup is according to accessible specs.

PR #1673, opened 2023-06-09
[terra-worklist-data-grid] Changed focus styling for cells

artpark:
### Summary

**What was changed:**
Updated the css stylings for cell focus to have a dashed black outline.

**Why it was changed:**
To match the other stylings in terra, such as terra-table.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9141 

---

@cerner/terra


chrismichalewicz:
Tested this with Microsoft Edge and it looked good to me.  Also did a quick test with JAWS to confirm nothing changed.  +1.  @eawww  once you've had a chance to test with your Mac and if it looks good please update the Accessibility label as well as the UX one. Thanks!

packages/terra-worklist-data-grid/src/WorklistDataGrid.module.scss
```diff
@@ -4,9 +4,9 @@
 
 :local {
   .worklist-data-grid [tabIndex='0']:focus {
-    background-color: var(--terra-worklist-data-grid-cell-focus-background-color, #cbe7fa);
-    box-shadow: var(--terra-worklist-data-grid-cell-focus-box-shadow, inset 0 0 0 1px rgba(38, 162, 229, 0.5), inset 0 0 0 4px rgba(76, 178, 233, 0.2));
-    outline: none;
+    outline: var(--terra-worklist-data-grid-cell-focus-outline, 2px dashed #000);
+    outline-offset: var(--terra-data-grid-cell-focus-outline-offset, -2px);
+    z-index: 1;
```
cm9361, line 9: I would not expect us to have to set the z-index for the implementation.

cm9361:
@artpark - It does not appear that you are covering the use case where the user clicks or presses space while on a cell.  In that scenario, my understanding is that we will preserve the styling that is used today.  The keyboard focus will move with the navigation.  However, the active selection state does not change.  There do not appear to be screen shots to show this behavior.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -144,15 +151,48 @@ function WorklistDataGrid(props) {
     grid.current.rows[focusedRow.current].cells[focusedCol.current].tabIndex = -1;
     focusedRow.current = clickedCell.parentElement.rowIndex;
     focusedCol.current = clickedCell.cellIndex;
+    setSelectedRow(clickedCell.parentElement.rowIndex);
+    setSelectedCol(clickedCell.cellIndex);
```
cm9361, line 155: It looks like you are duplicating the logic used for the focusedRow and focusedColumn ref variables.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -129,6 +135,7 @@ function WorklistDataGrid(props) {
     const elementWithTabStop = grid.current.rows[nextRow].cells[nextCol];
     elementWithTabStop.tabIndex = 0;
```
cm9361, line 136: Should this now be handled during rendering since you have a useState hook now?  Would the focus be set in the useEffect hook?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
artpark: I've removed the cell select functionality so this is resolved

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
artpark: I've removed the cell select functionality so this is resolved

cm9361:
@artpark - With this update, we would never have our blue background when clicking a cell, correct?  Would we want to merge this state?

packages/terra-worklist-data-grid/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 * Changed
   * Removed ability to set variable row heights
+  * Updated styling on cell focus
```
sdadn, line 6:
```suggestion
  * Updated styling on cell focus.
```

sdadn (approved):

artpark:
> @artpark - With this update, we would never have our blue background when clicking a cell, correct? Would we want to merge this state?

I think it's alright since https://github.com/cerner/terra-framework/pull/1676 should be adding the blue background, as well as the cell selected functionality.

packages/terra-worklist-data-grid/src/WorklistDataGrid.module.scss
artpark: Removed

packages/terra-worklist-data-grid/CHANGELOG.md
artpark: Updated

mjpalazzo:
The focus styling looks good. I tested with Chrome, Edge and Safari. I also did some screen reader testing to see how it looked with the screen reader. JAWS worked as expected. With VoiceOver, the dashed border displayed when a cell had focus. However, VoiceOver is not announcing row and column headers as I navigate the table cells. Not a show stopper, but something we need to address in the future.

sycombs (approved):

eawww:
Looks great in the body. 
I'm noticing with the header row cells the focus indicator is getting chopped a bit. ☹️ 
image

artpark:
> Looks great in the body. I'm noticing with the header row cells the focus indicator is getting chopped a bit. ☹️ image
Screenshot 2023-06-16 at 2 12 51 PM

eawww:
@artpark yeah that looks good. Thank you.

mjpalazzo:
> > The focus styling looks good. I tested with Chrome, Edge and Safari. I also did some screen reader testing to see how it looked with the screen reader. JAWS worked as expected. With VoiceOver, the dashed border displayed when a cell had focus. However, VoiceOver is not announcing row and column headers as I navigate the table cells. Not a show stopper, but something we need to address in the future.
> 
> Hey @mjpalazzo. VoiceOver has some known weirdness with grids. You should be able to use normal table navigation with VO+Arrow Keys and select a cell with VO+Space. The APG example has similarly unideal behavior.

Thanks for the reminder about the VO+arrow and VO+space @eawww. 2 weeks vaca cleared that from my mind.

PR #1697, opened 2023-07-12
[terra-abstract-modal] Contain focus in modal content

smason0:
### Summary

**What was changed:**
~~Added `trapFocus` prop to AbstractModal and passing the prop down as "true" from ModalManager. The _ModalContent component now uses a FocusTrapWrapper to conditionally trap focus.~~ Update: AbstractModal now always traps focus in its content, no longer conditional (see discussion below).

**Why it was changed:**
This change makes AbstractModal* modals trap a keyboard user's focus within the modal rather than focus jumping to outside of the modal, for improved a11y. ~~This also allows other Terra components that consume AbstractModal to optionally add this focus-trapping behavior. The new `trapFocus` prop defaults to "false" to remain passive.~~

*Edited

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

Tested with dev site examples in Chrome, Firefox, Safari, and Edge browsers.

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9167 

---

@cerner/terra


packages/terra-modal-manager/tests/wdio/modal-manager-spec.js
```diff
@@ -336,27 +336,31 @@ Terra.describeViewports('ModalManager - Behaviors', ['large'], () => {
       });
     });
 
-    describe('Outside Focus Handling Before Modal', () => {
-      it('shifts focus before modal', () => {
+    describe('Modal Content Focus Shift Back', () => {
+      it('shifts focus to last element in modal content when shifting back from the first element', () => {
```
sdadn, line 340: Grammar nit:
```suggestion
      it('shifts focus to the last element in modal content when shifting back from the first element', () => {
```

packages/terra-modal-manager/tests/wdio/modal-manager-spec.js
```diff
@@ -336,27 +336,31 @@ Terra.describeViewports('ModalManager - Behaviors', ['large'], () => {
...
+        expect($('#DemoContainer-1 .maximize').isFocused()).toEqual(true);
+        Terra.validates.element('focused shifted to last', { selector });
         browser.keys('Escape');
       });
     });
 
-    describe('Outside Focus Handling After Modal', () => {
-      it('shifts focus after the modal', () => {
+    describe('Modal Content Focus Shift Forward', () => {
+      it('shifts focus to first element in modal content when shifting forward from the last element', () => {
```
sdadn, line 353: Grammar nit:
```suggestion
      it('shifts focus to the first element in modal content when shifting forward from the last element', () => {
```

packages/terra-abstract-modal/tests/wdio/abstract-modal-spec.js
```diff
@@ -150,6 +150,38 @@ Terra.describeViewports('Abstract Modal', ['medium'], () => {
         Terra.validates.element('focus on modal', { selector });
       });
     });
+
+    describe('Focus Trapped Content', () => {
+      it('clicks to open modal', () => {
+        browser.url('/raw/tests/terra-abstract-modal/abstract-modal/focus-trapped-abstract-modal');
+        $('button').click();
+      });
```
sdadn, line 158: This doesn't seem to be validating anything; is this intended to set up subsequent tests or is there a missing expect/validate statement?
```suggestion
      it('clicks to open modal', () => {
        browser.url('/raw/tests/terra-abstract-modal/abstract-modal/focus-trapped-abstract-modal');
        $('button').click();
      });
```

smason0: This is just a setup step that doesn't validate anything

trandrew1023 (approved):

packages/terra-abstract-modal/tests/wdio/abstract-modal-spec.js
sdadn: We can use `beforeAll` or `beforeEach` instead.

```suggestion
      beforeAll(() => {
        browser.url('/raw/tests/terra-abstract-modal/abstract-modal/focus-trapped-abstract-modal');
        $('button').click();
      });
```

sdadn (approved):

mjpalazzo:
I tested the focus trapped example keyboard navigation with Chrome, Edge and Safari. I was not able to Tab or Shift Tab off the modal dialog. I tried a few key combinations and could not exit the modal. Clicking the Escape key or the close button were the only ways close the modal. I could use the mouse to click an area outside the modal and that worked like clicking Escape.
I tested with Edge/JAWS and Safari/VoiceOver and the screen readers worked as expected.
Nice work, @smason0

eawww (commented):
One thing to discuss but otherwise looks good!

packages/terra-abstract-modal/src/_ModalContent.jsx
```diff
@@ -73,9 +78,20 @@ const defaultProps = {
   isScrollable: false,
   role: 'dialog',
   rootSelector: '#root',
+  trapFocus: false,
```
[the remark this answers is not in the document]
smason0: That's a good point. I made this passive so consumers like NotificationDialog, which has its own [focus trap behavior](https://github.com/cerner/terra-framework/blob/17f77f704c1c644b99b3b251a862ea4497032c7a/packages/terra-notification-dialog/src/NotificationDialog.jsx#L170) would not be affected. This might require further investigation about if NotificationDialog's focus trap behavior meets the a11y standards. Also, if today's non-focus trapping behavior of AbstractModal is considered a bug, we probably wouldn't even want to allow for a setting of 'false' for the `trapFocus` prop, and thus there would be no need for a `trapFocus` prop at all, we'd just always trap focus.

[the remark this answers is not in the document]
eawww: Yeah we can just remove the prop and always trap focus. Thanks Steven!

sycombs (approved):

eawww (approved):

PR #1700, opened 2023-07-18
[terra-slide-panel] Add focus check for disclosing node before focusing

trandrew1023:
### Summary

**What was changed:**
Added check for disclosing node's focus before calling focus. If the disclosing node has undefined focus, the main node will be focused.

This is no longer reproducible within Terra using IE11 and custom example with SVGs in buttons, but this change has still been added to prevent unknown issues with undefined focus on disclosing nodes. An example was added with just an SVG with onclick which would cause the crash. With the added check, the crash no longer occurs.

**Why it was changed:**
Issues with SVGs and browser (IE11) crashing due to focus not being defined.

### Testing

The example added to test these slide panel changes can be found under Tests -> Slide Panel -> Slide Panel SVG Toggle

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

Just SVG with onclick:
Before:
image

After:
image

Verified in IE11 with different buttons with Terra Icons and SVGs
image

### Reviews

- [x] UX review
- [ ] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9296

---

@cerner/terra

### DIFF
--- a/packages/terra-framework-docs/CHANGELOG.md
+++ b/packages/terra-framework-docs/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Added
   * Added `terra-worklist-data-grid` as dependency.
+  * Added example to test toggling and focusing disclosing node for `terra-slide-panel`.
 
 ## 1.25.0 - (July 19, 2023)
 
--- a/packages/terra-framework-docs/src/terra-dev-site/test/slide-panel/SlidePanelDocCommon.test.module.scss
+++ b/packages/terra-framework-docs/src/terra-dev-site/test/slide-panel/SlidePanelDocCommon.test.module.scss
@@ -30,4 +30,15 @@
   .button {
     background-color: #008a00;
   }
+
+  .circle-svg {
+    overflow: visible;
+    :hover {
+      cursor: pointer;
+    }
+  }
+
+  .p-clickable:hover {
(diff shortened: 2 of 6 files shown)


trandrew1023:
Updated to test example to include <p> to toggle and test focus

image

artpark (approved):

sdadn:
Does this fix result in any change in behavior for non-IE browsers?

trandrew1023:
> Does this fix result in any change in behavior for non-IE browsers?

No change in behavior from testing since SVGs in non-IE browsers don't crash with an undefined issue in non-IE so the focus check didn't change anything. 

One thing to note though is when disclosing from an SVG or even `<p>`, the focus returns to the page body instead of the disclosing node even though the disclosing node is set. This happens with/without the added focus check

smason0 (approved):

sdadn (approved):

sycombs (approved):

chrismichalewicz:
@trandrew1023  I have reviewed this using Edge in IE Mode and this looks good. Added the Functionally Reviewed label.

cm9361 (approved):

trandrew1023:
Removed UX review since this was mostly just a functional change

PR #1710, opened 2023-07-20
[terra-worklist-data-grid] Add support for pinned columns

sycombs:
### Summary

This change updates terra-worklist-data-grid to support pinned columns. This PR mainly focuses on adding the basic functionality to worklist-data-grid and isn't a comprehensive addition of the feature. Therefore, things like WDIO tests, edge cases around the use of the feature and finalized styling and visuals will added in subsequent PRs. The reason behind breaking up the change into multiple PRs is because terra-workist-data-grid is under active development and in a volatile state. Therefore we opted to merge finalized pieces as soon as possible without breaking the main build.

This PR also fixes the text alignment of the row header.

### Testing

This change was tested using:

- [ ] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

Known issues:
- There is a 1 pixel offset on the left side of the grid that show overflow content behind pinned columns.
- Pinned columns become transparent on hover and show the content of overflow columns. This is due to the hover CSS setting the background-color to be transparent. A fix would be to apply the exact shade as an opaque background color.
- When horizontally scrolling to the left, the focus indicator will be visible behind pinned columns.
- When navigating left on overflow columns, focus will move to a hidden overflow cell instead of scrolling left.

These issues will be resolved in follow-up PRs/Jiras.

**This PR resolves:**

UXPLATFORM-9041 

---

@cerner/terra


packages/terra-worklist-data-grid/src/todo
sycombs, line 1: Temporary file; will be removed later

packages/terra-worklist-data-grid/src/wdg-sticky.jsx
sycombs, line 1: Temporary file; will be removed later

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -250,7 +289,17 @@ function WorklistDataGrid(props) {
...
-  }, [columns]);
+  }, [pinnedColumns, overflowColumns]);
+
+  // useEffect for pinned column offset
+  useEffect(() => {
+    if (pinnedColumns.length > 0) {
+      calculateOffsets();
+    }
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [dataGridColumns]);
```
cm9361, line 300: You would want to leverage the useEffect that if dependent upon the columns prop.  You don't need an addition useEffect hook.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -151,6 +164,23 @@ function WorklistDataGrid(props) {
...
 
+  const [pinnedColumnOffsets, setPinnedColumnOffsets] = useState([0]);
+  const [pinnedColumnsTotalWidth, setPinnedColumnsTotalWidth] = useState(0);
+
+  const calculateOffsets = () => {
+    let cumulativeOffset = 0;
+    const offsetArray = [0];
+    const totalPinnedColumnsCount = hasSelectableRows ? pinnedColumns.length : pinnedColumns.length - 1;
+
+    dataGridColumns.slice(0, totalPinnedColumnsCount).map((pinnedCol) => {
```
cm9361, line 175: Wouldn't it be easier to use a simple for loop in this scenario?  I think that would more readable and efficient than slicing an array just to be able to use the map function..

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
```diff
@@ -1,13 +1,16 @@
...
 import { injectIntl } from 'react-intl';
+import * as KeyCode from 'keycode-js';
 import classNames from 'classnames/bind';
+
 import ThemeContext from 'terra-theme-context';
+
 import WorklistDataGridUtils from '../utils/WorklistDataGridUtils';
 import styles from './Cell.module.scss';
+import ColumnContext from '../utils/ColumnContext';
+import '../_elementPolyfill';
```
cm9361, line 13: I don't think you need to add this polyfill.  You are not using the closest method.

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
```diff
@@ -153,6 +159,7 @@ function Cell(props) {
 
   const className = cx('worklist-data-grid-cell', {
     masked: isMasked,
+    pinned: columnIndex < columnContext.pinnedColumnsLength,
```
cm9361, line 162: If you have an isPinned prop, why would you need this calculation.  I would expect for the parent Row component to handle determining if the cell is pinned.

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
```diff
@@ -168,6 +175,7 @@ function Cell(props) {
       className={className}
       onMouseDown={handleMouseDown}
       onKeyDown={handleKeyDown}
+      style={{ left: columnContext.pinnedColumnOffsets[columnIndex] }}
```
cm9361, line 178: The Row component should be able to provide this value to the cell.

packages/terra-worklist-data-grid/src/subcomponents/ColumnHeader.jsx
```diff
@@ -36,6 +38,11 @@ const propTypes = {
   onResizeMouseDown: PropTypes.func,
 };
 
+const defaultProps = {
+  pinnedColumns: [],
+  overflowColumns: [],
```
cm9361, line 43: These don't appear to be props of the ColumnHeader component.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -507,17 +566,25 @@ function WorklistDataGrid(props) {
...
-          headerHeight={columnHeaderHeight}
-          tableHeight={tableHeight}
-          tabStopColumnIndex={focusedRow.current === 0 ? focusedCol.current : undefined}
-          onColumnSelect={handleColumnSelect}
-          onResizeMouseDown={onResizeMouseDown}
-        />
-        <tbody>
-          {buildRows(rows)}
-        </tbody>
+        <ColumnContext.Provider
```
cm9361, line 569: You don't really need context for this use case.  To pass the value down one level to two components doesn't really seem to require the usage of context.

packages/terra-worklist-data-grid/src/subcomponents/ColumnHeader.jsx
```diff
@@ -1,13 +1,15 @@
...
 import ColumnHeaderCell from './ColumnHeaderCell';
 
 const propTypes = {
   /**
-   * Data for columns. By default, columns will be presented in the order given.
+   * Data for columns.
+   * Columns will be presented in the order given.
    */
-  columns: PropTypes.arrayOf(WorklistDataGridPropTypes.columnShape).isRequired,
+  columns: PropTypes.arrayOf(WorklistDataGridPropTypes.columnShape),
```
cm9361, line 12: Why would the columns array not be required?  You can't have a column header without columns.

packages/terra-worklist-data-grid/src/wdg-sticky.jsx
cm9361: So there is no need to review this file?  If so, can we just remove it from the draft PR?

packages/terra-worklist-data-grid/src/subcomponents/RowSelectionCell.jsx
```diff
@@ -1,8 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import '../_elementPolyfill';
 import { injectIntl } from 'react-intl';
+
 import Cell from './Cell';
+import '../_elementPolyfill';
```
cm9361, line 6: This polyfill is not necessary.

packages/terra-worklist-data-grid/src/subcomponents/Row.module.scss
```diff
@@ -7,6 +7,10 @@
     background-color: var(--terra-worklist-data-grid-row-striped-background-color, #f0f0f0);
   }
 
+  .worklist-data-grid-row:nth-child(odd) {
+    background-color: var(--terra-worklist-data-grid-row-striped-background-color, white);
+  }
```
cm9361, line 12: Why are we making the background color white?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
sycombs: We're using a useEffect that is dependent on the dataGridColumns variable to account for the change in column widths after resizing. When this is done in the useEffect that is dependent on the columns props, the pinned column offsets won't update when columns are resized.

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
sycombs: The polyfill was already being imported; we just reordered the imports.

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
sycombs: The `isPinned` prop above is leftover code we forgot to remove. There is no `isPinned` prop.

packages/terra-worklist-data-grid/src/subcomponents/ColumnHeader.jsx
sycombs: Most of the changes in this file were left over from our initial changes and have been reverted: [1a46886](https://github.com/cerner/terra-framework/pull/1710/commits/1a4688675452581e97d235eee07f053457a24860)

packages/terra-worklist-data-grid/src/subcomponents/Row.module.scss
sycombs: The background color was originally transparent and resulted in overflow columns being visible behind pinned columns. We tested with other themes and this does not seem to impact them.

packages/terra-worklist-data-grid/src/subcomponents/ColumnHeader.jsx
sycombs: You're right, most of the changes in this file were left over from our original changes. Reverted here: [1a46886](https://github.com/cerner/terra-framework/pull/1710/commits/1a4688675452581e97d235eee07f053457a24860)

packages/terra-worklist-data-grid/src/wdg-sticky.jsx
sycombs: This file contains our original prototype that we are using as reference for the new pinned column changes (updated for the new state of main). We will remove it once we finalize the changes before opening the PR.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
sycombs: We chose to use context for a few reasons:

- We can bypass the `Row` and `ColumnHeader` components, which don't need to know anything about pinned/overflow columns, and directly pass the data to `Cell` and `ColumnHeaderCell` (the only components that need to know whether or not they are pinned).
- We originally used separate `pinnedColumns` and `overflowColumns` props which were passed through the `Row` and `ColumnHeader` components. This created extra handling within those components that is largely irrelevant to the purpose of those components.
- Passing `pinnedColumns` and `overflowColumns` props down multiple levels of components requires additional consideration for handling cell selection and column resizing due to shared indexes, but using the existing `dataGridColumns` variable (which encompasses all columns) does not break existing functionality.

While it would be nice to have a simple `isPinned` prop on the `Cell` and `ColumnHeaderCell` components, we also need to pass the offsets that are calculated at the `WorklistDataGrid` level, and adding these props to `Row` and `ColumnHeader` serves only as a passthrough and unnecessarily bloats the props for those components.

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
sycombs: Can you elaborate on what you mean by this?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
sdadn: We used this to conform with functional programming patterns. Regarding efficiency, the complexity is O(N) either way so I don't think there's a huge difference. Readability is subjective and dependent upon the project's conventions.

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
cm9361: If the Row component receives the pinned column information, It can pass this value as a prop to each Cell component.

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
cm9361: I think it makes sense for the property to exist.  It's parent component can pass the value to it.

packages/terra-worklist-data-grid/src/subcomponents/Row.module.scss
cm9361, line 12: This is incorrect.  You can't use the same color for both even and odd rows.  You would need different variables.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
cm9361: How will this work properly once we memoize the ColumnHeader and Row objects.  The properties passed to these component will not change and thus not rerender.

packages/terra-worklist-data-grid/src/subcomponents/Cell.module.scss
```diff
@@ -30,12 +30,21 @@
...
+    z-index: 1;
+    box-shadow: inset 1px 0px #dedfe0;
+    border-width:  1px 0 1px 0;
+    border: inherit, 1px solid #dedfe0;
+  }
 
   // Hover state for selectable elements
   .selectable:hover {
-    background-color: var(--terra-worklist-data-grid-hover-background-color, rgba(0, 146, 224, 0.12));
+    background-color: var(--terra-worklist-data-grid-hover-background-color, #D8E5ED);
```
cm9361, line 47: Don't uppercase letters give you a linter error?

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
sdadn: The row component isn't receiving any pinned column information.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
sdadn: Row and ColumnHeader are already being bypassed so even now, they have no impact on re-rendering pinned-columns. Any changes to the context provider causes the children components to re-render and React.memo does not affect this behavior.

https://react.dev/reference/react/memo#updating-a-memoized-component-using-a-context

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
cm9361: It could.  It is just not designed that way..  The Row could receive the pinned column information and construct the Cell components accordingly.

sdadn:
This PR does not add WDIO tests. Those will be added in a separate PR after #1712 is merged.

eawww:
@sdadn @sycombs I'm not able to get non-pinned columns to scroll into view when focused on any browser. Is this planned to be addressed in the next PR?

sdadn:
> @sdadn @sycombs I'm not able to get non-pinned columns to scroll into view when focused on any browser. Is this planned to be addressed in the next PR?

If you're referring to keyboard navigation, then yes, we will address it in a separate PR so we can have a more focused review on those changes.

eawww:
@sdadn cool beans. If it's just to get the pinned columns API out there for alpha adoption, it can probs be merged without UX review.

artpark (approved):

packages/terra-worklist-data-grid/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -183,19 +187,28 @@ const ColumnHeaderCell = (props) => {
...
   const theme = useContext(ThemeContext);
 
+  const pinnedColumnsDivider = columnIndex === columnContext.pinnedColumnsLength - 1
+    ? (
+      <div
+        className={cx('pinned-columns-divider')}
+        style={{ height: tableHeight, left: width - 1 }} // eslint-disable-line react/forbid-dom-props
+      />
+    )
+    : null;
```
cm9361, line 197: It looks like you should be able to add this directly into the JSX of the return statement using conditional rendering.  You don't need this additional variable

packages/terra-worklist-data-grid/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -183,19 +187,28 @@ const ColumnHeaderCell = (props) => {
...
-      className={cx('column-header', theme.className, { selectable: isSelectable })}
+      className={cx('column-header', theme.className, { selectable: isSelectable, pinned: columnIndex < columnContext.pinnedColumnsLength })}
       tabIndex={isTabStop ? 0 : -1}
       role="columnheader"
       scope="col"
       aria-sort={sortIndicator}
       onMouseDown={(isSelectable && onColumnSelect) ? handleMouseDown : undefined}
       onKeyDown={(isSelectable && onColumnSelect) ? handleKeyDown : undefined}
-      style={{ width: `${width}px`, height: headerHeight }}
+      style={{ width: `${width}px`, height: headerHeight, left: columnContext.pinnedColumnOffsets[columnIndex] }} // eslint-disable-line react/forbid-dom-props
```
cm9361, line 211: I would think that the left style would not be added unless this is a pinned column.  Do we want to access invalid array indexes to provide this implementation?

packages/terra-worklist-data-grid/src/subcomponents/ColumnResizeHandle.jsx
```diff
@@ -4,7 +4,9 @@ import React, {
 import PropTypes from 'prop-types';
 import { injectIntl } from 'react-intl';
 import classNames from 'classnames/bind';
+
```
cm9361, line 7: There really is not a reason to change this file.

packages/terra-worklist-data-grid/src/subcomponents/Row.jsx
```diff
@@ -131,11 +133,11 @@ function Row(props) {
         rowIndex={cellRowIndex}
         columnIndex={cellColumnIndex}
         key={`${rowId}_${columnId}`}
-        isTabStop={tabStopColumnIndex === cellColumnIndex}
-        isSelected={!hasRowSelection && selectedCellColumnId === columnId}
         isMasked={cellData.isMasked}
-        isSelectable={cellData.isSelectable}
         isRowHeader={isRowHeader}
+        isSelectable={cellData.isSelectable}
```
cm9361, line 138: Why do these changes need to be made?  While we are in the midst of development, I think it would be best to limit these types of unnecessary changes to avoid merge conflicts.  It does not appear that any of these changes alter the logic.

packages/terra-worklist-data-grid/src/subcomponents/RowSelectionCell.jsx
cm9361: It does not appear that this change is necessary.  If we want to make a change here, we should remove the polyfill as it is not used.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -250,7 +282,28 @@ function WorklistDataGrid(props) {
...
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [columns]);
+  }, [pinnedColumns, overflowColumns]);
+
+  // useEffect to calculate pinned column offsets
+  useEffect(() => {
+    if (pinnedColumns.length > 0) {
+      let cumulativeOffset = 0;
+      const offsetArray = [0];
+      const totalPinnedColumnsCount = hasSelectableRows ? pinnedColumns.length : pinnedColumns.length - 1;
```
cm9361, line 292: Why would the total pinned column count be the length of the pinned columns array minus 1?  Why do we need to subtract one?  Maybe a comment would help.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -250,7 +282,28 @@ function WorklistDataGrid(props) {
...
+
+      // eslint-disable-next-line array-callback-return
+      dataGridColumns.slice(0, totalPinnedColumnsCount).map((pinnedCol) => {
```
cm9361, line 295: I would spell out pinnedColumn instead of using an abbreviation.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -250,7 +282,28 @@ function WorklistDataGrid(props) {
...
     setDataGridColumns(displayedColumns.map((column) => initializeColumn(column)));
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [columns]);
+  }, [pinnedColumns, overflowColumns]);
+
+  // useEffect to calculate pinned column offsets
+  useEffect(() => {
+    if (pinnedColumns.length > 0) {
+      let cumulativeOffset = 0;
+      const offsetArray = [0];
```
cm9361, line 291: What is the advantage of initializing the first element to 0?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -343,6 +396,9 @@ function WorklistDataGrid(props) {
     selectRows(false, rowId, rowIndex);
   };
 
+  // -------------------------------------
+  // event handlers
+
```
cm9361, line 401: We should have proper comments for the functions instead of these block comments.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -507,17 +568,25 @@ function WorklistDataGrid(props) {
...
-          onResizeMouseDown={onResizeMouseDown}
-        />
-        <tbody>
-          {buildRows(rows)}
-        </tbody>
+        <ColumnContext.Provider
+          value={{
+            pinnedColumnsLength: hasSelectableRows ? pinnedColumns.length + 1 : pinnedColumns.length,
+            pinnedColumnOffsets,
+            pinnedColumnsTotalWidth,
```
cm9361, line 575: Should this all be managed by one state variable object?  It looks like the column offsets and total width are usually calculated together.  The pinned columns length should be able to be part of the same object as they all depend on the number of pinned columns.

packages/terra-worklist-data-grid/src/subcomponents/ColumnHeaderCell.module.scss
```diff
@@ -38,8 +38,23 @@
...
   // Hover state for selectable elements
   .selectable:hover {
-    background-color: var(--terra-worklist-data-grid-hover-background-color, rgba(0, 146, 224, 0.12));
+    background-color: var(--terra-worklist-data-grid-header-hover-background-color, #e4f2f9);
     cursor: pointer;
   }
-}
 
+  .pinned {
+    border: inherit, 1px solid #fff;
```
cm9361, line 46: Don't you need to account for theming with your colors?  With this implementation, there is no option to change the color based on the theme.  Same for other newly introduced colors.

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/PinnedColumnsWithRowSelection.jsx
```diff
@@ -0,0 +1,146 @@
...
+          rows.forEach(element => {
+            if (element.isSelected) {
+              newRows.push(element.id);
+            }
+          });
+          setSelectedRows(determineSelectedRows(false, newRows));
+        }}
+        onRowSelectAll={() => {
+          const newRows = [];
+          rows.forEach(e => { e.isSelected = true; newRows.push(e.id); });
```
cm9361, line 132: We should have a better variable name than "e".

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/PinnedColumnsWithRowSelection.jsx
```diff
@@ -0,0 +1,146 @@
...
+        <label htmlFor="rowSelectionMode"><b>Row Selection Mode</b></label>
+        <input
+          id="rowSelectionMode"
+          type="checkbox"
+          ref={rowSelectionModeRef}
+          onChange={onRowSelectionModeChange}
+        />
+      </div>
+      <WorklistDataGrid
+        id="default-terra-worklist-data-grid"
```
cm9361, line 111: This example should have its own unique identifier.

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
```diff
@@ -168,6 +174,7 @@ function Cell(props) {
       className={className}
       onMouseDown={handleMouseDown}
       onKeyDown={handleKeyDown}
+      style={{ left: columnContext.pinnedColumnOffsets[columnIndex] }} // eslint-disable-line react/forbid-component-props
```
cm9361, line 177: We would only add this style if the cell is in a pinned column.  We should not rely on access elements beyond the bounds of the array for our implementation.

packages/terra-worklist-data-grid/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -183,19 +187,28 @@ const ColumnHeaderCell = (props) => {
   // Retrieve current theme from context
   const theme = useContext(ThemeContext);
 
+  const pinnedColumnsDivider = columnIndex === columnContext.pinnedColumnsLength - 1
+    ? (
+      <div
+        className={cx('pinned-columns-divider')}
+        style={{ height: tableHeight, left: width - 1 }} // eslint-disable-line react/forbid-dom-props
```
cm9361, line 194: You should be able to set the left style with CSS, right?  You don't need that here.

cm9361:
Your low-light theme implementation is incorrect.  It looks like you replaced the screen shots.  However, the new ones do not apply the colors properly.

packages/terra-worklist-data-grid/src/subcomponents/ColumnHeaderCell.jsx
sycombs: Right now using a variable clarifies the purpose of this div (that it is a pinned columns divider). Also, having a ternary operation within the return statement reduces readability and makes it more difficult to understand at a glance.

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
sycombs: [69036d9](https://github.com/cerner/terra-framework/pull/1710/commits/69036d994769ea38abc772f887a266894c8f9517)

packages/terra-worklist-data-grid/src/subcomponents/ColumnHeaderCell.jsx
sycombs: Updated: [9ac76fa](https://github.com/cerner/terra-framework/pull/1710/commits/9ac76faf9f340173a01fa8654ec5362ff045eaf1)

packages/terra-worklist-data-grid/src/subcomponents/ColumnResizeHandle.jsx
sycombs: Reverted: [69036d9](https://github.com/cerner/terra-framework/pull/1710/commits/69036d994769ea38abc772f887a266894c8f9517)

packages/terra-worklist-data-grid/src/subcomponents/Row.jsx
sycombs: Reverted: [d2870fe](https://github.com/cerner/terra-framework/pull/1710/commits/d2870fedaecf3d9ca23a1a7443b5b2575389ba88)

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
sycombs: This is because we are omitting the final pinned column from the offset arrays (since we only need to calculate the offset for the left edge of each column, and to calculate the edge of the last column, we only need the cumulative widths of all columns that came before it.).

Updated the variable name for clarity here: [b3b0076](https://github.com/cerner/terra-framework/pull/1710/commits/b3b00769ed5615dad6a7f14f771e655a44e05667)

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
sycombs: Updated: [c366625](https://github.com/cerner/terra-framework/pull/1710/commits/c3666253e0347c3c0e39171d02a5c0008a162975)

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
sycombs: This is because the left offset for the first column will always be 0, and this is an array of cumulative offsets. Also updated to use `cumulativeOffset` for the `offsetArray` initialization for more clarity: [80f7216](https://github.com/cerner/terra-framework/pull/1710/commits/80f72169793d55a4e2851b8b12d0051686391248)

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
sycombs: These are meant to be section headers for organization since there are a lot of existing functions. Each function can have additional separate comments, but that's not what these headers are meant for. If comments for individual functions are needed, they should be added in a separate PR.

packages/terra-worklist-data-grid/src/subcomponents/ColumnHeaderCell.module.scss
sycombs: You are correct that we need to account for theming; we will work with UX after this PR is merged to incorporate theming.

See Eric's comment: https://github.com/cerner/terra-framework/pull/1710#issuecomment-1650586532

CC: @eawww

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/PinnedColumnsWithRowSelection.jsx
sycombs: This example is mostly copied from the existing row selection example: https://github.com/cerner/terra-framework/blob/main/packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/RowSelection.jsx

Updated the variable names here: [beaa707](https://github.com/cerner/terra-framework/pull/1710/commits/beaa70782cbdf8729ccedd9c1bd337c9dab88314)

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/PinnedColumnsWithRowSelection.jsx
sycombs: Updated: [beaa707](https://github.com/cerner/terra-framework/pull/1710/commits/beaa70782cbdf8729ccedd9c1bd337c9dab88314)

packages/terra-worklist-data-grid/src/subcomponents/ColumnHeaderCell.jsx
sycombs: Can you elaborate? This is dynamically calculated based on the width of the last pinned column.

sycombs:
> Your low-light theme implementation is incorrect. It looks like you replaced the screen shots. However, the new ones do not apply the colors properly.

We have the variables created, just pending colors from UX:

- https://github.com/cerner/terra-framework/pull/1710/files#diff-6f5d751be4862d030a3fd116e1f14ccfee10502f37df242dd0b989c85046802cR6
- https://github.com/cerner/terra-framework/pull/1710/files#diff-9438a9395f34df30954d10be293421aab5f134100ec7952fdc33a06ee3d9187cR6

sdadn:
> > Your low-light theme implementation is incorrect. It looks like you replaced the screen shots. However, the new ones do not apply the colors properly.
> 
> We have the variables created, just pending colors from UX:
> 
> * https://github.com/cerner/terra-framework/pull/1710/files#diff-6f5d751be4862d030a3fd116e1f14ccfee10502f37df242dd0b989c85046802cR6
> * https://github.com/cerner/terra-framework/pull/1710/files#diff-9438a9395f34df30954d10be293421aab5f134100ec7952fdc33a06ee3d9187cR6

Fixed styling and updated wdio snapshots here: c3a6e5da3759dd2edfdf2e69c576feb54e4c9d62

packages/terra-worklist-data-grid/src/subcomponents/ColumnHeaderCell.jsx
cm9361: ok.  I misunderstood the code.

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/PinnedColumnsWithRowSelection.jsx
```diff
@@ -0,0 +1,147 @@
...
+          onChange={onRowSelectionModeChange}
+        />
+      </div>
+      <WorklistDataGrid
+        id="pinned-columns-with-row-selection"
+        pinnedColumns={cols.slice(0, 3)}
+        overflowColumns={cols.slice(3)}
+        rows={[...rows]}
+        rowHeaderIndex={rowHeaderIndex}
+        columnWidth="180px"
```
cm9361, line 116: The prop is defaultColumnWidth

packages/terra-worklist-data-grid/src/subcomponents/ColumnHeaderCell.module.scss
cm9361: Even if you work with UX later, shouldn't we establish the proper variables in this PR?

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/PinnedColumnsWithRowSelection.jsx
sdadn: Fixed here: 87abcfa6dc7cacf004d66941a3af33e0da86a752

sdadn: Used the correct proptype here: 013c4a4e79647edc07c2c7292a8d3d8c3e175c06

packages/terra-worklist-data-grid/src/subcomponents/ColumnHeaderCell.module.scss
sdadn: Added variables here: b358091d9295942c9a2bcc5ce4e02f8aa9113773

cm9361 (approved):

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
sdadn: Removed extra state variables here:
c7e169957a8ae9422a6857961fade208b4bc8a31
ccf753f2a86fa8e72b05eb5a4a4f19199510e76e

PR #1711, opened 2023-07-20
[terra-worklist-data-grid] Row selection for Worklist Data Grid

KoushikKalli:
### Summary
Select rows within the worklist data grid. 

### Testing

This change was tested using:

- [x] WDIO
- [X] Jest

### Reviews

- [x] UX review
- [x] Accessibility review

**This PR resolves:**

UXPLATFORM-9043


packages/terra-worklist-data-grid/src/subcomponents/Cell.module.scss
```diff
@@ -24,7 +24,8 @@
 
   // Create gradient background for masked cells
   .masked {
-    background: var(--terra-worklist-data-grid-mask-background, repeating-linear-gradient(-45deg, #dedfe0, #dedfe0 3px, transparent 3px, transparent 6px));
+    // background: var(--terra-worklist-data-grid-mask-background, repeating-linear-gradient(-45deg, #dedfe0, #dedfe0 3px, transparent 3px, transparent 6px));
+    background: var(--terra-worklist-data-grid-mask-background, repeating-linear-gradient(-45deg, #cfd2d3, #cfd2d3 3px, transparent 3px, transparent 6px));
```
cm9361, line 28: Why is this color changing?

packages/terra-worklist-data-grid/src/subcomponents/Row.jsx
```diff
@@ -159,7 +159,7 @@ function Row(props) {
 
   return (
     <tr
-      className={cx([isSelected ? 'row-selected' : 'worklist-data-grid-row', theme.className])}
+      className={cx([isSelected ? 'row-selected' : 'worklist-data-grid-row', hasRowSelection ? 'selectable' : null, theme.className])}
```
cm9361, line 162: You would want a pattern like this for your conditional class names

```
cx('worklist-data-grid-cell', {
    masked: isMasked,
    selectable: isSelectable && !isMasked,
    selected: isSelected && !isMasked,
    blank: !children,
  }, theme.className);
```

packages/terra-worklist-data-grid/src/subcomponents/Row.module.scss
```diff
@@ -3,12 +3,18 @@
 @import '../clinical-lowlight-theme/Row.module';
 
 :local {
+  tr.selectable:hover {
```
cm9361, line 6: I would use .worklsit-data-grid-row instead of the tr tag name.  Using the class name would be a better option.

cm9361, line 6: You could really create a parent .worklsit-data-grid-row selector to group all your styles and then use your parent operator (&) to handle these conditions.

packages/terra-worklist-data-grid/src/subcomponents/Row.module.scss
```diff
@@ -3,12 +3,18 @@
...
+    background-color: var(--terra-worklist-data-grid-hover-background-color, rgba(0, 146, 224, 0.12));
```
cm9361, line 7: If you are duplicating this style, does it need to be a mixin so that we can ensure consistency

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
```diff
@@ -117,8 +117,9 @@ function Cell(props) {
       event.stopPropagation();
       event.preventDefault();
     } else if (onCellSelect) {
-      onCellSelect({ rowId, columnId }, { row: rowIndex, col: columnIndex });
+      onCellSelect({ rowId, columnId }, { row: rowIndex, col: columnIndex }, !!event.shiftKey, false);
       event.stopPropagation();
+      event.preventDefault();
```
cm9361, line 122: Why do we need to prevent the default behavior on mouse down?  It should not cause issues, right?

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
```diff
@@ -130,7 +131,7 @@ function Cell(props) {
           event.stopPropagation();
           event.preventDefault();
         } else if (onCellSelect) {
-          onCellSelect({ rowId, columnId }, { row: rowIndex, col: columnIndex });
+          onCellSelect({ rowId, columnId }, { row: rowIndex, col: columnIndex }, !!event.shiftKey, true);
```
cm9361, line 134: It seems like you have too many unnamed objects here.  With having four arguments of multiple objects, it might be easier to create an object with named properties and have the objects as values.  That would probably be easier to maintain and follow.  Instead of returning whether the shift key was pressed, I would return the actual event as a separate argument.  So something like onCellSelect(event, object)

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
```diff
@@ -117,8 +117,9 @@ function Cell(props) {
       event.stopPropagation();
       event.preventDefault();
     } else if (onCellSelect) {
-      onCellSelect({ rowId, columnId }, { row: rowIndex, col: columnIndex });
+      onCellSelect({ rowId, columnId }, { row: rowIndex, col: columnIndex }, !!event.shiftKey, false);
```
cm9361, line 120: What is the last argument passed to onCellSelect?  Its purpose is not clear.

packages/terra-worklist-data-grid/src/subcomponents/Row.jsx
```diff
@@ -110,10 +110,10 @@ function Row(props) {
 
   const columnIndexOffSet = hasRowSelection ? 1 : 0;
 
-  const handleCellSelect = (rowIdColId, coordinates) => {
-    if (hasRowSelection) {
+  const handleCellSelect = (rowIdColId, coordinates, isShiftKeyPressed, isCellSelectedFromKeyboard) => {
+    if (hasRowSelection || isShiftKeyPressed) {
```
cm9361, line 114: If we are not in row selection mode, why do we care if the shift key has been pressed?  Using shift outside row selection mode would multi-select cells not rows, correct?

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/RowSelection.jsx
```diff
@@ -83,6 +145,16 @@ const RowSelection = () => {
     setSelectedRows([]);
   };
 
+  const enableSelectableRows = (rowIds) => {
+    const rowsToSelect = rows.filter(e => rowIds.indexOf(e.id) >= 0);
+    if (!rowSelectionModeRef.current.checked) {
+      rowSelectionModeRef.current.checked = true;
+    }
+    rowsToSelect.forEach(r => { r.isSelected = true; });
```
cm9361, line 153: Did you mean to use "==="?  If "isSelected" is a Boolean value, you don't need the extra comparison.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -257,28 +272,74 @@ function WorklistDataGrid(props) {
...
     } else if (onDisableSelectableRows) {
-      ariaLiveMsg.current = intl.formatMessage({ id: 'Terra.worklist-data-grid.row-selection-mode-disabled' });
+      multiRowSelection.current = null;
+      message = intl.formatMessage({ id: 'Terra.worklist-data-grid.row-selection-mode-disabled' });
       onDisableSelectableRows();
     }
+    setAriaLiveMsg(message);
+  };
+
+  const handleEnableSelectableRows = (newEndRowIndex, selectionData) => {
```
cm9361, line 291: I am a little confused by the method name.  Doesn't "hasRowSelection" enable the ability to select multiple rows?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -106,7 +106,11 @@ const propTypes = {
   onDisableSelectableRows: PropTypes.func,
 
   /**
-   * Boolean indicating whether or not the DataGrid should allow entire rows to be selected. An additional column will be
+   * Callback function to enable row selection mode and select a given set of rows. Parameters: `function(arrayOfRowIds)`
+   */
+  onEnableSelectableRows: PropTypes.func,
```
cm9361, line 111: Should this be a Boolean prop instead of a callback function?  I would think that onRowSelect would be updated to return an array of rows.  We don't want multiple options for returning the selected items.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -505,6 +624,7 @@ function WorklistDataGrid(props) {
         aria-label={ariaLabel}
         className={cx('worklist-data-grid', theme.className)}
         onKeyDown={handleKeyDown}
+        onKeyUp={handleKeyUp}
```
cm9361, line 627: I would not expect us to have to handle a key up event.  This makes me think there is an issue with the design.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -338,24 +399,70 @@ function WorklistDataGrid(props) {
     selectCell(cellRowIdColId, cellCoordinates);
   };
 
-  const handleRowSelection = (rowId, rowIndex, selectedCellCoordinates) => {
+  const handleRowSelection = (rowId, rowIndex, selectedCellCoordinates, enableSelectableRows, useLastSelectedRowAsAnchor) => {
```
cm9361, line 402: Who is passing these new arguments?  What does it mean to enable selectable rows?  Wouldn't the WorklistDataGrid be determining if the last selected row was used as an anchor since it is processing the key events?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -338,24 +399,70 @@ function WorklistDataGrid(props) {
...
+        }
         nextRow -= 1;
         break;
       case KeyCode.KEY_DOWN:
+        if (event.shiftKey) {
+          activateMultiRowSelection = true;
+        }
         nextRow += 1;
         break;
+      case KeyCode.KEY_SHIFT:
```
cm9361, line 461: The keydown event has a Boolean check for if the shift key is pressed.  You should not need a handler for the shift key.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -338,24 +399,70 @@ function WorklistDataGrid(props) {
...
+        if (cellCoordinates.row !== 0 && !multiRowSelection?.current?.anchorRow) {
+          multiRowSelection.current = { anchorRow: cellCoordinates.row, previousSelectionEndRow: cellCoordinates.row };
```
cm9361, line 463: Why do we need both an anchor row and previous selection end row?

packages/terra-worklist-data-grid/src/subcomponents/Cell.module.scss
KoushikKalli: Reverted

packages/terra-worklist-data-grid/src/subcomponents/Row.jsx
KoushikKalli: Updated

packages/terra-worklist-data-grid/src/subcomponents/Row.module.scss
KoushikKalli: Updated

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
KoushikKalli: It was added to prevent the default behavior of text selection. After our discussion offline, removed preventDefault() and used user-select to avoid shift click text selection.

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
KoushikKalli: Created an object with all the arguments. I refrained from passing event from Cell to Row as Cell itself is handling it's needs.

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
KoushikKalli: The anchor row for selecting multiple rows differs based on the selection made with click or space key. The purpose of the last argument is to recognize if the selection was initiated by keyboard.

packages/terra-worklist-data-grid/src/subcomponents/Row.jsx
KoushikKalli: The requirements demand us to enable row selection mode if it is off for Shift + Up/Down arrow keys, Shift + Space and Shift + mouse click.

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/RowSelection.jsx
KoushikKalli: No, it's assignment.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -338,24 +399,70 @@ function WorklistDataGrid(props) {
...
+      handleEnableSelectableRows(rowIndex, multiRowSelectionWithLastSelected.current);
+      multiRowSelectionWithLastSelected.current.previousSelectionEndRow = rowIndex;
+    }
+  };
+
+  const handleKeyUp = (event) => {
+    const key = event.keyCode;
+    switch (key) {
+      case KeyCode.KEY_SHIFT:
+        multiRowSelection.current = null;
```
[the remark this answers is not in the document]
KoushikKalli: Updated.

mjpalazzo:
@KoushikKalli - I encountered some issues with the Pinned Column with Row Selection example. I was in the grid navigating the cells and selecting/deselecting cells. As I used Right Arrow to navigate back to the pinned columns, the cell adjacent to the pin was partially covered and I could not figure out how to get it fully visible again.
I also noticed that cells in pinned columns cannot be selected. Is that intentional?
I used the mouse to turn on row selection and then navigated to the check box in row 3 and could not select it with Space or Shift Space. I will send a screenshot on the Teams channel.

packages/terra-worklist-data-grid/tests/jest/__snapshots__/WorklistDataGrid.test.jsx.snap
sdadn, line 1: Are additional Jest tests being added?

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
```diff
@@ -137,22 +137,24 @@ function Cell(props) {
...
     columnContext.setCellAriaLiveMessage(intl.formatMessage({ id: 'Terra.worklist-data-grid.resume-navigation' }));
   };
 
+  /**
+   * Handles mouse down event for cell
+   */
+  const onMouseDown = useCallback((event) => {
+    if (!isFocusTrapEnabled) {
+      onCellSelect({
+        rowId, columnId, rowIndex, columnIndex, multiSelect: !!event.shiftKey, selectedByKeyboard: false,
```
cm9361, line 153: Why are we using "!!" on event.shiftKey?  That is already a Booelan value.

packages/terra-worklist-data-grid/src/subcomponents/Cell.module.scss
```diff
@@ -2,6 +2,11 @@
 @import '../orion-fusion-theme/Cell.module';
 @import '../clinical-lowlight-theme/Cell.module';
 
+@mixin terra-worklist-data-grid-selectable-hover {
+  background-color: var(--terra-worklist-data-grid-hover-background-color, #d8e5ed);
+  cursor: pointer;
+}
```
cm9361, line 8: Why is the mixin being defined in the Cell component if it is used in the Cell and Row components.  Shouldn't it be in its own separate file that can be imported?

packages/terra-worklist-data-grid/src/subcomponents/Row.jsx
```diff
@@ -159,7 +159,10 @@ function Row(props) {
 
   return (
     <tr
-      className={cx([isSelected ? 'row-selected' : 'worklist-data-grid-row', theme.className])}
+      className={cx('worklist-data-grid-row', {
+        'row-selected': isSelected,
+        selectable: hasRowSelection,
```
cm9361, line 164: You can't use only the hasRowSelection prop.  You also have to ensure that the row is selectable.  There should be an isSelectable prop on the rowShape.  That needs to be added so that consumers can specify whether a row is selectable.

packages/terra-worklist-data-grid/src/subcomponents/Row.jsx
```diff
@@ -159,7 +159,10 @@ function Row(props) {
 
   return (
     <tr
-      className={cx([isSelected ? 'row-selected' : 'worklist-data-grid-row', theme.className])}
+      className={cx('worklist-data-grid-row', {
+        'row-selected': isSelected,
```
cm9361, line 163: I would make this class "selected" instead of "row-selected" to match the pattern for the Cell component.  This is already a Row component so we don't really need to prefix with "row-" with class mangling in place.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
cm9361: I still don't believe we should need a key up event.  We should not have manually track the Shift key.

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/RowSelection.jsx
```diff
@@ -146,6 +281,12 @@ const RowSelection = () => {
           disableSelectableRows();
         }}
         onColumnSelect={onColumnSelect}
+        onEnableRowSelection={() => {
```
cm9361, line 284: You would not want to use an anonymous function.  We would not want that pattern replicated.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -320,31 +334,95 @@ function WorklistDataGrid(props) {
   );
 
   const handleClearRowSelection = () => {
+    let message = '';
+    rowSelection.current.lastIndividuallySelectedRow = null;
     if (isAnyRowSelected()) {
-      setAriaLiveMessage(intl.formatMessage({ id: 'Terra.worklist-data-grid.all-rows-unselected' }));
+      message = intl.formatMessage({ id: 'Terra.worklist-data-grid.all-rows-unselected' });
```
cm9361, line 340: What required this change?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -398,13 +477,51 @@ function WorklistDataGrid(props) {
...
+  const handleCellSelection = (selectionDetails) => {
+    selectCell(selectionDetails);
   };
 
-  const handleRowSelection = (rowId, rowIndex, selectedCellCoordinates) => {
-    handleCellSelectionChange(null, null, selectedCellCoordinates);
-    selectRows(false, rowId, rowIndex);
+  const handleRowSelection = (selectionDetails) => {
+    handleCellSelectionChange(null, null, { row: selectionDetails.rowIndex, col: selectionDetails.columnIndex });
+    const selectionData = selectionDetails.selectedByKeyboard ? rowSelection.current.lastIndividuallySelectedRow : rowSelection?.current?.focussedRow;
```
cm9361, line 486: This variable name doesn't really seem to clearly describe the purpose of the variable.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -320,31 +334,95 @@ function WorklistDataGrid(props) {
...
+      id: hasSelectableRows
+        ? 'Terra.worklist-data-grid.row-selection-multiple-rows-selected'
+        : 'Terra.worklist-data-grid.row-selection-mode-enabled',
+    });
+
+    ariaMessage += intl.formatMessage({
+      id: rowIdsToSelect.length === 1
+        ? 'Terra.worklist-data-grid.row-selection-template'
+        : 'Terra.worklist-data-grid.row-selection-selected-rows-range',
+    }, { row: selectionData.anchorRow, endRow: newEndRowIndex });
```
cm9361, line 398: How are you separating these two sentences?  There is no punctuation between them.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -320,31 +334,95 @@ function WorklistDataGrid(props) {
...
+        msgId = 'Terra.worklist-data-grid.row-selection-cleared-template';
+      }
       rowLabel = rows[rowIndex - 1].ariaLabel || (rowIndex + 1);
     }
     setAriaLiveMessage(intl.formatMessage({ id: msgId }, { row: rowLabel }));
     if (selectAllRows && onRowSelectAll) {
       onRowSelectAll();
     } else if (onRowSelect) {
-      onRowSelect(rowId);
+      onRowSelect([{ rowId, isSelected: isSelectAction }]);
```
cm9361, line 425: For a single selection, I would expect that the consumer would be managing the selection state for the component.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -398,13 +477,51 @@ function WorklistDataGrid(props) {
...
+      } else {
+        // no anchor row
+        selectRows(false, selectionDetails.rowId, selectionDetails.rowIndex);
+      }
+    } else if (selectionDetails.selectedByKeyboard) {
+      // Shift + Space
+      rowSelection.current.isRowSelectionEnabledByGrid = !hasSelectableRows;
+      rowSelection.current.lastIndividuallySelectedRow = { anchorRow: selectionDetails.rowIndex };
+      selectMultipleRows(selectionDetails.rowIndex, rowSelection.current.lastIndividuallySelectedRow);
+      rowSelection.current.lastIndividuallySelectedRow.previousSelectionEndRow = selectionDetails.rowIndex;
```
cm9361, line 508: Why does the lastIndividuallySelectedRow need to be an object?  There seem to be quite a few layers to maintain this new logic.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -398,13 +477,51 @@ function WorklistDataGrid(props) {
...
+    const selectionData = selectionDetails.selectedByKeyboard ? rowSelection.current.lastIndividuallySelectedRow : rowSelection?.current?.focussedRow;
+
+    if (!selectionDetails.multiSelect) {
+      if (hasSelectableRows) {
+        // regular click or space key
+        selectRows(false, selectionDetails.rowId, selectionDetails.rowIndex);
+      }
+    } else if (hasSelectableRows) {
+      // Shift key is pressed
+      if (selectionData?.anchorRow) {
```
cm9361, line 495: How does an anchor row let us know that the shift key has been pressed?

packages/terra-framework-docs/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Changelog
 
 ## Unreleased
+* Added
+  * Added examples and tests for `terra-worklist-data-grid` to interact with cells containing focusable elements.
+
+* Added
+  * Added example and tests for `terra-worklist-data-grid` to select rows via mouse and keyboard.
```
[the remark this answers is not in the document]
KoushikKalli: Changed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -118,7 +118,11 @@ const propTypes = {
   onDisableSelectableRows: PropTypes.func,
 
   /**
-   * Boolean indicating whether or not the DataGrid should allow entire rows to be selected. An additional column will be
+   * Callback function to enable row selection mode. Parameters: `function()`
```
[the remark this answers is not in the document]
KoushikKalli: Changed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -320,31 +334,95 @@ function WorklistDataGrid(props) {
...
+    });
+
+    ariaMessage += intl.formatMessage({
+      id: rowIdsToSelect.length === 1
+        ? 'Terra.worklist-data-grid.row-selection-template'
+        : 'Terra.worklist-data-grid.row-selection-selected-rows-range',
+    }, { row: selectionData.anchorRow, endRow: newEndRowIndex });
+
+    setAriaLiveMessage(ariaMessage);
+    onRowSelect(rowIdsToSelect.concat(rowIdsToUnselect));
```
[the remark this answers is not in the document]
KoushikKalli: The intention was to send a list of row Ids and action to take for each rowId - select or unselect. I've changed the signature now to pass two arrays - rowIdsToSelect & rowIdsToUnselect

packages/terra-worklist-data-grid/src/subcomponents/Cell.module.scss
KoushikKalli: Changed

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/RowSelection.jsx
KoushikKalli: Changed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: The merge conflict initiated this change.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -320,31 +334,95 @@ function WorklistDataGrid(props) {
...
+      onEnableRowSelection();
+    }
+
+    // Rows in range to remain selected
+    const selectionStartRowIndex = Math.min(selectionData.anchorRow, newEndRowIndex);
+    const selectionEndRowIndex = Math.max(selectionData.anchorRow, newEndRowIndex);
+
+    // We are subtracting 1 to accommodate for the column header in the grid.
+    const rowsToSelect = rows.slice(selectionStartRowIndex - 1, selectionEndRowIndex);
+    const rowIdsToSelect = rowsToSelect.map(r => ({ rowId: r.id, isSelected: true }));
```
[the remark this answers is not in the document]
KoushikKalli: Changed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Added period between two sentences.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Updated

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -398,13 +477,51 @@ function WorklistDataGrid(props) {
...
+        selectMultipleRows(selectionDetails.rowIndex, selectionData);
+        selectionData.previousSelectionEndRow = selectionDetails.rowIndex;
+      } else {
+        // no anchor row
+        selectRows(false, selectionDetails.rowId, selectionDetails.rowIndex);
+      }
+    } else if (selectionDetails.selectedByKeyboard) {
+      // Shift + Space
+      rowSelection.current.isRowSelectionEnabledByGrid = !hasSelectableRows;
+      rowSelection.current.lastIndividuallySelectedRow = { anchorRow: selectionDetails.rowIndex };
```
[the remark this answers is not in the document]
KoushikKalli: Last individually selected row is the row the user selects via a single click or the space bar. The existing requirement called for the use of last individually selected row and not last selected row.
We do not store last selected row, only last individually selected row, which is used by Shift+Space as anchor for its selection.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: We updated the name of the variable to rangeUsingLastIndividuallySelectedRow since it contains the anchor for the range and the end point in the range.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: The anchor row does not. It is the condition in the _if_ block that allows us to make the determination.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Updated.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -320,31 +334,95 @@ function WorklistDataGrid(props) {
...
     if (isAnyRowSelected()) {
-      setAriaLiveMessage(intl.formatMessage({ id: 'Terra.worklist-data-grid.all-rows-unselected' }));
+      message = intl.formatMessage({ id: 'Terra.worklist-data-grid.all-rows-unselected' });
       // Esc (while in row selection mode and rows are selected): Clear selection
       if (onClearSelectedRows) {
         onClearSelectedRows();
       }
     } else if (onDisableSelectableRows) {
-      setAriaLiveMessage(intl.formatMessage({ id: 'Terra.worklist-data-grid.row-selection-mode-disabled' }));
+      rowSelection.current.focussedRow = null;
```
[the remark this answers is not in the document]
KoushikKalli: Updated

packages/terra-worklist-data-grid/src/subcomponents/Row.jsx
KoushikKalli: As we discussed offline, @eawww will follow up on this and let us know if its required.

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
KoushikKalli: Updated

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -398,13 +477,51 @@ function WorklistDataGrid(props) {
...
+
+  const handleKeyUp = (event) => {
+    const key = event.keyCode;
+    switch (key) {
+      case KeyCode.KEY_SHIFT:
+        rowSelection.current.focussedRow = null;
+        break;
+      default:
+        break;
+    }
```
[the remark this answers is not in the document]
KoushikKalli: It is deliberate. We thought it will be easier for additional keys to be added later.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -398,13 +477,51 @@ function WorklistDataGrid(props) {
...
+  const handleRowSelection = (selectionDetails) => {
+    handleCellSelectionChange(null, null, { row: selectionDetails.rowIndex, col: selectionDetails.columnIndex });
+    const selectionData = selectionDetails.selectedByKeyboard ? rowSelection.current.lastIndividuallySelectedRow : rowSelection?.current?.focussedRow;
+
+    if (!selectionDetails.multiSelect) {
+      if (hasSelectableRows) {
+        // regular click or space key
+        selectRows(false, selectionDetails.rowId, selectionDetails.rowIndex);
+      }
+    } else if (hasSelectableRows) {
```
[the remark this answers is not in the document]
KoushikKalli: The if-else block was coded to handle the different key and mouse combinations for row selection and it factors in if row selection is turned on or off. This logic could change with new requirements in an upcoming jira. We will consider simplifying the block in that jira.

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/RowSelection.jsx
```diff
@@ -104,6 +226,24 @@ const RowSelection = () => {
...
+      setSelectedRows(rowIdsToSelect);
+    }
+  };
+
+  const enableRowSelection = () => {
+    if (!rowSelectionModeRef.current.checked) {
+      rowSelectionModeRef.current.checked = true;
+    }
+    setHasSelectableRows(true);
+  };
```
cm9361, line 245: The functions in your example should probably leverage the useCallback hook.

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/RowSelection.jsx
```diff
@@ -146,6 +276,7 @@ const RowSelection = () => {
           disableSelectableRows();
```
cm9361, line 276: We would want our consumers to pass anonymous functions to us.  It will cause unnecessary rerenders.

packages/terra-framework-docs/src/terra-dev-site/test/worklist-data-grid/WorklistDataGridWithRowSelection.test.jsx
```diff
@@ -38,6 +38,17 @@ const WorklistDataGridWithRowSelection = () => {
...
 
+  const onRowSelect = (rowIdsToSelect, rowIdsToUnselect) => {
+    if (rowIdsToUnselect) {
+      rows.forEach((r, index) => { if (rowIdsToUnselect.indexOf(r.id) >= 0) { rows[index].isSelected = false; } });
+    }
+
+    if (rowIdsToSelect) {
+      rows.forEach((r, index) => { if (rowIdsToSelect.indexOf(r.id) >= 0) { rows[index].isSelected = true; } });
+      setSelectedRows(rowIdsToSelect);
+    }
```
cm9361, line 49: Shouldn't you be able to handle this logic in one forEach loop?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -320,31 +335,94 @@ function WorklistDataGrid(props) {
...
     if (isAnyRowSelected()) {
-      setAriaLiveMessage(intl.formatMessage({ id: 'Terra.worklist-data-grid.all-rows-unselected' }));
+      message = intl.formatMessage({ id: 'Terra.worklist-data-grid.all-rows-unselected' });
       // Esc (while in row selection mode and rows are selected): Clear selection
       if (onClearSelectedRows) {
         onClearSelectedRows();
       }
     } else if (onDisableSelectableRows) {
-      setAriaLiveMessage(intl.formatMessage({ id: 'Terra.worklist-data-grid.row-selection-mode-disabled' }));
+      rowSelection.current.rangeUsingFocusedRow = null;
```
cm9361, line 347: What happens if the consumer doesn't clear the selection?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -320,31 +335,94 @@ function WorklistDataGrid(props) {
...
+      message = intl.formatMessage({ id: 'Terra.worklist-data-grid.row-selection-mode-disabled' });
```
cm9361, line 348: Is row selection mode being disabled or turned off?  Is this text from our requirements?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -320,31 +335,94 @@ function WorklistDataGrid(props) {
...
     }
     setAriaLiveMessage(intl.formatMessage({ id: msgId }, { row: rowLabel }));
     if (selectAllRows && onRowSelectAll) {
       onRowSelectAll();
     } else if (onRowSelect) {
-      onRowSelect(rowId);
+      if (isSelectAction) {
+        onRowSelect([rowId]);
+      } else {
+        onRowSelect([], [rowId]);
```
cm9361, line 424: Why are we passing an empty array for the first argument?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -320,31 +335,94 @@ function WorklistDataGrid(props) {
...
   const selectRows = (selectAllRows, rowId, rowIndex) => {
     let rowLabel;
+    let isSelectAction = true;
     let msgId = 'Terra.worklist-data-grid.all-rows-selected';
+    // Reset last selected row when all rows are selected.
+    rowSelection.current.rangeUsingLastIndividuallySelectedRow = null;
     if (!selectAllRows) {
-      const isSelectAction = !rows[rowIndex - 1].isSelected; // Determine if this is select or unselected.
-      msgId = isSelectAction ? 'Terra.worklist-data-grid.row-selection-template' : 'Terra.worklist-data-grid.row-selection-cleared-template';
+      isSelectAction = !rows[rowIndex - 1].isSelected; // Determine if this is select or unselected.
```
cm9361, line 407: We should not be controlling the selection state.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -320,31 +335,94 @@ function WorklistDataGrid(props) {
...
+        ? 'Terra.worklist-data-grid.row-selection-multiple-rows-selected'
+        : 'Terra.worklist-data-grid.row-selection-mode-enabled',
+    })}. ${intl.formatMessage({
+      id: rowIdsToSelect.length === 1
+        ? 'Terra.worklist-data-grid.row-selection-template'
+        : 'Terra.worklist-data-grid.row-selection-selected-rows-range',
+    }, { row: rangeInfo.anchorRow, endRow: newEndRowIndex })}`;
+
+    setAriaLiveMessage(ariaMessage);
+    onRowSelect(rowIdsToSelect, rowIdsToUnselect);
```
cm9361, line 397: I would expect that we would only be passing one array back to consumers.  We should not need to pass two separate arrays.

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/RowSelection.jsx
KoushikKalli: Could you elaborate why and it's need?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: That should be okay because there aren't any rows selected and any further selection will reconfigure the range information. As we discussed yesterday in the meeting with Eric & Chris, the issue around the consumer not honoring the intended action of the user needs more evaluation especially as it relates to aria messages.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: The requirements do not specify any messages and we did not hear any objections to the current messages during our demos or UX reviews.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: We need not, but we chose to pass two separate arrays to make it clearer for the consumers.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: onRowSelect accepts two arguments to explicitly communicate the user's intention. The first being rows they would like to select and the second are the rows they would like to unselect.

packages/terra-worklist-data-grid/src/subcomponents/Mixins.scss
```diff
@@ -0,0 +1,4 @@
+@mixin terra-worklist-data-grid-selectable-hover {
+  background-color: var(--terra-worklist-data-grid-hover-background-color, #d8e5ed);
+  cursor: pointer;
+}
```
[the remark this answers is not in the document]
sdadn: That should be fixed by UXPLATFORM-9473.

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/RowSelection.jsx
KoushikKalli: Updated

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/RowSelection.jsx
KoushikKalli: Discussed offline. Updated

mjpalazzo:
@KoushikKalli - The row selection capabilities are working well. I tested keyboard navigation with Chrome and Edge and I tested with JAWS on Edge. I was able to use Shift Space to select a range of rows. JAWS announced the range selected. I also used Control Space to add non-adjacent rows to the selection. I also deselected with Shift Space and Control Space. I would like to do a little more testing with Safari and VoiceOver tomorrow. Great work!

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
cm9361: But the whole strategy is supposed to be updating such that we are providing message based on input from consumers.  Should we be adding more of these messages?  What does it mean for the row selection mode to be disabled?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -346,35 +437,95 @@ function WorklistDataGrid(props) {
...
-    setFocusedCol(selectedCellCoordinates.col);
-    setCurrentSelectedCell(null);
+  const handleRowSelection = useCallback((selectionDetails) => {
+    if (!selectionDetails?.selectedByKeyboard) {
+      handleFocus.current = false;
+    }
 
-    if (onRowSelect) {
-      onRowSelect(rowId);
+    handleCellSelectionChange(null, null, { row: selectionDetails.rowIndex, col: selectionDetails.columnIndex });
```
cm9361, line 488: handleRowSelection should not be calling handleCellSelectionChange.  We just removed that method in a recent PR.  The function is not clear when you are passing null for the row id and column id.  We are trying to do too much with one method.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -328,6 +347,78 @@ function WorklistDataGrid(props) {
...
+    onRowSelect(rowIdsToSelect, rowIdsToUnselect);
+  }, [hasSelectableRows, intl, onEnableRowSelection, onRowSelect, rows]);
+
+  const selectRows = useCallback((rowId, rowIndex) => {
+    let isSelectAction = true;
+    let msgId = 'Terra.worklist-data-grid.all-rows-selected';
+    // Reset last selected row when all rows are selected.
+    rowSelection.current.rangeUsingLastIndividuallySelectedRow = null;
+
+    isSelectAction = !rows[rowIndex - 1].isSelected; // Determine if this is select or unselected.
```
cm9361, line 402: We are not managing the selection state in our control.  We should not be determining the selection action.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -328,6 +347,78 @@ function WorklistDataGrid(props) {
...
+      msgId = 'Terra.worklist-data-grid.row-selection-cleared-template';
+    }
+    const rowLabel = rows[rowIndex - 1].ariaLabel || (rowIndex + 1);
+
+    setAriaLiveMessage(intl.formatMessage({ id: msgId }, { row: rowLabel }));
+    if (onRowSelect) {
+      if (isSelectAction) {
+        onRowSelect([rowId]);
+      } else {
+        onRowSelect([], [rowId]);
```
cm9361, line 417: The selection state should not change how we call the callback.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -346,35 +437,95 @@ function WorklistDataGrid(props) {
...
-    const rowLabel = rows[rowIndex - 1].ariaLabel || (rowIndex + 1);
+    const cellGridCoordinates = { row: selectionDetails.rowIndex, col: selectionDetails.columnIndex };
+    // If current cell is selected, do nothing.
+    if (isCellSelected(cellGridCoordinates.row, cellGridCoordinates.col)) {
+      return;
+    }
+    // Determine if the cell selection should proceed.
+    const cellDataCoordinates = mapGridCellToDataCell(cellGridCoordinates);
+    const cell = rows[cellDataCoordinates.row].cells[cellDataCoordinates.col];
+    if ((cell.isSelectable === false) || cell.isMasked) {
```
cm9361, line 471: selectionDetails has an isCellSelectable property.  You can use that as was done in the previous logic.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -346,35 +437,95 @@ function WorklistDataGrid(props) {
...
+      setFocusedRowCol(cellGridCoordinates.row, cellGridCoordinates.col, true);
+      return;
+    }
 
-    setAriaLiveMessage(intl.formatMessage({ id: rowSelectionMessageId }, { row: rowLabel }));
+    // Make note of cell that is currently selected.
+    handleCellSelectionChange(selectionDetails.rowId, selectionDetails.columnId, cellGridCoordinates);
```
cm9361, line 477: Why did we need to change from our use of the state variables.  We had eliminated this method.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -346,35 +437,95 @@ function WorklistDataGrid(props) {
...
   }, [hasSelectableRows, intl, onColumnSelect]);
 
-  const handleCellSelection = useCallback((cellRowIdColId, cellCoordinates, cellSelectable) => {
+  const handleCellSelectionChange = useCallback((rowId, columnId, cellCoordinates) => {
     if (!hasSelectableRows) {
-      setAriaLiveMessage(intl.formatMessage({ id: 'Terra.worklist-data-grid.cell-selection-template' },
-        { row: cellCoordinates.row + 1, column: cellCoordinates.col + 1 }));
+      setAriaLiveMessage(intl.formatMessage({
+        id: rowId ? 'Terra.worklist-data-grid.cell-selection-template' : 'Terra.worklist-data-grid.cell-selection-cleared',
+      }, { row: cellCoordinates.row + 1, column: cellCoordinates.col + 1 }));
```
cm9361, line 444: Looks like we added back logic that we removed to simplify the implementation and make more readable.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
cm9361: I still don't think having the two array is going to work out well, especially since we don't provide a selection state for a single row selection.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -346,35 +437,95 @@ function WorklistDataGrid(props) {
...
+      onCellSelect(selectionDetails.rowId, selectionDetails.columnId);
+    }
+  }, [handleCellSelectionChange, isCellSelected, mapGridCellToDataCell, onCellSelect, rows, setFocusedRowCol]);
 
-    setFocusedRow(selectedCellCoordinates.row);
-    setFocusedCol(selectedCellCoordinates.col);
-    setCurrentSelectedCell(null);
+  const handleRowSelection = useCallback((selectionDetails) => {
+    if (!selectionDetails?.selectedByKeyboard) {
+      handleFocus.current = false;
```
cm9361, line 485: This should only be set to false by the mouse click

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
```diff
@@ -129,22 +129,24 @@ function Cell(props) {
...
     columnContext.setCellAriaLiveMessage(intl.formatMessage({ id: 'Terra.worklist-data-grid.resume-navigation' }));
   };
 
+  /**
+   * Handles mouse down event for cell
+   */
+  const onMouseDown = ((event) => {
+    if (!isFocusTrapEnabled) {
+      onCellSelect({
+        rowId, columnId, rowIndex, columnIndex, multiSelect: event.shiftKey, selectedByKeyboard: false, isCellSelectable: (!isMasked && isSelectable),
```
cm9361, line 145: Did the linter not ask for this to be formatted differently such as a prop on each line?

packages/terra-worklist-data-grid/src/orion-fusion-theme/ColumnHeaderCell.module.scss
```diff
@@ -8,12 +8,12 @@
...
-    
+
     // Pinned Columns divider
     --terra-worklist-data-grid-pinned-columns-divider-background-color: #808080;
     --terra-worklist-data-grid-pinned-columns-divider-box-shadow: 0;
-    
+
     // Hover
-    --terra-worklist-data-grid-hover-background-color: #d5e7f2;
+    --terra-worklist-data-grid-header-hover-background-color: #d5e7f2;
```
cm9361, line 17: Is the header supposed to have a different hover color?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -410,15 +561,26 @@ function WorklistDataGrid(props) {
...
         nextRow -= 1;
         break;
       case KeyCode.KEY_DOWN:
+        if (event.shiftKey) {
+          activateMultiRowSelection = true;
+        }
         nextRow += 1;
         break;
+      case KeyCode.KEY_SHIFT:
+        if (cellCoordinates.row !== 0 && !rowSelection?.current?.rangeUsingFocusedRow?.anchorRow) {
```
cm9361, line 580: How could rowSelection be null if it is a reference variable?  Do we need the "?" after it?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -492,6 +654,18 @@ function WorklistDataGrid(props) {
...
+      // Handle the row selection case.
+      if (nextRow > rows.length) {
+        nextRow = rows.length;
+      } else if (nextRow <= 0) {
+        nextRow = 1; // Only non-header rows can be selected.
+      }
+      rowSelection.current.isRowSelectionModeToggledByGrid = !hasSelectableRows;
+      selectMultipleRows(nextRow, rowSelection.current.rangeUsingFocusedRow);
+      rowSelection.current.rangeUsingFocusedRow.previousSelectionEndRow = nextRow;
+    }
```
cm9361, line 667: This event does bubble.  What happens if the users is in an input or text area element or contenteditable div?  We would not want to execute this new logic on each key press, right?

mjpalazzo:
@KoushikKalli - I did some testing with Safari and VoiceOver.
Keyboard navigation with Safari works as required.
VoiceOver generally works well, but I struggle with the difference in VoiceOver navigation and visual focuswith Up/Down Arrow versus Control+Option+Up/Down Arrow. When VoiceOver focus indicator is on one cell but visual focus is on another Space selects the cell with visual focus, not what VoiceOver announced. This could be user error, but we need to discuss this behavior. It is confusing!

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -346,35 +437,95 @@ function WorklistDataGrid(props) {
...
+    setFocusedRowCol(cellCoordinates.row, cellCoordinates.col, true);
+    if ((rowId !== currentSelectedCell?.rowId) || (columnId !== currentSelectedCell?.columnId)) {
+      setCurrentSelectedCell((rowId && columnId) ? { rowId, columnId } : null);
     }
+  }, [currentSelectedCell?.columnId, currentSelectedCell?.rowId, hasSelectableRows, intl, setFocusedRowCol]);
 
-    setFocusedRow(cellCoordinates.row);
-    setFocusedCol(cellCoordinates.col);
-    setCurrentSelectedCell({ rowId: cellRowIdColId.rowId, columnId: cellRowIdColId.columnId });
+  const mapGridCellToDataCell = useCallback((cellGridCoordinates) => (
```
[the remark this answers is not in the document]
KoushikKalli: Removed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Updated

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Removed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Removed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -346,35 +437,95 @@ function WorklistDataGrid(props) {
...
+    // The grid has an additional row for the heading and
+    // may have an additional column when when selection is active.
+    { row: cellGridCoordinates.row - 1, col: cellGridCoordinates.col + (hasSelectableRows ? -1 : 0) }
+  ), [hasSelectableRows]);
 
-    if (cellSelectable && onCellSelect) {
-      onCellSelect(cellRowIdColId.rowId, cellRowIdColId.columnId);
+  const handleCellSelection = useCallback((selectionDetails) => {
+    if (!selectionDetails?.selectedByKeyboard) {
+      handleFocus.current = false;
```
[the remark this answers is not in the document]
KoushikKalli: No longer necessary. Removed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: No longer necessary. Removed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Removed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -597,6 +771,7 @@ function WorklistDataGrid(props) {
         aria-label={ariaLabel}
         className={cx('worklist-data-grid', theme.className)}
         onKeyDown={handleKeyDown}
+        onKeyUp={handleKeyUp}
```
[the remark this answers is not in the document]
KoushikKalli: The events will still fire. As discussed, we would have to devise a solution to handle events that originate from consumer components on the grid. We will also have to figure what should happen when the Worklist Data Grid keyboard shortcuts are triggered from within consumer components on the grid. For example, what happens when Ctrl+A is pressed from inside the input textbox within the grid. We may have to open another story if necessary.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: The events will still fire. As discussed, we would have to devise a solution to handle events that originate from consumer components on the grid. We will also have to figure what should happen when the Worklist Data Grid keyboard shortcuts are triggered from within consumer components on the grid. For example, what happens when Ctrl+A is pressed from inside the input textbox within the grid. We may have to open another story if necessary.

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/RowSelection.jsx
```diff
@@ -104,6 +226,25 @@ const RowSelection = () => {
...
+        rows[index].isSelected = false;
+      }
+
+      if (rowIdsToSelect && rowIdsToSelect.indexOf(row.id) >= 0) {
+        rows[index].isSelected = true;
+      }
+    });
+  }, [rows]);
+
+  const enableRowSelection = () => {
```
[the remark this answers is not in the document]
KoushikKalli: Updated

packages/terra-worklist-data-grid/src/orion-fusion-theme/ColumnHeaderCell.module.scss
KoushikKalli: The variable was renamed by a different story in terra-default-theme. This change ensures same variable is used in the other two themes.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -328,6 +347,78 @@ function WorklistDataGrid(props) {
...
+      id: rowIdsToSelect.length === 1
+        ? 'Terra.worklist-data-grid.row-selection-template'
+        : 'Terra.worklist-data-grid.row-selection-selected-rows-range',
+    }, { row: rangeInfo.anchorRow, endRow: newEndRowIndex })}`;
+
+    setAriaLiveMessage(ariaMessage);
+    onRowSelect(rowIdsToSelect, rowIdsToUnselect);
+  }, [hasSelectableRows, intl, onEnableRowSelection, onRowSelect, rows]);
+
+  const selectRows = useCallback((rowId, rowIndex) => {
```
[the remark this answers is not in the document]
KoushikKalli: Updated

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: As we discussed offline, not making further change.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: As we discussed offline, not making further change.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Removed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -328,6 +347,78 @@ function WorklistDataGrid(props) {
...
+        // The range was moved up towards the anchor so rows that no longer qualify for the range need to be unselected. New endRangeIndex becomes ordered End
+        rowIdsToUnselect = rows.slice(Math.max(rangeInfo.anchorRow, newEndRowIndex), rangeInfo.previousSelectionEndRow).map(row => row.id);
+      }
+    }
+
+    const ariaMessage = `${intl.formatMessage({
+      id: hasSelectableRows
+        ? 'Terra.worklist-data-grid.row-selection-multiple-rows-selected'
+        : 'Terra.worklist-data-grid.row-selection-mode-enabled',
+    })}. ${intl.formatMessage({
```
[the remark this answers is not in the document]
KoushikKalli: Updated

sdadn:
The _Row Selection with Pinned Columns_ example doesn't seem to be working. Can you take a look at it?

cm9361:
> The _Row Selection with Pinned Columns_ example doesn't seem to be working. Can you take a look at it?

I observed this issue as well.  It would seem that some unit test should have failed since this functionality is not working.  Do we need better test coverage?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -271,8 +276,17 @@ function WorklistDataGrid(props) {
...
+
+    if (!hasSelectableRows) {
+      rowSelection.current.focussedRow = null;
+      rowSelection.current.lastIndividuallySelectedRow = null;
+    }
+
     // Since the row selection mode has changed, the row selection mode needs to be updated.
-    setAriaLiveMessage(intl.formatMessage({ id: hasSelectableRows ? 'Terra.worklist-data-grid.row-selection-mode-enabled' : 'Terra.worklist-data-grid.row-selection-mode-disabled' }));
+    if (!rowSelection.current.isRowSelectionEnabledByGrid) {
+      setAriaLiveMessage(intl.formatMessage({ id: hasSelectableRows ? 'Terra.worklist-data-grid.row-selection-mode-enabled' : 'Terra.worklist-data-grid.row-selection-mode-disabled' }));
```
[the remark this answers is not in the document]
cm9361: why do we care if a row selection was toggled by the grid?  It really isn't clear what the purpose of this variable is by looking at the name.

KoushikKalli:
> > The _Row Selection with Pinned Columns_ example doesn't seem to be working. Can you take a look at it?
> 
> I observed this issue as well. It would seem that some unit test should have failed since this functionality is not working. Do we need better test coverage?



> > The _Row Selection with Pinned Columns_ example doesn't seem to be working. Can you take a look at it?
> 
> I observed this issue as well. It would seem that some unit test should have failed since this functionality is not working. Do we need better test coverage?

We have made the change to the example to correct the issue. The functionality is not broken as evident by the fact that no changes were required in the grid code. The pinned column with row selection example had to be updated to implement row selection according to the latest contract.

packages/terra-worklist-data-grid/src/subcomponents/Row.jsx
```diff
@@ -103,15 +103,15 @@ function Row(props) {
 
   const columnIndexOffSet = hasRowSelection ? 1 : 0;
 
-  const handleCellSelect = useCallback((rowIdColId, coordinates, cellSelectable = true) => {
-    if (hasRowSelection) {
+  const handleCellSelect = (selectionDetails) => {
```
cm9361, line 106: You removed the useCallback function on this method.  That needs to remain in place or it will cause unnecessary rerenders.

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/RowSelection.jsx
```diff
@@ -78,31 +200,50 @@ const RowSelection = () => {
...
+    rows.forEach((row, index) => {
+      if (rowIdsToUnselect && rowIdsToUnselect.indexOf(row.id) >= 0) {
+        rows[index].isSelected = false;
+      }
+
+      if (rowIdsToSelect && rowIdsToSelect.indexOf(row.id) >= 0) {
+        rows[index].isSelected = true;
+      }
+    });
+  }, [rows]);
```
cm9361, line 239: Seems like this logic is a little inefficient.  You have to loop though every row of data even if you only have one row to update and it is the first one.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
cm9361: Ctrl+A should select all the text inside the input textbox in that scenario.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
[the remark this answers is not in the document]
KoushikKalli: In the current implementation, when the user turns on row selection mode using the grid's keyboard shortcuts, the announcement is being made before the consumer is notified. This check prevents a repetition of the same announcement after the state change by the consumer. The UXPLATFORM-9507 story will change all announcements to come after the consumer changes the state. It will make this variable & check no longer necessary but such change should be part of that story.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -199,7 +205,7 @@ function WorklistDataGrid(props) {
   const [focusedCol, setFocusedCol] = useState(0);
   const [ariaLiveMessage, setAriaLiveMessage] = useState(null);
   const [cellAriaLiveMessage, setCellAriaLiveMessage] = useState(null);
-
+  const rowSelection = useRef({ isRowSelectionModeToggledByGrid: false });
```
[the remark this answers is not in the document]
KoushikKalli: With the requirements simplified, we expect the logic to be simplified.

packages/terra-worklist-data-grid/src/subcomponents/Row.jsx
KoushikKalli: Added back

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -328,6 +345,83 @@ function WorklistDataGrid(props) {
...
+
+    // const ariaMessage = `${intl.formatMessage({
+    //   id: hasSelectableRows
+    //     ? 'Terra.worklist-data-grid.row-selection-multiple-rows-selected'
+    //     : 'Terra.worklist-data-grid.row-selection-mode-enabled',
+    // })}. ${intl.formatMessage({
+    //   id: rowIdsToSelect.length === 1
+    //     ? 'Terra.worklist-data-grid.row-selection-template'
+    //     : 'Terra.worklist-data-grid.row-selection-selected-rows-range',
+    // }, { row: rangeInfo.anchorRow, endRow: newEndRowIndex })}`;
```
cm9361, line 388: Should this commented code be removed?

eawww (approved):
I know this will likely be addressed by the upcoming rework of the announcement logic but just noting that when enabling row selection via Shift Space, screen reader announces an index one less than what it announces when you select it after that.

Still this is coming along wonderfully!

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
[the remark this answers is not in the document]
cm9361: So should we just remove those messages for now or in the next JIRA to reduce the complexity of what we are reviewing and have to update in the subsequent JIRA?  I question if it is worth the extra effort to manage messages that we know are going away.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -210,11 +216,11 @@ function WorklistDataGrid(props) {
...
   // functions
 
-  const isRowSelectionCell = (columnIndex) => (
+  const isRowSelectionCell = useCallback((columnIndex) => (
     hasSelectableRows && columnIndex < displayedColumns.length && displayedColumns[columnIndex].id === WorklistDataGridUtils.ROW_SELECTION_COLUMN.id
-  );
+  ), [displayedColumns, hasSelectableRows]);
 
-  const setFocusedRowCol = (newRowIndex, newColIndex, makeActiveElement) => {
+  const setFocusedRowCol = useCallback((newRowIndex, newColIndex, makeActiveElement) => {
```
cm9361, line 223: We should not have added useCallback to this function.  There is not a need to memoize this function.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -210,11 +216,11 @@ function WorklistDataGrid(props) {
   // -------------------------------------
   // functions
 
-  const isRowSelectionCell = (columnIndex) => (
+  const isRowSelectionCell = useCallback((columnIndex) => (
```
cm9361, line 219: We should not need to use useCallback on this function.  There is not a need to memoize the function.

packages/terra-worklist-data-grid/tests/jest/WorklistDataGrid.test.jsx
```diff
@@ -199,13 +199,13 @@ describe('basic grid', () => {
     );
 
     // Find column headers
-    const nonSelectableCell = wrapper.find(Row).at(0).find('.selectable');
+    const nonSelectableCell = wrapper.find(Row).at(0).find('th:not(.selectable)');
```
cm9361, line 202: Why did this have to change in the test?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Removed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Removed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
KoushikKalli: Removed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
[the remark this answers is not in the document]
KoushikKalli: The code matches the current requirement. All changes should be addressed holistically by the story that has the new requirements.

packages/terra-worklist-data-grid/tests/jest/WorklistDataGrid.test.jsx
KoushikKalli: We believe this line was inconsistent with the intention of the test and needed to be fixed. The intention of the test is to verify selection of a **non-selectable** cell in row selection mode. Though clicking both selectable & non-selectable cell will result in row-selection, we changed this line to be consistent with the intention of the test.

cm9361:
> @KoushikKalli - I did some testing with Safari and VoiceOver. Keyboard navigation with Safari works as required. VoiceOver generally works well, but I struggle with the difference in VoiceOver navigation and visual focuswith Up/Down Arrow versus Control+Option+Up/Down Arrow. When VoiceOver focus indicator is on one cell but visual focus is on another Space selects the cell with visual focus, not what VoiceOver announced. This could be user error, but we need to discuss this behavior. It is confusing!

You can configure VoiceOver such that it follows the keyboard focus.  You have to make additional configuration changes for tables.  I believe that I sent those preferences in a chat with you and Michael once.

packages/terra-worklist-data-grid/tests/jest/WorklistDataGrid.test.jsx
KoushikKalli: Based on the example worklist data grid used by the test, if you don't use the _th_, the find will return many things such as _FocusTrap, div, span, td_ & _th_ that do not have the selectable class. Using the _th_ in the selector allows you to get to the non-selectable cell directly and makes the test cleaner.

cm9361: That could only happen if the selector is incorrect.  If you replaced th is the class associated with the th and td elements, then you would still get that element.  I can approve for now, and we can update later, if necessary.

cm9361 (approved):

mjpalazzo (approved):

packages/terra-framework-docs/src/terra-dev-site/doc/worklist-data-grid/Examples.2/RowSelection.jsx
kenk2: Question: Why are we relying on `indexOf`? Instrinctually if I had a row selection, shouldn't I get this row when I select the row? This could as @cm9361 says also have unintended side effects because we're updating every row, not just the one we select.

packages/terra-framework-docs/src/terra-dev-site/test/worklist-data-grid/PinnedColumnsWithRowSelection.test.jsx
```diff
@@ -99,16 +99,16 @@ const RowSelection = () => {
...
       ariaLabel="Worklist Data Grid with Pinned Columns and Row Selection"
       hasSelectableRows={hasSelectableRows}
-      onRowSelect={(rowId) => {
-        const newRows = [];
-        const selectedRow = rows.find(row => row.id === rowId);
-        selectedRow.isSelected = !selectedRow.isSelected;
-        rows.forEach(element => {
-          if (element.isSelected) {
-            newRows.push(element.id);
+      onRowSelect={(rowIdsToSelect, rowIdsToUnselect) => {
```
kenk2, line 102: Question: Could we possibly not inline this? This seems to have enough going on that it should be a separate function space.

packages/terra-framework-docs/src/terra-dev-site/test/worklist-data-grid/WorklistDataGridWithRowSelection.test.jsx
```diff
@@ -38,6 +38,18 @@ const WorklistDataGridWithRowSelection = () => {
     clearRowSelection();
   };
 
+  const onRowSelect = (rowIdsToSelect, rowIdsToUnselect) => {
```
kenk2, line 41: I'm noticing we're doing this quite a lot repeatedly. Is there possibly a way we could specify this as a util function and then just repeat that everywhere we need this?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
kenk2: Concurring with @cm9361 Because we could just manage the selected state with 1 array. Only care about that one, the others we could just assume are now unchecked. I think that's a simpler way to approach this.

kenk2 (approved):

PR #1714, opened 2023-07-26
[terra-tabs] Added drag and drop feature

saket2403:
### Summary

The [react-beautiful-dnd](https://github.com/atlassian/react-beautiful-dnd) was used to implement the drag and drop functionality in tabs. This library internally takes care of the keyboard navigation as well as the Assistive technology response when a drag-drop is performed.

**What was changed:**
As requested by the MPage team 2 props have been added to support the drag and drop feature:

- **isDraggable**: Boolean Prop to make the tabs draggable
- **onTabOrderChange**: Callback function which is triggered when a drop event is completed

**Why it was changed:**
Drag and Drop was identified as a gap for the MPage passthrough.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [x] Functional review

### Additional Details

For Testing: 

- Open the Deployment Link 
- Navigate to Tests Tab 
- Click on Controlled Tabs.

**This PR resolves:**

UXPLATFORM-8867 

---

@cerner/terra


packages/terra-tabs/src/Tabs.jsx
```diff
@@ -54,11 +54,20 @@ const propTypes = {
    * Key of the pane that should be open initially.
    */
   defaultActiveKey: PropTypes.string,
+  /**
+   * Whether or not the tab is draggable.
+   */
+  isDraggable: PropTypes.bool,
+  /**
+   * Callback function triggered when tab is drag and dropped .
```
supreethmr, line 62:
```suggestion
   * Callback function triggered when tab is drag and dropped.
```

packages/terra-tabs/src/common-tabs/CommonTabs.jsx
```diff
@@ -50,6 +50,14 @@ const propTypes = {
    * Parameters: 1. Event 2. Selected pane's key
    */
   onChange: PropTypes.func,
+  /**
+   * Whether or not the tab is draggable.
+   */
+  isDraggable: PropTypes.bool,
+  /**
+     * Callback function triggered when tab is drag and dropped .
+     */
```
supreethmr, line 59:
```suggestion
  /**
   * Callback function triggered when tab is drag and dropped.
   */
```

supreethmr:
I think only active style is being carried out with current when we are in process of drag and drop using keyboard as shown in below clip. I think while dragging either we need to carry the focus (dotted line) outline with current tab or we might need to clear that and it back when it is dropped using space key (on mac).

https://github.com/cerner/terra-framework/assets/12869809/eaba9d98-8251-4c67-a861-fd72e92f937e

saket2403:
> I think only active style is being carried out with current when we are in process of drag and drop using keyboard as shown in below clip. I think while dragging either we need to carry the focus (dotted line) outline with current tab or we might need to clear that and it back when it is dropped using space key (on mac).
> 
>  Screen.Recording.2023-07-31.at.4.54.37.PM.mov

Fixed [65f7651](https://github.com/cerner/terra-framework/pull/1714/commits/65f7651ad47332052eb991b3a928fd529b032b0d)

packages/terra-tabs/src/common-tabs/_Tab.jsx
```diff
@@ -168,6 +178,43 @@ const Tab = ({
...
+            data-terra-tabs-show-focus-styles
+            data-terra-tab-draggable
+          >
+            <div>
+
+              <IconKnurling className={cx('icon-knurling')} />
+              <VisuallyHiddenText aria-hidden id={responseId} text={onFocusResponse} />
+              {customDisplay}
+              {customDisplay ? null : icon}
+              {customDisplay || isIconOnly ? null : <span className={cx('label')}>{label}</span>}
```
supreethmr, line 210: nit: Can we update the above condition like below to avoid null 

```suggestion
        {customDisplay || icon}
        {(!customDisplay || !isIconOnly) && <span className={cx('label')}>{label}</span>}
```

rahulkumar8cs:
+1 for accessibility changes, we need to add translation for drop.
UX review required again after translation.

packages/terra-tabs/src/common-tabs/_Tabs.jsx
```diff
@@ -240,22 +314,22 @@ class Tabs extends React.Component {
 
   render() {
     const {
-      tabData, ariaLabel, variant, onChange,
+      tabData, ariaLabel, variant, onChange, isDraggable,
```
supreethmr, line 317: I think we need to replace all the reference of `tabdata` in render function to use State.visibileTabData to prevent possible difference in values between state and prop value

PR #1745, opened 2023-08-14
[terra-tabs] Allow programmatic activation of Tabs 

saket2403:
### Summary

Allow programmatic activation of Tabs.

**What was changed:**

The active key of tabs could only be set initially following which it only changed on clicking the tab. Mpages require the tab to be activated programmatically( by clicking another button). So a condition was added to honor the prop set by the consumer to set the active key along with the existing state variable.

**Why it was changed:**
Gap in functionality identified by MPages.

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9513 

---

@cerner/terra


packages/terra-tabs/src/Tabs.jsx
```diff
@@ -115,7 +115,7 @@ class Tabs extends React.Component {
     React.Children.forEach(children, child => {
       let content;
       let tabContent;
-      if (child.key === this.state.activeKey) {
+      if (child.key === this.props.activeKey || (!this.props.activeKey && child.key === this.state.activeKey)) {
```
sugan2416, line 118: Do we need to check `!this.props.activeKey` ?

packages/terra-tabs/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
```
sdadn, line 5:
```suggestion
* Added
```

packages/terra-tabs/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
...
+  * Add support for programmatic activation of tabs.
```
sdadn, line 6:
```suggestion
  * Added support for programmatic activation of tabs.
```

sdadn (changes requested):

packages/terra-tabs/src/Tabs.jsx
supreethmr: Can we simplify this condition for better readability 
```suggestion
      const activeKey = this.props.activeKey || this.state.activeKey;
      if (child.key === activeKey) {
```

packages/terra-tabs/src/Tabs.jsx
```diff
@@ -138,7 +138,7 @@ class Tabs extends React.Component {
     return (
       <CommonTabs
         id={customProps.id}
-        activeItemKey={this.state.activeKey}
+        activeItemKey={this.props.activeKey || this.state.activeKey}
```
supreethmr, line 141: we can make use of previously initialised activeKey variable instead of this condition

supreethmr (approved):
+1 with above comments addressed

packages/terra-tabs/src/Tabs.jsx
saket2403: Done https://github.com/cerner/terra-framework/pull/1745/commits/9491ceeefbb8edcd28558754f97d71b9291318ec

packages/terra-tabs/src/Tabs.jsx
saket2403: Done https://github.com/cerner/terra-framework/pull/1745/commits/9491ceeefbb8edcd28558754f97d71b9291318ec

sdadn (approved):

PR #1755, opened 2023-08-22
[terra-worklist-data-grid] Fix height of pinned column border

cm9361:
### Summary

**What was changed:**
Modified WorklistDataGrid logic such that the border height for the pinned column separator adjust to table height changes.

**Why it was changed:**
The change was made to ensure the separator is visible and looks correct to users.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

Screen Shot 2023-08-21 at 5 54 13 PM

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details
N/A

**This PR resolves:**

UXPLATFORM-9526 

---

@cerner/terra

### DIFF
--- a/packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
+++ b/packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
@@ -7,6 +7,7 @@ import PropTypes from 'prop-types';
 import { injectIntl } from 'react-intl';
 import classNames from 'classnames/bind';
 import * as KeyCode from 'keycode-js';
+import ResizeObserver from 'resize-observer-polyfill';
 
 import ThemeContext from 'terra-theme-context';
 import VisuallyHiddenText from 'terra-visually-hidden-text';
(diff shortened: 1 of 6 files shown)


packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -243,10 +244,19 @@ function WorklistDataGrid(props) {
...
+      return;
+    }
+
     grid.current = node;
 
-    // Update table height state variable
-    setTableHeight(grid.current.offsetHeight - 1);
+    const resizeObserver = new ResizeObserver(() => {
+      // Update table height state variable
+      setTableHeight(grid.current.offsetHeight - 1);
```
sycombs, line 255: Is it possible to add a WDIO test where the table height changes?

cm9361: I am not aware of a way to do that.  Do we have an example of re-rendering a component with different prop values?

sdadn: Won't the component automatically rerender if we update the input data?

sycombs (approved):

PR #1762, opened 2023-08-25
WorklistDataGrid Updated the functionality for selecting a range of rows

brithom:
### Summary
Updated the functionality for selecting a range of rows. The change was necessary to make the range selection based on selected row for all keyboard and mouse shortcuts.

This change was tested using:

- [x] WDIO
- [x] Jest

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [x] Functional review

### This PR resolves
UXPLATFORM-9499


packages/terra-worklist-data-grid/src/subcomponents/Row.jsx
```diff
@@ -98,7 +98,7 @@ function Row(props) {
   const columnIndexOffSet = hasRowSelection ? 1 : 0;
 
   const handleCellSelect = useCallback((selectionDetails) => {
-    if (hasRowSelection || selectionDetails.multiSelect) {
+    if (hasRowSelection || selectionDetails.isShiftKeyPressed) {
       if (onRowSelect) {
         onRowSelect(selectionDetails);
       }
```
cm9361, line 104: As we look to modularize the functionality for our upcoming enhancements, I am wondering if we should allow the WorklistDataGrid to handle this logic.  Moving the handler logic there removes the grid specific logic from the Row subcomponent.

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
```diff
@@ -142,7 +142,7 @@ function Cell(props) {
   const onMouseDown = ((event) => {
     if (!isFocusTrapEnabled) {
       onCellSelect({
-        rowId, columnId, rowIndex, columnIndex, multiSelect: event.shiftKey, selectedByKeyboard: false, isCellSelectable: (!isMasked && isSelectable),
+        rowId, columnId, rowIndex, columnIndex, isShiftKeyPressed: event.shiftKey, isCellSelectable: (!isMasked && isSelectable),
```
cm9361, line 145: Should we simplify the name to isShiftPressed?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -399,22 +408,26 @@ function WorklistDataGrid(props) {
...
     const rowsToUnSelect = [];
-    rowSelection.current.rangeUsingLastIndividuallySelectedRow = { anchorRow: rowIndex };
     const rowLabel = rows[rowIndex - 1].ariaLabel || (rowIndex + 1);
 
     if (!rows[rowIndex - 1].isSelected) {
+      multiSelectRange.current = { start: rowIndex, end: null }; // Establish new starting point for future range.
       setAriaLiveMessage(intl.formatMessage({ id: 'Terra.worklist-data-grid.row-selection-template' }, { row: rowLabel }));
       rowsToSelect.push(rowId);
     } else {
+      if (rowIndex === multiSelectRange.current?.start) {
```
cm9361, line 427: Based on your logic, the current property of mutliSelectRange would never be null.  There does not seem to be a need for the question mark.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -279,15 +281,14 @@ function WorklistDataGrid(props) {
...
 
     if (!hasSelectableRows) {
-      rowSelection.current.rangeUsingFocusedRow = null;
-      rowSelection.current.rangeUsingLastIndividuallySelectedRow = null;
+      multiSelectRange.current = {};
     }
 
     // Since the row selection mode has changed, the row selection mode needs to be updated.
-    if (!rowSelection.current.isRowSelectionModeToggledByGrid) {
+    if (!isRowSelectionModeToggledByGrid.current) {
```
cm9361, line 288: Why would the row selection being toggled by the grid impact this message?  Aren't we still enabling or disabling row-selection mode?  It really seems like we should be able to get rid of this variable.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -357,37 +357,46 @@ function WorklistDataGrid(props) {
     }
   };
 
-  const selectMultipleRows = useCallback((newEndRowIndex, rangeInfo) => {
-    if (!onRowSelect || !rangeInfo) {
-      return;
+  const getRowsInRangeToUnselect = useCallback((startOfRange, prevEndOfRange, newEndOfRange) => {
```
cm9361, line 360: Is the startOfRange the anchor or is the prevEndOfRange the anchor.  I also don't think we should abbreviate names when possible.  We should spell out "previous".  It would be helpful to have documentation for this method.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -357,37 +357,46 @@ function WorklistDataGrid(props) {
...
+    let rowIdsToUnselect = [];
+    if (!prevEndOfRange) {
+      // short-circuit since there was no previous range, there is nothing to unselect.
+      return rowIdsToUnselect;
+    }
+
+    if (startOfRange > prevEndOfRange) {
```
cm9361, line 367: Aren't these row indexes?  The variable names could probably be a little more clear.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -357,37 +357,46 @@ function WorklistDataGrid(props) {
     }
   };
 
-  const selectMultipleRows = useCallback((newEndRowIndex, rangeInfo) => {
-    if (!onRowSelect || !rangeInfo) {
-      return;
+  const getRowsInRangeToUnselect = useCallback((startOfRange, prevEndOfRange, newEndOfRange) => {
+    let rowIdsToUnselect = [];
+    if (!prevEndOfRange) {
```
cm9361, line 362: Shouldn't this also return if the start and end of the range are equal?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
cm9361, line 360: This method is only used in one place. It seems like the logic is small enough to just be included in the caller and this method removed.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -357,37 +357,46 @@ function WorklistDataGrid(props) {
...
+      return rowIdsToUnselect;
+    }
+
+    if (startOfRange > prevEndOfRange) {
+      // The range extends upward from the anchor row
+      if (newEndOfRange > prevEndOfRange) {
+        // The range was moved down towards the anchor so rows that no longer qualify for the range need to be unselected.
+        rowIdsToUnselect = rows.slice(prevEndOfRange - 1, Math.min(newEndOfRange, startOfRange)).map(row => row.id);
+      }
+    } else if (startOfRange < prevEndOfRange) {
```
cm9361, line 373: I would think that this could be an else condition.  We already checked for greater than in the if condition.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -357,37 +357,46 @@ function WorklistDataGrid(props) {
...
+      // The range extends downward from the anchor row
+      if (newEndOfRange < prevEndOfRange) {
+        // The range was moved up towards the anchor so rows that no longer qualify for the range need to be unselected. New endRangeIndex becomes ordered End
+        rowIdsToUnselect = rows.slice(Math.max(startOfRange, newEndOfRange), prevEndOfRange).map(row => row.id);
+      }
     }
+    return rowIdsToUnselect;
+  }, [rows]);
 
+  const selectMultipleRows = useCallback((startOfRange, prevEndOfRange, newEndOfRange) => {
```
cm9361, line 383: What is newEndOfRange?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -459,41 +472,29 @@ function WorklistDataGrid(props) {
...
-        rangeInfo.previousSelectionEndRow = selectionDetails.rowIndex;
-      } else {
-        // no anchor row
-        selectRow(selectionDetails.rowId, selectionDetails.rowIndex);
+    if (!hasSelectableRows) {
+      if (selectionDetails.isShiftKeyPressed) {
+        // Shift+Space, Shift+Click, Shift+Up, Shift+Down
+        isRowSelectionModeToggledByGrid.current = !hasSelectableRows;
+        multiSelectRange.current = { start: selectionDetails.rowIndex, end: null };
+        selectMultipleRows(multiSelectRange.current.start, multiSelectRange.current.end, selectionDetails.rowIndex);
```
cm9361, line 480: multiSelectRange is a ref hook.  Why do you need to pass its values to the selectMultipleRows function.  It seems like that function could obtain the values directly.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -212,7 +212,9 @@ function WorklistDataGrid(props) {
   const [focusedCol, setFocusedCol] = useState(0);
   const [ariaLiveMessage, setAriaLiveMessage] = useState(null);
   const [cellAriaLiveMessage, setCellAriaLiveMessage] = useState(null);
-  const rowSelection = useRef({ isRowSelectionModeToggledByGrid: false });
+  const isRowSelectionModeToggledByGrid = useRef(false);
+  const inShiftUpDownMode = useRef(false);
+  const multiSelectRange = useRef({ start: null, end: null });
```
cm9361, line 217: We should create more meaningful variable names than "start" and "end"

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -212,7 +212,9 @@ function WorklistDataGrid(props) {
   const [focusedCol, setFocusedCol] = useState(0);
   const [ariaLiveMessage, setAriaLiveMessage] = useState(null);
   const [cellAriaLiveMessage, setCellAriaLiveMessage] = useState(null);
-  const rowSelection = useRef({ isRowSelectionModeToggledByGrid: false });
+  const isRowSelectionModeToggledByGrid = useRef(false);
+  const inShiftUpDownMode = useRef(false);
```
cm9361, line 216: I cannot understand the purpose of this variable based on its name.  What is the shift up/down mode?  Could there be a more clear name that can be used?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
cm9361, line 217: Maybe it would be easier to maintain the logic if we used two reference variables instead of an object with properties.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -357,37 +357,46 @@ function WorklistDataGrid(props) {
...
+    return rowIdsToUnselect;
+  }, [rows]);
 
+  const selectMultipleRows = useCallback((startOfRange, prevEndOfRange, newEndOfRange) => {
     if (!hasSelectableRows) {
       onEnableRowSelection();
     }
+    if (!onRowSelect) {
+      return;
+    }
```
cm9361, line 389: If the consumer does not have an onRowSelect callback, do we need to enable row selection?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
cm9361, line 217: If the start index and end index are numbers, should the unset value be -1 instead of null?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -399,22 +408,26 @@ function WorklistDataGrid(props) {
       id: rowIdsToSelect.length === 1
         ? 'Terra.worklist-data-grid.row-selection-template'
         : 'Terra.worklist-data-grid.row-selection-selected-rows-range',
-    }, { row: rangeInfo.anchorRow, endRow: newEndRowIndex })}`;
+    }, { row: startOfRange, endRow: newEndOfRange })}`;
 
     setAriaLiveMessage(ariaMessage);
     onRowSelect(rowIdsToSelect, rowIdsToUnselect);
```
cm9361, line 414: I still think we should just have one array for our API.  As a consumer, I would not want to have to deal with two separate arrays to manage my state.  I don't think we should introduce the two array approach to our consumers.

packages/terra-worklist-data-grid/tests/jest/WorklistDataGrid.test.jsx
```diff
@@ -238,25 +247,29 @@ describe('basic grid', () => {
 
 describe('with pinned columns', () => {
   it('sets pinnedColumns as pinned', () => {
+    const pinnedColumns = dataFile.cols.slice(0, 2);
```
sycombs, line 250: Since we're pulling out the `pinnedColumns` into its own variable, can we do the same with `overflowColumns` since they're slicing the same base array? Also, is it possible to move them up a block into the `describe` so that the variables can be shared by each test and don't need to be redeclared in each `it`?

packages/terra-worklist-data-grid/tests/jest/WorklistDataGrid.test.jsx
```diff
@@ -324,3 +337,473 @@ describe('Error handling - prop types', () => {
     expect(console.error).toHaveBeenCalledWith(expect.stringContaining(ERRORS.ROW_HEADER_INDEX_EXCEEDS_PINNED)); // eslint-disable-line no-console
   });
 });
+
+describe('Row selection', () => {
+  it('verifies only onRowSelect is called when space is used to select a row.', () => {
+    const mockOnCellSelect = jest.fn();
+    const mockOnRowSelect = jest.fn();
+    const mockOnEnableRowSelection = jest.fn();
```
sycombs, line 345: Should these be in some kind of `before` block since they are being declared for every test in this `describe`?

sdadn: +1

These should be in the `beforeAll()` as they don't need to be redefined in every test. We can use [mockClear](https://jestjs.io/docs/mock-function-api#mockfnmockclear) to reset the implementation between test runs. However we already have that option enabled in our [jest config](https://github.com/cerner/terra-toolkit/blob/fdb6c307005eb96fad4a9b468dd17f2012d69050/packages/jest-config-terra/jest.config.js#L6C1-L6C1) so we don't even need to do that!

packages/terra-worklist-data-grid/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 * Changed
   * Removed roving tabIndex and resolved focus management issues within cells.
+  * Updated the functionality for selecting a range of rows.
```
sdadn, line 13: Are there API changes for row selection?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
sdadn: Doesn't the variable name `multiSelectRange` already add meaning to these variables?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
sdadn: What about the case when they are equal?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
brithom: We will accept suggestions for this variable name. When the user starts to select a range of rows using the Shift+Up or the Shift+Down, it establishes a new anchor. As long as the user does not release the shift key, another Shift+Up/Down continues the selection or unselection using the same anchor established by the initial Shift+Up/Down. We needed a variable to distinguish between the first Shift+Up/Down and subsequent ones.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
brithom: The multiSelectRange gives the context and meaning to the start and end. The start and end are the start and end of the range.

brithom: I think that in this case, it is better to keep them together. A range should have a start and an end when one is established.

brithom: Yes, they are indexes. The null makes it clear that the value is not defined.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
brithom: Once we start deriving the messages after the consumer responds then this will go away. The requirements states that when the row selection is enabled or disabled from outside the grid, the grid also needs to announce the change in row selection state. That is currently being done on the re-render based on change to the hasSelectableRows prop. However, when row selection is enabled or disabled from within the grid, the message is immediately announced. Therefore on the re-render, we needed a way to determined if the change was triggered from inside or outside the grid so that the message was not announce multiple times.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
brithom: We think so since row selection can still be enabled from outside of the grid with/without onRowSelect.

packages/terra-worklist-data-grid/tests/jest/WorklistDataGrid.test.jsx
brithom: These existing tests broke because we added 1 additional row to the test data. We only made this simple tweak to make the tests less fragile. Should a similar change be necessary in the future, these tests will no longer break.

packages/terra-worklist-data-grid/src/subcomponents/Row.jsx
brithom: Point taken but this Jira did not change that logic. As a result, the suggestion you made should be handled in the Jira that creates the base data grid component.

packages/terra-worklist-data-grid/src/subcomponents/Cell.jsx
brithom: Done

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
brithom: The start of the range is the anchor. Since anchor is not a globally defined term, we went with startOfRange. It is best to explain with an example:

- User clicks row 4 (This establishes the anchor or the starting point for range selection. As there is no range as yet, there is no end of range)
- User Shift+Click row 6 (This means we now have a range of selected rows. Since no range existed before the previousEnd is not defined. For the newly established range, the end is row 6)
- User Shift+Click row 8 (This means that the range is being extended from it previous end at row 6 to the new end at row 8. That is, the range is now from row 4 to row 8)

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
brithom: Since we are no longer passing these as parameter, multiSelectRange ref is directly used.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
brithom: This is handled by the if/else block below.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
brithom: We have kept the method separate for readability.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
brithom: Saad is correct, if we use an else, the equal case would get merged into the else block which technically would not be correct. The explicit checks are also there to show the direction of the range from the anchor.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
brithom: Changed

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
brithom: Changed

packages/terra-worklist-data-grid/tests/jest/WorklistDataGrid.test.jsx
brithom: Since we created a `describe` for the rowSelection tests, we moved the mocks to `beforeEach` block instead.

packages/terra-worklist-data-grid/CHANGELOG.md
brithom: Yes, based on discussion from post scrum. We made the update.

chrismichalewicz:
Functionally reviewed, looks good. Great job!

packages/terra-worklist-data-grid/tests/jest/WorklistDataGrid.test.jsx
```diff
@@ -324,3 +337,423 @@ describe('Error handling - prop types', () => {
...
+describe('Row selection', () => {
+  let mockOnCellSelect = null;
+  let mockOnRowSelect = null;
+  let mockOnEnableRowSelection = null;
+
+  beforeEach(() => {
+    mockOnCellSelect = jest.fn();
+    mockOnRowSelect = jest.fn();
+    mockOnEnableRowSelection = jest.fn();
+  });
```
sdadn, line 350: Is `beforeEach` necessary? I would think that you need this only once in a `beforeAll` block?

sdadn: Apparently, this doesn't work in `beforeAll`, disregard this.

sdadn (approved):

eawww (approved):

mjpalazzo:
Approving with the understanding that the inconsistent announcement issue will be fixed with a Jira Eric is opening.

PR #1768, opened 2023-08-30
[terra-slide-group] Fix VoiceOver not announcing title of nested panel

smason0:
### Summary

**What was changed:**
Added VisuallyHiddenText to Slide component within SlideGroup to be conditionally visible to screen readers when the slide is nested. This text should be the aria-label of the slide (typically the title).

**Why it was changed:**
When using the VoiceOver screen reader, the aria-label of the initial slide in the SlideGroup was announced, but any nested slide was not announced upon being opened. This change allows the aria-label of nested slides to be announced to provide better context for screen reader users.

### Testing

This change was tested using:

- [ ] WDIO
- [x] Jest
- [ ] No tests are needed

Tested using VoiceOver Utility version 10 (869.25) on MacOS Ventura 13.5 in Safari browser. Also tested with JAWS screen reader in Edge browser to ensure passive behavior.

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-8971 

---

@cerner/terra

### DIFF
--- a/packages/terra-slide-group/src/Slide.jsx
+++ b/packages/terra-slide-group/src/Slide.jsx
@@ -32,8 +32,7 @@ const propTypes = {
   slideAriaLabel: PropTypes.string,
 
   /**
-   * @private
-   * Indicates if the slide is nested within a SlideGroup.
+   * If true, the Slide is nested within a SlideGroup.
    */
   isNested: PropTypes.bool,
 };
--- a/packages/terra-slide-group/src/Slide.jsx
+++ b/packages/terra-slide-group/src/Slide.jsx
@@ -35,7 +35,7 @@ const propTypes = {
    * @private
    * Indicates if the slide is nested within a SlideGroup.
    */
-  isNested: PropTypes.func,
+  isNested: PropTypes.bool,
 };
 
 const defaultProps = {
--- a/packages/terra-slide-group/src/Slide.jsx
+++ b/packages/terra-slide-group/src/Slide.jsx
@@ -2,6 +2,7 @@ import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
 import ThemeContext from 'terra-theme-context';
+import VisuallyHiddenText from 'terra-visually-hidden-text';
 import styles from './Slide.module.scss';
 
 const cx = classNames.bind(styles);
@@ -29,6 +30,12 @@ const propTypes = {
    * The aria label for the Slide.
    */
   slideAriaLabel: PropTypes.string,
+
+  /**
+   * @private
+   * Indicates if the slide is nested within a SlideGroup.
+   */
+  isNested: PropTypes.func,
 };
 
 const defaultProps = {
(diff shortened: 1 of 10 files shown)


packages/terra-slide-group/src/Slide.jsx
```diff
@@ -29,6 +30,12 @@ const propTypes = {
    * The aria label for the Slide.
    */
   slideAriaLabel: PropTypes.string,
+
+  /**
+   * @private
```
kenk2, line 35: Question: Why is this a private method?

kenk2 (commented):
Just one question, otherwise seems good and I can approve it once answered.

packages/terra-slide-group/src/Slide.jsx
smason0: Good point. When I originally added this, I was under the impression that Slide would possibly be consumed externally, but it is only used as an internal component within SlideGroup. Therefore, I can remove this tag since the props for Slide are all essentially "private" anyway and never exposed to clients via the docs.

packages/terra-slide-group/tests/jest/SlideGroup.test.jsx
```diff
@@ -14,11 +14,39 @@ const slideGroup = (
 
 it('should render a default SlideGroup', () => {
   const wrapper = shallow(slideGroup);
+
+  const cssTransitions = wrapper.find('CSSTransition');
+  expect(cssTransitions.at(0).key()).toBe('1');
+  expect(cssTransitions.at(0).prop('enter')).toBe(false);
+  expect(cssTransitions.at(0).prop('enter')).toBe(false);
```
kenk2, line 21: One last question: Why are these lines being double-repeated, I shouldn't expect a change here when I call the same `expect` shouldn't I?

packages/terra-slide-group/tests/jest/SlideGroup.test.jsx
```diff
@@ -14,11 +14,39 @@ const slideGroup = (
...
+  expect(cssTransitions.at(2).find('Slide').prop('isNested')).toBe(true);
+
   expect(wrapper).toMatchSnapshot();
 });
 
 it('should render a SlideGroup with animation enabled', () => {
   const wrapper = shallow(React.cloneElement(slideGroup, { isAnimated: true }));
+
+  const cssTransitions = wrapper.find('CSSTransition');
+  expect(cssTransitions.at(0).prop('enter')).toBe(true);
```
kenk2, line 43: One last nit to add some whitespace between each `item`

kenk2 (commented):
Just a few nitpicks, otherwise LGTM

packages/terra-slide-group/tests/jest/SlideGroup.test.jsx
smason0: Oops, yeah the other ones should be the "exit" prop. Copy/paste error I will fix now. I will change the grouping also

kenk2 (approved):
LGTM!

trandrew1023 (approved):

eawww (approved):
VO announces name of nested disclosures.
JAWS announces name of nested disclosures and previous disclosures when nested ones are dismissed.

PR #1771, opened 2023-08-31
Terra File Path component setup

artpark:
### Summary

**What was changed:**
Created all the necessary files for creating a new component. This includes the documentation as well.

**Why it was changed:**
To prepare for development of the new Terra File Path component.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9588 

---

@cerner/terra


artpark:
Tests will fail due to empty CSS files, etc.

kenk2:
> Tests will fail due to empty CSS files, etc.

Since much of this work is placeholder, is it possible that we can have dummy CSS rules so we don't introduce these builds failures into the CI pipeline?

sdadn:
Since we're already working allowing for individual package development like in https://github.com/cerner/terra-toolkit/pull/811, can we also add all the appropriate devDependencies within this package as well?

sdadn:
> Tests will fail due to empty CSS files, etc.

Will this break the main build?

adoroshk (approved):
I checked that project installs and runs without problems, test and lint issues are only those that are expected due to empty css files and placeholder variables in the component itself.

adoroshk:
> > Tests will fail due to empty CSS files, etc.
> 
> Will this break the main build?

It will due to lint issues. Should we have a feature branch created for that component for now, before some actual work is done? 

It probab

adoroshk (changes requested):
Should we change the destination to be a feature branch instead of main for now and merge to main once it doesn't break the build? (or some other solution that doesn't break the build would work)

sdadn:
> Should we change the destination to be a feature branch instead of main for now and merge to main once it doesn't break the build? (or some other solution that doesn't break the build would work)

We have several teams working on this repo so we shouldn't break main.

adoroshk (approved):

packages/terra-file-path/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-framework/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-framework#readme",
+  "peerDependencies": {
+    "react": "^16.8.5",
+    "react-dom": "^16.8.5",
+    "react-intl": ">=2.8.0, <6.0.0"
```
sdadn, line 28: Can we change the format for clarity?
```suggestion
    "react-intl": "^2.8.0 || 5"
```

packages/terra-file-path/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "terra-file-path",
+  "main": "lib/FilePath.js",
```
sdadn, line 3: Will this be the only component exported or do we know at this time if there will be additional exports?

packages/terra-file-path/package.json
artpark: Wouldn't versions 3, and 4 be excluded with the new format?

packages/terra-file-path/package.json
artpark: Just this I believe!

artpark:
> Fails
> * terra-framework-docs
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [ed6054c](https://github.com/cerner/terra-framework/commit/ed6054c8c3031ac3714501197de860e3b6ac33d7)

Docs are not ready for File Path so I'm going to ignore this

packages/terra-file-path/package.json
sdadn: According to https://semver.npmjs.com, both of these are identical:

**Before:**
CleanShot 2023-09-01 at 15 02 31@2x

**After:**
CleanShot 2023-09-01 at 15 01 17@2x

(if anything, it considered the comma in the existing syntax to be incorrect)

packages/terra-file-path/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "classnames": "^2.2.5",
+    "focus-trap-react": "^6.0.0",
+    "keycode-js": "^3.1.0",
+    "prop-types": "^15.5.8"
+  },
+  "scripts": {
+    "compile": "babel --root-mode upward src --out-dir lib --copy-files",
+    "lint": "npm run lint:js && npm run lint:scss",
+    "lint:js": "eslint --ext .js,.jsx . --ignore-path ../../.eslintignore",
+    "lint:scss": "stylelint src/**/*.scss",
```
sdadn, line 40: Do these scripts work at the package level?

packages/terra-file-path/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "react-intl": ">=2.8.0, <6.0.0"
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "focus-trap-react": "^6.0.0",
+    "keycode-js": "^3.1.0",
+    "prop-types": "^15.5.8"
+  },
+  "scripts": {
+    "compile": "babel --root-mode upward src --out-dir lib --copy-files",
```
sdadn, line 37: Can we add clean scripts e.g.:
https://github.com/cerner/terra-toolkit/blob/440ead889dfd75b31a3535e67d76179b0836dd71/packages/terra-enzyme-intl/package.json#L48-L51

packages/terra-file-path/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "lint": "npm run lint:js && npm run lint:scss",
+    "lint:js": "eslint --ext .js,.jsx . --ignore-path ../../.eslintignore",
```
sdadn, line 39: can we add a `lint:js:fix` script?

packages/terra-file-path/package.json
artpark: Saad and I talked offline about this. In summary, I was able to get everything except WDIO to work locally. There is some extra work that is not in scope to get WDIO to work, so those will be added in at a later time.

packages/terra-file-path/package.json
artpark: [b8a0750](https://github.com/cerner/terra-framework/pull/1771/commits/b8a075007e53efdc47e7cfb0034d0ac7cb89bd5f)

packages/terra-file-path/package.json
artpark: [b8a0750](https://github.com/cerner/terra-framework/pull/1771/commits/b8a075007e53efdc47e7cfb0034d0ac7cb89bd5f)

packages/terra-file-path/package.json
artpark: [6bc9e07](https://github.com/cerner/terra-framework/pull/1771/commits/6bc9e0793b1875c8d9188e6f9360e7a6b27f2719)

packages/terra-file-path/jest.config.js
```diff
@@ -0,0 +1,11 @@
...
+
+module.exports = {
+  ...jestConfig,
+  setupFiles: [
+    './jest.enzymeSetup.js',
+  ],
+  snapshotSerializers: [
+    'enzyme-to-json/serializer',
+  ],
+};
```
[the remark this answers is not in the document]
artpark: To get the jest / lint scripts to work locally, these are required unfortunately :( same goes with .eslintrc.js and jest.enzymeSetup.js

packages/terra-file-path/.npmignore
```diff
@@ -0,0 +1,5 @@
+.eslintrc.js
+*.log
+jest.config.js
+jest.enzymeSetup.js
+tests
```
sdadn, line 5: nit:
```suggestion
*.log
.eslintrc.js
jest.config.js
jest.enzymeSetup.js
tests
```

packages/terra-file-path/tests/jest/FilePath.test.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+beforeAll(() => {
+  jest.spyOn(console, 'error').mockImplementation();
+  jest.spyOn(console, 'warn').mockImplementation();
+});
+
+afterEach(() => {
+  console.error.mockClear(); // eslint-disable-line no-console
+  console.warn.mockClear(); // eslint-disable-line no-console
+});
```
sdadn, line 14: Since, this is enabled in [jest-config-terra](https://github.com/cerner/terra-toolkit/blob/440ead889dfd75b31a3535e67d76179b0836dd71/packages/jest-config-terra/jest.config.js#L6), we no longer need to explicitly do a mockClear according to the docs:

https://jestjs.io/docs/mock-function-api#mockfnmockclear

sdadn (approved):
I have some minor comments but looks good if those get resolved!

packages/terra-file-path/.npmignore
artpark: I did a sort alphabetically and this was the order, so I will keep it like this

sdadn: Ah okay, makes sense

PR #1772, opened 2023-09-04
[terra-worklist-data-grid] Fix scroll issues

cm9361:
### Summary

**What was changed:**
The WorklistDataGrid was updated such that element given focus scroll into view properly when pinned columns are present 
during horizontal scrolling as well as the fixed column header row during vertical scrolling.

**Why it was changed:**
The change was made so that content given focus is made visible to users.

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details
N/A

**This PR resolves:**

UXPLATFORM-9438 

---

@cerner/terra


packages/terra-worklist-data-grid/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
 ## Unreleased
+* Fixed
```
kenk2, line 4: Quick question: Should there be a whitespace here or is this okay for the NPM release script?

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -439,6 +440,35 @@ function WorklistDataGrid(props) {
...
+    const nextCellRect = grid.current.rows[toCell.row].cells[toCell.col].getBoundingClientRect();
+
+    // Calculate horizontal scroll offset for right boundary
+    if (nextCellRect.right > gridContainerRect.right) {
+      gridContainerRef.current.scrollBy(nextCellRect.right - gridContainerRect.right, 0);
+    } else {
+      // Calculate horizontal scroll offset for left boundary
+      let scrollOffsetX = 0;
+      if (pinnedColumnOffsets.length > 0) {
+        const lastPinnedColumnRect = grid.current.rows[toCell.row].cells[pinnedColumnOffsets.length - 1].getBoundingClientRect();
```
kenk2, line 456: Not really a comment but I'm wondering if at some point we could refactor this logic to more simplify specify the currently selected row and the newly selected row.

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -734,7 +764,7 @@ function WorklistDataGrid(props) {
 
   // -------------------------------------
   return (
-    <div className={cx('worklist-data-grid-container')}>
+    <div ref={gridContainerRef} className={cx('worklist-data-grid-container')}>
```
kenk2, line 767: Extended to my previous comment: This component's getting really long. Maybe in the future we could look at refactoring a lot of this logic or even splitting up some of the shared responsibilities of this component.

kenk2 (approved):
Just a few questions and concerns, otherwise LGTM!

packages/terra-worklist-data-grid/src/WorklistDataGrid.jsx
cm9361: @kenk2 - We will be creating the DataGrid component that we have been discussing.  That will break up the logic.

packages/terra-worklist-data-grid/CHANGELOG.md
cm9361: @kenk2 - I am not aware of an issue.  If there is an issue, let me know, and I will update accordingly.

eawww:
When navigating with arrow keys on a very squished grid, there seems to be some inconsistency with which edge of the cell is aligned with the edge of the viewable area.
Ex: When navigating from the pinned columns into the first floating column, the column is aligned so the right edge is visible. But a subsequent press of the right arrow key aligns the left edge of the next column.
image
*[right arrow key is pressed]*
image

Ex2: When navigating across subsequent floating columns, they are consistently aligned to the left edge, but navigating up or down aligns the column to the right edge.
image
*[down arrow key is pressed]*
image

I know this is a bit of an edge case and everything works nicely when the floating area is larger than the column widths, but if we could make it consistently align with the left edge while still having the last column show the whole column when the grid isn't squished, that would be ideal.

packages/terra-worklist-data-grid/CHANGELOG.md
sdadn: It shouldn't affect the script.

packages/terra-worklist-data-grid/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
...
+  * Fixed scroll issues releated to pinned columns and fixed column header row.
```
sdadn, line 5: Are these 2 entries? then these should be on separate lines.

cm9361:
> When navigating with arrow keys on a very squished grid, there seems to be some inconsistency with which edge of the cell is aligned with the edge of the viewable area. Ex: When navigating from the pinned columns into the first floating column, the column is aligned so the right edge is visible. But a subsequent press of the right arrow key aligns the left edge of the next column. image _[right arrow key is pressed]_ image
> 
> Ex2: When navigating across subsequent floating columns, they are consistently aligned to the left edge, but navigating up or down aligns the column to the right edge. image _[down arrow key is pressed]_ image
> 
> I know this is a bit of an edge case and everything works nicely when the floating area is larger than the column widths, but if we could make it consistently align with the left edge while still having the last column show the whole column when the grid isn't squished, that would be ideal.

@eawww - I think I might have some insight here.  So my logic only executes after the browser attempt to adjust.  For pinned columns, the difference is large enough that my logic would always be needed.  However, browser have built-in auto scroll logic that I could not determine how to override.

packages/terra-worklist-data-grid/CHANGELOG.md
cm9361: It is all one change.  I just called out both issues explicitly.  Is that acceptable?

eawww (approved):
Tested again after async conversation with @cm9361 from which he improved the way cells are scrolled into view. Now grid ensures right edge of focused column is visible when focusing from left to right and left edge is visible when focusing from right to left. This will allow keyboard users visual access to the entire content, even at impractically small widths.

No negative effects to the JAWS or VO experiences.

It's... beautiful 🥹

PR #1774, opened 2023-09-05
[terra-date-picker] Screen reader updates for invalid dates and calendar date selection

sugan2416:
### Summary

**What was changed:**

- Updated screen reader response to announce when invalid dates are entered.
- Updated screen reader response to announce when calendar popup dates are selected.
- Updated example for invalid date entry.

**Why it was changed:**

- Screen reader currently does not provide context when non-number inputs OR invalid dates are entered in Date Picker fields.
- Screen reader fails to announce the selected date under calendar popup. 

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9479 

---

@cerner/terra

### DIFF
--- a/packages/terra-date-picker/CHANGELOG.md
+++ b/packages/terra-date-picker/CHANGELOG.md
@@ -8,7 +8,7 @@
 * Changed
   * Visual focus changed to dashed lines for individual fields.
   * Screen reader announcement for visual label.
-  * Updated screen response to announce when invalid characters
+  * Updated screen response to announce when invalid dates
     are entered.
   * Updated screen response to announce when calendar popup dates
     are selected.
--- a/packages/terra-date-picker/CHANGELOG.md
+++ b/packages/terra-date-picker/CHANGELOG.md
@@ -5,18 +5,18 @@
 * Fixed
   * Fixed DatePicker scroll on smaller viewports.
 
-* Updated
+* Changed
+  * Visual focus changed to dashed lines for individual fields.
   * Screen reader announcement for visual label.
+  * Updated screen response to announce when invalid characters
+    are entered.
+  * Updated screen response to announce when calendar popup dates
+    are selected.
 
 ## 4.89.0 - (September 5, 2023)
 
 * Changed
-  * Visual focus changed to dashed lines for individual fields.
   * Fixed the date format being read 3 times when using SR
-  * Updated screen response to announce when invalid characters 
-    are entered.
-  * Updated screen response to announce when calendar popup dates 
-    are selected.
 
 ## 4.88.0 - (August 31, 2023)
 
--- a/packages/terra-date-picker/CHANGELOG.md
+++ b/packages/terra-date-picker/CHANGELOG.md
@@ -13,6 +13,10 @@
 * Changed
   * Visual focus changed to dashed lines for individual fields.
   * Fixed the date format being read 3 times when using SR
+  * Updated screen response to announce when invalid characters 
+    are entered.
+  * Updated screen response to announce when calendar popup dates 
+    are selected.
 
 ## 4.88.0 - (August 31, 2023)
 
--- a/packages/terra-date-picker/CHANGELOG.md
+++ b/packages/terra-date-picker/CHANGELOG.md
@@ -8,8 +8,7 @@
 * Changed
   * Visual focus changed to dashed lines for individual fields.
   * Screen reader announcement for visual label.
-  * Updated screen response to announce when invalid dates
-    are entered.
+  * Updated screen response to announce when invalid date is entered.
   * Updated screen response to announce when calendar popup dates
     are selected.
 
--- a/packages/terra-date-picker/src/DateInput.jsx
+++ b/packages/terra-date-picker/src/DateInput.jsx
@@ -136,6 +136,11 @@ const propTypes = {
    * An ISO 8601 string representation of the minimum date that can be selected. The value must be in the `YYYY-MM-DD` format. Must be on or after `01/01/1900`
    */
   minDate: PropTypes.string,
+  /**
+   * A function that gets called for each date in the picker to evaluate which date should be disabled.
+   * A return value of true will be enabled and false will be disabled.
+   */
+  filterDate: PropTypes.func,
 };
 
 const defaultProps = {
@@ -159,6 +164,7 @@ const defaultProps = {
   includeDates: undefined,
   maxDate: '2100-12-31',
   minDate: '1900-01-01',
+  filterDate: undefined,
 };
 
 const DatePickerInput = (props) => {
@@ -186,6 +192,7 @@ const DatePickerInput = (props) => {
     includeDates,
     maxDate,
     minDate,
+    filterDate,
     ...customProps
   } = props;
 
@@ -816,11 +823,12 @@ const DatePickerInput = (props) => {
   }
   let calendarDate = inputDate ? `${inputDate} ${intl.formatMessage({ id: 'Terra.datePicker.selected' })}` : '';
 
-  // Check if date is excluded or out or range or not included
+  // Check if date is excluded or out of range or not included or filtered
   let invalidEntry = '';
   if (DateUtil.isDateExcluded(DateUtil.createSafeDate(dateValue, initialTimeZone), props.excludeDates)
       || DateUtil.isDateOutOfRange(DateUtil.createSafeDate(dateValue, initialTimeZone), DateUtil.createSafeDate(DateUtil.getMinDate(props.minDate), initialTimeZone), DateUtil.createSafeDate(DateUtil.getMaxDate(props.maxDate), initialTimeZone))
-      || DateUtil.isDateNotIncluded(DateUtil.createSafeDate(dateValue, initialTimeZone), props.includeDates)) {
+      || DateUtil.isDateNotIncluded(DateUtil.createSafeDate(dateValue, initialTimeZone), props.includeDates)
+      || (props.filterDate && !props.filterDate(DateUtil.createSafeDate(dateValue, initialTimeZone)))) {
     invalidEntry = intl.formatMessage({ id: 'Terra.datePicker.invalidDate' });
     calendarDate = '';
   }
--- a/packages/terra-date-picker/src/DateInput.jsx
+++ b/packages/terra-date-picker/src/DateInput.jsx
@@ -165,7 +165,7 @@ const DatePickerInput = (props) => {
   const [dayInitialFocused, setDayInitialFocused] = useState(false);
   const [monthInitialFocused, setMonthInitialFocused] = useState(false);
   const [yearInitialFocused, setYearInitialFocused] = useState(false);
-  // TODO:  Added below states for invalid input message for SR
+  // Added below state for invalid input message for SR
   const [invalidInput, SetInvalidInput] = useState(false);
   const editOnkeyDown = useRef(false);
   const theme = React.useContext(ThemeContext);
@@ -219,12 +219,6 @@ const DatePickerInput = (props) => {
     }
   }, [shouldShowPicker, onClick]);
 
-  const setFocus = (node) => {
-    if (node) {
-      node.focus();
-    }
-  };
-
   /**
    * Moves focus to the correct input depending on date ordering. Focus changing is
    * disabled if a complete date has been entered in order to make single input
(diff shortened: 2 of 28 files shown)


supreethmr:
@sugan2416  Can you update the Jira number on Description

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -165,13 +165,15 @@ const DatePickerInput = (props) => {
   const [dayInitialFocused, setDayInitialFocused] = useState(false);
   const [monthInitialFocused, setMonthInitialFocused] = useState(false);
   const [yearInitialFocused, setYearInitialFocused] = useState(false);
+  // Added below state for invalid input message for SR
+  const [invalidInput, SetInvalidInput] = useState(false);
```
supreethmr, line 169: I think we already have `isInvalid` prop for date-picker. so we should make use of this prop to identify the invalid state of the date-picker component.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -270,6 +272,18 @@ const DatePickerInput = (props) => {
...
   };
 
+  const handleInvalidInputChange = (val) => {
+    visuallyHiddenComponent.innerText = '';
+    SetInvalidInput(val);
+    if (val) {
+      visuallyHiddenComponent.innerText = intl.formatMessage({ id: 'Terra.datePicker.invalidDate' });
+    }
+  };
+
```
supreethmr, line 282:
```suggestion
```
Using `isInvalid` prop can replace this function and visuallyHiddenText can be updated to 
```js
   <VisuallyHiddenText text={(isInvalid) ? intl.formatMessage({ id: 'Terra.datePicker.invalidDate' }) : '' }/>
```

packages/terra-date-picker/src/DateInput.jsx
MadanKumarGovindaswamy: This isInvalid Prop is used for css style internally. But we added this invalidInput for screen reader when they enter character.

supreethmr: I don't see any usage of the state `invalidInput` in DateInput. it's just being set to true and false

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -270,6 +272,18 @@ const DatePickerInput = (props) => {
     }
   };
 
+  const handleInvalidInputChange = (val) => {
```
supreethmr, line 275: `handleInvalidInputChange` is always called with value true in that case do you still need the parameter `val`

packages/terra-date-picker/src/DateInput.jsx
MadanKumarGovindaswamy: we are validating on every change triggered and if it is not a valid character we use this state to set a text and toggle the state for the inner text on visually hidden element.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -765,6 +790,10 @@ const DatePickerInput = (props) => {
     { 'is-invalid': isInvalid },
   ]);
 
+  // Indicates selected date from calendar popup for SR
+  const calendarDate = DateUtil.isValidDate(value, momentDateFormat) ? `${getLocalizedDateForScreenReader(DateUtil.createSafeDate(dateValue, initialTimeZone), { intl, locale: intl.locale })} ${intl.formatMessage({ id: 'Terra.datePicker.selected' })}` : label;
+  const inputDate = DateUtil.isValidDate(value, momentDateFormat) ? `${getLocalizedDateForScreenReader(DateUtil.createSafeDate(dateValue, initialTimeZone), { intl, locale: intl.locale })}` : '';
```
supreethmr, line 795: what is difference between calendar date and input date.

packages/terra-date-picker/src/DateInput.jsx
MadanKumarGovindaswamy: isInvalid prop  is used to highlight the date picker component.
Screenshot 2023-09-08 at 5 03 46 PM

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -794,7 +829,7 @@ const DatePickerInput = (props) => {
         <Button
           data-terra-open-calendar-button
           className={buttonClasses}
-          text={intl.formatMessage({ id: 'Terra.datePicker.openCalendar' })}
+          text={`${calendarDate} ${intl.formatMessage({ id: 'Terra.datePicker.openCalendar' })}`}
```
supreethmr, line 832: this should be set as aria-label instead to prevent tooltip from showing the selected date along with text `open calendar` when hover on button

packages/terra-date-picker/src/DateInput.jsx
MadanKumarGovindaswamy: changes done. Thanks.

packages/terra-date-picker/src/DateInput.jsx
supreethmr: you are calling this function only when there is invalid input. not being called when the valid character is entered. In this case you don't need the state variable as well as the parameter in function

MadanKumarGovindaswamy: Updated the handle function.
Screenshot 2023-09-08 at 5 43 08 PM
And we are toggling to false on valid change.
Screenshot 2023-09-08 at 5 44 52 PM

packages/terra-date-picker/src/DateInput.jsx
sugan2416: calendarDate is used to indicate the selected date in calendar popup (will be passed as part of aria-label for button)
 calendarDate value would be like "<date> selected"
 inputDate has the entered date (if valid) and it was added to convey the date along with format. It provides context when hot keys are used as well. It's value is only the date.

packages/terra-date-picker/src/DateInput.jsx
sugan2416: updated in https://github.com/cerner/terra-framework/pull/1774/commits/67e9043fe4a041fcc8573ca4f0cb82cc709ae204

packages/terra-date-picker/src/DateInput.jsx
sugan2416: refactored in https://github.com/cerner/terra-framework/pull/1774/commits/67e9043fe4a041fcc8573ca4f0cb82cc709ae204

packages/terra-date-picker/src/DateInput.jsx
MadanKumarGovindaswamy: Removed state dependency on Invalid key entry. Thanks

packages/terra-date-picker/src/DateInput.jsx
supreethmr: after above commit was this change tested on JAWS

supreethmr:
Can we write test to validate Button arialabel is being updated with expected value we can ignore the screenshot testing as we are only making changes to screenshot responses. attribute matchers can be used to validate the value. 
Reference :  `expect(button).toHaveAttribute('aria-label', 'value'); https://webdriver.io/docs/selectors/#link-text

sugan2416:
> Can we write test to validate Button arialabel is being updated with expected value we can ignore the screenshot testing as we are only making changes to screenshot responses. attribute matchers can be used to validate the value. Reference : `expect(button).toHaveAttribute('aria-label', 'value'); https://webdriver.io/docs/selectors/#link-text

Added tests in https://github.com/cerner/terra-framework/pull/1774/commits/c8e525783d2f4eec0d2833ce8cae0e08bbea5bd7

packages/terra-date-picker/src/DateInput.jsx
sugan2416: Tested after changes

supreethmr (approved):

packages/terra-date-picker/CHANGELOG.md
```diff
@@ -13,6 +13,10 @@
 * Changed
   * Visual focus changed to dashed lines for individual fields.
   * Fixed the date format being read 3 times when using SR
+  * Updated screen response to announce when invalid characters 
+    are entered.
+  * Updated screen response to announce when calendar popup dates 
+    are selected.
```
saket2403, line 19: Move this to unreleased section

saket2403 (approved):

packages/terra-date-picker/CHANGELOG.md
sugan2416: https://github.com/cerner/terra-framework/pull/1774/commits/657893f07bc56471de199d189c83edd40a751f4d

rahulkumar8cs:
+1 for invalid input and calendar date selection

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -116,6 +116,31 @@ const propTypes = {
    * The selected or entered date value to display in the date input.
    */
   value: PropTypes.string,
+  /**
+   * @private
```
sdadn, line 120: Why are these props private?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -116,6 +116,31 @@ const propTypes = {
...
+   * An ISO 8601 string representation of the minimum date that can be selected. The value must be in the `YYYY-MM-DD` format. Must be on or after `01/01/1900`
+   */
+  minDate: PropTypes.string,
+  /**
+   * A function that gets called for each date in the picker to evaluate which date should be disabled.
+   * A return value of true will be enabled and false will be disabled.
+   */
+  filterDate: PropTypes.func,
```
sdadn, line 143: We should mention this prop addition in the CHANGELOG.

sdadn (changes requested):

sugan2416:
> When I entered Excluded date 09/15/2023 VO did not announced it has invalid date. Same happened even when I re-visted the field.
> 
>  date-picker2.mov 
> [date-picker.mov.zip](https://github.com/cerner/terra-framework/files/12609619/date-picker.mov.zip)
> 
> Can you check this before merging the PR

It is announcing here. It can be seen in "more content available" description.
https://github.com/cerner/terra-framework/assets/40137430/aed3c9b3-8910-4722-b396-acb493e9e80f

packages/terra-date-picker/src/DateInput.jsx
sugan2416: These props are only internally used by DateInput component for date validation

packages/terra-date-picker/src/DateInput.jsx
sugan2416: Would it be required as we won't make it available for consumers, only being used as a passed down prop as mentioned above

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -765,6 +816,23 @@ const DatePickerInput = (props) => {
...
+  }
+  let calendarDate = inputDate ? `${inputDate} ${intl.formatMessage({ id: 'Terra.datePicker.selected' })}` : '';
+
+  // Check if date is excluded or out of range or not included or filtered
+  let invalidEntry = '';
+  if (DateUtil.isDateExcluded(DateUtil.createSafeDate(dateValue, initialTimeZone), props.excludeDates)
+      || DateUtil.isDateOutOfRange(DateUtil.createSafeDate(dateValue, initialTimeZone), DateUtil.createSafeDate(DateUtil.getMinDate(props.minDate), initialTimeZone), DateUtil.createSafeDate(DateUtil.getMaxDate(props.maxDate), initialTimeZone))
+      || DateUtil.isDateNotIncluded(DateUtil.createSafeDate(dateValue, initialTimeZone), props.includeDates)
+      || (props.filterDate && !props.filterDate(DateUtil.createSafeDate(dateValue, initialTimeZone)))) {
+    invalidEntry = intl.formatMessage({ id: 'Terra.datePicker.invalidDate' });
```
supreethmr, line 832: what happens if we call `handleInvalidInputChange` instead of this here

sugan2416: Used `invalidEntry`  to allow invalid date to be notified as and when the complete date is entered.  `handleInvalidInputChange`  does not take effect in this case, in spite of being re-written.

PR #1776, opened 2023-09-05
[terra-embedded-content-consumer] Add title prop to component so that title can be easily set into the frame

kolkheang:
### Summary

Add title prop to the component so that it's face up and easier to set the title of the content in the frame

**What was changed:**
Add title prop to the component so consumer can easily set the title of the iframe

**Why it was changed:**
For better accessibility, having a title text of the content in the frame is critical. This change is to make setting the title easier.

### Testing

This change was tested using:

- [ ] WDIO
- [x] Jest
- [x] Visual testing (please attach a screenshot or recording)
- [x] Other (please describe below)
- [ ] No tests are needed

The change can also be tested by inspecting the source code by looking for the `title` attribute for the `<iframe>` using the examples site: https://engineering.cerner.com/terra-framework/pull/1776.

### Reviews

In addition to engineering reviews, this PR needs:

- [ ] UX review
- [x] Accessibility review
- [x] Functional review

### Additional Details

VoiceOver announcing the title of the frame:
https://github.com/cerner/terra-framework/assets/3170314/01c0088d-9dbf-4255-95ae-f697350dc482

JAWS announcing the title of the frame:
https://github.com/cerner/terra-framework/assets/3170314/47ad7d1f-168a-4a73-8c11-104344a34b11

**This PR resolves:**

UXPLATFORM-9419 

---

Thank you for contributing to Terra.
@cerner/terra

PR #84, opened 2018-02-16
Add webdriver.io tests for terra-application-menu-layout

rakshith91:
### Summary
Consuming wdio tests for the package and getting rid of the nightwatch tests.
Resolves #83 

Thanks for contributing to Terra. 
@cerner/terra

[CONTRIBUTORS.md]: ../blob/master/CONTRIBUTORS.md
[License]: ../blob/master/LICENSE

PR #1777, opened 2023-09-05
[terra-embedded-content-consumer] Update example to show how to pass inline html content to be set in the frame

kolkheang:
### Summary

Add support for `srcdoc` prop to allow inline html content to be set into the frame

**What was changed:**

The `srcdoc` attribute to set inline html content for the frame can be done today without this change via the `options` prop. However, having a first class prop for this purpose would make it easier for consumers.

**Why it was changed:**

Support first class `srcdoc` prop to make it easier for consumers to set the inline html content for `srcdoc` frame attribute.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9424 

---

@cerner/terra


packages/terra-embedded-content-consumer/tests/wdio/embedded-content-consumer-spec.js
```diff
@@ -79,4 +79,13 @@ Terra.describeViewports('Embedded Content Consumer', ['tiny', 'large'], () => {
       expect($('#Authorized').isExisting());
     });
   });
+
+  describe('consumer with srcdoc prop', () => {
+    it('renders content inside of an iframe using content sets in srcdoc', () => {
```
zxeleven, line 84: small nit:
```suggestion
    it('renders content inside of an iframe using content set in srcdoc', () => {
```

zxeleven (approved):

chenyixuan625 (approved):

sycombs:
> The srcdoc attribute to set inline html content for the frame can be done today without this change via the options prop. However, having a first class prop for this purpose would make it easier for consumers.

If consumers are able to achieve the same functionality without adding a prop, is updating the API really necessary?

kolkheang:
> > The srcdoc attribute to set inline html content for the frame can be done today without this change via the options prop. However, having a first class prop for this purpose would make it easier for consumers.
> 
> If consumers are able to achieve the same functionality without adding a prop, is updating the API really necessary?

I think it helps to make `srcdoc` as a first class prop so consumers know up front. Without this, consumers do not know if it's even possible to pass this attribute to the frame today.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -37,8 +75,24 @@ const propTypes = {
 
 class EmbeddedContentConsumer extends React.Component {
   componentDidMount() {
+    // Merging the iframe options props
+    const frameOptions = { ...this.props.options };
+
+    if (this.props.srcdoc) {
+      if (frameOptions.iframeAttrs == null) {
```
supreethmr, line 82: to be more specific
```suggestion
      if (frameOptions.iframeAttrs === null) {
```

kolkheang: https://github.com/cerner/terra-framework/pull/1777/commits/5ceeb6dddfd25c257fde0c83c7fee19477b4a497

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -37,8 +75,26 @@ const propTypes = {
 
 class EmbeddedContentConsumer extends React.Component {
   componentDidMount() {
+    // Merging the iframe options props
+    const frameOptions = { ...this.props.options };
+
+    if (this.props.srcdoc) {
+      if (frameOptions.iframeAttrs === null
```
supreethmr, line 82: since options is not required prop it can be undefined as well so accessing property from undefined object (frameOptions.iframeAttrs) will throw up errors. Either we can use default props to set value for options prop to be `{}` so that it does not throw up an error while accessing the property `iframeAttrs`

kolkheang: There is a check for `frameOptions.iframeAttrs === undefined`.

sdadn:
> I think it helps to make `srcdoc` as a first class prop so consumers know up front. Without this, consumers do not know if it's even possible to pass this attribute to the frame today.

This sounds like a documentation issue, not an API one. Looking at your change, it doesn't reduce the lines of code a consumer has to write, just where they have to write it. From your example (and my understanding), it seems like the change will result in:

Before
```js
 const frameOptions = { 
  iframeAttrs: { 
    title: 'inline html content', 
    width: '100%', height: '100px', 
    srcdoc="<p><b>Inline HTML Content</b></p><p>This is an inline html content that can be used to render the content into the frame.</p>"
    } 
  };
  const embeddedContentConsumer = (
     <div>
       <EmbeddedContentConsumer
         options={frameOptions}
       />
    </div>
  );
```

After:
```js
 const frameOptions = { 
  iframeAttrs: { 
    title: 'inline html content', 
    width: '100%', height: '100px', 
    } 
  };
  const embeddedContentConsumer = (
     <div>
       <EmbeddedContentConsumer
         srcdoc="<p><b>Inline HTML Content</b></p><p>This is an inline html content that can be used to render the content into the frame.</p>"
         options={frameOptions}
       />
    </div>
  );
```

If consumers are already using the first method then why would they need to make a code change? This will also split consumers into using functionality two different ways. 

The `frameOptions`object is still being passed anyway so there's no technical advantage to using a dedicated prop for this feature.

If the motivation behind this change is lack of consumer awareness, then this is something that can be resolved through better documentation.

cc: @JessieRandle , @mjpalazzo , @sycombs

kolkheang:
https://github.com/cerner/terra-framework/pull/1777#issuecomment-1717756734

@sdadn - yeah, that is right. The `iframeAttrs` object is still used to send the `srcdoc` attribute to the iframe. If everyone thinks that we don't need to add `srcdoc` prop to this component, I can close this PR and update the documentation as part of the other PRs related to this component.

kolkheang:
> > I think it helps to make `srcdoc` as a first class prop so consumers know up front. Without this, consumers do not know if it's even possible to pass this attribute to the frame today.
> 
> This sounds like a documentation issue, not an API one. Looking at your change, it doesn't reduce the lines of code a consumer has to write, just where they have to write it. From your example (and my understanding), it seems like the change will result in:
> 
> Before
> 
> ```js
>  const frameOptions = { 
>   iframeAttrs: { 
>     title: 'inline html content', 
>     width: '100%', height: '100px', 
>     srcdoc="<p><b>Inline HTML Content</b></p><p>This is an inline html content that can be used to render the content into the frame.</p>"
>     } 
>   };
>   const embeddedContentConsumer = (
>      <div>
>        <EmbeddedContentConsumer
>          options={frameOptions}
>        />
>     </div>
>   );
> ```
> 
> After:
> 
> ```js
>  const frameOptions = { 
>   iframeAttrs: { 
>     title: 'inline html content', 
>     width: '100%', height: '100px', 
>     } 
>   };
>   const embeddedContentConsumer = (
>      <div>
>        <EmbeddedContentConsumer
>          srcdoc="<p><b>Inline HTML Content</b></p><p>This is an inline html content that can be used to render the content into the frame.</p>"
>          options={frameOptions}
>        />
>     </div>
>   );
> ```
> 
> If consumers are already using the first method then why would they need to make a code change? This will also split consumers into using functionality two different ways.
> 
> The `frameOptions`object is still being passed anyway so there's no technical advantage to using a dedicated prop for this feature.
> 
> If the motivation behind this change is lack of consumer awareness, then this is something that can be resolved through better documentation.
> 
> cc: @JessieRandle , @mjpalazzo , @sycombs

The other PRs to add `title` prop, and `isScrollable` prop for the iframe are also using the existing `iframeAttrs` within `options` prop to pass `title` and `scrolling` attributes to the iframe. 

cc: @scottwilmarth

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -37,8 +75,26 @@ const propTypes = {
...
+        || frameOptions.iframeAttrs === undefined) {
+        frameOptions.iframeAttrs = {};
+      }
```
kenk2, line 85:
```suggestion
   frameOptions.iframeAttrs = this.props.srcdoc && iframeOptions.attrs ? frameOptions.iframeAttrs : {}
```

Think this could be simplified this way.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -37,8 +75,26 @@ const propTypes = {
...
+
+      Object.assign(frameOptions.iframeAttrs, { srcdoc: this.props.srcdoc });
```
kenk2, line 87:
```suggestion
frameOptions.iframeAttrs = {
    ...frameOptions.iframeAttrs,
    srcdoc: this.props.srcdoc
}
```

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -37,8 +75,26 @@ const propTypes = {
...
+      Object.keys(frameOptions.iframeAttrs).forEach(key => {
```
kenk2, line 88: Question: Why are we filtering out null-or-undefined keys? Theoretically they shouldn't matter to the outcome if they don't have real values should they?

kenk2 (changes requested):

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
supreethmr: what happens when `frameOptions` is undefined ?

supreethmr:
> > I think it helps to make `srcdoc` as a first class prop so consumers know up front. Without this, consumers do not know if it's even possible to pass this attribute to the frame today.
> 
> This sounds like a documentation issue, not an API one. Looking at your change, it doesn't reduce the lines of code a consumer has to write, just where they have to write it. From your example (and my understanding), it seems like the change will result in:
> 
> Before
> 
> ```js
>  const frameOptions = { 
>   iframeAttrs: { 
>     title: 'inline html content', 
>     width: '100%', height: '100px', 
>     srcdoc="<p><b>Inline HTML Content</b></p><p>This is an inline html content that can be used to render the content into the frame.</p>"
>     } 
>   };
>   const embeddedContentConsumer = (
>      <div>
>        <EmbeddedContentConsumer
>          options={frameOptions}
>        />
>     </div>
>   );
> ```
> 
> After:
> 
> ```js
>  const frameOptions = { 
>   iframeAttrs: { 
>     title: 'inline html content', 
>     width: '100%', height: '100px', 
>     } 
>   };
>   const embeddedContentConsumer = (
>      <div>
>        <EmbeddedContentConsumer
>          srcdoc="<p><b>Inline HTML Content</b></p><p>This is an inline html content that can be used to render the content into the frame.</p>"
>          options={frameOptions}
>        />
>     </div>
>   );
> ```
> 
> If consumers are already using the first method then why would they need to make a code change? This will also split consumers into using functionality two different ways.
> 
> The `frameOptions`object is still being passed anyway so there's no technical advantage to using a dedicated prop for this feature.
> 
> If the motivation behind this change is lack of consumer awareness, then this is something that can be resolved through better documentation.
> 
> cc: @JessieRandle , @mjpalazzo , @sycombs

+1 If we are changing a way to pass down srcDoc then this qualifies for breaking change unless the existing way of using iframeAttrs still works with this change

eawww (approved):
+1 for no newly introduced UX or a11y issues. Though I do agree with @sdadn that this might be better addressed via documentation.

kolkheang:
https://github.com/cerner/terra-framework/pull/1777#issuecomment-1719909001

@supreethmr 

> +1 If we are changing a way to pass down srcDoc then this qualifies for breaking change unless the existing way of using iframeAttrs still works with this change

Yeah, it will be passive. Existing flow still works (when this new prop is not set). If the new prop is set, then that will take precedence.

Since this is really not about accessibility, I fine with not adding another prop to this component because there is already a way to pass `srcdoc` attribute to the iframe via `options` prop. My initial thought was that consumers would be able to quickly see the available option as supported method to render embedded content and easily set the content without having to construct the `iframeAttrs` object.

scottwilmarth:
Adding the same comment from issue #1776 for those with questions on "why the title attribute":
@sdadn and all, for an iFrame to be accessible it must have the title attribute. By including "title" as a first class prop we can implement measures to ensure the title is provided and eventually make it required. As is, it is not guaranteed that any engineer will add it. Yes we can beef up our documentation, but that also requires people to read which they do not regularly do. I am specifically asking for this prop to be added to ensure we help enforce best practices. While we have to make it a passive change for now, I will push to make it required as soon as possible.

scottwilmarth:
> Adding the same comment from issue #1776 for those with questions on "why the title attribute":
> @sdadn and all, for an iFrame to be accessible it must have the title attribute. By including "title" as a first class prop we can implement measures to ensure the title is provided and eventually make it required. As is, it is not guaranteed that any engineer will add it. Yes we can beef up our documentation, but that also requires people to read which they do not regularly do. I am specifically asking for this prop to be added to ensure we help enforce best practices. While we have to make it a passive change for now, I will push to make it required as soon as possible.

Please disregard this comment @sdadn  and all. @kolkheang  and I discussed this PR and it had failed A11y tests due to not having a Title attribute and I didn't fully read all the details. My bad... This is NOT about adding Title attribute, but instead SRCDOC. Forry for the confusion.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
kolkheang: Yeah, on line #84, an empty object is initialized and assigned to `frameOptions.iframeAttrs`.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
kolkheang: I'll re-work this logic a bit, but I get your point. We don't need to include `this.props.srcdoc` in the check.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
kolkheang: My initial thought was that we don't want to set bad values for the keys. However, I think by not filtering out the null or undefined, that would give the consumers more flexibility if they need to set something to null or undefined.

I will go ahead and remove this logic.

kolkheang:
If you all still think that we don't need to provide a first class `srcdoc` prop for Embedded Content Consumer and instead just use the `options` prop to pass in `srcdoc` with `{ iframeAttrs: { srcdoc: 'html content' } }`, then I will re-work this PR to make only the doc updates.

kolkheang:
I re-worked this PR to remove the `srcdoc` prop but keep the examples and using the `options` prop to pass in the `srcdoc` attribute to the `iframeAttrs` object.

sycombs (approved):

sdadn (dismissed):

sdadn (approved):
Since this PR is mainly a documentation updates, I'm fine with merging this for now and not keeping it indefinitely blocked. However, I don't think the work on the component is complete and there may be follow up PRs to complete the accessibility updates to this project.
 
cc: @kolkheang , @sycombs

PR #1784, opened 2023-09-07
Rename terra-worklist-data-grid package

smason0:
### Summary

**What was changed:**

**Why it was changed:**

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-XXXX 

---

@cerner/terra


packages/terra-data-grid/CHANGELOG.md
sdadn, line 1: Can we add a line above the 0.5 version stating that the component was formerly `terra-worklist-data-grid`?

smason0: Added

artpark (approved):

packages/terra-data-grid/src/WorklistDataGrid.module.scss
```diff
@@ -1,3 +1,4 @@
+/* stylelint-disable terra/custom-property-namespace */
```
sycombs, line 1: Should the classnames in this file be updated rather than disabling this rule? Is that something that will be done in a future story?

smason0: I was thinking these might need to be updated, but also to avoid breaking this rule, the names might get long/redundant (ex: `--terra-data-grid-worklist-data-grid-cell-focus-outline`). If we do prefer these be renamed, it should probably be done as part of this story.

trandrew1023 (approved):

packages/terra-data-grid/CHANGELOG.md
```diff
@@ -9,6 +14,9 @@
 * Changed
   * Updated the `terra-worklist-data-grid` component to provide row selection assistive technology messages based on the properties provided by the consumer.
 
+___
+All release notes below apply to `terra-worklist-data-grid` package (renamed with following release).
```
sdadn, line 18:
```suggestion
All release notes below apply to `terra-worklist-data-grid` package (renamed with the `0.6` release).
```

packages/terra-data-grid/src/WorklistDataGrid.module.scss
sdadn: We can look at refining the variable names when we refactor the implementation. But to simply changes, we should stick to renaming the package and imports in this PR/story.

packages/terra-data-grid/src/index.js
```diff
@@ -0,0 +1,5 @@
+import WorklistDataGrid from './WorklistDataGrid';
+
+// TODO: Re-enable linter rule when FlowsheetDataGrid is added.
```
sdadn, line 3: Since the package is data-grid, I think it would make sense for the default export to be the DataGrid component. However it is undecided whether or not we should expose component at this time. So we can remove this comment for now.

packages/terra-framework-docs/package.json
sdadn, line 1: we should probably remove the `terra-worklist-data-grid` dependency.

packages/terra-data-grid/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Changelog
 
 ## Unreleased
+
+* Breaking Changes
+  * Renamed `terra-worklist-data-grid` package to `terra-data-grid`.
+  * Changed WorklistDataGrid component from default to named export.
```
sdadn, line 7: Let's also indicate the new import method here:

````suggestion
  * Changed WorklistDataGrid component from default to named export. WorklistDataGrid can now be imported as `import { WorklistDataGrid } from 'terra-data-grid'`.
  
````

File: packages/terra-data-grid/package.json [hunk 5]
```diff
@@ -1,18 +1,18 @@
...
     "type": "git",
     "url": "git+https://github.com/cerner/terra-framework.git",
-    "directory": "packages/terra-worklist-data-grid"
+    "directory": "packages/terra-data-grid"
   },
   "keywords": [
     "Cerner",
     "Terra",
     "Framework",
-    "terra-worklist-data-grid",
```
sdadn, line 15: We can keep this keyword

sdadn (approved):

PR #1788, opened 2023-09-11
[terra-slider] New Terra slider component creation

nikhitasharma:
### Summary

**What was changed:**
Initial setup for terra slider component creation

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9555 

---

@cerner/terra


packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+    const sliderClassNames = cx([
+      'slider',
+      theme.className,
+    ]);
+
+    return (
+      <div className={sliderClassNames}>
+        <label className={cx('label')} aria-label={labelName}>{labelName}</label>
+        <input type="range" name="slider" aria-label="slider" disabled={isDisabled} min={minimumValue} max={maximumValue} step="any" />
```
cm9361, line 67: I would not expect the aria-label to be hard-coded to "slider".  Is there not a requirement to allow this value to be passed by consumers?

cm9361, line 67: Shouldn't the step be a prop that can be set by consumers?  It does not seem correct that it is being hard-coded to "any".

cm9361, line 67: I don't see a prop to allow consumers to specify the actual value for the slider.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,77 @@
...
+   */
+  intl: PropTypes.shape({ formatMessage: PropTypes.func }).isRequired,
+};
+
+const defaultProps = {
+  isDisabled: false,
+};
+
+/* eslint-disable-next-line react/prefer-stateless-function */
+class Slider extends React.Component {
```
cm9361, line 43: We should be creating a functional component not a React class component.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,77 @@
...
+    const theme = this.context;
+
+    const sliderClassNames = cx([
+      'slider',
+      theme.className,
+    ]);
+
+    return (
+      <div className={sliderClassNames}>
+        <label className={cx('label')} aria-label={labelName}>{labelName}</label>
```
cm9361, line 66: Do we want to use a label component in this scenario?  Would it be better to use a span element?

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,7 @@
+@import '~terra-mixins/lib/Mixins';
```
cm9361, line 1: Why do we need to import these mixins?

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,77 @@
...
+  /**
+  * Slider label value.
+  */
+  labelName: PropTypes.string.isRequired,
+
+  /**
+   * @private
+   * The intl object containing translations. This is retrieved from the context automatically by injectIntl.
+   */
+  intl: PropTypes.shape({ formatMessage: PropTypes.func }).isRequired,
```
cm9361, line 35: This is not being used.  Why do we need to add intl support at this point?

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,77 @@
...
+    const {
+      isDisabled,
+      minimumValue,
+      maximumValue,
+      labelName,
+    } = this.props;
+
+    const theme = this.context;
+
+    const sliderClassNames = cx([
```
cm9361, line 59: This variable isn't necessary.  It can all be done directly on line 65.

cm9361:
How are consumers notified of slider value changes?

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Maximum value in the range.
+  */
+  maximumValue: PropTypes.number.isRequired,
+
+  /**
+  * Slider label value.
+  */
+  labelName: PropTypes.string.isRequired,
```
cm9361, line 29: I thought the minimum and maximum values would be used for the labels based on the callouts?

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,77 @@
...
+   * The intl object containing translations. This is retrieved from the context automatically by injectIntl.
+   */
+  intl: PropTypes.shape({ formatMessage: PropTypes.func }).isRequired,
+};
+
+const defaultProps = {
+  isDisabled: false,
+};
+
+/* eslint-disable-next-line react/prefer-stateless-function */
```
cm9361, line 42: This rule had to be disabled because you did not create a functional component.  However, it should not have been disabled.

PR #1791, opened 2023-09-13
Fix for Active Key Indicator.

ST063655:
### Summary

**What was changed:**

If an active tab key is present in the tab data, the active key variable is set to prop active key else state the active key.

**Why it was changed:**

On Selecting the tab, the Dotted Focus is added but the underlined active key tab indicator is missing.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9640 

---

@cerner/terra

### DIFF
--- a/packages/terra-tabs/CHANGELOG.md
+++ b/packages/terra-tabs/CHANGELOG.md
@@ -4,7 +4,9 @@
 
 * Added
   * Added prop to support the focus traverse within the tab content.
-  * Fix for Active Key Style issue.
+
+* Fixed
+  * Fixed Active Key Style issue.
 
 ## 7.6.0 - (September 5, 2023)
 
--- a/packages/terra-tabs/CHANGELOG.md
+++ b/packages/terra-tabs/CHANGELOG.md
@@ -6,7 +6,7 @@
   * Added prop to support the focus traverse within the tab content.
 
 * Fixed
-  * Fixed Active Key Style issue.
+  * Fixed tab active indicator issue.
 
 ## 7.6.0 - (September 5, 2023)
 
--- a/packages/terra-tabs/CHANGELOG.md
+++ b/packages/terra-tabs/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Added
   * Added prop to support the focus traverse within the tab content.
+  * Fix for Active Key Style issue.
 
 ## 7.6.0 - (September 5, 2023)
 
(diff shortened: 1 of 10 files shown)


packages/terra-tabs/src/Tabs.jsx
```diff
@@ -137,7 +137,15 @@ class Tabs extends React.Component {
...
-    const activeTabKey = this.state.activeKey || this.props.activeKey;
+    const tabKeys = [];
+
+    if (this.props.children && this.props.children.length > 0) {
+      this.props.children.forEach(child => {
+        tabKeys.push(child.key);
+      });
+    }
+    const foundActiveTabKey = tabKeys.find(element => element === this.props.activeKey);
+    const activeTabKey = foundActiveTabKey ? this.props.activeKey : this.state.activeKey;
```
supreethmr, line 148: this can be simplified to : 
```suggestion
        const activeTab = React.Children.toArray(children).filter(child => child.key === this.props.activeKey);
        const activeTabKey = (activeTab) ? activeTab.key : this.state.activeKey
```

packages/terra-tabs/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Added
   * Added prop to support the focus traverse within the tab content.
+  * Fix for Active Key Style issue.
```
supreethmr, line 7: can we move this entry under Fixed section instead of Added
 * Fixed
   *  Fixed tab active indicator issue

supreethmr (approved):
+1 with above comments addressed

packages/terra-tabs/CHANGELOG.md
sdadn: The entry is too vague. What is the issue? We should write it as:

`Fixed <description of bug>`

sdadn (changes requested):

packages/terra-tabs/src/Tabs.jsx
ST063655: Added.

packages/terra-tabs/CHANGELOG.md
ST063655: Added.

packages/terra-tabs/src/Tabs.jsx
supreethmr: Curious !! why it didn't work.

ST063655: Using the above Suggestion, initially the active key is not getting set.

ST063655: Implemented the above suggested code.

saket2403 (approved):

PR #1795, opened 2023-09-14
Created DataGrid component and refactored WDG to use it

KoushikKalli:
### Summary
Create the DataGrid component in the terra-data-grid package and refactor WorklistDataGrid component to use the new component.

**Why it was changed:**
The evolution of design necessitated the creation of DataGrid to allow other grids such as WorklistDataGrid, Flowsheet etc. to consume DataGrid.

### Testing
This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

**This PR resolves:**

UXPLATFORM-9539


packages/terra-data-grid/CHANGELOG.md
sdadn, line 1: There was a recent release, can you update it with main?

packages/terra-data-grid/src/index.js
```diff
@@ -1,4 +1,5 @@
 import WorklistDataGrid from './WorklistDataGrid';
+import DataGrid from './DataGrid';
 
 // eslint-disable-next-line import/prefer-default-export
-export { WorklistDataGrid };
```
sdadn, line 4: I don't think we're exporting DataGrid for consumers at this time.

packages/terra-data-grid/tests/jest/Cell.test.jsx
```diff
@@ -109,7 +109,7 @@ describe('Cell', () => {
     expect(blankCell).toHaveLength(1);
 
     const cellContent = blankCell.find('div');
-    expect(cellContent.text()).toBe('Terra.worklistDataGrid.blank');
+    expect(cellContent.text()).toBe('Terra.data-grid.blank');
```
adoroshk, line 112: - I noticed that here we switched from camelCase to snake-case naming in translation variables. I checked with some other components and it looks like for consistancy it's better to keep it lowCamelCase (`Terra.dataGrid.blank` in this case) through all translations.

packages/terra-data-grid/src/subcomponents/Cell.jsx
```diff
@@ -197,20 +197,20 @@ function Cell(props) {
...
-        {intl.formatMessage({ id: 'Terra.worklistDataGrid.blank' })}
+        {intl.formatMessage({ id: 'Terra.data-grid.blank' })}
       </span>
     );
   } else {
     cellContent = children;
   }
 
-  const className = cx('worklist-data-grid-cell', {
+  const className = cx('data-grid-cell', {
```
cm9361, line 213: I would just use a class name of cell instead of data-grid-cell.  This component will be part of the Terra Table component.  This will just lead to the need for an additional refactor.

packages/terra-data-grid/src/proptypes/DataGridPropTypes.js
```diff
@@ -2,8 +2,8 @@ import rowShape from './rowShape';
 import cellShape from './cellShape';
 import { columnShape, SortIndicators } from './columnShape';
 
-const WorklistDataGridPropTypes = {
+const DataGridPropTypes = {
   cellShape, columnShape, rowShape, SortIndicators,
 };
 
-export default WorklistDataGridPropTypes;
+export default DataGridPropTypes;
```
cm9361, line 9: Do we really need this object?  It is just grouping import statements for us, correct?  These prop types will not live with the data grid long term so it would be nice to be able to remove this additional object to maintain.  Once we create the table, this would be another object that we would need to rename.

packages/terra-data-grid/src/subcomponents/Cell.module.scss
```diff
@@ -4,15 +4,15 @@
...
-    padding: var(--terra-worklist-data-grid-cell-content-padding, 7px);
+    padding: var(--terra-data-grid-cell-content-padding, 7px);
     position: relative;
 
     &:first-child {
-      border-left: var(--terra-worklist-data-grid-cell-border-left, 1px solid #dedfe0);
+      border-left: var(--terra-data-grid-cell-border-left, 1px solid #dedfe0);
     }
 
     input {
```
cm9361, line 18: The Cell component should not be controlling the styles for the input element.  This is a RowSelectionCell concern.  It should not be part of the Cell component sass.

packages/terra-data-grid/src/subcomponents/Row.jsx
```diff
@@ -141,13 +124,13 @@ function Row(props) {
...
       ariaLabel={ariaLabel}
-      onCellSelect={handleCellSelect}
+      onCellSelect={onCellSelect}
     />
   ) : null;
 
   return (
     <tr
-      className={cx('worklist-data-grid-row', {
+      className={cx('data-grid-row', {
```
cm9361, line 133: I would just make the class for this element, "row", instead of "data-grid-row".  The Row component will be a subcomponent of Terra Table.  With class mangling, there is no need for a data-grid prefix and will lead to another rename.

packages/terra-data-grid/src/DataGrid.module.scss
```diff
@@ -0,0 +1,25 @@
...
+    overflow: auto;
+    width: 100%;
+  }
+
+  .data-grid :focus {
+    outline: var(--terra-data-grid-cell-focus-outline, 3px dashed #000);
+    outline-offset: var(--terra-data-grid-cell-focus-outline-offset, -2px);
+  }
+
+  .data-grid {
```
cm9361, line 17: These two can be combined using nested styles.

packages/terra-data-grid/src/WorklistDataGrid.module.scss
```diff
@@ -1,29 +1,7 @@
...
-
 :local {
-  .worklist-data-grid :focus {
-    outline: var(--terra-worklist-data-grid-cell-focus-outline, 3px dashed #000);
-    outline-offset: var(--terra-worklist-data-grid-cell-focus-outline-offset, -2px);
-  }
-
   .worklist-data-grid-container {
     height: 100%;
     overflow: auto;
```
cm9361, line 4: Are you just creating another container for event listeners?  I don't think the WorklistDataGrid needs an overflow both vertically and horizontally.  That is being handled by the DataGrid component.  You would be introducing the possibility of multiple scrollbars.

packages/terra-data-grid/src/index.js
cm9361: Correct, we are not exposing the DataGrid component to consumers.

cm9361:
It looks like we changed some of the source data for some of our tests.  However, it does not seem like that should be necessary.

packages/terra-data-grid/src/subcomponents/Row.jsx
```diff
@@ -1,12 +1,12 @@
-import React, { useCallback, useContext, useState } from 'react';
+import React, { useContext, useState } from 'react';
 import PropTypes from 'prop-types';
 import ThemeContext from 'terra-theme-context';
 import classNames from 'classnames/bind';
 import styles from './Row.module.scss';
 import RowSelectionCell from './RowSelectionCell';
```
cm9361, line 6: Do we have a separate JIRA for removing the RowSelectionCell logic from this component?  The Row should not be dealing with the RowSelectionCell.  This is a DataGrid concept, and the logic should be moved to the appropriate component.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -0,0 +1,583 @@
...
+  /**
+   * Callback function that is called when all selected cells need to be unselected. Parameters: none.
+   */
+  onClearSelection: PropTypes.func,
+
+  /**
+   * Callback function that is called when a range selection occurs. Parameters:
+   * @param {number} rowIndex rowIndex
+   * @param {number} columnIndex columnIndex
+   * @param {number} direction direction
```
cm9361, line 105: We should have better descriptions for our function parameters.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -0,0 +1,583 @@
...
+   */
+  onRangeSelection: PropTypes.func,
+
+  /**
+   * Boolean indicating whether or not the DataGrid should allow entire rows to be selectable. An additional column will be
+   * rendered to allow for row selection to occur.
+   */
+  hasSelectableRows: PropTypes.bool,
```
cm9361, line 113: This is a WorklistDataGrid concept, right?

packages/terra-data-grid/src/subcomponents/Cell.jsx
cm9361, line 1: The isSelectable default prop should no longer be set by the Cell component.  This is a behavior of a grid component.  The logic should be handled at the DataGrid level.

packages/terra-data-grid/tests/jest/DataGrid.test.jsx
```diff
@@ -0,0 +1,286 @@
...
+        id="test-terra-data-grid"
+        pinnedColumns={dataFile.cols.slice(0, 2)}
+        overflowColumns={dataFile.cols.slice(2)}
+        rows={dataFile.rows}
+        onCellSelect={mockCellSelect}
+      />,
+    );
+
+    // Find column headers
+    const maskedCell = wrapper.find(Row).at(0).find('.masked');
```
kenk2, line 158: nit: Can we use `first()` instead of `at(0)` in all the instances in this file?

kenk2 (changes requested):

packages/terra-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -372,23 +230,23 @@ function WorklistDataGrid(props) {
...
+  const handleClearSelection = useCallback(() => {
+    if (!hasSelectableRows) {
+      if (onClearSelectedCells) {
+        onClearSelectedCells();
+      }
+    } else {
+      multiSelectRange.current = {}; // Clear the information used for selecting multiple rows.
+      if (rows.find(row => row.isSelected)) {
+        // Esc (while in row selection mode and rows are selected): Clear selection
+        if (onClearSelectedRows) {
```
cm9361, line 242: If this method is not defined, is there any need to check for selected rows?  It seems like moving the check up would make the logic in that scenario more efficient.

packages/terra-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -460,63 +318,25 @@ function WorklistDataGrid(props) {
...
-          const lastPinnedColumnRect = grid.current.rows[toCell.row].cells[pinnedColumnOffsets.length - 1].getBoundingClientRect();
-          scrollOffsetX = nextCellRect.left - lastPinnedColumnRect.right;
-        }
-      } else {
-        scrollOffsetX = nextCellRect.left - gridContainerRect.left;
-      }
-
-      if (scrollOffsetX < 0) {
-        gridContainerRef.current.scrollBy(scrollOffsetX, 0);
+  const onRangeSelection = useCallback((rowIndex, columnIndex, direction) => {
```
cm9361, line 321: You should add method documentation here to ensure that the logic is easier to maintain.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -0,0 +1,583 @@
...
+  const initializeColumn = (column) => {
+    const newColumn = { ...column };
+    newColumn.width = column.width || defaultColumnWidth;
+    newColumn.minimumWidth = column.minimumWidth || defaultColumnMinimumWidth;
+    newColumn.maximumWidth = column.maximumWidth || defaultColumnMaximumWidth;
+
+    return newColumn;
+  };
+
+  const displayedColumns = (hasSelectableRows ? [WorklistDataGridUtils.ROW_SELECTION_COLUMN] : []).concat(pinnedColumns).concat(overflowColumns);
```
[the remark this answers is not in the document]
sdadn: Are there any performance differences in spread vs concat? I think the readability issue would have been fixed if the concats were split off into separate lines.

packages/terra-data-grid/CHANGELOG.md
sdadn, line 1: Since this is an internal change that does not affect the API, we do not need a CHANGELOG entry.

packages/terra-data-grid/src/DataGrid.jsx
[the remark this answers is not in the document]
kenk2: @sdadn Concat is actually more performant. We could even newline this (in which case, we should think about a 
 prettier/eslint config because we currently don't seem to implement maxline limits). I suggested it this way because chaining many concat calls together can make us lose track of what's being tacked onto the array.

packages/terra-data-grid/src/subcomponents/Cell.module.scss
```diff
@@ -4,15 +4,15 @@
 @import '../clinical-lowlight-theme/Cell.module';
```
smason0, line 4: Should now be able to remove disabling the linter rule in the scss files in this directory

packages/terra-data-grid/src/DataGrid.jsx
KoushikKalli: Updated

packages/terra-data-grid/src/DataGrid.jsx
KoushikKalli: Yes and as we discussed for scoping reasons, a new JIRA will be created to remove hasSelectableRows from components other than WDG.

packages/terra-data-grid/src/proptypes/DataGridPropTypes.js
KoushikKalli: Removed

packages/terra-data-grid/src/subcomponents/Cell.jsx
KoushikKalli: Updated

packages/terra-data-grid/src/subcomponents/Row.jsx
KoushikKalli: Updated

packages/terra-data-grid/tests/jest/Cell.test.jsx
KoushikKalli: Updated

packages/terra-data-grid/CHANGELOG.md
KoushikKalli: Updated

KoushikKalli: Reverted

packages/terra-data-grid/src/subcomponents/Cell.jsx
cm9361: We would want the data grid to handle that because cells in a table role are not selectable.  The grid being created first should not control the base behavior of the table subcomponents.  Could you add the JIRAs to the PR so that we can ensure none are missed?

kenk2 (commented):
Code Quality Wise: Accepting that some re-factors will be implemented later on as part of other JIRA's and tasks. Otherwise LGTM for now. But will defer to others once they have had a chance to review this PR.

smason0 (approved):

sdadn (approved):

PR #1797, opened 2023-09-19
[terra-collapsible-menu-view] Changes for the terra-filepath component

artpark:
### Summary

**What was changed:**
- Fixes for lowlight and fusion themes icons. They were rendering horizontal ellipses when it should've been vertical
- Pulled in terra-hyperlink & terra-icons
- Added a private prop for horizontal ellipses in the responsive view
- Added a private prop to use a chevron right icon (">") instead of a vertical divider for CollapsibleMenuView.Divider
- Added a new item type called CollapsibleMenuView.Hyperlink
- Various CSS changes to get ready for terra-filepath

Screenshot 2023-09-25 at 10 06 54 AM

**Why it was changed:**
The new terra-filepath component will use CollapsibleMenuView as the underlying component.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9589 

---

@cerner/terra


packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
```diff
@@ -5,10 +5,12 @@ import classNames from 'classnames';
 import classNamesBind from 'classnames/bind';
 import ThemeContext from 'terra-theme-context';
 import { injectIntl } from 'react-intl';
+import IconEllipses from 'terra-icon/lib/icon/IconEllipses';
```
adoroshk, line 8:
```suggestion
import { IconEllipses } from 'terra-icon';
```
There is no problem with functionality, but as per @sdadn initiative we are trying to move to a proper import pattern and terra-icon@3.57.0 that makes it possible was just released, would be nice to implement it that way.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewDivider.jsx
```diff
@@ -2,22 +2,45 @@ import React from 'react';
...
+
 const contextTypes = {
   isCollapsibleMenuItem: PropTypes.bool,
 };
 
-const CollapsibleMenuViewDivider = (props, { isCollapsibleMenuItem }) => {
+const CollapsibleMenuViewDivider = ({ useChevronRightIcon }, { isCollapsibleMenuItem }) => {
   const theme = React.useContext(ThemeContext);
+  if (useChevronRightIcon) {
+    if (isCollapsibleMenuItem) {
```
smason0, line 26: I understand why this is here, but without the knowledge that the right chevron divider is made for the Filepath component, this is maybe unintuitive that it affects how the menu looks. A clarifying comment here might help. I'm also wondering if naming the prop differently (ex: isFilepath) might make it clear it changes more than just the divider styling while also describing its intended use case.

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
```diff
@@ -158,6 +181,7 @@ class CollapsibleMenuView extends React.Component {
       { 'is-calculating': this.isCalculating },
       { 'collapsible-menu-view-flex-end': !isStartAligned },
       { 'collapsible-menu-view-flex-start': isStartAligned },
+      { reverse: isReversed },
```
smason0, line 184: I noticed in testing the example that the tab order of these items is reversed (i.e. hitting Tab goes from right to left). Will this be addressed in a future change?

packages/terra-collapsible-menu-view/tests/jest/CollapsibleMenuViewDivider.test.jsx
```diff
@@ -16,6 +16,17 @@ describe('CollapsibleMenuViewDivider', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
+  it('should render a chevron right icon', () => {
+    const wrapper = shallowWithIntl(<CollapsibleMenuViewDivider useChevronRightIcon />);
```
smason0, line 20: nit: Can we add assertions for these tests? Like for this one actually checking for the existence of the icon programmatically

cm9361:
It looks like the orientation of the ellipsis changed for all scenarios.  If so, is this desired by UX.  A non-passive change would require a major version bump.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
```diff
@@ -0,0 +1,132 @@
...
+        onMouseDown={onMouseDown}
+        variant={variant}
+      />
+    </div>
+  );
+};
+
+CollapsibleMenuViewHyperlink.propTypes = propTypes;
+CollapsibleMenuViewHyperlink.defaultProps = defaultProps;
+CollapsibleMenuViewHyperlink.contextTypes = contextTypes;
```
cm9361, line 130: This looks like a concept on a class component instead of your new functional component.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
```diff
@@ -0,0 +1,132 @@
...
+  EXTERNAL: 'external',
+  IMAGE: 'image',
+  VIDEO: 'video',
+  AUDIO: 'audio',
+  DOCUMENT: 'document',
+};
+
+const contextTypes = {
+  isCollapsibleMenuItem: PropTypes.bool,
+};
```
cm9361, line 21: What is a contextType?

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
```diff
@@ -0,0 +1,132 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import Hyperlink from 'terra-hyperlink';
+import Menu from './_CollapsibleMenu';
+import styles from './CollapsibleMenuView.module.scss';
```
cm9361, line 6: You are pulling styles from a different component module into your component.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
```diff
@@ -0,0 +1,132 @@
...
+   * Sets the hyperlink variant. One of `default`, `external`, `image`, `video`, `audio`, `document`.
+   */
+  variant: PropTypes.oneOf(['default', 'external', 'image', 'video', 'audio', 'document']),
+};
+
+const defaultProps = {
+  variant: variants.DEFAULT,
+};
+
+const CollapsibleMenuViewHyperlink = ({
```
cm9361, line 73: What is the purpose of this component?  It is not documented.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
```diff
@@ -0,0 +1,132 @@
...
+{ isCollapsibleMenuItem }) => {
+  const handleOnClick = (event) => {
+    if (href) {
+      window.location.href = href;
+    } else {
+      onClick(event);
+    }
+  };
+
+  if (isCollapsibleMenuItem) {
```
cm9361, line 95: If this is a collapsible menu hyperlink, why would isCollapsibleMenuItem be false?

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewDivider.jsx
```diff
@@ -2,22 +2,45 @@ import React from 'react';
...
+   */
+  useChevronRightIcon: PropTypes.bool,
+};
+
 const contextTypes = {
   isCollapsibleMenuItem: PropTypes.bool,
 };
 
-const CollapsibleMenuViewDivider = (props, { isCollapsibleMenuItem }) => {
+const CollapsibleMenuViewDivider = ({ useChevronRightIcon }, { isCollapsibleMenuItem }) => {
```
cm9361, line 23: Should we change this from props?  What if we want to add additional props in the future?  Do we want to have a default prop value for this?

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewDivider.jsx
cm9361, line 26: Seems like these two conditions can be combined into one.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewDivider.jsx
```diff
@@ -2,22 +2,45 @@ import React from 'react';
...
   const theme = React.useContext(ThemeContext);
+  if (useChevronRightIcon) {
+    if (isCollapsibleMenuItem) {
+      return null;
+    }
+
+    return (
+      <div className={cx(['chevron-right', 'face-up-item', theme.className])}>
```
cm9361, line 31: Why do we need an additional div element around the icon?

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
```diff
@@ -45,6 +42,24 @@ const propTypes = {
...
+   * @private
+   * Switches to using horizontal ellipses instead of vertical ones when menu items are truncated.
+   */
+  useHorizontalIcon: PropTypes.bool,
+
+  /**
+   * @private
+   * Reverses the order of the rendered items.
+   */
+  isReversed: PropTypes.bool,
```
cm9361, line 56: I don't know that "isReversed" is clear enough for consumers of the API.  For example, which direction is the normal direction.  Do we need to make the property about the collapse direction?  Maybe that would be more clear.

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
```diff
@@ -45,6 +42,24 @@ const propTypes = {
    * Aligns the menu to the start of the container
    */
   isStartAligned: PropTypes.bool,
+
+  /**
+   * @private
```
cm9361, line 47: Should this be marked a private?  isn't it used externally in your tests?

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
```diff
@@ -45,6 +42,24 @@ const propTypes = {
...
+   * Switches to using horizontal ellipses instead of vertical ones when menu items are truncated.
+   */
+  useHorizontalIcon: PropTypes.bool,
```
cm9361, line 50: If we have this prop, why did the screen shots change for the ellipsis?  Shouldn't that only happen for your FilePath test?

packages/terra-framework-docs/src/terra-dev-site/test/collapsible-menu-view/Filepath.test.jsx
```diff
@@ -0,0 +1,92 @@
+import React from 'react';
+import CollapsibleMenuView from 'terra-collapsible-menu-view';
+
+const FilepathTestDemo = () => (
```
cm9361, line 4: CollapsibleMenu shouldn't know about the Filepath component.  You should probably give the test a generic name.

cm9361:
Should there be Readme or GitHub page updates since you are adding new API props?

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewDivider.jsx
artpark: I can add a comment saying that it will not be rendered in the collapsed menu view! However, I'm not sure about naming it `isFilepath` because all these changes try to not couple collapsible menu view and filepath as much as possible.

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
artpark: The order will be reversed again in `terra-filepath`, I tried to not add another set of logic within here to not clutter it with filepath code

artpark: I didn't catch that, thanks. I didn't realize flex-direction also affected tab order :(

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
artpark: I'm not sure what you mean? Defining types is the same between a class and a functional component right?

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
artpark: It's used for checking whether the item should be displayed or truncated into the collapsed menu

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
artpark: With the way that collapsible menu view was, all the scss was in one file. Unfortunately I don't have the time to split out all the scss into separate files for time sake

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
artpark: It can either be displayed outside the dropdown menu or not

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewDivider.jsx
artpark: I had a css issue where the svg of the icon was getting squished by the padding size. This div ensures that the svg will always be at its appropriate size

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
artpark: This is a private prop that is not intended to be used by consumers

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
artpark: Yes since it's not intended to be used by consumers. I still would like to test that it works though, I'm not sure what the standard is but can you not have private props that are still tested?

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
artpark: This was a bug discovered when I was working on this component, all themes other than default was rendering the wrong icon. This PR also fixes it to be consistent across all themes

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
sdadn: If it's only for internal use then also using them for tests are fine imo.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
sdadn, line 1: Is this an internal component or an exported one?

artpark: Exported one, I was debating on whether or not to private it but I think it should be alright for consumers to use

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
artpark: Just added a brief documentation about the hyperlink to the terra docs

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
sdadn: If it's a component we want to advertise and have consumers to use, then we should create an `index.js` file and have it as a named export instead of relying on paths.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
cm9361: No, it is not.  You would use the useContext hook for functional components.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewDivider.jsx
cm9361: Yes, you can destructure.  i get how it is working.  I guess my concern is for readability.  You could just as easily destructure from props.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewDivider.jsx
cm9361: Did you try using margin instead of padding?

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
cm9361: Even if it is private, I think is should be renamed for better readability and maintenance in the future.

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.module.scss
```diff
@@ -73,14 +62,20 @@
     height: var(--terra-collapsible-menu-view-divider-height, 2.143rem);
   }
 
+  .chevron-right {
+    padding: 0;
+    padding-left: var(--terra-collapsible-menu-view-chevron-right-padding-left, 0.714rem);
```
cm9361, line 67: You are setting the padding left to 0 and then changing it on the next line.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
cm9361: This does not seem correct.

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
```diff
@@ -171,26 +219,46 @@ class CollapsibleMenuView extends React.Component {
...
+      if (this.props.isReversed) {
+        hiddenChildren = visibleChildren.splice(0, this.hiddenStartIndex + 2).concat(hiddenChildren);
+      } else {
+        hiddenChildren = this.collapsedMenuAlwaysShown
+          ? visibleChildren.splice(this.hiddenStartIndex).concat(prepopulatedBaseDivider).concat(hiddenChildren)
+          : visibleChildren.splice(this.hiddenStartIndex).concat(hiddenChildren);
+      }
+    }
+
+    const icon = useHorizontalIcon ? <IconEllipses /> : <IconEllipses className={cx('vertical-icon')} />;
```
cm9361, line 232: You would conditionally render the class name not the entire React component.  It should be easy with the classnames module.

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.module.scss
```diff
@@ -22,6 +22,10 @@
     justify-content: flex-start;
   }
 
+  .vertical-icon {
```
cm9361, line 25: This class name might be a little vague.  What icon is vertical?

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewDivider.jsx
artpark: Margin works however with the way collapsible menu view is setup, it calls getBoundingClientRect on each child, which ignores margins. So it was throwing off resizing calculations

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
artpark: Same as above, the legacy API https://legacy.reactjs.org/docs/legacy-context.html#referencing-context-in-stateless-function-components

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
artpark: @cm9361 What do you think about `isReversedChildrenOrder`? I'm struggling to come up with a good name for it

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
artpark: It's not something we're actively advertising though, so now I'm wondering if we _should_ private it

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
cm9361: We shouldn't be using a legacy API for new development.  The page that you linked says that it is going to be removed.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
cm9361: We should not be using a legacy API.

artpark: Discussion continued here: https://github.com/cerner/terra-framework/pull/1797#discussion_r1338751838

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
trandrew1023: That name makes sense to me

packages/terra-collapsible-menu-view/src/index.js
sdadn, line 1: +1

packages/terra-collapsible-menu-view/tests/jest/CollapsibleMenuViewHyperlink.test.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+describe('CollapsibleMenuViewHyperlink', () => {
+  // Snapshot Tests
+  it('should render a default component', () => {
+    const wrapper = shallowWithIntl(<CollapsibleMenuViewHyperlink text="Testing" href="https://www.oracle.com/" />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  // Props
+  it('should merge custom props', () => {
```
sdadn, line 14: What does this test do? I'm confused at what the before/after would be.

packages/terra-collapsible-menu-view/tests/jest/CollapsibleMenuViewHyperlink.test.jsx
```diff
@@ -0,0 +1,26 @@
...
+    const wrapper = shallowWithIntl(<CollapsibleMenuViewHyperlink text="Testing" href="https://www.oracle.com/" className="Testing" />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  describe('Collapsible Menu Context', () => {
+    it('should render a menu item', () => {
+      const context = { isCollapsibleMenuItem: true };
+      const wrapper = shallowWithIntl(<CollapsibleMenuViewHyperlink text="Testing" href="https://www.oracle.com/" />, { context }).dive();
+      expect(wrapper).toMatchSnapshot();
```
sdadn, line 23: we should assert for the existence of a menu item.

packages/terra-collapsible-menu-view/tests/jest/CollapsibleMenuViewDivider.test.jsx
```diff
@@ -16,6 +17,18 @@ describe('CollapsibleMenuViewDivider', () => {
...
+    expect(wrapper.find(<IconChevronRight />)).toBeTruthy();
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should not render a menu divider if using the chevron right icon in the context of the collapsible menu', () => {
+    const context = { isCollapsibleMenuItem: true };
+    const wrapper = shallowWithIntl(<CollapsibleMenuViewDivider useChevronRightIcon />, { context });
+    expect(wrapper).toMatchSnapshot();
```
sdadn, line 29: We should also assert that a menu divider does not exist

trandrew1023 (approved):

smason0 (approved):

packages/terra-collapsible-menu-view/tests/jest/CollapsibleMenuViewHyperlink.test.jsx
artpark: It was an existing test that the other collapsible menu view components used, but I think it doesn't really test anything so I'll remove it

packages/terra-collapsible-menu-view/tests/jest/CollapsibleMenuViewHyperlink.test.jsx
artpark: The wrapper _is_ the menu item, so the snapshot will cover this

artpark: Also for whatever reason, I can't seem to do wrapper.exists(Menu.Item) here, it just can't find it even though it is clearly rendering it in the snapshot :/

sdadn: > The wrapper _is_ the menu item, so the snapshot will cover this
Ah okay, I thought it was a nested element. That should be fine then.

sdadn (approved):

adoroshk (approved):

cm9361 (approved):

PR #1805, opened 2023-09-21
Focus should remain on Grid when Row Selection Mode is turned off using Grid shortcut

brithom:
### Summary
Fixed issue where WorklistDataGrid loses focus when Row Selection Mode is turned off from a Row Selection Cell.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

**This PR resolves:**

UXPLATFORM-9655


packages/terra-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -244,6 +253,9 @@ function WorklistDataGrid(props) {
           onClearSelectedRows();
         }
       } else if (onDisableSelectableRows) {
+        // In order to set focus to the right cell after row selection is turned off, keep
+        // track of cell in grid that had focus when the call was initiated.
+        gridFocusedWhenRowSelectionDisabled.current = true;
```
cm9361, line 258: Shouldn't we be able to handle this with the useEffect for the row selection prop?  It doesn't seem like we should need an extra ref variable.

packages/terra-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -392,9 +404,18 @@ function WorklistDataGrid(props) {
     }
   };
 
+  const onFocus = () => {
+    gridReceivedFocus.current = true;
```
cm9361, line 408: This is going to trigger a lot.  You don't have the logic to ensure that you are not changing focus between children like the DataGrid component.

packages/terra-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -192,6 +195,12 @@ function WorklistDataGrid(props) {
     // Since the row selection mode has changed, the row selection mode needs to be updated.
     setRowSelectionModeAriaLiveMessage(intl.formatMessage({ id: hasSelectableRows ? 'Terra.worklistDataGrid.row-selection-mode-enabled' : 'Terra.worklistDataGrid.row-selection-mode-disabled' }));
 
+    if (gridReceivedFocus.current) {
+      let newFocusCell = dataGridFuncRef.current.getFocusedCell();
+      newFocusCell = { row: newFocusCell.row, col: Math.max(newFocusCell.col + (hasSelectableRows ? 1 : -1), 0) };
+      dataGridFuncRef.current.setFocusedRowCol(newFocusCell.row, newFocusCell.col, gridFocusedWhenRowSelectionDisabled.current);
+    }
```
cm9361, line 202: Do we really need to track focus?  I believe the DataGrid already calls the method to set focus when it is received.  All we would need to do is ensure the proper state is set before focus is received.  That should be able to be handled in the useEffect.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -208,6 +207,20 @@ function DataGrid(props) {
...
+  // allow those components to request focus change.
+  useImperativeHandle(
+    props.focusFuncRef,
+    () => ({
+      setFocusedRowCol,
+      getFocusedCell() { return { row: focusedRow, col: focusedCol }; },
+    }),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [focusedCol, focusedRow],
+  );
```
cm9361, line 222: Are we making this more complicated than necessary with the useImperativeHandle hook?

packages/terra-data-grid/src/WorklistDataGrid.jsx
cm9361: You have a focus event.  That would let you know if you received focus.

packages/terra-data-grid/src/WorklistDataGrid.jsx
cm9361: Then should you remove the event listener after the first trigger?  Is there a reason to maintain the listener afterwards?

brithom: Done

packages/terra-data-grid/src/WorklistDataGrid.jsx
brithom: Discussed offline and came up with alternate solution of using the dom.

packages/terra-data-grid/tests/wdio/worklist-data-grid-spec.js
```diff
@@ -174,7 +174,14 @@ Terra.describeViewports('WorklistDataGrid', ['medium', 'large'], () => {
       rowSelectionNavigateToCell(4, 1);
       browser.keys(['Escape']); // clears row selection mode
 
-      Terra.validates.element('row-selection-mode-off-cell-4-1-focused', { selector: defaultSelector });
+      Terra.validates.element('row-selection-mode-off-cell-4-0-focused', { selector: defaultSelector });
+    });
+
+    it('clears row selection mode when focus is on a row selection cell', () => {
```
sdadn, line 180: This reads as "row selection mode is cleared due to receiving focus on a row selection cell". Based on the PR description, I would suggest:

```suggestion
    it('does not change focus when row selection mode is cleared when focused on a row selection cell', () => {
```

I would also suggest asserting that focus is on the right element.

sdadn (approved):

brithom:
@sdadn, you added "UX Review Required" but we do not think this is needed so I am going to remove it. We did not change any of the screenshots or added any new ones. With the exception of the bug, everything should behave the same. I am adding @eawww just in case he thinks otherwise.

artpark (approved):

PR #1810, opened 2023-09-27
[terra-data-grid] Create base FlowsheetDataGrid component

smason0:
### Summary

**What was changed:**
Added new FlowsheetDataGrid component, using newly created DataGrid component as a base. This component accepts one row header column that is always pinned, and does not allow resizing of columns. Also added FlowsheetDataGrid as an export from the main entrypoint (index.js) of the package.

**Why it was changed:**
This will serve as the starter component needed to begin implementing an accessible flowsheet-style data grid. More features are to be added in future updates.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9401 

---

@cerner/terra


packages/terra-data-grid/tests/wdio/flowsheet-data-grid-spec.js
```diff
@@ -0,0 +1,16 @@
+const defaultSelector = '#default-terra-flowsheet-data-grid';
+
+Terra.describeViewports('FlowsheetDataGrid', ['medium', 'large'], () => {
+  describe('Grid focus', () => {
+    beforeEach(() => {
+      browser.url('/raw/tests/cerner-terra-framework-docs/data-grid/flowsheet-data-grid/default-flowsheet-data-grid');
+    });
+
+    it('validates that the first cell gains initial focus when Tab is pressed', () => {
```
cm9361, line 9: Isn't this test already covered by the DataGrid component?  Are we worried that we introduced logic that requires us to retest this functionality?

packages/terra-data-grid/tests/jest/__snapshots__/FlowsheetDataGrid.test.jsx.snap
```diff
@@ -0,0 +1,99 @@
...
+    ariaLabel="Test Flowsheet Data Grid"
+    columnHeaderHeight="2.5rem"
+    defaultColumnWidth={200}
+    id="test-terra-flowsheet-data-grid"
+    overflowColumns={
+      Array [
+        Object {
+          "displayName": "March 16",
+          "id": "Column-1",
+          "isResizable": false,
```
cm9361, line 17: We should test not passing a value for isResizable as well as passing a value of true.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * The column containing row headers. This column will be pinned to the leftmost side of the grid.
+   */
+  rowHeaderColumn: columnShape,
+
+  /**
+   * Data for overflow columns. Overflow columns are rendered in the Flowsheet Data Grid's horizontal overflow.
+   * Columns will be presented in the order given.
+   */
+  overflowColumns: PropTypes.arrayOf(columnShape),
```
cm9361, line 42: If we cannot have pinned columns in the flowsheet component, then I would just name this prop, "columns", for simplicity.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -0,0 +1,102 @@
...
+    <div className={cx('flowsheet-data-grid-container')}>
+      <DataGrid
+        id={id}
+        ariaLabel={ariaLabel}
+        ariaLabelledBy={ariaLabelledBy}
+        rows={rows}
+        rowHeight={rowHeight}
+        rowHeaderIndex={0}
+        pinnedColumns={[{ ...rowHeaderColumn, isResizable: false }]}
+        overflowColumns={overflowColumns.map(column => ({ ...column, isResizable: false }))}
```
cm9361, line 91: Does this spread operator approach overwrite existing isResizable values?  If someone passed true for the prop, would this overwrite it to false?

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -0,0 +1,102 @@
...
+
+  /**
+   * Data for content in the body of the Grid. Rows will be rendered in the order given. The first cell in each row will be the row header.
+   */
+  rows: PropTypes.arrayOf(rowShape),
+
+  /**
+   * The column containing row headers. This column will be pinned to the leftmost side of the grid.
+   */
+  rowHeaderColumn: columnShape,
```
cm9361, line 36: Instead of having a separate prop, can we just enforce that the first column of the columns array is the pinned column?  Then consumers just need to create a single array.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -0,0 +1,102 @@
...
+  rowHeight: PropTypes.string,
+};
+
+const defaultProps = {
+  defaultColumnWidth: 200,
+  columnHeaderHeight: '2.5rem',
+  rowHeight: '2.5rem',
+  rowHeaderColumn: {},
+  overflowColumns: [],
+};
```
cm9361, line 66: You will probably want a default value for the rows prop as well.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * String that specifies the column height. Any valid CSS height value is accepted.
+   */
+  columnHeaderHeight: PropTypes.string,
+
+  /**
+   * String that specifies the height for the rows in the grid. Any valid CSS value is accepted.
+   */
+  rowHeight: PropTypes.string,
+};
```
cm9361, line 58: I think you still need the onCellSelect callback function for the Flowsheet.

packages/terra-data-grid/tests/jest/FlowsheetDataGrid.test.jsx
```diff
@@ -0,0 +1,83 @@
...
+        overflowColumns={dataFile.cols.slice(1)}
+        rows={dataFile.rows}
+        ariaLabel="Test Flowsheet Data Grid"
+      />,
+    );
+
+    const dataGrid = wrapper.find('InjectIntl(DataGrid)');
+    const pinnedColumns = dataGrid.prop('pinnedColumns');
+    expect(pinnedColumns).toEqual(
+      [{ ...dataFile.cols[0], isResizable: false }],
```
cm9361, line 62: This test isn't really clear to me.  We should directly test that the columns have isResizable set to false.

packages/terra-data-grid/tests/jest/FlowsheetDataGrid.test.jsx
```diff
@@ -0,0 +1,83 @@
...
+      <FlowsheetDataGrid
+        id="test-terra-flowsheet-data-grid"
+        rowHeaderColumn={dataFile.cols[0]}
+        overflowColumns={dataFile.cols.slice(1)}
+        rows={dataFile.rows}
+        ariaLabel="Test Flowsheet Data Grid"
+      />,
+    );
+
+    const dataGrid = wrapper.find('InjectIntl(DataGrid)');
```
cm9361, line 59: Can't we just find the DataGrid component?

packages/terra-data-grid/tests/jest/FlowsheetDataGrid.test.jsx
```diff
@@ -0,0 +1,83 @@
...
+    const pinnedColumns = dataGrid.prop('pinnedColumns');
+    expect(pinnedColumns).toEqual(
+      [{ ...dataFile.cols[0], isResizable: false }],
+    );
+
+    const overflowColumns = dataGrid.prop('overflowColumns');
+    expect(overflowColumns).toEqual(
+      [
+        { ...dataFile.cols[1], isResizable: false },
+        { ...dataFile.cols[2], isResizable: false },
```
cm9361, line 69: If you only have two columns and the array is 0 based, how does this work?

packages/terra-framework-docs/src/terra-dev-site/test/data-grid/flowsheet-data-grid/DefaultFlowsheetDataGrid.test.jsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const { cols, rows } = gridDataJSON;
+
+  return (
+    <FlowsheetDataGrid
+      id="default-terra-flowsheet-data-grid"
+      rowHeaderColumn={cols[0]}
+      overflowColumns={cols.slice(1)}
+      rows={rows}
+      columnHeaderHeight="50px"
```
cm9361, line 69: The "default" test would not have the column header height as a prop.  That prop is not required.

packages/terra-data-grid/tests/wdio/flowsheet-data-grid-spec.js
smason0: Correct, I can probably change this to simply display the component rather than testing the tab stop. That should be covered by the DataGrid test

packages/terra-framework-docs/src/terra-dev-site/test/data-grid/flowsheet-data-grid/DefaultFlowsheetDataGrid.test.jsx
smason0: I only added this because I see that the other "Default" tests also use this, (WDG, and DataGrid), so it is visually consistent. I agree though, we might want to consider removing this from the others as well

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
smason0: Yes, it overrides whatever is set for isResizable (if set to true, it will be false)

packages/terra-data-grid/tests/jest/FlowsheetDataGrid.test.jsx
smason0: Is there an advantage to using the actual component? We should be able to use this selector to access what is needed for the test the same way, without requiring an additional import.

packages/terra-data-grid/tests/jest/FlowsheetDataGrid.test.jsx
smason0: I can break this out into what is actually in the object array rather than using spread

packages/terra-data-grid/tests/jest/FlowsheetDataGrid.test.jsx
smason0: The "dataFile" object has three columns, with the overflow being indices 1 and 2

packages/terra-data-grid/tests/jest/__snapshots__/FlowsheetDataGrid.test.jsx.snap
smason0: Yes, that is what I tried to do by passing 'true' for col[0] and col[2] and nothing for col[1] in the test data. The end result in this snapshot is that they are all 'false' as we would expect.

packages/terra-data-grid/src/index.js
```diff
@@ -1,4 +1,4 @@
 import WorklistDataGrid from './WorklistDataGrid';
+import FlowsheetDataGrid from './FlowsheetDataGrid';
 
-// eslint-disable-next-line import/prefer-default-export
```
sdadn, line 3: Do we still need this linter disable line?

packages/terra-framework-docs/src/terra-dev-site/test/data-grid/flowsheet-data-grid/DefaultFlowsheetDataGrid.test.jsx
sdadn: Let's remove it then and make a note to update other tests to match this.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import DataGrid from './DataGrid';
+import rowShape from './proptypes/rowShape';
+import { columnShape } from './proptypes/columnShape';
+import styles from './FlowsheetDataGrid.module.scss';
```
sdadn, line 7: Nit: but I would prefer grouping imports into 3rd party, terra and then internal imports for better readability.
```suggestion
import React from 'react';
import PropTypes from 'prop-types';
import classNames from 'classnames/bind';

import DataGrid from './DataGrid';
import rowShape from './proptypes/rowShape';
import { columnShape } from './proptypes/columnShape';
import styles from './FlowsheetDataGrid.module.scss';
```

packages/terra-data-grid/src/index.js
smason0: It isn't needed, since there are now multiple exports

packages/terra-framework-docs/src/terra-dev-site/test/data-grid/flowsheet-data-grid/DefaultFlowsheetDataGrid.test.jsx
```diff
@@ -0,0 +1,75 @@
...
+          { content: 'Cardiac Index (L/min/m2)' },
+          { content: '2.25' },
+          { content: '2.28' },
+          { content: '2.8' },
+        ],
+      },
+      {
+        id: '4',
+        cells: [
+          { content: 'ICP (mmHg)', isMasked: true },
```
eawww, line 43: Note: Probably fine for a test but when we get to doc site examples, those should not include masked row headers.

eawww (commented):
- Not sure if it was introduced in this PR but the cells seem to be clipping a bit of the bottom of the text 
image
- Since this is providing the base for upcoming development it should include at least a base doc page template.

packages/terra-framework-docs/src/terra-dev-site/test/data-grid/flowsheet-data-grid/DefaultFlowsheetDataGrid.test.jsx
smason0: Yes, this was a carryover from the other data grids. I can exclude this for the consumer-facing examples

smason0:
> * Not sure if it was introduced in this PR but the cells seem to be clipping a bit of the bottom of the text
I'm noticing this as well with Worklist Data Grid when I make the column small enough
Screenshot 2023-09-29 at 3 33 43 PM
This seems to be because the line-height doesn't evenly fit within the available space to fit in two lines. Would we possibly want to change the default row height of DataGrid to better accommodate two rows of text?

packages/terra-data-grid/tests/jest/FlowsheetDataGrid.test.jsx
cm9361: I am not sure if there is an advantage.  It just seems like your test is wanting to find the DataGrid and InjectIntl is an implementation detail.  Maybe it is not a big deal.

cm9361 (approved):

packages/terra-data-grid/tests/jest/FlowsheetDataGrid.test.jsx
smason0: I see other examples where using the actual component might be beneficial, like if we are validating props of subcomponents, but I think in this case it shouldn't matter whether using this enzyme selector or the component itself.

eawww (approved):
Not gonna do a full a11y or UX review since this is mostly just a the broth for the Flowsheet soup.

Logged a jira for the text being chopped off
UXPLATFORM-9736

sdadn (approved):

packages/terra-framework-docs/src/terra-dev-site/test/data-grid/data-grid/DefaultDataGrid.test.jsx
```diff
@@ -68,6 +68,7 @@ const DefaultDataGrid = () => {
         overflowColumns={cols}
         rows={rows}
         rowHeaderIndex={0}
+        // TODO: Remove columnHeaderHeight from example
```
KoushikKalli, line 71: Is this comment for future JIRAs?

KoushikKalli: Sounds good. I'm approving the PR.

KoushikKalli (approved):

trandrew1023 (approved):

chrismichalewicz:
@smason0  Added the Functional Review label, right now looks good but like Eric indicated this is just the beginning with more functionality to be added to it so it definitely an incomplete review at this point. Will perform additional functional reviews on future PR's

PR #1813, opened 2023-09-28
[terra-worklist-data-grid] added keyboard support for column resizing

sdadn:
### Summary

This PR adds keyboard column support for the datagrid. Users can now navigate column headers and resize them using keyboard shortcuts.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**Note:** due to the architecture changes to datagrid, this PR will not be merged into main. Instead, it will be used as a reference to integrate it into table & datagrid in a separate PR. This will still need review to finalize the design & implementation.

**This PR resolves:**

UXPLATFORM-9147 

---

@cerner/terra


packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -153,14 +189,46 @@ const ColumnHeaderCell = (props) => {
...
+      const parentTable = columnHeaderCellRef.current.closest('table');
+
+      // Update resize handle height to match parent table height
+      if (parentTable) {
+        return `${parentTable.offsetHeight}px`;
+      }
+    }
+
+    return null;
+  };
```
cm9361, line 230: You should no longer need this logic with the ResizeObserver being in the DataGrid component.  You should be able to accept the table height as a prop and use directly.  It will also allow you to remove the polyfill.

packages/terra-data-grid/src/subcomponents/ColumnResizeHandle.jsx
```diff
@@ -85,28 +114,115 @@ const ColumnResizeHandle = (props) => {
...
+  const [isAriaValueText, setIsAriaValueText] = useState(false);
+
+  const fitToTable = () => {
+    // Find parent table element
+    const parentTable = resizeHandleRef.current.closest('table');
+
+    // Update resize handle height to match parent table height
+    if (parentTable) {
+      resizeHandleRef.current.style.height = `${parentTable.offsetHeight}px`;
+    }
```
cm9361, line 131: You should be able to get the table height from the prop passed by the DataGrid component down to the subcomponents.

packages/terra-data-grid/src/subcomponents/ColumnResizeHandle.jsx
```diff
@@ -85,28 +114,115 @@ const ColumnResizeHandle = (props) => {
...
+
   return (
-    /* eslint-disable react/forbid-dom-props */
-    // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/interactive-supports-focus
+    // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions, jsx-a11y/no-static-element-interactions
     <div
       ref={resizeHandleRef}
       draggable
-      role="slider"
+      role={!isNavigationEnabled ? 'slider' : 'divider'}
```
cm9361, line 204: Should we use isNavigationEnabled instead of "!isNavigationEnabled" to deal with the positive case in the comparison?

packages/terra-data-grid/src/subcomponents/ColumnResizeHandle.module.scss
```diff
@@ -4,22 +4,23 @@
...
+    
     &:focus {
-      border-color: var(--terra-data-grid-resize-handle-focus-border-color, rgba(34, 42, 46, 1)) !important;
+      border-color: var(--terra-data-grid-resize-handle-focus-border-color, #007cc3) !important;
       outline: 0 !important;
     }
 
     &:hover {
-      border-color: var(--terra-data-grid-resize-handle-hover-border-color, rgba(34, 42, 46, 0.6));
+      border-color: var(--terra-data-grid-resize-handle-hover-border-color, #909496);
```
cm9361, line 23: There should not be a need to change from RGBA to HEX.  Also, I am not sure that this conversion is correctly accounting for opacity.

sdadn: I thought it was odd that this was using RGBA vs HEX in the other stylesheets but you're right about the opacity. Reverting it for now.

packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.jsx
sdadn: Oh i didn't commit those changes. Committed here: 82615a08708ec4735399361487866e70b9601cff

packages/terra-data-grid/src/subcomponents/ColumnResizeHandle.module.scss
sdadn: Removed here: 010c3f9825fddd05ed9c04a188bf0c8a422f0df3

packages/terra-data-grid/src/subcomponents/ColumnResizeHandle.jsx
sdadn: reversed condition here: 04a321ed7dfea78806cae4df33e09d43c17bb022

packages/terra-data-grid/src/subcomponents/ColumnResizeHandle.jsx
sdadn: Changed here:

sdadn: and removed `if` block here: 9569340cc2fa8ff84ac8eb88dff2325fb14cd05c

packages/terra-data-grid/src/subcomponents/ColumnResizeHandle.jsx
```diff
@@ -1,80 +1,110 @@
...
-  useContext, useRef, useCallback, useState,
+  useContext, useRef, useEffect, useState,
 } from 'react';
 import PropTypes from 'prop-types';
 import { injectIntl } from 'react-intl';
+import * as KeyCode from 'keycode-js';
 import classNames from 'classnames/bind';
 import ThemeContext from 'terra-theme-context';
+import ColumnContext from '../utils/ColumnContext';
+import '../_elementPolyfill';
```
kenk2, line 10: Question about this: What part of the code is using this polyfill?

packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -153,10 +196,28 @@ const ColumnHeaderCell = (props) => {
...
+        if (isResizable && isResizeHandleActive) {
+          columnHeaderCellRef.current.focus();
+          setResizeHandleActive(false);
+
+          event.stopPropagation();
+          event.preventDefault();
+        }
+        break;
+      case KeyCode.KEY_RIGHT:
+        if (isResizable && !isResizeHandleActive) {
```
brithom, line 215: The _isResizeHandleActive_ is not getting reset if you tab in and out of the grid from the resizeHandler. 
As a result when you tab back into the grid and navigate right, resizeHandler is skipped and focus goes to the next cell. 
Not sure of the required behavior but when we tab back into the grid either
-  focus goes back to the resizeHandler and the right arrow then goes to the next cell or
- focus goes to the cell and the right arrow goes to the resizeHandler

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -65,6 +65,11 @@ const propTypes = {
    */
   columnHeaderHeight: PropTypes.string,
 
+  /**
+   * Numeric increment in pixels to adjust column width when resizing via the keyboard
+   */
+  columnResizeIncrement: PropTypes.number,
```
KoushikKalli, line 71: With the introduction of this prop, is the `onColumnResize` prop still needed? I could not find it in any test or example.

sdadn: @cm9361 thoughts?

supreethmr:
Can we get PR description updated it still says this as Draft PR

packages/terra-data-grid/src/DataGrid.module.scss
```diff
@@ -13,6 +13,8 @@
     // Style for the grid itself
     border-collapse: separate;
     border-spacing: 0;
+    // Allow for resizing of last grid column
+    margin-right: 30px;
```
supreethmr, line 17: does this holds good for all the viewports

sdadn:
> Can we get PR description updated it still says this as Draft PR

Forgot to update that, thanks!!

packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.jsx
sdadn: Fixed in 74fb315bfbf49d46e4c4fb4dfe7a02fb165df62b. Tabbing back will now return focus to the cell and right arrow will go to the resizeHandler. Currently working on adding tests for it.

packages/terra-data-grid/src/DataGrid.jsx
cm9361: Consumers new the onColumnResize so that they can respond to the column being resized.  They might want to store preferences or make modifications to the new size or update the size of other columns, for example.  Yes, we need the onColumnResize prop.

packages/terra-data-grid/src/DataGrid.module.scss
cm9361: This provides extra space so that the final column can be resized.  Do you have an example in mind where maybe more room is required?  We also did not want to take too much room from the table itself.

supreethmr: Since the value has provided with `px` this would remain same on all the viewports. There might be few scenarios where data-grid might have minimum columns or it might get required space to get displayed without scroll in such scenarios due to this margin we might notice scroll because of the additional space added by margin OR Extra white space only on right-end part of grid. Curious to know if this change was suggested / approved by UX team.

Screenshot 2023-10-19 at 6 36 46 PM

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -536,8 +575,12 @@ const DataGrid = injectIntl((props) => {
             columns={dataGridColumns}
             headerHeight={columnHeaderHeight}
             tableHeight={tableHeight}
+            activeColumnIndex={focusedRow === 0 ? focusedCol : undefined}
+            activeColumnResizing={focusedRow === 0 && checkResizable}
+            columnResizeIncrement={columnResizeIncrement}
             onColumnSelect={handleColumnSelect}
             onResizeMouseDown={onResizeMouseDown}
+            onResizeHandleChange={onResizeHandleChange}
```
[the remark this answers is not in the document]
sdadn: Do we only want to enable resizing if a consumer provides the the`onColumnsResize` prop?

cc: @cm9361

packages/terra-framework-docs/src/terra-dev-site/test/data-grid/data-grid/DataGridWithColumnResizing.test.jsx
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
```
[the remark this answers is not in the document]
sdadn: Updated tests to utilize this file here: 
7cd98b589a489e5490f845202dd1d840554ae18e

packages/terra-data-grid/src/subcomponents/ColumnResizeHandle.jsx
sdadn: Looks like we aren't so removed it here: 9340a4ceec311a633eef24d9bca8e6a00c211c75

eawww:
Works great!
As far as the styling of the line goes, it's looking like the focus state was used in place of the active state so now when we focus it with the arrow keys it looks as though it's activated and then there's no visible change when the slider is active. Can that be reswizzled so the styling for the keyboard interaction matches the styling for the mouse interaction?

sdadn:
> Works great! As far as the styling of the line goes, it's looking like the focus state was used in place of the active state so now when we focus it with the arrow keys it looks as though it's activated and then there's no visible change when the slider is active. Can that be reswizzled so the styling for the keyboard interaction matches the styling for the mouse interaction?

@eawww I updated the styles here: 01783570c0c966ed387a0ea606a605ce1e5521a2

KoushikKalli (approved):

chrismichalewicz:
@sdadn  I tested for accessibility. It worked really well, see my results below. Marking this as Accessibility reviewed. Great job! 

Tested for accessibility
Keyboard: 
Was able to enable column resize on the handle by using the arrow keys to the handle and pressing enter.  At which point was able to resize the column with the left/right arrow keys. Pressed esc to resume grid navigation.  All worked well

Screen reader:
Tested with Edge + JAWS Version 2022.2202.3
When navigating to the resize handle it is properly communicated that I am on the resize handle and what column it pertains to. Instructions provided "Press Enter" worked. 

When I resized with the arrow keys, the new width was communicated appropriately. Pressing the arrows repeatedly quickly worked well in that it only communicated width after I was completed pressing the arrow keys 

When Esc was pressed it communicated I am no longer resizing the column and resuming the grid navigation.

brithom (approved):

packages/terra-data-grid/tests/jest/ColumnHeader.test.jsx
```diff
@@ -84,10 +83,6 @@ describe('ColumnHeader', () => {
     expect(columnHeader).toHaveLength(1);
     expect(columnHeader.props().height).toBe('3rem');
 
-    // Validate ColumnHeaderCell React component
```
kenk2, line 87: Why was this removed? Shouldn't this exist?

packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -196,8 +254,8 @@ const ColumnHeaderCell = (props) => {
       role="columnheader"
       scope="col"
       aria-sort={sortIndicator}
-      onMouseDown={(isSelectable && onColumnSelect) ? handleMouseDown : undefined}
-      onKeyDown={(isSelectable && onColumnSelect) ? handleKeyDown : undefined}
+      onMouseDown={onColumnSelect ? handleMouseDown : null}
```
kenk2, line 257: Why `null` and not `undefined` like we have before? I believe if these are set `null` then these will show up in  snapshots versus undefined.

packages/terra-data-grid/src/subcomponents/ColumnHeader.jsx
```diff
@@ -14,31 +14,48 @@ const propTypes = {
...
   tableHeight: PropTypes.number,
-
+  /**
+   * Column index for cell that can receive tab focus.
+   */
+  activeColumnIndex: PropTypes.number,
+  /**
+   * Specifies if resize handle should be active.
+   */
+  activeColumnResizing: PropTypes.bool,
```
kenk2, line 28: nit: If this is a bool, it feels weird that we don't prefix this as `is`. Non-blocking though.

kenk2 (commented):
Just a few comments, not blocking really. Otherwise LGTM.

packages/terra-data-grid/src/subcomponents/ColumnResizeHandle.jsx
```diff
@@ -85,28 +119,109 @@ const ColumnResizeHandle = (props) => {
...
-      aria-valuemin={minimumWidth}
-      aria-valuenow={columnWidth}
-      aria-valuemax={maximumWidth}
-      aria-label={columnText}
-      aria-valuetext={intl.formatMessage({ id: 'Terra.dataGrid.resizeHandleValueText' }, { columnWidth })}
-      style={{ height: `${height}px` }}
+      aria-valuemin={isActive ? minimumWidth : null}
+      aria-valuenow={isActive ? columnWidth : null}
+      aria-valuemax={isActive ? maximumWidth : null}
+      aria-label={isAriaLabel ? intl.formatMessage({ id: 'Terra.worklist-data-grid.resize-handle-template' }, { columnText }) : null}
```
eawww, line 209: For some reason this 'Terra.worklist-data-grid.resize-handle-template' is being set as the actual aria-label upon focusing the resize handle.

packages/terra-data-grid/tests/jest/ColumnResizeHandle.test.jsx
```diff
@@ -40,7 +65,131 @@ describe('ColumnResizeHandle', () => {
...
+  });
+
+  it('sets aria-label on focus', async () => {
+    const wrapper = mountWithIntl(
+      <ColumnResizeHandle />,
+    );
+
+    wrapper.find('.resize-handle').simulate('focus');
+
+    expect(wrapper.find('.resize-handle').props()['aria-label']).toBe('Terra.worklist-data-grid.resize-handle-template');
```
eawww, line 111: We're testing for the bug value?

packages/terra-data-grid/tests/jest/ColumnResizeHandle.test.jsx
```diff
@@ -40,7 +65,131 @@ describe('ColumnResizeHandle', () => {
...
+      <ColumnContext.Provider value={{ setColumnHeaderAriaLiveMessage: jest.fn() }}>
+        <ColumnResizeHandle />
+      </ColumnContext.Provider>,
+    );
+
+    wrapper.find('.resize-handle').simulate('focus');
+    wrapper.find('.resize-handle').simulate('keydown', { keyCode: 32 }); // space
+
+    expect(wrapper.find('.resize-handle').props()['aria-label']).toBe(null);
+    expect(wrapper.find('.resize-handle').props()['aria-valuetext']).toBe('Terra.worklist-data-grid.resize-handle-value-text');
```
eawww, line 125: Here too

packages/terra-data-grid/tests/jest/ColumnResizeHandle.test.jsx
sdadn: Correct, that should be the default value when no locale is set.

eawww: This is the i18n key rather than the string it's supposed to look up. It shouldn't be the aria-label.

packages/terra-data-grid/tests/jest/ColumnHeader.test.jsx
sdadn: Good catch, added these back here: 9cb7c6db2d658fe3b335b1ed87c0c867b887ad13

packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.jsx
sdadn: Yeah, you're right. I originally changed them for consistency but we don't want these props to show up if they are not used. Reverted them back here: b00d53fbb56b9221531f03703da298489563f9d4

packages/terra-data-grid/src/subcomponents/ColumnHeader.jsx
sdadn: Updated variable name here: 3abc11520c3c9130410fc26d759cea04bf1d41bf

eawww (approved):
⭐️⭐️⭐️⭐️⭐️... ⭐️

packages/terra-data-grid/src/DataGrid.module.scss
sdadn: I think it's for giving the last resize handle a but more room to be grabbed. Do we still want it? 
cc: @cm9361

packages/terra-data-grid/src/subcomponents/ColumnResizeHandle.jsx
sdadn: This got magically fixed on its own.

packages/terra-data-grid/src/DataGrid.module.scss
eawww: Oh yeah is this the little bit of extra space that the handle needs to be able to expand? If so, is that the smallest it can be?

cm9361: Yes, this is the extra space that the handle needs to expand for the last row.  I tried to pick a value that would be comfortable for users.  I am sure we can select a smaller value and it still work.

packages/terra-data-grid/src/DataGrid.jsx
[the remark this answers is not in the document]
cm9361: I don't think we want to prevent the resize just because they don't want to handle the callback.  The grid is actually resized at this point.

packages/terra-data-grid/src/subcomponents/ColumnResizeHandle.module.scss
```diff
@@ -1,25 +1,35 @@
...
+      border-color: var(--terra-data-grid-resize-handle-hover-border-color, rgba(34, 42, 46, 0.6));
       outline: 0 !important;
     }
 
     &:hover {
       border-color: var(--terra-data-grid-resize-handle-hover-border-color, rgba(34, 42, 46, 0.6));
     }
   }
+
+  .resize-handle-selected {
```
cm9361, line 32: Should this class just be "selected" instead of "resize-handle-selected" since it is localized?  Also, we should be able to add this under the .resize-handle style block.

sdadn: Updated here: 46c86b0cfd4e7b4787ba5483f5141f63b1099093

packages/terra-data-grid/src/DataGrid.module.scss
sdadn: Reduced to 15px here: 17fa632348ba196e03112e7502690d002bbc1c68

cm9361 (approved):

packages/terra-data-grid/tests/jest/ColumnHeader.test.jsx
```diff
@@ -86,6 +86,7 @@ describe('ColumnHeader', () => {
 
     // Validate ColumnHeaderCell React component
     const columnHeaderCells = columnHeader.find(ColumnHeaderCell);
+    // console.log(columnHeaderCells.first().props())
```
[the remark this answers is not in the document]
sdadn: Oh good catch. Removed here: 429c45056e4d10e8b347f7d10da950384fde7f86

kenk2 (approved):

supreethmr (approved):

PR #1822, opened 2023-10-03
[terra-slider] New Terra slider component creation 

nikhitasharma:
### Summary

**What was changed:**
Adding new Terra Slider component along with documentation

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9555
UXPLATFORM-9600

---

@cerner/terra


packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,110 @@
...
+
+  return (
+    <div className={sliderClassNames}>
+      <label className={cx('label')} id="sliderLabel">
+        {labelName}
+      </label>
+      <span className={cx('min-label')}>
+        {minLabel}
+      </span>
+      <input className={cx('input-range')} type="range" ref={sliderRef} aria-labelledby="sliderLabel" aria-valuemin={minimumValue} aria-valuemax={maximumValue} aria-valuenow={defaultValue} defaultValue={defaultValue} disabled={isDisabled} min={minimumValue} max={maximumValue} onChange={handleValueChange} />
```
sycombs, line 99: This is a lot of attributes on one line, can we break them up so it's more readable?

```suggestion
      <input
        className={cx('input-range')}
        type="range"
        ref={sliderRef}
        aria-labelledby="sliderLabel"
        aria-valuemin={minimumValue}
        aria-valuemax={maximumValue}
        aria-valuenow={defaultValue}
        defaultValue={defaultValue}
        disabled={isDisabled}
        min={minimumValue}
        max={maximumValue}
        onChange={handleValueChange}
      />
```

packages/terra-slider/translations/en-US.json
```diff
@@ -0,0 +1,3 @@
+{
+  "Terra.slider.Slider": "Slider"
```
sycombs, line 2: Is there a JIRA logged to add the translations for other locales?

packages/terra-slider/tests/jest/Slider.test.jsx
```diff
@@ -0,0 +1,25 @@
...
+      <Slider minimumValue={0} maximumValue={100} labelName="testLabel" defaultValue={50} />,
+    );
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render a disabled slider', () => {
+    const wrapper = shallow(
+      <Slider isDisabled minimumValue={0} maximumValue={100} labelName="testLabel" defaultValue={50} />,
+    );
+    expect(wrapper).toMatchSnapshot();
```
sycombs, line 16: Can we add an assertion to check the value of the `disabled` attribute on the `input` element?

packages/terra-slider/tests/jest/Slider.test.jsx
```diff
@@ -0,0 +1,25 @@
...
+  });
+
+  it('should render a slider with custom min and max labels', () => {
+    const wrapper = shallow(
+      <Slider minimumValue={0} maximumValue={100} labelName="testLabel" minimumLabel="testMinimumLabel" maximumLabel="testMaximumLabel" defaultValue={50} />,
+    );
+    expect(wrapper).toMatchSnapshot();
```
sycombs, line 23: Similar to the above comment, can we add an assertion to check the values of the minimum and maximum labels?

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+  * Maximum value in the range.
+  */
+  maximumValue: PropTypes.number.isRequired,
+
+  /**
+  * Default or initial value for the slider.
+  */
+  defaultValue: PropTypes.number.isRequired,
```
cm9361, line 28: This property should probably be called, value.  We want the consumer to be able to specify the value of the component not the default value.

cm9361:
We should probably create a test and example where the consumer is changing the slider value via the prop.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  .max-label {
+    color: #6f7477;
+    font-size: var(--terra-slider-range-label-font-size, 12px);
+    font-weight: 400;
+    margin-left: 7px;
+    text-align: left;
+    white-space: initial;
+  }
```
cm9361, line 39: You have the same values for both classes.  You should just create one class for both the min and max labels.

cm9361, line 39: You have the same values for both classes.  You should just create one class for both the min and max labels.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,212 @@
...
+  .slider {
+    align-items: center;
+    display: inline-flex;
+    min-height: 44px;
+  }
+
+  .label {
+    color: var(--terra-slider-label-color, #1c1f21);
+    font-weight: 700;
+    margin-right: 7px;
```
cm9361, line 17: You should have theme variables from the font weight and margin

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,212 @@
...
+
+// Themes
+@import './orion-fusion-theme/Slider.module';
+@import './clinical-lowlight-theme/Slider.module';
+
+:local {
+  .slider {
+    align-items: center;
+    display: inline-flex;
+    min-height: 44px;
```
cm9361, line 11: You should have a theme variable for this value.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,212 @@
...
+    color: var(--terra-slider-label-color, #1c1f21);
+    font-weight: 700;
+    margin-right: 7px;
+    text-align: center;
+    white-space: initial;
+    width: 20%;
+  }
+
+  .min-label {
+    color: #6f7477;
```
cm9361, line 24: This makes me think the lowlight theme is not being handled properly.  There should be a theme variable as I would not expect the font color to be the same for lowlight and light themes.  It should also be configurable for other themes.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,212 @@
...
+    font-size: var(--terra-slider-range-label-font-size, 12px);
+    font-weight: 400;
+    margin-right: 7px;
```
cm9361, line 27: You should have theme variables for these styles.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,212 @@
...
+
+  .input-range:disabled::-webkit-slider-thumb {
+    -webkit-appearance: none; /* stylelint-disable-line property-no-vendor-prefix, plugin/no-unsupported-browser-features */
+    background: #007cc3;
+  }
+
+  /* For IE/Edge */
+  .input-range::-ms-track {
+    -webkit-appearance: none; /* stylelint-disable-line property-no-vendor-prefix, plugin/no-unsupported-browser-features */
+    background: linear-gradient(to right, var(--terra-slider-track-selected-range, #007cc3) 0%, var(--terra-slider-track-selected-range, #007cc3) var(--terra-slider-progress-status, 50%), var(--terra-slider-track-unselected-range, #d3d4d5) var(--terra-slider-progress-status, 50%), var(--terra-slider-track-unselected-range, #d3d4d5) 100%);
```
cm9361, line 113: I would expect this to be one single theme variable.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,212 @@
...
+    border-radius: 3px;
+    height: 6px;
+    width: 6px;
```
cm9361, line 116: You should have theme variables for these values.

packages/terra-slider/src/Slider.module.scss
cm9361, line 1: You have a lot of styles with no theme variables.  You need to identify the set of variables necessary and ensure that the styles can be themed.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  .input-range::-webkit-slider-thumb {
+    -webkit-appearance: none; /* stylelint-disable-line property-no-vendor-prefix, plugin/no-unsupported-browser-features */
+    background: var(--terra-slider-thumb-background, #007cc3);
+    border-color: var(--terra-slider-thumb-border-color, #0065a3);
+    border-radius: 50%;
```
cm9361, line 68: Did the border radius of 50% come from a UX callout?  Usually the value is in rem units.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,212 @@
...
+    border-style: solid;
+    border-width: 1px;
+    box-shadow: 0 0 4px 0 rgba(0, 0, 0, 0.25);
+    cursor: pointer;
+    height: 24px;
+    margin-top: -8px;
+    width: 24px;
```
cm9361, line 75: These should be themeable.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  .input-range:focus::-webkit-slider-thumb {
+    border-color: var(--terra-slider-thumb-focus-border-color, #007cc3);
+    border-width: 1px;
```
cm9361, line 80: Why not just use the border style?
border: 1px solid var(...)

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,212 @@
...
+    outline: var(--terra-slider-thumb-focus-outline, 2px dashed #000);
+    outline-offset: 1px;
+  }
+
+  .input-range:hover::-webkit-slider-thumb {
+    background: var(--terra-slider-thumb-hover-background, #0065a3);
+    border-color: var(--terra-slider-thumb-hover-border-color, #0065a3);
```
cm9361, line 87: In your logic the background color and border color are the same.  Do you just not need a border?

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  .input-range:active::-webkit-slider-thumb {
+    background: var(--terra-slider-thumb-active-background, #007cc3);
+    border-color: var(--terra-slider-thumb-active-border-color, #0065a3);
+  }
+
+  .input-range:hover::-webkit-slider-runnable-track {
+    background: linear-gradient(to right, var(--terra-slider-track-selected-range-hover, #0065a3) 0%, var(--terra-slider-track-selected-range-hover, #0065a3) var(--terra-slider-progress-status, 50%), var(--terra-slider-track-unselected-range-hover, #bcbfc0) var(--terra-slider-progress-status, 50%), var(--terra-slider-track-unselected-range-hover, #bcbfc0) 100%);
```
cm9361, line 96: I would expect this to be a single theme variable.  Same comment for the other gradient backgrounds.  The current API is too complicated.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,212 @@
...
+    outline-offset: 1px;
+  }
+
+  .input-range:hover::-ms-thumb {
+    background: var(--terra-slider-thumb-hover-background, #0065a3);
+  }
+
+  .input-range:active::-ms-thumb {
+    background: var(--terra-slider-thumb-active-background, #007cc3);
+  }
```
cm9361, line 146: The colors for focus and active are the same.  You should be able to combine those styles in your stylesheet.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+  * Minimum value label to display on the left side of the slider.
+  */
+  minimumLabel: PropTypes.string,
+
+  /**
+  * Maximum value label to display on the right side of the slider.
+  */
+  maximumLabel: PropTypes.string,
```
cm9361, line 43: Do we need a minimum label and maximum label?  Isn't the requirement to show the actual values?  We already have props for the minimum and maximum value.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,110 @@
...
+
+  const sliderRef = useRef();
+
+  useEffect(() => {
+    sliderRef.current.style.setProperty('--terra-slider-progress-status', `${sliderRef.current.value}%`);
+  }, []);
+
+  function handleValueChange(event) {
+    event.currentTarget.style.setProperty('--terra-slider-progress-status', `${event.currentTarget.value}%`);
+    event.currentTarget.setAttribute('aria-valuenow', `${event.currentTarget.value}`);
```
cm9361, line 75: This is not the proper way to update the attributes in React.

cm9361:
You have a blue border around your slider when it has focus.  The UX callouts do not specify this behavior.  It seems like the border should not exists and the outline just around the slider thumb.

packages/terra-slider/src/Slider.jsx
cm9361, line 99: Should we be leveraging our Terra Arrange component for the min label, slider, max label layout?

cm9361, line 99: You shouldn't need aria-valuemin and aria-valuemax since you are using the native input control.  We shouldn't need aria-valuenow once we update the implementation to allow consumers to pass the value.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,212 @@
...
+    min-height: 44px;
+  }
+
+  .label {
+    color: var(--terra-slider-label-color, #1c1f21);
+    font-weight: 700;
+    margin-right: 7px;
+    text-align: center;
+    white-space: initial;
+    width: 20%;
```
cm9361, line 20: Why is the width 20%.  Wouldn't a label just take up the space necessary?  This seems like an i18n issue.

packages/terra-slider/src/Slider.module.scss
cm9361, line 1: A lot of your style values are in pixels.  However, we usually are supposed to leverage rem units for accessibility.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: We do not have same values for these classes, the margin needs to be on the right for min label and on the left for max label.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,212 @@
...
+    -webkit-appearance: none; /* stylelint-disable-line property-no-vendor-prefix, plugin/no-unsupported-browser-features */
+    background-color: transparent;
+    margin-left: 7px;
+    margin-right: 7px;
+    min-height: 44px;
+    min-width: 240px;
+    width: 70%;
+  }
+  
+  .input-range:disabled {
```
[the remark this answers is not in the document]
nikhitasharma: This particular disabled css can be nested under input-range and I will nest them. But the rest of the browser specific css need to be separate as Firefox gives compatibility issues  when seeing styles for chrome (webkit styles). Thats why those are written separately.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: These values are same in all themes, I don't see a need to have theme variables for these.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: These are infact given same values by UX. I agree with you though. I will double check with them on this and if we need to change them to be different.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: these were same for hover for some reason in the call-outs. I will double check with UX on this.

packages/terra-slider/src/Slider.jsx
nikhitasharma: It was agreed that the minimum labels and maximum labels would be optional ways to display customized or more readable labels. These are needed to achieve that.

packages/terra-slider/src/Slider.jsx
nikhitasharma: Do you have a suggested way of updating the slider progress here.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,110 @@
...
+    'slider',
+    theme.className,
+  ]);
+
+  const minLabel = minimumLabel || minimumValue;
+  const maxLabel = maximumLabel || maximumValue;
+
+  return (
+    <div className={sliderClassNames}>
+      <label className={cx('label')} id="sliderLabel">
```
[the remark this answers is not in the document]
nikhitasharma: Sorry, this was meant to be removed in the finalized code.

packages/terra-slider/src/Slider.jsx
nikhitasharma: It was confirmed in a UX/ accessibility meeting that these values need to be present for the slider. I will double check if these are required.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: I will reach out to UX and see if these can be updated in the callouts to use rems.

packages/terra-slider/src/Slider.jsx
nikhitasharma: Will update this.

packages/terra-slider/translations/en-US.json
nikhitasharma: There is no Jira logged for this currently, maybe because it was not a requirement for the passthrough. I agree that one should be logged and will follow-up on this and post here.

chrismichalewicz:
@nikhitasharma  I performed functional testing and a11y testing and have the below comments. Items bolded I think might be issues. Then I have some general questions/comments at the bottom. 

Functionally:
- Was able to move the slider thumb as expected with mouse and keyboard arrow keys, page up/down 
- Was unable to manipulate the disabled slider.
- **In the disabled slider example when I hovered over the thumb, my mouse pointer changed to the pointing hand cursor indicating I could click on it. Since it's disabled I wouldn't expect the mouse cursor to change. @eawww please confirm.** 

Accessibility:
Keyboard:
- Was able to navigate appropriately via keyboard.
Screen Reader: (tested with JAWS in Chrome)
- Communicated "default label 0 50 100" @eawww should we indicate for 0 and 100 those are the min/max values? 
- **If I hold down the arrow keys the slides increments continuously but when I lift the key it repeates the position I am on multiple times. This seems like an issue.** 
- **In the custom slider example it communicates "0% 20 100%" I would think we'd also want the thumb value to be a percentage in this case.**  

General Question:

- Is the custom slider example to illustrate the consumer can define the slider to be something other than just integers in this example percentages? Is there other functionality we want to call out. 
- If possible I'd update the Labels to be something not generic, currently it reads as "Default Label"
- We probably need another Jira for "Implementation Guide creation" to beef up the Doc site. I can get that logged.

packages/terra-slider/src/Slider.module.scss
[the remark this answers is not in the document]
cm9361: I am not clear on how that would impact the nesting approach.  It would give the exact same CSS.

packages/terra-slider/src/Slider.module.scss
cm9361: These are not the only themes.  We have the Redwood theme coming up as well.  There is also the cerner-clinical-theme.  We cannot limit to just designing for lowlight and fusion.

packages/terra-slider/src/Slider.module.scss
cm9361: I think some of the newer PRs used "background".  You definitely don't need both styles though.  We don't need to consult with UX on that.  Using background would allow other themes to use a gradient color, if desired.

packages/terra-slider/src/Slider.jsx
cm9361: Yes, use state variables.

packages/terra-slider/src/Slider.jsx
cm9361: We should be able to read the ARIA and APG documentation for this.  I would expect to get this for free for a semantic HTML element.

packages/terra-slider/src/Slider.module.scss
cm9361: Then you should pull out that one style instead of duplicating all of them.

packages/terra-slider/src/Slider.jsx
cm9361: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/slider_role

HTML's <input type="range"> implicitly has the role of slider. Do not use aria-valuemax or aria-valuemin attributes on <input type="range"> elements; use min and max instead. Otherwise, any global aria-* attributes and any other aria-* attributes applicable to the slider role.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  .input-range:disabled::-moz-range-track {
+    -moz-appearance: none; /* stylelint-disable-line property-no-vendor-prefix, plugin/no-unsupported-browser-features */
+    background: linear-gradient(to right, var(--terra-slider-track-disabled-selected-range, #007cc3) 0%, var(--terra-slider-track-disabled-selected-range, #007cc3)  50%, var(--terra-slider-track-disabled-unselected-range, #d3d4d5)  50%, var(--terra-slider-track-disabled-unselected-range, #d3d4d5) 100%);
+  }
+
+  .input-range:disabled::-moz-range-thumb {
+    -moz-appearance: none; /* stylelint-disable-line property-no-vendor-prefix, plugin/no-unsupported-browser-features */
+    background: #007cc3;
```
cm9361, line 210: There should be a theme variable for this color.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,110 @@
...
+import ThemeContext from 'terra-theme-context';
+import styles from './Slider.module.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /**
+  * Slider label value.
+  */
+  labelName: PropTypes.string.isRequired,
```
cm9361, line 13: Should this prop be named something like labelText?  Don't we have that same name for some other Terra components?

packages/terra-slider/src/Slider.jsx
nikhitasharma: @eawww @chrismichalewicz  Can you confirm that these aria props can be removed.

packages/terra-slider/src/Slider.jsx
nikhitasharma: I am sure I saw labelName somewhere and decided to use that. But I did see labelText as well in terra-core and other repos. So, I will update this.

packages/terra-slider/translations/en-US.json
nikhitasharma: Thank you @chrismichalewicz!

packages/terra-slider/src/Slider.jsx
nikhitasharma: I will do a quick change and see if it effects the screenreader behavior after removing them and post here.

cm9361: The documentation states that they should not be there.  We should not need additional confirmation.

packages/terra-slider/src/Slider.module.scss
eawww: @cm9361 this is probably to ensure the border color is overridden for the hover style. There's several ways this could be accomplished but this seems alright.

packages/terra-slider/src/Slider.jsx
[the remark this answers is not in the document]
eawww: Agreed. This is causing all of the sliders on the page to read with the value of the first label.

packages/terra-framework-docs/src/terra-dev-site/doc/slider/Examples.2/CustomSlider.jsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import Slider from 'terra-slider';
+
+const CustomSlider = () => (
+  <Slider minimumValue={0} maximumValue={100} labelName="Field Label" minimumLabel="0%" maximumLabel="100%" defaultValue={20} />
```
eawww, line 5: We should avoid using dummy values like "Label" in our examples. Something like "Zoom", "Size", "Intensity" or something more natural like that would be better.

packages/terra-framework-docs/src/terra-dev-site/doc/slider/Examples.2/DefaultSlider.jsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import Slider from 'terra-slider';
+
+const DefaultSlider = () => (
+  <Slider minimumValue={0} maximumValue={100} labelName="Label" defaultValue={50} />
```
eawww, line 5: At least one of these examples should show how to use the callback and value props to manage state.

eawww (commented):
- When disabled, the slider shouldn't get the pointer cursor. It should have the default cursor.
- There should be an example that doesn't display the min/max values.
- At least one example should show how the component is used with the `onChange` and `value` props to manage application state.

chrismichalewicz:
> * When disabled, the slider shouldn't get the pointer cursor. It should have the default cursor.
> * There should be an example that doesn't display the min/max values.
> * At least one example should show how the component is used with the `onChange` and `value` props to manage application state.

@eawww Maybe I misinterpretted the designs that Dee put together for the min/max values I was under the impression we should always be displaying those. If that is the case would we still need an example without them?   I could see an example with non-numeric values maybe XS->XL

packages/terra-slider/src/Slider.module.scss
cm9361: @eawww You only need one of the styles.  The background style can be set to a color.  We don't want to have two styles that can conflict each other with new code.

packages/terra-framework-docs/src/terra-dev-site/doc/slider/Examples.2/CustomSlider.jsx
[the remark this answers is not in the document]
nikhitasharma: Updated.

packages/terra-slider/src/Slider.jsx
nikhitasharma: Updated.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: Updated.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: Updated.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: Updated to rems.

packages/terra-slider/src/Slider.jsx
nikhitasharma: Updated.

packages/terra-slider/src/Slider.jsx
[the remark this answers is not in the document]
nikhitasharma: Removed.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: Updated to rems.

packages/terra-slider/src/Slider.jsx
nikhitasharma: Updated to labelText.

packages/terra-framework-docs/src/terra-dev-site/doc/slider/Examples.2/CustomSlider.jsx
nikhitasharma: Updated.

packages/terra-framework-docs/src/terra-dev-site/doc/slider/Examples.2/DefaultSlider.jsx
nikhitasharma: Added an example.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: I have removed the border color wherever they are same as the background color. I checked that there are no visual problems or thumb size changes after removing this.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: Updated these colors. They are different now.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,110 @@
...
+  }
+  const theme = React.useContext(ThemeContext);
+
+  const sliderClassNames = cx([
+    'slider',
+    theme.className,
+  ]);
+
+  const minLabel = minimumLabel || minimumValue;
+  const maxLabel = maximumLabel || maximumValue;
```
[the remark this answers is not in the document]
nikhitasharma: @eawww  Yes, they are controlled separately. The underlying values are still numeric. 
@cm9361  We need this to have the label customizations.

nikhitasharma:
> @nikhitasharma I performed functional testing and a11y testing and have the below comments. Items bolded I think might be issues. Then I have some general questions/comments at the bottom.
> 
> Functionally:
> 
> * Was able to move the slider thumb as expected with mouse and keyboard arrow keys, page up/down
> * Was unable to manipulate the disabled slider.
> * **In the disabled slider example when I hovered over the thumb, my mouse pointer changed to the pointing hand cursor indicating I could click on it. Since it's disabled I wouldn't expect the mouse cursor to change. @eawww please confirm.**
> 
> Accessibility: Keyboard:
> 
> * Was able to navigate appropriately via keyboard.
>   Screen Reader: (tested with JAWS in Chrome)
> * Communicated "default label 0 50 100" @eawww should we indicate for 0 and 100 those are the min/max values?
> * **If I hold down the arrow keys the slides increments continuously but when I lift the key it repeates the position I am on multiple times. This seems like an issue.**
> * **In the custom slider example it communicates "0% 20 100%" I would think we'd also want the thumb value to be a percentage in this case.**
> 
> General Question:
> 
> * Is the custom slider example to illustrate the consumer can define the slider to be something other than just integers in this example percentages? Is there other functionality we want to call out.
> * If possible I'd update the Labels to be something not generic, currently it reads as "Default Label"
> * We probably need another Jira for "Implementation Guide creation" to beef up the Doc site. I can get that logged.

- I have updated the disabled slider to not show the cursor pointer when hovered. 
- The repetitions should not happen after removing the aria attributes. Please have a look. 
- The percentage was added only to show the customization of min-max labels as that example was discussed in the meeting with UX. The underlying actual values are still integers. I am not sure there is a simple way of announcing the slider values as percentages based on custom input labels. 
- Updated labels to better values instead of "Default label"

nikhitasharma:
> You have a blue border around your slider when it has focus. The UX callouts do not specify this behavior. It seems like the border should not exists and the outline just around the slider thumb.

There is a consistent border color for all states, including focus. I have removed styling wherever overriding is unnecessary and if background color and border colors are the same.

eawww:
> In the disabled slider example when I hovered over the thumb, my mouse pointer changed to the pointing hand cursor indicating I could click on it. Since it's disabled I wouldn't expect the mouse cursor to change. Please confirm.

This is correct. Disabled slider should have the default cursor so people don't get the impression it's interactable.

packages/terra-slider/src/Slider.module.scss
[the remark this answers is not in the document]
nikhitasharma: Came across this when firefox styling were not working when nested with chrome webkit stylings: https://stackoverflow.com/questions/48648530/moz-styling-doesnt-work-with-additional-webkit-selector

nikhitasharma:
> We should probably create a test and example where the consumer is changing the slider value via the prop.

Added.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: Mentioned in callouts. The label will wrap to the next line in case of long label text. Like here:
https://github.com/cerner/terra-framework/pull/1822/files#diff-20e91930c1eda6dd71cdd1e5ab6f0cf0631a00ffe4ea4c2dc02f2a9c06eaaf0d

image

eawww:
I chatted with @scottwilmarth about some ideas to improve the screen reader experience and here's what we came up with:

What we should do is apply aria-hidden to all the visible labels including the min and max because it's all just kind of noise since we have the aria-label on the slider itself.

Then below the slider have visually hidden text that says something like

"Adjust slider to select a value between [min] ([min value]) and [max] ([max value])"

for example "Adjust slider with arrow keys to select a value between cold (0) and hot (100)" then have an aria-describedby on the slider with the id of that text.

If there are no custom labels, it can just use the min and max numbers: "Adjust slider with arrow keys to select a value between 0 and 20"

packages/terra-slider/README.md
```diff
@@ -0,0 +1,38 @@
+# Terra Slider
+
+[![NPM version](https://badgen.net/npm/v/terra-slider)](https://www.npmjs.com/package/terra-slider)
+[![Build Status](https://badgen.net/travis/cerner/terra-framework)](https://travis-ci.com/cerner/terra-framework)
+
+Terra Slider component provides users a way to select a value from a given range by sliding a thumb along a bar to track the value.
```
sycombs, line 6: small nit on wording:

```suggestion
Terra Slider provides a way to select a value from a given range by sliding a thumb along a bar to track the value.
```

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,124 @@
...
+import ThemeContext from 'terra-theme-context';
+import styles from './Slider.module.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /**
+  * Slider label value.
+  */
+  labelText: PropTypes.string.isRequired,
```
sycombs, line 13: Should this just be called `label` since "text" would be implied?

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,124 @@
...
+  */
+  onChange: PropTypes.func,
+};
+
+const defaultProps = {
+  isDisabled: false,
+  onChange: null,
+};
+
+class Slider extends React.Component {
```
sycombs, line 56: Is there a reason this is a class component? I was under the impression that any new Terra components would be functional components.

sycombs (changes requested):

packages/terra-framework-docs/src/terra-dev-site/doc/slider/Examples.2/SliderWithOnChangeCallback.jsx
```diff
@@ -0,0 +1,26 @@
...
+    <div>
+      <Slider
+        minimumValue={0}
+        maximumValue={100}
+        labelText="Zoom"
+        value={20}
+        onChange={handleChange}
+      />
+      <p>{message}</p>
+    </div>
```
sycombs, line 22: nit: Can we use a React Fragment here instead of a div?

```suggestion
    <>
      <Slider
        minimumValue={0}
        maximumValue={100}
        labelText="Zoom"
        value={20}
        onChange={handleChange}
      />
      <p>{message}</p>
    </>
```

nikhitasharma:
> I chatted with @scottwilmarth about some ideas to improve the screen reader experience and here's what we came up with:
> 
> What we should do is apply aria-hidden to all the visible labels including the min and max because it's all just kind of noise since we have the aria-label on the slider itself.
> 
> Then below the slider have visually hidden text that says something like
> 
> "Adjust slider to select a value between [min] ([min value]) and [max] ([max value])"
> 
> for example "Adjust slider with arrow keys to select a value between cold (0) and hot (100)" then have an aria-describedby on the slider with the id of that text.
> 
> If there are no custom labels, it can just use the min and max numbers: "Adjust slider with arrow keys to select a value between 0 and 20"

@eawww  I have added these, please have a look.

packages/terra-framework-docs/src/terra-dev-site/doc/slider/Examples.2/SliderWithOnChangeCallback.jsx
nikhitasharma: Updated.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,124 @@
...
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import ThemeContext from 'terra-theme-context';
+import styles from './Slider.module.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /**
+  * Slider label value.
```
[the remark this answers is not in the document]
nikhitasharma: Updated.

packages/terra-slider/src/Slider.jsx
nikhitasharma: Updated.

packages/terra-slider/README.md
nikhitasharma: Updated.

packages/terra-slider/src/Slider.jsx
cm9361: +1.  This should definitely be a functional component.  I believe this was a comment from the previous review as well.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,131 @@
...
+
+  /**
+  * Function to trigger when user changes the input value. Sends parameter {Event} event.
+  */
+  onChange: PropTypes.func,
+};
+
+const defaultProps = {
+  isDisabled: false,
+  onChange: null,
```
cm9361, line 55: There is no need to set the default prop to null.  An undefined value would be falsey as well.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,131 @@
...
+    const sliderClassNames = cx([
+      'slider',
+      theme.className,
+    ]);
+
+    const minLabel = minimumLabel || minimumValue;
+    const maxLabel = maximumLabel || maximumValue;
+    const VisuallyHiddenTextValue = (minimumLabel || maximumLabel)
+      ? `Adjust slider to select a value between ${minimumLabel} (${minimumValue}) and ${maximumLabel} (${maximumValue})`
+      : `Adjust slider to select a value between ${minimumValue} and ${maximumValue}`;
```
cm9361, line 105: You have user-visible text but no internationalization.

packages/terra-slider/src/Slider.module.scss
cm9361: You don't have the percentages set for the other components though.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,131 @@
...
+    const descriptionId = uuidv4();
+
+    return (
+      /* eslint-disable-next-line react/forbid-dom-props */
+      <div style={{ '--terra-slider-progress-status': `${this.state.value}%` }} className={sliderClassNames}>
```
cm9361, line 110: You should not need a CSS variable here.  You would set the appropriate style directly.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,131 @@
...
+        </span>
+        <VisuallyHiddenText id={descriptionId} text={VisuallyHiddenTextValue} />
+      </div>
+    );
+  }
+}
+
+Slider.propTypes = propTypes;
+Slider.defaultProps = defaultProps;
+Slider.contextType = ThemeContext;
```
cm9361, line 129: You should be leveraging useContext instead of contextType to get the ThemeContext once you create your functional component.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,131 @@
...
+          {label}
+        </span>
+        <span className={cx('slider-label', 'slider-min-label')} aria-hidden="true">
+          {minLabel}
+        </span>
+        <input className={cx('input-range')} type="range" ref={this.sliderRef} aria-label={label} aria-describedby={descriptionId} value={this.state.value} disabled={isDisabled} min={minimumValue} max={maximumValue} onChange={this.handleOnChange} />
+        <span className={cx('slider-label', 'slider-max-label')} aria-hidden="true">
+          {maxLabel}
+        </span>
+        <VisuallyHiddenText id={descriptionId} text={VisuallyHiddenTextValue} />
```
cm9361, line 121: We should not need an id attribute for the VisuallyHiddenText element.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,131 @@
...
+
+    const theme = this.context;
+    const sliderClassNames = cx([
+      'slider',
+      theme.className,
+    ]);
+
+    const minLabel = minimumLabel || minimumValue;
+    const maxLabel = maximumLabel || maximumValue;
+    const VisuallyHiddenTextValue = (minimumLabel || maximumLabel)
```
cm9361, line 103: A capital letter for the beginning of a variable name is reserved for React components.

packages/terra-slider/src/Slider.module.scss
cm9361: It doesn't appear that this was updated.

packages/terra-slider/src/Slider.module.scss
cm9361: There just seems to be too many variables here.  A consumer would just set the linear gradient.  They don't need all of these variables to do that.  It should be done with one variable.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,203 @@
...
+    box-shadow: var(--terra-slider-thumb-box-shadow, 0 0 0.286rem 0 rgba(0, 0, 0, 0.25));
+    cursor: pointer;
+    height: var(--terra-slider-thumb-height, 1.715rem);
+    margin-top: -0.57rem;
+    width: var(--terra-slider-thumb-width, 1.715rem);
+  }
+
+  .input-range:focus::-moz-range-thumb {
+    outline: var(--terra-slider-thumb-focus-outline, 0.143rem dashed #000);
+    outline-offset: 0.071rem;
```
cm9361, line 178: There should be a CSS variable for this value.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,203 @@
...
+
+  .input-range:disabled::-ms-track {
+    -webkit-appearance: none; /* stylelint-disable-line property-no-vendor-prefix, plugin/no-unsupported-browser-features */
+    background: linear-gradient(to right, var(--terra-slider-track-disabled-selected-range, #007cc3) 0%, var(--terra-slider-track-disabled-selected-range, #007cc3)  50%, var(--terra-slider-track-disabled-unselected-range, #d3d4d5)  50%, var(--terra-slider-track-disabled-unselected-range, #d3d4d5) 100%);
+  }
+
+  .input-range:disabled::-ms-thumb {
+    -webkit-appearance: none; /* stylelint-disable-line property-no-vendor-prefix, plugin/no-unsupported-browser-features */
+    background: var(--terra-slider-thumb-disabled-background, #007cc3);
+    cursor: default;
```
cm9361, line 151: I would expect the cursor to be on the input component not the thumb.  The disabled selector should also be on the input component instead of the thumb.  You can click on the track and cause behaviors to occur so the current design does not seem accurate in that aspect.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,203 @@
...
+    background-color: transparent;
+    margin-left: var(--terra-slider-input-range-margin-left, 0.5rem);
+    margin-right: var(--terra-slider-input-range-margin-right, 0.5rem);
+    min-height: 3.143rem;
+    min-width: 17.143rem;
+    width: 70%;
+
+    &:disabled {
+      opacity: 30%;
+    }
```
cm9361, line 50: All of these values should have CSS variables.  We have more themes than called out in this component so we cannot base the logic off that.  We must account for the Cerner clinical theme and upcoming Redwood theme.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,203 @@
...
+
+  /* For Firefox */
+  .input-range::-moz-range-track {
+    -moz-appearance: none; /* stylelint-disable-line property-no-vendor-prefix, plugin/no-unsupported-browser-features */
+    background: linear-gradient(to right, var(--terra-slider-track-selected-range, #007cc3) 0%, var(--terra-slider-track-selected-range, #007cc3) var(--terra-slider-progress-status, 50%), var(--terra-slider-track-unselected-range, #d3d4d5) var(--terra-slider-progress-status, 50%), var(--terra-slider-track-unselected-range, #d3d4d5) 100%);
+    border-radius: var(--terra-slider-track-border-radius, 0.214rem);
+    height: var(--terra-slider-track-height, 0.43rem);
+  }
+
+  .input-range::-moz-range-thumb {
```
cm9361, line 162: Aren't you able to combine some of these selectors and their corresponding styles so that there is not so much duplication?

packages/terra-slider/translations/en-US.json
cm9361: Where would we be using this translation?

cm9361:
There is still a blue border around the slider when using tab to focus the control.

packages/terra-slider/tests/wdio/__snapshots__/reference/terra-default-theme/en/chrome_large/slider-spec/keyboard-focus.png
cm9361, line 1: This screen shot does not seem to match the UX callouts.

cm9361:
> > You have a blue border around your slider when it has focus. The UX callouts do not specify this behavior. It seems like the border should not exists and the outline just around the slider thumb.
> 
> There is a consistent border color for all states, including focus. I have removed styling wherever overriding is unnecessary and if background color and border colors are the same.

I think it might be consistent but also incorrect.  You can update the CSS styles so that this border does not exists.  It does not seem to follow our design patterns.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,203 @@
...
+    white-space: initial;
+  }
+
+  .slider-min-label {
+    margin-right: var(--terra-slider-min-label-margin-right, 0.5rem);
+  }
+
+  .slider-max-label {
+    margin-left: var(--terra-slider-max-label-margin-left, 0.5rem);
+  }
```
cm9361, line 37: Instead of creating these two classes and styles, why not just add the margin or padding to the slider element to the left and right?  That way you only have to apply the style to one element.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,131 @@
...
+        <span className={cx('label')} aria-hidden="true">
+          {label}
+        </span>
+        <span className={cx('slider-label', 'slider-min-label')} aria-hidden="true">
+          {minLabel}
+        </span>
+        <input className={cx('input-range')} type="range" ref={this.sliderRef} aria-label={label} aria-describedby={descriptionId} value={this.state.value} disabled={isDisabled} min={minimumValue} max={maximumValue} onChange={this.handleOnChange} />
+        <span className={cx('slider-label', 'slider-max-label')} aria-hidden="true">
+          {maxLabel}
+        </span>
```
cm9361, line 120: Wouldn't this be a good place to use the Terra Arrange component?  Could it help simplify your styles?

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,131 @@
...
+      return;
+    }
+    // eslint-disable-next-line react/no-did-update-set-state
+    this.setState({
+      value: this.props.value,
+    });
+  }
+
+  handleOnChange(event) {
+    this.setState({ value: event.currentTarget.value });
```
cm9361, line 79: If the value has already been changed in the component, why do we need to set state?

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,131 @@
...
+    return (
+      /* eslint-disable-next-line react/forbid-dom-props */
+      <div style={{ '--terra-slider-progress-status': `${this.state.value}%` }} className={sliderClassNames}>
+        <span className={cx('label')} aria-hidden="true">
+          {label}
+        </span>
+        <span className={cx('slider-label', 'slider-min-label')} aria-hidden="true">
+          {minLabel}
+        </span>
+        <input className={cx('input-range')} type="range" ref={this.sliderRef} aria-label={label} aria-describedby={descriptionId} value={this.state.value} disabled={isDisabled} min={minimumValue} max={maximumValue} onChange={this.handleOnChange} />
```
cm9361, line 117: The value should not be a state variable.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,203 @@
...
+
+  .input-range {
+    -webkit-appearance: none; /* stylelint-disable-line property-no-vendor-prefix, plugin/no-unsupported-browser-features */
+    background-color: transparent;
+    margin-left: var(--terra-slider-input-range-margin-left, 0.5rem);
+    margin-right: var(--terra-slider-input-range-margin-right, 0.5rem);
+    min-height: 3.143rem;
+    min-width: 17.143rem;
```
cm9361, line 45: Do we really want the minimum width to be so large?  Are we considering mobile devices?  It seems like the consumer should be able to control the width by the size of the parent container.

packages/terra-slider/package.json
```diff
@@ -0,0 +1,67 @@
...
+  },
+  "homepage": "https://github.com/cerner/terra-framework#readme",
+  "peerDependencies": {
+    "react": "^16.8.5",
+    "react-dom": "^16.8.5",
+    "react-intl": "^2.8.0 || 3 || 4 || 5"
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "keycode-js": "^3.1.0",
```
cm9361, line 32: It does not look like this dependency is being used.

packages/terra-slider/src/Slider.jsx
cm9361, line 121: Is this element actually required if we use an aria label?

packages/terra-slider/src/Slider.jsx
nikhitasharma: UXPLATFORM-9721 logged to take care of internationalization.

eawww (approved):
Perfecto!

packages/terra-slider/src/Slider.jsx
sycombs: The text should still be internationalized and these strings should be added in the en.json and en-US.json files. That Jira is for updating the translations files for the other locales.

packages/terra-slider/src/Slider.jsx
nikhitasharma: Updated.

packages/terra-slider/src/Slider.jsx
nikhitasharma: Discussed this offline, the id attribute is required when using aria-describedby.

packages/terra-slider/src/Slider.jsx
nikhitasharma: Updated this.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: Added

packages/terra-slider/src/Slider.module.scss
nikhitasharma: Updated cursor to be on input component.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: Added

packages/terra-slider/src/Slider.module.scss
nikhitasharma: Logged UXPLATFORM-9778 for future css optimizations.

packages/terra-slider/tests/wdio/__snapshots__/reference/terra-default-theme/en/chrome_large/slider-spec/keyboard-focus.png
nikhitasharma: Updated to remove focus outline.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: Discussed offline, and concluded that its best to keep it this way.

packages/terra-slider/src/Slider.jsx
nikhitasharma: Logged UXPLATFORM-9778 for future css optimizations.

packages/terra-slider/src/Slider.jsx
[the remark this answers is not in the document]
nikhitasharma: aria-describedby seems to be required based on UX inputs.

packages/terra-slider/package.json
nikhitasharma: removed

packages/terra-slider/src/Slider.jsx
nikhitasharma: Discussed offline and this is a required part of slider progress logic.

packages/terra-slider/src/Slider.jsx
nikhitasharma: My bad, I thought these updates would take place as part of that Jira. I will get this updated soon.

packages/terra-slider/src/Slider.jsx
nikhitasharma: Still working on converting this to a functional component. Will get this updated soon.

nikhitasharma:
> > > You have a blue border around your slider when it has focus. The UX callouts do not specify this behavior. It seems like the border should not exists and the outline just around the slider thumb.
> > 
> > 
> > There is a consistent border color for all states, including focus. I have removed styling wherever overriding is unnecessary and if background color and border colors are the same.
> 
> I think it might be consistent but also incorrect. You can update the CSS styles so that this border does not exists. It does not seem to follow our design patterns.

You are right. I initially thought you were talking about the thumb border but I now understand. I have updated this.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: Discussed this offline and explained the progress status updates for slider css. Concluded that this cannot be a single theme variable because of dynamic updates and will keep this as is. Any further css optimizations will be investigated in UXPLATFORM-9778.

chrismichalewicz:
Looks good, great work! @nikhitasharma

packages/terra-slider/src/Slider.jsx
nikhitasharma: Updated

packages/terra-slider/src/Slider.jsx
nikhitasharma: Added.

packages/terra-slider/src/Slider.jsx
nikhitasharma: Updated.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: Requested @eawww  to double check on this. But I am guessing this is what they need on callouts.

nikhitasharma: Looks like the concern was around the width being too small and the thumb movement being minimal those cases. That's why an optimal min-width is suggested here.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,122 @@
...
+  /**
+  * @private
+  * The intl object containing translations. This is retrieved from the context automatically by injectIntl.
+  */
+  intl: PropTypes.shape({ formatMessage: PropTypes.func, locale: PropTypes.string }),
+};
+
+const defaultProps = {
+  isDisabled: false,
+  onChange: undefined,
```
cm9361, line 64: You should not need this as a default value.  If the prop is not passed, the onChange function will be undefined.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,122 @@
...
+  /**
+  * Function to trigger when user changes the input value. Sends parameter {Event} event.
+  */
+  onChange: PropTypes.func,
+
+  /**
+  * @private
+  * The intl object containing translations. This is retrieved from the context automatically by injectIntl.
+  */
+  intl: PropTypes.shape({ formatMessage: PropTypes.func, locale: PropTypes.string }),
```
cm9361, line 59: I think we usually use this format for intl
intl: PropTypes.shape({ formatMessage: PropTypes.func }).isRequired,

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,122 @@
...
+  const sliderRef = useRef();
+  const minLabel = minimumLabel || minimumValue;
+  const maxLabel = maximumLabel || maximumValue;
+  const visuallyHiddenTextValue = (minimumLabel && maximumLabel)
+    ? intl.formatMessage({ id: 'Terra.slider.ariaDescribedByTextWithLabels' }, {
+      minimumLabel, maximumLabel, minimumValue, maximumValue,
+    })
+    : intl.formatMessage({ id: 'Terra.slider.ariaDescribedByTextWithoutLabels' }, { minimumValue, maximumValue });
+  const descriptionId = uuidv4();
+  const [value, setValue] = useState(props.value);
```
cm9361, line 92: This should not be a state variable.  We want to consumers to be able to change the value as necessary.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,122 @@
...
+
+const Slider = props => {
+  const {
+    isDisabled,
+    intl,
+    minimumValue,
+    maximumValue,
+    label,
+    minimumLabel,
+    maximumLabel,
```
cm9361, line 75: onChange should be part of this deconstruction.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,122 @@
...
+      minimumLabel, maximumLabel, minimumValue, maximumValue,
+    })
+    : intl.formatMessage({ id: 'Terra.slider.ariaDescribedByTextWithoutLabels' }, { minimumValue, maximumValue });
+  const descriptionId = uuidv4();
+  const [value, setValue] = useState(props.value);
+
+  const handleOnChange = (event) => {
+    setValue(event.currentTarget.value);
+    if (props.onChange) {
+      props.onChange(event);
```
cm9361, line 97: I wouldn't expect us to return the entire event to our consumers.  We should be returning specific information.  In this case, we probably want to return the value.  This could be an object that consumers use to obtain the value and update their state.

packages/terra-slider/src/Slider.jsx
```diff
@@ -0,0 +1,122 @@
...
+    theme.className,
+  ]);
+  const sliderRef = useRef();
+  const minLabel = minimumLabel || minimumValue;
+  const maxLabel = maximumLabel || maximumValue;
+  const visuallyHiddenTextValue = (minimumLabel && maximumLabel)
+    ? intl.formatMessage({ id: 'Terra.slider.ariaDescribedByTextWithLabels' }, {
+      minimumLabel, maximumLabel, minimumValue, maximumValue,
+    })
+    : intl.formatMessage({ id: 'Terra.slider.ariaDescribedByTextWithoutLabels' }, { minimumValue, maximumValue });
```
cm9361, line 90: Wouldn't the calculation of this message need to be in a useEffect?  If consumers update the value, minimum value, maximum value, etc, I would think the screen reader should announce that.

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,203 @@
...
+    min-height: var(--terra-slider-input-range-min-height, 3.143rem);
+    min-width: var(--terra-slider-input-range-min-width, 17.143rem);
+    width: var(--terra-slider-input-range-width, 70%);
+
+    &:disabled {
+      cursor: default;
+      opacity: var(--terra-slider-input-range-disabled-opacity, 30%);
+    }
+    
+    &:focus {
```
cm9361, line 54: Wasn't this actually an issue for the :focus-visible state?

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,203 @@
...
+  .slider-min-label {
+    margin-right: var(--terra-slider-min-label-margin-right, 0.5rem);
+  }
+
+  .slider-max-label {
+    margin-left: var(--terra-slider-max-label-margin-left, 0.5rem);
+  }
+
+  .input-range {
+    -webkit-appearance: none; /* stylelint-disable-line property-no-vendor-prefix, plugin/no-unsupported-browser-features */
```
cm9361, line 40: I see us disabling linter warnings.  Why do we need to change the webkit appearance for the input?

packages/terra-slider/src/Slider.module.scss
```diff
@@ -0,0 +1,203 @@
...
+  .input-range::-moz-range-thumb {
+    -moz-appearance: none; /* stylelint-disable-line property-no-vendor-prefix, plugin/no-unsupported-browser-features */
+    background: var(--terra-slider-thumb-background, #007cc3);
+    border-color: var(--terra-slider-thumb-border-color, #0065a3);
+    border-radius: var(--terra-slider-thumb-border-radius, 1.715rem);
+    border-style: solid;
+    border-width: var(--terra-slider-thumb-border-width, 0.071rem);
+    box-shadow: var(--terra-slider-thumb-box-shadow, 0 0 0.286rem 0 rgba(0, 0, 0, 0.25));
+    height: var(--terra-slider-thumb-height, 1.715rem);
+    margin-top: -0.57rem;
```
cm9361, line 173: Looks like you missed the theme variable here.

cm9361:
We should have an example and/or test that covers the consumer change the value of the Slider component via its value prop.

packages/terra-slider/src/Slider.jsx
nikhitasharma: Done

packages/terra-slider/src/Slider.jsx
nikhitasharma: Updated

packages/terra-slider/src/Slider.jsx
nikhitasharma: Added

packages/terra-slider/src/Slider.jsx
nikhitasharma: I see the norm of passing the entire event in multiple terra components. But I changed this to make it more efficient.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: This works as well.

packages/terra-slider/src/Slider.jsx
nikhitasharma: No, because the values used in calculation would remain the same when user interacts with slider.

packages/terra-slider/src/Slider.module.scss
nikhitasharma: To remove the default webkit styling and add in our custom styling.

packages/terra-slider/src/Slider.jsx
nikhitasharma: Discussed offline and concluded that the state is required to manage sliders internal updates, but a useEffect was added to update state when props change externally.

nikhitasharma:
> We should have an example and/or test that covers the consumer change the value of the Slider component via its value prop.

Added.

cm9361 (approved):

PR #1825, opened 2023-10-04
[terra-file-path] File Path Implementation

artpark:
### Summary

**What was changed:**
Initial implementation of the file path component 🥳

Collapsible Menu View has 1 new private prop for setting the menu icon's label that will be read by screen readers
Collapsible Menu View Item has 1 new private prop that renders items as just texts rather than buttons

Since all changes to Collapsible Menu View should not affect consumers, no CHANGELOG update is needed for it.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9551 

---

@cerner/terra


artpark:
> Fails
> * terra-collapsible-menu-view
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [de8a303](https://github.com/cerner/terra-framework/commit/de8a3039c48bda44084f7c010a773dcbd2db8224)

No changes that would affect consumers have been made, so no changelogs necessary for terra-collapsible-menu-view

packages/terra-file-path/src/FilePath.jsx
```diff
@@ -16,9 +26,43 @@ const propTypes = {
...
+      onClick={item.onClick}
+      onBlur={item.onBlur}
+      onFocus={item.onFocus}
+      onKeyDown={item.onKeyDown}
+      onKeyUp={item.onKeyUp}
+      onMouseDown={item.onMouseDown}
+      variant={item.variant}
+    />,
+    <CollapsibleMenuView.Divider useChevronRightIcon key={`divider-${item.key}`} />,
+  ]).flat();
```
artpark, line 52: flat() has to be used here instead of a React fragment due to Collapsible Menu View not spreading out the children.

chrismichalewicz:
@artpark  Functional review looked good. Updated the Functional Review label.

For a11y review: 
Keyboard -
Able to tab through each link
Selecting Enter opened each link to oracle.com
- Unable to select it via Spacebar but in referencing the Hyperlink button exampels that appears to be expected. 
- with the On Click File Path example both Spacebar and Enter works for selecting an item.
I am not sure if this is expected or not?  @artpark @scottwilmarth @eawww  would appreciate some input on that. 

Screen reader tested Jaws + Edge
- Each item communicates as a link but not how the user should interact with it.  @scottwilmarth @eawww  Would that be expected for a link?  I know it's technically a button but as it is acting like a link I am calling it that.   I did verify that the Hyperlinkbutton examples also do this so not necessarily an issue with the File Patch component. 

@jeremyfuksa  tagging you on this as Eric is out of office this week, is there someone we can have cover his UX reviews for the team while he is gone? I didn't want to just default to Scott know how busy he is.

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
```diff
@@ -53,6 +53,12 @@ const propTypes = {
    */
   isReversedChildrenOrder: PropTypes.bool,
 
+  /**
+   * @private
+   * The text for the menu icon. This is read by screenreaders.
+   */
+  iconText: PropTypes.string,
```
cm9361, line 60: This looks like the text for the More button based on the code.  Why is the prop private?  Do we not expect it to be set by consumers?

artpark: We don't expect it to be set by consumers yes, it should only read out "Parent Folders" when I discussed it with @eawww

packages/terra-file-path/src/FilePath.jsx
cm9361: Could you elaborate on this more.  This also sounds like it would be good as a comment in the code.

packages/terra-file-path/src/FilePath.jsx
```diff
@@ -1,12 +1,59 @@
...
+     */
+    onFocus: PropTypes.func,
+    /**
+     * Callback function triggered when key is pressed.
+     */
+    onKeyDown: PropTypes.func,
+    /**
+     * Callback function triggered when key is released.
+     */
+    onKeyUp: PropTypes.func,
```
cm9361, line 46: What action are we expecting consumers to take on a key up event?

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
```diff
@@ -110,6 +110,7 @@ const CollapsibleMenuViewHyperlink = ({
   return (
     <div className={cx(['face-up-item'])}>
       <Hyperlink
+        {...customProps}
```
cm9361, line 113: Why did we have to add custom props to this component?

packages/terra-framework-docs/src/terra-dev-site/test/file-path/OnClickFilePath.test.jsx
cm9361, line 1: Your tests are not covering your other events such as keydown, keyup, focus, and blur.  I don't think we need them, but if we do, it should be covered by testing.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
artpark: `key` was not being passed in when using hyperlink from the filepath level. So this was added for that

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItem.jsx
```diff
@@ -64,6 +64,12 @@ const propTypes = {
    * 160, 240, 320, 640, 960, 1280, 1760, or auto
    */
   menuWidth: PropTypes.oneOf(['160', '240', '320', '640', '960', '1280', '1760', 'auto']),
+
+  /**
+   * @private
+   * Renders the item as just text.
+   */
+  isTextOnly: PropTypes.bool,
```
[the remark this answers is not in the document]
artpark: Yes we need this, items rendered by CollapsibleMenuViewItem are buttons by default, there was no option to render them as just text. So this isTextOnly is a non-passive way to render just text. It's also private cause I didn't think collapsible menu view needed to render just text in any scenarios

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
cm9361: Can we not just update to pass the key directly?

packages/terra-framework-docs/src/terra-dev-site/doc/file-path/Examples.2/OnClickFilePath.2.doc.mdx
```diff
@@ -0,0 +1,3 @@
+import OnClickFilePath from './OnClickFilePath?dev-site-example';
+
+<DefaultFilePath />
```
smason0, line 3: This should be OnClickFilePath

cm9361:
> @cm9361 This is a reply to all the comments about the different props added like onMouseDown, onBlur, etc.
> 
> I added those because the `terra-hyperlink` has all those fields already https://engineering.cerner.com/terra-ui/components/cerner-terra-core-docs/hyperlink/about#hyperlink-props. And since this is just a menu of hyperlinks, I figured it would be the most flexible if consumers would just treat each item as a `terra-hyperlink` alongside all of its existing props

I think the question we have to ask is what is the use in your component.  Even if they are exposed in other components, it doesn't mean this component needs to expose them.

packages/terra-file-path/src/FilePath.jsx
```diff
@@ -1,12 +1,59 @@
...
+    /**
+     * @private Callback function not intended for use with this API, but if set pass it through to the element regardless.
+     */
+    onMouseDown: PropTypes.func,
+    /**
+     * Sets the hyperlink variant. One of `default`, `external`, `image`, `video`, `audio`, `document`.
+     */
+    variant: PropTypes.oneOf(['default', 'external', 'image', 'video', 'audio', 'document']),
```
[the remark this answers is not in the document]
smason0: +1. Even though terra-hyperlink supports these variants, I'm not sure it makes sense for file path to, unless it's something we specifically need to support. If it is needed, examples of each should be provided.

artpark:
@cm9361 That's a fair point yea, this component doesn't need all of those props. The two necessary props are `href` and `onClick`. I can remove the rest then

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItem.jsx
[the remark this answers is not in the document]
cm9361: Shouldn't the FilePath options be buttons as well.  Aren't they clickable?

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
artpark: I had it be a prop so that file path can have screen readers read "Parent Folders" while collapsible menu view remains the same and reads "More".

I can change it to menuIconText if that would make it clearer

cm9361: Maybe menuButtonText?  Or is it really just a clickable icon?

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewHyperlink.jsx
artpark: We could yes, but other custom props such as `id` might be needed in certain situations too. Plus this has been the pattern for all our terra components as well, I just forgot to add it with the last PR

packages/terra-framework-docs/src/terra-dev-site/doc/file-path/Examples.2/OnClickFilePath.2.doc.mdx
artpark: Thanks, I missed this

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
artpark: Just a clickable icon

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItem.jsx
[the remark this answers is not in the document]
artpark: They are hyperlinks, if onClick is provided they will be rendered as buttons, otherwise rendered as anchor tags

packages/terra-file-path/src/FilePath.jsx
artpark: Added a comment in the code itself. Here's what I added:
.flat() is used here instead of a react fragment on purpose. The children that get passed down to CollapsibleMenuView will not work with react fragments. They will be passed down without being spread. This results in pairs of hyperlinks and dividers being grouped together as one, breaking how width calculation is handled for resizing.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItem.jsx
```diff
@@ -151,6 +159,14 @@ class CollapsibleMenuViewItem extends React.Component {
           {subMenuItems}
         </Menu>
       );
+    } else if (isTextOnly) {
+      item = (
+        <div className={cx('face-up-item')}>
+          <p>
```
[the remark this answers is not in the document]
artpark: Removed

scottwilmarth:
@artpark and I chatted after I reviewed the code and experience. While there are some improvements we can make to ensure hierarchical relationships are understood, this implementation is a good start to providing the basic structure. A few key things I would like to see addressed in future work (added into backlog) are:

1. An accessible name for the group of controls (contextual menu and string of links) to ensure they are understood as a group.
2. More context is injected to convey the relationship of all "links" within the path regardless of how content is visually presented.

More research and input from people with disabilities will be needed to provide more clarification of the context needed.

cm9361 (approved):

smason0 (approved):

PR #1828, opened 2023-10-05
[collapsible-menu-view]- Added `aria-label` prop to subcomponents to announce group information.

PK106552:
### Summary

**What was changed:**
Added `aria-label` prop to Collapsible-menu-view.

**Why it was changed:**
Collapsible-menu-view subcomponents already setting the `role=group` on the container. However, the screen readers are not announcing this because there is no `aria-label `containing an accessible name for the group. To resolve this, both of those components will need to be updated with an `ariaLabel` prop that allows consumers to provide an accessible group name

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9625 

---

@cerner/terra

Screenshot 2023-10-05 at 4 08 34 PM

Screenshot 2023-10-05 at 4 12 15 PM


saket2403 (approved):

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItemGroup.jsx
```diff
@@ -90,7 +90,7 @@ class CollapsibleMenuViewItemGroup extends React.Component {
         });
         return (
           <li role="none">
-            <div {...customProps} role="group">
+            <div {...customProps} aria-label={ariaLabel} role="group">
```
supreethmr, line 93: ariaLabel is not being added as first class prop here. Then why do we need explicit prop when Consumer can still pass in aria-label through custom props.

packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItemGroup.jsx
```diff
@@ -107,7 +107,7 @@ class CollapsibleMenuViewItemGroup extends React.Component {
       }
       return (
         <li role="none">
-          <div {...customProps} role="group">
+          <div {...customProps} aria-label={ariaLabel} role="group">
```
supreethmr, line 110: Same here the example updated with aria-label with this PR will work without this change. 
`aria-label={ariaLabel}`

PR #1830, opened 2023-10-10
[terra-data-grid] Flowsheet Data Grid - No Result Cell

smason0:
### Summary

**What was changed:**
Added support for "No result" cells, displayed as "--" in the Flowsheet Data Grid.

**Why it was changed:**
This adds the proper visual indication for when a (non-header) Flowsheet cell has no results, and allows it to be read as "No result" by AT.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9405 

---

@cerner/terra


packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -112,6 +112,11 @@ const propTypes = {
    * rendered to allow for row selection to occur.
    */
   hasSelectableRows: PropTypes.bool,
+
+  /**
+   * Boolean indicating that cells without any content will be displayed as a dash, meaning "No Result."
+   */
+  hasNoResultCells: PropTypes.bool,
```
trandrew1023, line 119: Do we need to specify the default prop here as well?

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -103,6 +103,7 @@ function FlowsheetDataGrid(props) {
         columnHeaderHeight={columnHeaderHeight}
         onCellSelect={onCellSelect}
         onClearSelectedCells={onClearSelectedCells}
+        hasNoResultCells
```
cm9361, line 106: The Flowsheet should be creating the no result cells.  There should not be any changes required at another level.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -86,19 +95,42 @@ function FlowsheetDataGrid(props) {
...
+              <span aria-hidden>{intl.formatMessage({ id: 'Terra.flowsheetDataGrid.no-result-display' })}</span>
+              <VisuallyHiddenText text={intl.formatMessage({ id: 'Terra.flowsheetDataGrid.no-result' })} />
+            </>
+          ),
+        } : cell),
+      ),
+    }),
+  );
+
+  // Make all columns not resizable.
```
smason0, line 119: These couple lines were just a refactor, nothing changed functionally

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
smason0: Done

packages/terra-data-grid/src/DataGrid.jsx
smason0: This prop has been removed as part of addressing Charles' comment

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -86,19 +95,42 @@ function FlowsheetDataGrid(props) {
...
+  const nonResizablePinnedColumns = columns.length ? [{ ...columns[0], isResizable: false }] : [];
+  const nonResizableOverflowColumns = columns.length > 1 ? columns.slice(1).map(column => ({ ...column, isResizable: false })) : [];
```
cm9361, line 121: Maybe I am not thinking properly here, but couldn't we just map all the columns the same way in one call and then separate as needed for the props?

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -86,19 +95,42 @@ function FlowsheetDataGrid(props) {
     rowHeight,
     onCellSelect,
     onClearSelectedCells,
+    intl,
   } = props;
 
+  // Replace each non-header cell that contains no content with a dash indicating "No results".
+  const parsedRows = rows.map(
```
cm9361, line 102: Maybe we could rename to something like "generateFlowsheetRows" or "getFlowsheetRows" or something similar?

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -86,19 +95,42 @@ function FlowsheetDataGrid(props) {
...
+    (row) => ({
+      ...row,
+      cells: row.cells.map(
+        (cell, index) => ((!cell.content && index !== 0) ? {
```
cm9361, line 106: Is the index really the "column"?  Are we excluding that column because it is the row header?  If so, it would probably be good to add a comment.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -86,19 +95,42 @@ function FlowsheetDataGrid(props) {
     rowHeight,
     onCellSelect,
     onClearSelectedCells,
+    intl,
   } = props;
 
+  // Replace each non-header cell that contains no content with a dash indicating "No results".
```
cm9361, line 101: I think this comment exposes too much of the method implementation.  Also, it doesn't fully describe the current implementation or its future state.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -86,19 +95,42 @@ function FlowsheetDataGrid(props) {
...
+      cells: row.cells.map(
+        (cell, index) => ((!cell.content && index !== 0) ? {
+          ...cell,
+          content: (
+            <>
+              <span aria-hidden>{intl.formatMessage({ id: 'Terra.flowsheetDataGrid.no-result-display' })}</span>
+              <VisuallyHiddenText text={intl.formatMessage({ id: 'Terra.flowsheetDataGrid.no-result' })} />
+            </>
+          ),
+        } : cell),
```
cm9361, line 114: This logic might benefit from the introduction of some variables and an if statement.  The code is logically correct.  It just seems like there is a decent amount of chaining.  For example, maybe there should be variables to store the cloned row and cell objects.  Then you could access the cells and content properties from those objects in your logic.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
smason0: Right, that makes more sense to do the operation all at once and then dividing it. Done.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
smason0: I went with 'flowsheetRows' since it is just the array itself.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
smason0: Added a comment explaining this

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
smason0: Removed, as the above comment explains it better anyway

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -86,19 +95,48 @@ function FlowsheetDataGrid(props) {
...
+
+  const flowsheetRows = useMemo(() => {
+    const noResultCellContent = (
+      <>
+        <span aria-hidden>{intl.formatMessage({ id: 'Terra.flowsheetDataGrid.no-result-display' })}</span>
+        <VisuallyHiddenText text={intl.formatMessage({ id: 'Terra.flowsheetDataGrid.no-result' })} />
+      </>
+    );
+
+    const newRows = rows;
```
cm9361, line 113: You still need to clone on this line and the new cells line to not impact the source prop.
const newRows = [...rows]
const newCells = [...cells]

smason0: Good catch, fixed

cm9361 (approved):

sdadn (approved):

chrismichalewicz:
Reviewed for a11y. Tested in Edge/JAWS Version 2022.2202.38

- When the cell with -- is selected Jaws communicates "No Result" 

This looks good. I updated Accessibility Review label
cc: @eawww

eawww:
Works great! Just needs something on the doc page explaining when this is automatically displayed and an example showing the functionality.

smason0:
> Works great! Just needs something on the doc page explaining when this is automatically displayed and an example showing the functionality.

Added an example containing "No result" cells, with a description

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -86,19 +95,48 @@ function FlowsheetDataGrid(props) {
...
+        <VisuallyHiddenText text={intl.formatMessage({ id: 'Terra.flowsheetDataGrid.no-result' })} />
+      </>
+    );
+
+    const newRows = [...rows];
+    newRows.forEach((row, rowIndex) => {
+      const newCells = [...row.cells];
+      newCells.forEach((cell, cellIndex) => {
+        // Cell has no content and is not a row header (first column), set content to "No result".
+        if (!cell.content && cellIndex !== 0) {
```
eawww, line 118: Wouldn't this also display cells with a 0 value as No Result? 

That, along with things like `NaN` and `undefined` seem like this could be dangerous. A better approach might be to check for explicitly defined values like `null`, `'--'`, and `''`

eawww (commented):
A couple nits but also caught one probably major concern.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
smason0: Good catch! A numeric 0 value does make this No Result. I can add explicit cases rather than a falsy check, assuming I know what exactly we want to check for. Are we not including `undefined` as a case to show No Result? Also, if the consumer explicitly passes a `'--'` do we want to treat it as No Result?

smason0: I added a check for specifically `null`, `''`, and `'--'`.

eawww (approved):
A few wording changes but otherwise looks great!

PR #1836, opened 2023-10-16
UXPLATFORM-9611: Enhance screen reader support to announce column headers are interactable.

KoushikKalli:
### Summary
Added additional screen reader support to announce that column headers are interactable upon selection by giving focus to the button control that allows column headers to be selected.

**What was changed:**
Focus of a column header cell is changed from the cell to the button control inside the cell that allows selection.

**Why it was changed:**
To ensure screen readers announce that column headers are interactable upon selection.

### Testing
This change was tested using:
- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews
- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

**This PR resolves:** UXPLATFORM-9611


sdadn (approved):

cm9361:
What is the plan to get these changes into the new Terra Table component?

packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -259,9 +266,10 @@ const ColumnHeaderCell = (props) => {
       aria-sort={sortIndicator}
       onMouseDown={onColumnSelect ? handleMouseDown : null}
       onKeyDown={(isSelectable || isResizable) ? handleKeyDown : null}
+      onFocus={handleFocus}
       style={{ width: `${width}px`, height: headerHeight, left: cellLeftEdge }} // eslint-disable-line react/forbid-dom-props
     >
-      <div className={cx('header-container')} role={displayName && 'button'}>
+      <div ref={columnHeaderCellButtonRef} tabIndex="-1" className={cx('header-container')} role={displayName && 'button'}>
```
cm9361, line 272: The role should only be a button if the column header is selectable.

packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -222,6 +223,12 @@ const ColumnHeaderCell = (props) => {
     }
   };
 
+  const handleFocus = () => {
+    if (isResizable && !isResizeHandleActive) {
+      columnHeaderCellButtonRef.current?.focus();
+    }
+  };
```
cm9361, line 230: Instead of adding a new focus event listener, I would expect that we could leverage the useEffect logic to make this work properly.

packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -180,7 +181,7 @@ const ColumnHeaderCell = (props) => {
 
   // Restore focus to column header after resize action is completed.
   const onResizeHandleMouseUp = useCallback(() => {
-    columnHeaderCellRef.current.focus();
+    columnHeaderCellButtonRef.current.focus();
```
cm9361, line 184: What happens if there is no button?

packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.jsx
cm9361, line 272: You most likely only want this button ref variable if the header is selectable.

packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.module.scss
```diff
@@ -23,6 +23,11 @@
       height: 100%;
       padding: var(--terra-data-grid-header-cell-padding, 5px);
       width: 100%;
+
+      &:focus {
+        outline: var(--terra-data-grid-cell-focus-outline, 3px dashed #000);
+        outline-offset: var(--terra-data-grid-cell-focus-outline-offset, -2px);
+      }
```
cm9361, line 30: Don't we already have this styling for the DataGrid as a whole?  I know that it existed for the WorklistDataGrid.  We should not have to duplicate the style, correct?

packages/terra-data-grid/tests/wdio/worklist-data-grid-spec.js
```diff
@@ -498,5 +498,51 @@ Terra.describeViewports('WorklistDataGrid', ['medium', 'large'], () => {
       expect($('tr.column-header-row').$('//th[1]').getCSSProperty('width').parsed.value).toBe(180);
       Terra.validates.element('column-resize-decrease-keyboard', options);
     });
+
+    it('focuses on the resize handle in the column header when navigating right from the column header', () => {
+      browser.url('/raw/tests/cerner-terra-framework-docs/data-grid/worklist-data-grid/default-worklist-data-grid');
+
+      browser.keys(['Tab', 'ArrowRight']); // Focus is on (0,0)
+      expect(browser.$('tr.column-header-row th:nth-child(1) div[draggable=true]').isFocused()).toBe(true);
```
cm9361, line 506: Should we verify by the resize handle class name instead of depending on a draggable attribute being present?

sdadn:
> What is the plan to get these changes into the new Terra Table component?

The plan is to merge these into #1813 for now and then integrate that PR into the Table component once that is ready. This way, there will be a single PR that needs to be refactored.

eawww (approved):

packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.jsx
KoushikKalli: Removed.

packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.module.scss
KoushikKalli: Removed

chrismichalewicz:
@KoushikKalli  @brithom Tested for accessibility. Looks good, great job! Updating the label to a11y reviewed. 
Tested with Edge and JAWS 2022.2202.38
The screen reader communicated appropriately on a sortable column header 'to activate press enter', upon pressing enter the sort updated appropriately

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -584,12 +585,19 @@ const DataGrid = injectIntl((props) => {
...
     }
 
     handleFocus.current = true;
   };
 
+  const onBlur = (event) => {
+    if (!event.currentTarget.contains(event.relatedTarget)) {
+      setGridHasFocus(false);
+    }
+  };
```
cm9361, line 599: We should not need to add new event handler logic to determine if the grid has focus.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -584,12 +585,19 @@ const DataGrid = injectIntl((props) => {
       // Not triggered when swapping focus between children
       if (handleFocus.current) {
         setFocusedRowCol(focusedRow, focusedCol, true);
+        setGridHasFocus(true);
```
cm9361, line 588: I would think that we could determine this information without needing to alter the focus event handler logic.

packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -159,6 +159,7 @@ const ColumnHeaderCell = (props) => {
 
   const columnContext = useContext(ColumnContext);
   const columnHeaderCellRef = useRef();
+  const columnHeaderCellButtonRef = useRef();
```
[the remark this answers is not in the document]
cm9361: I that name was used because it was the name of the component

packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -159,16 +159,24 @@ const ColumnHeaderCell = (props) => {
...
 
   const [isResizeHandleActive, setResizeHandleActive] = useState(false);
 
+  const columnHeaderFocusArea = useCallback(() => (columnHeaderCellButtonRef.current ? columnHeaderCellButtonRef.current : columnHeaderCellRef.current), []);
```
cm9361, line 166: Do we really need a useCallback here?  This should really just be a simple or statement to obtain this value, when needed.

packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -258,7 +266,10 @@ const ColumnHeaderCell = (props) => {
       onKeyDown={(isSelectable || isResizable) ? handleKeyDown : null}
       style={{ width: `${width}px`, height: headerHeight, left: cellLeftEdge }} // eslint-disable-line react/forbid-dom-props
     >
-      <div className={cx('header-container')} role={displayName && 'button'}>
+      <div
+        className={cx('header-container')}
+        {...isSelectable && displayName && { ref: columnHeaderCellButtonRef, tabIndex: '-1', role: 'button' }}
```
cm9361, line 271: This won't work properly for the Terra Table.  In that scenario, the tabIndex would be 0.  Also, should the cell tabIndex be updated when this one is set?  Would the cell ever have focus in that use case?

cm9361: We might be able to get away with it since the button is taking up most of the area.  Ideally, the cell itself would not get focus when there is a button.

packages/terra-data-grid/src/subcomponents/ColumnHeaderCell.jsx
cm9361: Why can't we just use a ref variable?

packages/terra-data-grid/src/DataGrid.jsx
cm9361: But the column header cell would lose focus itself.  We should not need a state update because the CSS is based on focus, right?

PR #1838, opened 2023-10-17
Release 10/17/23

sycombs:
The following packages will be released:

  - terra-collapsible-menu-view@6.82.0
  - terra-data-grid@0.8.1
  - terra-file-path@0.3.0
  - @cerner/terra-framework-docs@1.41.0
  - terra-slide-panel-manager@5.82.0
  - terra-table@4.37.0


packages/terra-collapsible-menu-view/CHANGELOG.md
sycombs, line 1: It looks like there were code changes in this component that are needed for `terra-file-path`, but because they are not consumer-facing, no changelog entry was added. However, these still should be released in order to be consumed by `terra-file-path`.

adoroshk (approved):

packages/terra-file-path/CHANGELOG.md
```diff
@@ -2,4 +2,6 @@
 
 ## Unreleased
 
+## 0.3.0 - (October 17, 2023)
+
```
sdadn, line 6: It's odd that the entries for 0.1 & 0.2 are missing but not sure if it's worth correcting.

sycombs: Yeah I wasn't sure why but the previous entries were removed here: https://github.com/cerner/terra-framework/pull/1825/files#diff-877f2e53487931f7411ca4d8207fdc643647e2c6f97f8d5b68b821c70add4302R5

packages/terra-table/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "terra-table",
   "main": "lib/index.js",
-  "version": "4.36.0",
+  "version": "4.37.0",
```
sdadn, line 4: This needs to be a major alpha release. So 5.0.0-alpha (or however the correct format is).

sycombs: Should be correct now [73f2111](https://github.com/cerner/terra-framework/pull/1838/commits/73f21115e3c5e98c7d9ea6f365fffa77ed4a87b4)

sdadn (approved):

packages/terra-table/package.json
sdadn: Looks good!

PR #1839, opened 2023-10-18
Zebra stripe table

cm9361:
### Summary

**What was changed:**
The Terra Table was updated so that consumers could specify whether zebra striping is implemented in the table.

**Why it was changed:**
The change was made as table implementations vary on the usage of zebra striping.  Moreover, the data grid components that leverage the table require zebra striping.

### Testing

This change was tested using:

- [X] WDIO
- [X] Jest
- [ ] No tests are needed

### Reviews

- [X] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9700 

---

@cerner/terra


packages/terra-framework-docs/src/terra-dev-site/doc/table/Examples/ZebraStripedTable.jsx
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
+import Table from 'terra-table';
+
+const gridDataJSON = {
```
kenk2, line 4: Non-blocking nit: This is table but here and other places using this data this is referencing grid.

packages/terra-table/src/subcomponents/Row.jsx
```diff
@@ -38,6 +38,11 @@ const propTypes = {
    */
   isSelected: PropTypes.bool,
 
+  /**
+   * Boolean indicating whether zebra striping is enabled
```
sdadn, line 42:
```suggestion
   * Boolean indicating whether zebra striping is enabled.
```

sdadn (approved):
Small nit but otherwise looks good!

packages/terra-table/src/subcomponents/Row.jsx
```diff
@@ -99,6 +105,7 @@ function Row(props) {
       className={cx('row', {
         selected: isSelected,
         selectable: hasRowSelection,
+        'striped-table-row': isTableStriped,
```
kenk2, line 108: I think we could name this `striped`? This also feels awkward compared to the other class names that we use here.

packages/terra-table/src/subcomponents/Row.jsx
```diff
@@ -81,6 +86,7 @@ function Row(props) {
     hasRowSelection,
     id,
     isSelected,
+    isTableStriped,
```
kenk2, line 89: nit: Like my other comment, could we just name this `isStriped`? The naming feels weird compared to the other variable names.

kenk2 (approved):
Only a few tiny comments, but otherwise LGTM!

packages/terra-table/src/subcomponents/Row.jsx
cm9361: This naming convention is similar.  The row may or may not be striped.  It is the parent table that has striping enabled.  We have CSS to control this at the row component level though.

packages/terra-table/src/subcomponents/Row.jsx
kenk2: I think in that case, we can leave this as-is as was the intention that the table is the component that is passing this down to the component.

packages/terra-table/src/Table.jsx
```diff
@@ -101,6 +101,11 @@ const propTypes = {
    * rendered to allow for row selection to occur.
    */
   hasSelectableRows: PropTypes.bool,
+
+  /**
+   * Boolean indicating whether or not the table should have zebra striping for rows.
```
eawww, line 106:
```suggestion
   * Boolean specifying whether or not the table should have zebra striping for rows.
```
I know we use "indicating" in some places, but "specifying" or "defining" is more accurate.

packages/terra-table/src/subcomponents/Row.jsx
```diff
@@ -38,6 +38,11 @@ const propTypes = {
    */
   isSelected: PropTypes.bool,
 
+  /**
+   * Boolean indicating whether zebra striping is enabled.
```
eawww, line 42:
```suggestion
   * Boolean specifying whether zebra striping is enabled.
```

eawww (approved):
Lil nit but overall looks good

PR #1843, opened 2023-10-18
[terra-table] Allow consumers to hide table column headers

cm9361:
### Summary

**What was changed:**
The Terra Table component was updated to allow consumers to control the visibility of column headers in the table.

**Why it was changed:**
The change was made to maintain passivity with existing implementations that requires the column headers to be hidden such as the Stella Timeline.

### Testing

This change was tested using:

- [X] WDIO
- [X] Jest
- [ ] No tests are needed

### Reviews

- [X] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details
N/A

**This PR resolves:**

UXPLATFORM-9638 

---

@cerner/terra


packages/terra-framework-docs/src/terra-dev-site/doc/table/Examples/TableWithoutHeaders.jsx
```diff
@@ -0,0 +1,84 @@
...
+    { id: 'Column-1', displayName: 'Location' },
+    { id: 'Column-2', displayName: 'Illness Severity' },
+    { id: 'Column-3', displayName: 'Visit' },
+    { id: 'Column-4', displayName: 'Allergy' },
+    { id: 'Column-5', displayName: 'Primary Contact' },
+    { id: 'Column-6', displayName: 'Generic Order Counts' },
+    { id: 'Column-7', displayName: 'Patient Age' },
+    { id: 'Column-8', displayName: 'Medication History' },
+    { id: 'Column-9', displayName: 'My Relationship' },
+    { id: 'Column-10', displayName: 'Not Selectable', isSelectable: false },
```
adoroshk, line 16: Is the `isSelectable` here by accident, or does it have some effect?

packages/terra-table/src/Table.jsx
```diff
@@ -291,19 +299,28 @@ function Table(props) {
...
+        className={cx('table', theme.className, { headerless: !hasColumnHeaders })}
         {...(activeIndex != null && { onMouseUp, onMouseMove, onMouseLeave: onMouseUp })}
       >
         <ColumnContext.Provider
           value={columnContextValue}
         >
+          <colgroup>
+            {tableColumns.map((column) => (
+              // eslint-disable-next-line react/forbid-dom-props
+              <col style={{ width: `${column.width}px` }} />
```
adoroshk, line 311: I guess we need to add a `key` prop:
Screenshot 2023-10-19 at 10 57 51 AM

packages/terra-framework-docs/src/terra-dev-site/doc/table/Examples/TableWithoutHeaders.jsx
cm9361: We are setting the value to false.  This will make it not selectable.

adoroshk (approved):

kenk2 (approved):

chrismichalewicz:
@cm9361  Functionally it looks good, confirmed no headers are read in JAWS as well.

eawww (commented):
This wasn't clear in the jira but for the Stella Timeline case where the headers need to be visually hidden, they still need to exist to give that context to screen readers. A screen reader won't know to associate the external timeline as the column headers.

sdadn (approved):

eawww:
It gets the header info now! Thanks Charles. One last thing is the header cells shouldn't have buttons since they can't be accessed by anyone without a screen reader.

cm9361:
> It gets the header info now! Thanks Charles. One last thing is the header cells shouldn't have buttons since they can't be accessed by anyone without a screen reader.

Thank you for the feedback, @eawww.  I need the logic from upcoming PRs to remove the button.  I would not want to add that to this PR to prevent merge conflicts.  What I can do is make the cell be 'not selectable'.  This should give the desired behavior once the JIRAs in-flight are merged.  Would that be acceptable?

eawww:
@cm9361 could you point me to the jira or PR removal of the button is slated on?

cm9361:
> @cm9361 could you point me to the jira or PR removal of the button is slated on?

https://github.com/cerner/terra-framework/pull/1836/files#diff-6e67623f4efab392deb44bd2b292882500459c61651ae61d911cd8dbd01bc775R271

eawww (approved):
🥶 Cool beans! 🫘
Approving with the understanding that the removal of the button is being accomplished elsewhere.

PR #1848, opened 2023-10-20
[terra-tabs] Fixed keyboard navigation for draggable tabs

saket2403:
### Summary
Fixed keyboard navigation for draggable tabs failing due to draggable props from react beautiful dnd.

**What was changed:**
Tab was focused programmatically to fix the issue.

**Why it was changed:**
Arrow Key navigation was failing because of the focus not being on the tab after mouse click.

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9761 

---

@cerner/terra


packages/terra-tabs/src/common-tabs/_Tab.jsx
```diff
@@ -230,6 +230,10 @@ const Tab = ({
         onChange(event, itemKey);
       }
     }
+    // Fix for keyboard navigation after mouse click which was failing due to draggable props.
+    if (isDraggable) {
+      event.currentTarget.focus();
```
ShettyAkarsh, line 235: so now the focus always remains on the dragged tab ? and once dropped the keyboard navigation starts from that position ?

ShettyAkarsh:
@saket2403  does the arrow navigation works fine when the isdraggable prop is false ?

packages/terra-tabs/src/common-tabs/_Tab.jsx
saket2403: Before the change when a draggable tab was clicked, the focus was going onto the main or the body div due to the draggable props of the dnd library. So I have put the focus on the clicked tab. Change has nothing to do with whether the tab has been dragged or not.

saket2403:
> @saket2403 does the arrow navigation works fine when the isdraggable prop is false ?

Yes it works fine in non draggable tabs because when tab is clicked the focus is on the tab.

supreethmr (approved):

PR #1851, opened 2023-10-20
[terra-data-grid] DataGrid consumes Table

kenk2:
### Summary

**What was changed:**
DataGrid consumes table.

**Why it was changed:**
To eliminate redundant code and follow the DataGrid roadmap

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9543 

---

@cerner/terra


packages/terra-data-grid/src/DataGrid.module.scss
```diff
@@ -1,25 +1,11 @@
...
-    // Style for the grid itself
-    border-collapse: separate;
-    border-spacing: 0;
-    table-layout: fixed;
-    user-select: none;
-    width: 1px;
 
-    :focus {
-      outline: var(--terra-data-grid-cell-focus-outline, 3px dashed #000);
-      outline-offset: var(--terra-data-grid-cell-focus-outline-offset, -2px);
```
cm9361, line 22: This focus selector should not have been removed.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -48,6 +48,11 @@ const propTypes = {
    */
   columnHeaderHeight: PropTypes.string,
 
+  /**
+   * Boolean specifying whether or not the component should have zebra striping for rows.
+   */
+  isStriped: PropTypes.bool,
```
cm9361, line 54: This should not be a prop for the Flowsheet.  The Flowsheet should always be striped.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -94,6 +101,7 @@ function FlowsheetDataGrid(props) {
         id={id}
         ariaLabel={ariaLabel}
         ariaLabelledBy={ariaLabelledBy}
+        isStriped={isStriped}
```
cm9361, line 104: We should always be passing true.

packages/terra-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -33,6 +33,11 @@ const propTypes = {
    */
   id: PropTypes.string.isRequired,
 
+  /**
+   * Boolean specifying whether or not the component should have zebra striping for rows.
+   */
+  isStriped: PropTypes.bool,
```
cm9361, line 39: This should not be a prop for the WorklistDataGrid.  The WorklistDataGrid should always be striped.

packages/terra-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -418,11 +425,26 @@ function WorklistDataGrid(props) {
         id={id}
         ariaLabel={ariaLabel}
         ariaLabelledBy={ariaLabelledBy}
-        rows={rows}
+        isStriped={isStriped}
```
cm9361, line 428: This value should always be true.

packages/terra-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -418,11 +425,26 @@ function WorklistDataGrid(props) {
...
         rowHeight={rowHeight}
         rowHeaderIndex={rowHeaderIndex}
-        pinnedColumns={pinnedColumns}
-        overflowColumns={overflowColumns}
+        pinnedColumns={pinnedColumns.map((columns) => ({
+          ...columns,
+          isSelectable: columns.isSelectable !== false,
+          isResizable: columns.isResizable !== false,
+        }))}
+        overflowColumns={overflowColumns.map((columns) => ({
```
cm9361, line 443: Wouldn't the iterator variable name be "column"?  Isn't this an individual column that is being passed to the function?

packages/terra-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -418,11 +425,26 @@ function WorklistDataGrid(props) {
...
         ariaLabelledBy={ariaLabelledBy}
-        rows={rows}
+        isStriped={isStriped}
+        rows={rows.map((row) => ({
+          ...row,
+          cells: row.cells.map(cell => ({
+            ...cell,
+            isSelectable: cell.isSelectable !== false,
+          })),
+        }))}
```
cm9361, line 435: It might be better for maintenance if we create a function for this.

packages/terra-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -418,11 +425,26 @@ function WorklistDataGrid(props) {
...
-        overflowColumns={overflowColumns}
+        pinnedColumns={pinnedColumns.map((columns) => ({
+          ...columns,
+          isSelectable: columns.isSelectable !== false,
+          isResizable: columns.isResizable !== false,
+        }))}
+        overflowColumns={overflowColumns.map((columns) => ({
+          ...columns,
+          isSelectable: columns.isSelectable !== false,
+          isResizable: columns.isResizable !== false,
```
cm9361, line 446: Looks like we are duplicating the same logic for columns.  Maybe it should be a function.

packages/terra-table/src/Table.jsx
```diff
@@ -187,6 +182,13 @@ function Table(props) {
   // -------------------------------------
   // useEffect Hooks
 
+  useEffect(() => {
+    if (pinnedColumns.length === 0) {
+      // eslint-disable-next-line no-console
+      console.warn(ERRORS.PINNED_COLUMNS_UNDEFINED);
+    }
+  }, [pinnedColumns.length]);
```
cm9361, line 190: We would not want the update on the length of the columns.  We just need it to update when pinnedColumns does.

packages/terra-table/src/Table.jsx
```diff
@@ -243,9 +245,9 @@ function Table(props) {
   // -------------------------------------
   // event handlers
 
-  const handleColumnSelect = useCallback((columnId) => {
+  const handleColumnSelect = useCallback((columnId, cellCoordinates) => {
```
cm9361, line 248: Why do we need this new cellCoordinates argument for a column selection?

packages/terra-table/src/Table.jsx
```diff
@@ -243,9 +245,9 @@ function Table(props) {
...
     if (onColumnSelect) {
-      onColumnSelect(columnId);
+      onColumnSelect(columnId, cellCoordinates);
```
cm9361, line 250: I would not expect consumers to get a cell coordinates argument in their onColumnSelect callback.

cm9361:
The WDIO snapshots should not be changing based on this integration.  The presence of so many changes means that there are issues that need to be resolved.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -145,62 +145,24 @@ const DataGrid = injectIntl((props) => {
...
-  const grid = useRef();
+  const tableRef = useRef();
   const gridContainerRef = useRef();
 
   const handleFocus = useRef(true);
 
-  const [focusedRow, setFocusedRow] = useState(0);
-  const [focusedCol, setFocusedCol] = useState(0);
+  const focusedRow = useRef(0);
+  const focusedCol = useRef(0);
```
cm9361, line 154: This should not have been changed to use a ref instead of state variable.  It alters the entire implementation.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -449,7 +365,7 @@ const DataGrid = injectIntl((props) => {
           }
         } else {
           // Right key
-          nextCol += 1;
+          nextCol = Math.min(nextCol + 1, columns.length - 1);
```
cm9361, line 368: It does not seem like we should need such changes in this PR.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -554,49 +431,35 @@ const DataGrid = injectIntl((props) => {
...
+    <div
+      // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+      tabIndex={0}
+      ref={gridContainerRef}
+      onKeyDown={handleKeyDown}
+      onMouseDown={onMouseDown}
+      onFocus={onFocus}
+      className={cx('data-grid-container')}
+    >
+      <GridContext.Provider value={{ role: 'grid' }}>
```
cm9361, line 444: I thought we had a constant for the role of "grid" that you have been using.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -554,49 +431,35 @@ const DataGrid = injectIntl((props) => {
...
+        <Table
+          id={id}
```
cm9361, line 446: The id of the table should not be the same as the DataGrid component.  You would have two components with the same id and would break the DOM.  At a minimum, I would expect us to add a suffix.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -358,11 +273,12 @@ const DataGrid = injectIntl((props) => {
       && !!(element.offsetWidth || element.offsetHeight || element.getClientRects().length)
       && window.getComputedStyle(element).visibility !== 'hidden'
       && element.closest('[inert]') === null
-      && (element.id === id || !grid.current.contains(element)),
+      && (element.id === id || !tableRef.current.contains(element)),
```
cm9361, line 276: We should not need to leverage the tableRef in this scenario.  We can check if the gridContainer contains the element.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -411,7 +327,7 @@ const DataGrid = injectIntl((props) => {
             || (targetElement.hasAttribute('contentEditable') && targetElement.getAttribute('contentEditable') !== false))) {
       return;
     }
-
+    const columns = tableRef.current.tHead.rows[0].cells;
```
cm9361, line 330: We should not need to access tHead.  Also, we already have a reference to our number of columns in state.  We don't need to access the DOM to get this information as we have a React component.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -217,124 +179,77 @@ const DataGrid = injectIntl((props) => {
...
-    // if grid has selectable rows but no pinned columns, then set the offset of the first column to 0
-    if (hasSelectableRows && pinnedColumns.length === 0) {
-      lastPinnedColumnIndex = 0;
-      offsetArray.push(cumulativeOffset);
-      setPinnedColumnOffsets(offsetArray);
-      return;
+    if (gridContainerRef) {
+      const table = gridContainerRef.current.getElementsByTagName('table')[0];
+      tableRef.current = table;
     }
```
cm9361, line 198: Instead of querying the DOM, shouldn't the Table be able to accept a prop where this ref can be set?  I would not think that we would need to query the DOM to get this information.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -217,124 +179,77 @@ const DataGrid = injectIntl((props) => {
...
-          scrollOffsetX = nextCellRect.left - lastPinnedColumnRect.right;
-        }
-      } else {
-        scrollOffsetX = nextCellRect.left - gridContainerRect.left;
-      }
-
-      if (scrollOffsetX < 0) {
-        gridContainerRef.current.scrollBy(scrollOffsetX, 0);
+    // Obtain coordinate rectangles for scrollable container, column header, and new cell selection
+    // TODO: Implementation is not easy to understand, get suggestions on how to clean this logic up.
```
cm9361, line 205: We should not be leaving a TODO in our code.

packages/terra-data-grid/tests/jest/DataGrid.test.jsx
```diff
@@ -130,13 +82,13 @@ describe('DataGrid', () => {
...
-    const columnHeader = wrapper.find(ColumnHeaderCell);
+    const columnHeader = wrapper.find('th.selectable').first();
 
     // Simulate onMouseDown event on row selection column header
-    columnHeader.at(0).simulate('mouseDown');
+    columnHeader.simulate('mouseDown');
 
     // Validate mock function was called from simulated click event
-    expect(mockColumnSelect).toHaveBeenCalledWith(WorklistDataGridUtils.ROW_SELECTION_COLUMN.id);
+    expect(mockColumnSelect).toHaveBeenCalled();
```
cm9361, line 91: Why did this change.  We explicitly wanted to test that this value was passed.

packages/terra-data-grid/tests/jest/WorklistDataGrid.test.jsx
```diff
@@ -79,10 +77,10 @@ describe('WorklistDataGrid', () => {
     );
 
     // Find column headers
-    const nonSelectableCell = wrapper.find(Row).at(0).find('th:not(.selectable)');
+    const nonSelectableCell = wrapper.find('th:not(.selectable)').first();
```
cm9361, line 80: We should have still be able to find the React components being rendered.  We would just have to use a string instead of using the import.  This could keep our tests pretty much the same.  I don't think we should be changing to using first and other concepts in this PR.  We just need to make sure everything is working.  Then we can look at minor updates.

packages/terra-data-grid/tests/wdio/data-grid-spec.js
```diff
@@ -113,24 +113,6 @@ Terra.describeViewports('DataGrid', ['medium', 'large'], () => {
...
-      Terra.validates.element('hover-odd-row-cell-selection', { selector: defaultSelector });
-    });
-
-    it('hovers over selectable cell on an even row', () => {
-      browser.$$('tbody tr')[3].$$('td')[0].moveTo();
-      browser.pause(2000);
-      Terra.validates.element('hover-even-row-cell-selection', { selector: defaultSelector });
-    });
-  });
-
```
cm9361, line 133: Why was this test removed?

packages/terra-table/tests/jest/Table.test.jsx
```diff
@@ -161,7 +161,7 @@ describe('Table', () => {
     columnHeader.at(0).simulate('mouseDown');
 
     // Validate mock function was called from simulated click event
-    expect(mockColumnSelect).toHaveBeenCalledWith(RowSelectionUtils.ROW_SELECTION_COLUMN.id);
+    expect(mockColumnSelect).toHaveBeenCalledWith(RowSelectionUtils.ROW_SELECTION_COLUMN.id, { col: 0, row: 0 });
```
cm9361, line 164: This change would be incorrect.

packages/terra-data-grid/src/DataGrid.module.scss
kenk2: Restored this.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
kenk2: Moved the striped logic out of the Flowsheet and WDG. Now only specified in DataGrid.

packages/terra-data-grid/src/WorklistDataGrid.jsx
kenk2: Created some mapper functions to apply this to Datagrid and WDG.

packages/terra-table/src/Table.jsx
kenk2: Removed this and the sending of the cell coordinates.

packages/terra-data-grid/src/DataGrid.jsx
kenk2: Remove the tHead reference.

packages/terra-data-grid/src/DataGrid.jsx
kenk2: Had a discussion, it turns out that we can't use forwardRef with components wrapped with react-intl: https://formatjs.io/docs/react-intl/upgrade-guide-3x/#migrate-withref-to-forwardref

So we'll just have to query the DOM. It's not the ideal solution but we can hopefully solve this unideal behavior with our tech currency initiative.

packages/terra-data-grid/src/DataGrid.jsx
kenk2: Took this out, this was doing too much.

packages/terra-data-grid/tests/jest/WorklistDataGrid.test.jsx
kenk2: Reverted this and other changes in the test files. Only implement minimally needed changes.

packages/terra-data-grid/tests/wdio/data-grid-spec.js
kenk2: I ended up restoring this because I thought that flowsheet, datagrid and WDG didn't have these stylings but they do per UX. Restored all of these.

packages/terra-table/tests/jest/Table.test.jsx
kenk2: Removed.

packages/terra-data-grid/src/DataGrid.jsx
kenk2: I know this isn't ideal, but I have not been able to get the focus logic working unless we move the tracking from state to refs. This also gets us closest to the original behavior of the WDG, DG and Flowsheet.

packages/terra-table/src/subcomponents/Row.module.scss
```diff
@@ -3,6 +3,10 @@
 
 
 :local {
+  .striped-table-row:nth-child(even) {
+    background-color: var(--terra-table-row-striped-background-color, #f0f0f0);
+  }
+
```
kenk2, line 9: For some reason, this was causing issues with row selection. I had to move this up to the same level as `.row` because CSS specificity would always use the striped table row style instead of the selected style.

packages/terra-data-grid/tests/jest/DataGrid.test.jsx
```diff
@@ -69,21 +67,21 @@ afterAll(() => {
...
         pinnedColumns={dataFile.cols.slice(0, 2)}
         overflowColumns={dataFile.cols.slice(2)}
         rows={dataFile.rows}
       />,
-    ).dive().dive();
+    );
 
     // One row used for the header.
-    const columnHeader = wrapper.find(ColumnHeader);
+    const columnHeader = wrapper.find('Memo(ColumnHeader)');
```
kenk2, line 80: This will look out of place, but shallow doesn't expose the deeply nested component and memoized components don't have display name to be found if you do not have the actual component to compare it with.

packages/terra-data-grid/src/DataGrid.jsx
kenk2: I believe this is just a simple passthrough. There won't be 2 elements with the same ID unless I'm mistaken.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -217,124 +180,75 @@ const DataGrid = injectIntl((props) => {
...
+    const [table] = gridContainerRef.current.getElementsByTagName('table');
+    grid.current = table;
+  }, [gridContainerRef]);
 
   // -------------------------------------
 
   const handleMoveCellFocus = (fromCell, toCell) => {
     // Obtain coordinate rectangles for grid container, column header, and new cell selection
-    const gridContainerRect = gridContainerRef.current.getBoundingClientRect();
+    const tableContainer = gridContainerRef.current.firstChild;
```
kenk2, line 203: The weird thing about this implementation is that the scroll is controlled by the table's container, not the datagrid's. So we have to search for the table container here.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -217,124 +180,75 @@ const DataGrid = injectIntl((props) => {
...
-
   // -------------------------------------
   // useEffect Hooks
 
-  // useEffect for row selection
   useEffect(() => {
-    setDataGridColumns(displayedColumns.map((column) => initializeColumn(column)));
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [hasSelectableRows]);
+    // We have to search the dom for the table ref because react-intl v2 does not support forwardRef.
```
kenk2, line 193: Discussed offline: See comment about react-intl 2 not supporting forwardRef (but supported in v3 which we currently do not support)

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -362,7 +276,7 @@ const DataGrid = injectIntl((props) => {
     );
 
     // Identify index of the active element in the DOM excluding data grid children
-    const index = focusableElements.indexOf(grid.current);
+    const index = focusableElements.indexOf(gridContainerRef.current);
```
kenk2, line 279: About this: The selector was trying to find `[tabindex]:not([tabindex='-1'])` as part of the query. However with this new implementation, the `tabindex={0}` is now on the DataGrid container (and not on the table). So we have to change the reference from the grid to the container.

kenk2:
Note: Currently https://github.com/cerner/terra-framework/pull/1813 is also in flight. I'd like to delete the subcomponents as part of this PR but since that PR relies on the subcomponents, I'm not inclined to delete them as part of this PR.

packages/terra-data-grid/src/utils/dataGridMappers.js
```diff
@@ -0,0 +1,15 @@
+const mapDataGridRow = (row) => ({
```
cm9361, line 1: Let's not create a new file.  I think we c

packages/terra-data-grid/src/DataGrid.module.scss
```diff
@@ -9,13 +9,9 @@
...
+  table {
     // Style for the grid itself
-    border-collapse: separate;
-    border-spacing: 0;
-    table-layout: fixed;
     user-select: none;
-    width: 1px;
 
     :focus {
       outline: var(--terra-data-grid-cell-focus-outline, 3px dashed #000);
```
cm9361, line 17: Is this style already handled by the table?  Does the DataGrid have a component that would require us to apply this style separately?

packages/terra-data-grid/src/DataGrid.jsx
cm9361: I am pretty sure that we have logic that depends on this, especially in the upcoming PRs.  We should be able to make the logic work without making this change.  How did it work before?

packages/terra-data-grid/src/DataGrid.jsx
cm9361, line 203: Assuming that the table container is the first child would allow the logic to be easily broken.  If another element is added, it would break the implementation.  We probably need a better way to find the reference to this element.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -217,124 +180,75 @@ const DataGrid = injectIntl((props) => {
...
     if (nextCellRect.right > gridContainerRect.right) {
-      gridContainerRef.current.scrollBy(nextCellRect.right - gridContainerRect.right, 0);
+      tableContainer.scrollBy(nextCellRect.right - gridContainerRect.right, 0);
     } else {
       // Calculate horizontal scroll offset for left boundary
       let scrollOffsetX = 0;
-      if (pinnedColumnOffsets.length > 0) {
-        if (toCell.col > pinnedColumnOffsets.length - 1) {
-          const lastPinnedColumnRect = grid.current.rows[toCell.row].cells[pinnedColumnOffsets.length - 1].getBoundingClientRect();
+      if (pinnedColumns.length > 0 || hasSelectableRows) {
```
cm9361, line 214: I would not expect us to have the "|| hasSelectableRows" change.  This condition should not require a change.

packages/terra-data-grid/src/DataGrid.jsx
cm9361: Then we should also update the focusableElements logic to use the proper container in the check

packages/terra-data-grid/src/DataGrid.jsx
cm9361: We should not pass the same id to such a field.  If someone sees that, they would assume it could be applied to the DOM and is unique.  It would be best to do a minor adjustment to ensure uniqueness.

packages/terra-data-grid/src/utils/WorklistDataGridUtils.js
```diff
@@ -2,7 +2,7 @@
  * The Object representation of the row selection column.
  */
 const ROW_SELECTION_COLUMN = {
-  id: 'WorklistDataGrid-rowSelectionColumn',
+  id: 'table-rowSelectionColumn',
```
cm9361, line 5: This will require a major version bump for the WorklistDataGrid.  We also need to update the appropriate examples and communicate the change to our consumers for Fusion Passthrough.

packages/terra-data-grid/tests/jest/DataGrid.test.jsx
cm9361: I don't think this is correct.

packages/terra-data-grid/tests/jest/DataGrid.test.jsx
```diff
@@ -98,19 +96,19 @@ describe('DataGrid', () => {
...
 
-    const rows = wrapper.find(Row);
+    const rows = wrapper.find('Row');
     expect(rows).toHaveLength(dataFile.rows.length);
-    verifyRow(0, rows.get(0), dataFile.rows[0], dataFile.cols);
-    verifyRow(1, rows.get(1), dataFile.rows[1], dataFile.cols);
-    verifyRow(2, rows.get(2), dataFile.rows[2], dataFile.cols);
+    verifyRow(0, rows.get(0), mapDataGridRow(dataFile.rows[0]), dataFile.cols.map(columns => mapDataGridColumn(columns)));
+    verifyRow(1, rows.get(1), mapDataGridRow(dataFile.rows[1]), dataFile.cols.map(columns => mapDataGridColumn(columns)));
+    verifyRow(2, rows.get(2), mapDataGridRow(dataFile.rows[2]), dataFile.cols.map(columns => mapDataGridColumn(columns)));
```
cm9361, line 111: I would not expect you to have to call the map function in your tests.  You should be testing for a known value not the result of a function.

packages/terra-framework-docs/src/terra-dev-site/doc/table/Examples/PinnedColumnsTable.jsx
```diff
@@ -54,6 +54,7 @@ const PinnedColumnsTable = () => (
     pinnedColumns={tableData.cols.slice(0, 2)}
     overflowColumns={tableData.cols.slice(2)}
     rows={tableData.rows}
+    ariaLabel="Table"
```
cm9361, line 57: Shouldn't our aria labels describe the example.  They are the same value for every example.

mjpalazzo:
Great work on Table @kenk2 
Below are a few observations regarding accessibility:
Keyboard navigation - Tables do not have keyboard interaction requirements so the only thing I validated here is that sighted users could scroll to view the right most cells. I observed browser-dependent differences.
With Chrome when focus was on the Table example I was able to use Down Arrow to bring right most cells into view for the Default Table and Pinned Columns Table examples. I observed one minor issue with the Pinned Columns Table example - when I used Up Arrow some content remained hidden behind the pinned columns.
With Edge I was not able to scroll at all using the Down and Up Arrow keys. This is a problem for Tables that are wider than the viewable area. It may be an issue with Terra Scroll and not the Table component. The scroll bar and its arrow buttons are visible but not interactable with the keyboard. If Terra Scroll supported keyboard navigation, then this issue would resolved. @eawww - Please share your thoughts.
For the Sortable Table example, I was able to Tab to the column headers of sortable columns. However, the sort icons are not displaying until I press Enter. I checked the ARIA APG sortable table example and it appears to be optional to display the sort icon in headers of columns that are not sorted. Here is what the ARIA example says:
"Adds a diamond shaped icon (e.g. ♢) to the header of each column that can be sorted but is not currently sorted. Some sortable tables add an icon to unsorted columns to help users distinguish sortable columns from columns that cannot be sorted. It is important that the shape of the unsorted icon differ in more than just color and size from the icons that indicate sort direction (e.g. '▼' and '▲') so people with visual impairments can easily distinguish them."
@eawww Please comment on whether we should add icons to sortable columns that are not sorted. If we don't, how does the user know that the column is sortable?
Table with Focusable Cell Elements example - I was able to use Tab and Shift Tab to navigate to the interactive elements. The elements themselves could be interacted with using their required keyboard interactions. For example, I could activate the button by pressing Enter. type text in the input field, expand the listbox and select items, launch the link, etc.
I will post screen reader observations in a separate comment.

mjpalazzo:
Screen reader observations
JAWS - All examples:
- When I pressed Tab to navigate to the table. JAWS announced "Table" When I pressed 'T' to navigate to the table, JAWS announced "Table with 10 columns and 3 rows column 1 row 1 patient button" It appears to be a quirk of JAWS that the number of columns and rows are not announced when Tab is used to navigate to the table because "Table with 10 columns and 3 rows " is displayed in the JAWS transcript bar at the top of the screen.
- For the column header cells, JAWS announced "button" even though the header cell is not interactable. JAWS should announce "column header" This may be related to [https://jira2.cerner.com/browse/UXPLATFORM-9804](url)
- I pressed Down arrow to navigate to each cell. JAWS announced the cell content. 
- Up Arrow navigates to the previous cell.
- When a navigated to the last cell and then pressed Down Arrow one more time JAWS announced "Table end" as expected.
Sortable Table example:
- When I pressed 'T' to navigate to the table, JAWS announced "Table with 10 columns and 3 rows column 1 row 1 patient button sorted ascending"
- Pressed Enter and JAWS announced "Patient button sorted ascending Fleck Arthur" This is odd behavior. The content of the column 1 row 2 should not be announced when focus is on the column header.
- Pressed Tab to navigate to the column 2 header. JAWS announced "Location column 2 row 1" No sort icon displays. JAWS should announce that the column is sortable. 
- Pressed Enter and the column sorted. JAWS announced "Enter" but not the sort order. Note that the APG Sortable Table example behaves the same way.
- When I use Tab to navigate to the sortable column headers the Down and Up Arrow keys stop working. Screen reader cannot announce contents of other column header cells and data cells. This is an issue that must be fixed.

Note that the ARIA APG Sortable Table example makes the following recommendation and I suggest implementing it in our sortable table example:
"To help screen reader users understand the purpose of the buttons in the column headers, an off-screen description of the sort functionality of the buttons is appended to the caption text. The description is added to the caption instead of to each button to prevent repetitious verbosity that could interfere with understanding of the column titles."

@eawww - Please provide guidance on this recommendation.

Table Focusable Cell example
I pressed 'T' to navigate to the table and used Tab and Shift Tab to navigate to the interactive elements. JAWS announced the name, role, value and instructions for each as expected.

eawww:
@mjpalazzo 

> With Edge I was not able to scroll at all using the Down and Up Arrow keys. This is a problem for Tables that are wider than the viewable area. It may be an issue with Terra Scroll and not the Table component. The scroll bar and its arrow buttons are visible but not interactable with the keyboard. If Terra Scroll supported keyboard navigation, then this issue would resolved. @eawww - Please share your thoughts.

For Table I believe we are leaving that up to the consumer. For the grids, we've already implemented logic to scroll selected cells into view. If that's broken when this PR is up for review, it'll need to be fixed.

> @eawww Please comment on whether we should add icons to sortable columns that are not sorted. If we don't, how does the user know that the column is sortable?

Previously, we didn't offer anything like this as a way of reducing visual clutter, but you do make a good point. It would be good to offer some affordance to sighted users that the column is sortable.
I've run this by some other UX folks and I think it'll require further investigation.

> "To help screen reader users understand the purpose of the buttons in the column headers, an off-screen description of the sort functionality of the buttons is appended to the caption text. The description is added to the caption instead of to each button to prevent repetitious verbosity that could interfere with understanding of the column titles."

This couldn't hurt but, since a button column header almost always means "sort", I wouldn't call it a dealbreaker.

mjpalazzo:
@kenk2 - Thanks for addressing the items I observed. For the table examples that are not sortable, JAWS no longer announces "button" for the column header cells. 
For the sortable table example, JAWS announces the column header cell content and role "button" and sort direction and helper text.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -151,39 +146,20 @@ const DataGrid = injectIntl((props) => {
...
   const displayedColumns = (hasSelectableRows ? [WorklistDataGridUtils.ROW_SELECTION_COLUMN] : []).concat(pinnedColumns).concat(overflowColumns);
-  const [dataGridColumns, setDataGridColumns] = useState(displayedColumns.map((column) => initializeColumn(column)));
 
-  // Manage column resize
-  const [tableHeight, setTableHeight] = useState(0);
-  const [activeIndex, setActiveIndex] = useState(null);
-  const activeColumnPageX = useRef(0);
-  const activeColumnWidth = useRef(200);
-  const tableWidth = useRef(0);
+  const dataGridRows = rows.map((row) => ({
```
cm9361, line 151: You should add a comment here.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -234,114 +208,60 @@ const DataGrid = injectIntl((props) => {
...
+    if (onColumnSelect) {
       onColumnSelect(columnId);
     }
+  }, [onColumnSelect, displayedColumns]);
+
+  const handleRowSelectionHeaderSelect = useCallback(() => {
+    setFocusedCol(0);
+    setFocusedRow(0);
+    if (onColumnSelect) {
+      onColumnSelect(WorklistDataGridUtils.ROW_SELECTION_COLUMN.id);
```
cm9361, line 263: You would only want this id to be sent from the WorklistDataGrid component.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -601,54 +461,43 @@ const DataGrid = injectIntl((props) => {
...
+          rowHeight={rowHeight}
+          rowHeaderIndex={rowHeaderIndex}
+          onColumnResize={onColumnResize}
+          onColumnSelect={handleColumnSelect}
+          onCellSelect={handleCellSelection}
+          onRowSelectionHeaderSelect={handleRowSelectionHeaderSelect}
+          hasSelectableRows={hasSelectableRows}
+          isStriped
+          gridRef={grid}
+          gridContainerRef={tableContainerRef}
```
cm9361, line 498: You should probably just use the React Context Provider value if using this approach.  Then you don't require the props to be passed.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -98,7 +98,11 @@ function FlowsheetDataGrid(props) {
     intl,
   } = props;
 
-  const flowsheetColumns = useMemo(() => columns.map(column => ({ ...column, isResizable: false })), [columns]);
+  const flowsheetColumns = useMemo(() => columns.map(column => ({
+    ...column,
+    isSelectable: column.isSelectable !== false,
```
cm9361, line 103: I think that you are doing this in the DataGrid component as well.

packages/terra-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -178,6 +178,22 @@ function WorklistDataGrid(props) {
   const gridReceivedFocus = useRef(false);
   const gridHasFocus = document.getElementById(`${id}-worklist-data-grid-container`)?.contains(document.activeElement);
 
+  const makeWorklistDataGridColumns = (column) => ({
+    ...column,
+    isResizable: column.isResizable !== false,
+    isSelectable: column.isSelectable !== false,
```
cm9361, line 184: It looks like this might be duplicated in DataGrid.  If so, I think we have to choose one approach.  If we are going to update in the subcomponents, which is ok, we would remove the logic from the DataGrid.

packages/terra-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -178,6 +178,22 @@ function WorklistDataGrid(props) {
   const gridReceivedFocus = useRef(false);
   const gridHasFocus = document.getElementById(`${id}-worklist-data-grid-container`)?.contains(document.activeElement);
 
+  const makeWorklistDataGridColumns = (column) => ({
```
cm9361, line 181: This function should be capable of iterating over the map as well. Then all you have to do is pass the necessary column array.

packages/terra-table/src/subcomponents/ColumnHeader.jsx
```diff
@@ -19,9 +19,21 @@ const propTypes = {
   headerHeight: PropTypes.string.isRequired,
 
   /**
-   * Number that specifies the height of the table in pixels.
-   */
+  * Number that specifies the height of the data grid in pixels.
```
cm9361, line 22: This is the Terra Table component.  I would not expect the prop to reference the data grid.

eawww (approved):
Aside from things mentioned above, seems comparable functionally to how the components were before.

packages/terra-data-grid/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Fixed
   * Fixed the column header background color for the Orion Fusion theme.
   * Fixed pinned column divider issue during keyboard resize.
+  * Fixed left column header border for components.
```
cm9361, line 8: Did you rebase with main?  This code should already be present and not show up as a change.

packages/terra-data-grid/src/DataGrid.module.scss
```diff
@@ -9,19 +9,7 @@
     width: 100%;
```
cm9361, line 9: I would not expect the data-grid-container element to have an overflow.  The Terra Table already had an element with an overflow.

packages/terra-data-grid/src/DataGrid.module.scss
cm9361: It appears that this is still removed.

packages/terra-data-grid/src/utils/WorklistDataGridUtils.js
```diff
@@ -8,8 +8,16 @@ const ROW_SELECTION_COLUMN = {
...
+  width: 40,
+  isSelectable: true,
+  isResizable: false,
+};
+
 const WorklistDataGridUtils = {
   ROW_SELECTION_COLUMN,
+  TABLE_ROW_SELECTION_COLUMN,
 };
```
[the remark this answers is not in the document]
kenk2: We do because we're using different ID's for the row selection column in each component. Table uses `table-rowSelectionColumn` while WDG uses `worklistDataGrid-rowSelectionColumn`.

packages/terra-data-grid/CHANGELOG.md
kenk2: This change was not made in datagrid, so I had to fix it in this PR. We're going to be inheriting styling from table from now on so the fix was needed here.

packages/terra-data-grid/src/DataGrid.module.scss
kenk2: So it turns out that this is in table, so we don't actually need this here after all.

packages/terra-framework-docs/src/terra-dev-site/test/table/mockSortableTableData.json
```diff
@@ -1,8 +1,8 @@
 {
   "cols": [
-    { "id": "Column-0", "displayName": "Vitals", "isSelectable": true, "sortIndicator": "ascending" },
-    { "id": "Column-1", "displayName": "March 16", "isSelectable": true },
-    { "id": "Column-2", "displayName": "March 17" },
+    { "id": "Column-0", "displayName": "Vitals", "isSelectable": true, "sortIndicator": "ascending", "isResizable": true },
+    { "id": "Column-1", "displayName": "March 16", "isSelectable": true, "isResizable": true },
+    { "id": "Column-2", "displayName": "March 17", "isSelectable": true },
```
cm9361, line 5: Why did this need to change for the Terra Table?

packages/terra-framework-docs/src/terra-dev-site/test/table/SortableTable.test.jsx
```diff
@@ -42,6 +42,7 @@ const SortableTable = () => {
       id="sortable-table"
       overflowColumns={tableColumns}
       rows={tableRows}
+      columnResizeIncrement={5}
```
cm9361, line 45: This should not have required a change since we have a default value defined.

packages/terra-framework-docs/src/terra-dev-site/test/table/mockSortableTableData.json
kenk2: This was just to provide a way for people to test column resizing and selectable columns in table.

packages/terra-table/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 * Fixed
   * Fixed the column header background color for the Orion Fusion theme.
+  * Fixed an issue with row selection integrating with table zebra stripes.
```
cm9361, line 11: This should not show up as your change.

packages/terra-data-grid/CHANGELOG.md
cm9361: There is no need to update the data grid logic.  We won't be using it.

packages/terra-data-grid/src/utils/WorklistDataGridUtils.js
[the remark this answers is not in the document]
cm9361: You don't need this entire structure just for an id.

kenk2:
> Fails
> * terra-data-grid
> * terra-table
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [c0e1e10](https://github.com/cerner/terra-framework/commit/c0e1e104fb4f816f7ca4fda4593b2981f8e41c23)

These do not involve consumer-facing changes. Therefore, no changelogs will be needed for these packages.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -615,55 +481,47 @@ const DataGrid = injectIntl((props) => {
...
+      onBlur={onBlur}
+      id={id}
+      className={cx('data-grid-container')}
+    >
+      <GridContext.Provider value={{
+        role: GridConstants.GRID,
+        setCellAriaLiveMessage,
+        tableRef: grid,
+        tableContainerRef,
+      }}
```
cm9361, line 501: You did not memoize your provider value.  This will cause many unnecessary rerenders.

cm9361 (approved):

sdadn (approved):

packages/terra-data-grid/tests/jest/DataGrid.test.jsx
```diff
@@ -64,21 +66,22 @@ afterAll(() => {
 
 describe('DataGrid', () => {
   it('verifies that the grid created is consistent with the rows and overflowColumns props', () => {
-    const wrapper = shallowWithIntl(
+    const wrapper = mountWithIntl(
```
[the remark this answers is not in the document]
kenk2: I attempted to make this work with the existing shallow render but was not able to find the components with `shallow` alone as Table is now embedded in a Host Component, thus making us unable to find the component without fully mounting it.

artpark (approved):

smason0 (approved):

PR #1852, opened 2023-10-22
[terra-table] Add sections to the Terra Table component

cm9361:
### Summary

**What was changed:**
The Terra Table was updated to support sections.

**Why it was changed:**
Consumers require the ability to group data in sections to present data to users in groupings.

### Testing

This change was tested using:

- [X] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [X] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9402 
UXPLATFORM-9536
---

@cerner/terra


mjpalazzo:
@cm9361 Thanks for adding the section functionality. I have a couple of observations:
The focus styling is the light blue solid line. Should it be the black dashed line?
In the examples, the first 3 header cells can get focus with Tab and the first header cell has a sort order icon, but none of these columns are actually sortable. I imagine sorting is not desired for tables with sections.

cm9361:
> @cm9361 Thanks for adding the section functionality. I have a couple of observations: The focus styling is the light blue solid line. Should it be the black dashed line? In the examples, the first 3 header cells can get focus with Tab and the first header cell has a sort order icon, but none of these columns are actually sortable. I imagine sorting is not desired for tables with sections.

@mjpalazzo - The focus styling will be fixed here: https://github.com/cerner/terra-core/pull/3956

eawww:
I didn't notice this before but the cells in the lower section aren't reading their respective top level column header in JAWS, only the section header. This may actually need to be manually set using the `headers` attribute.

Example 4 on [this page](https://www.davidmacd.com/blog/csun/files/complex-tables.html) reads the appropriate headers for all cells and appears to accomplish that using the `headers` attribute.

packages/terra-framework-docs/src/terra-dev-site/doc/table/About.1.doc.mdx
```diff
@@ -64,6 +64,17 @@ A column specifies the data to render a cell in the header row of the Worklist D
 |**width**|number|optional|none|A number (in px) specifying the width of the column. If not provided, the Data Grid's default column width will be used.|
 |**sortIndicator**|custom|optional|none|A string indicating which sorting indicator should be rendered. If not provided, no sorting indicator will be rendered. One of `ascending`, `descending`.
 
+### Section
+A section defines the rows rendered within the section.
+
+|Name|Type|Is Required|Default Value|Description|
+|---|---|---|---|---|
+|**id**|string|<span style={{color:'#d53040'}}>required</span>|none|An identifier to uniquely identify the row within the grid.|
```
smason0, line 72: nit: ... to uniquely identify the **section**

packages/terra-table/tests/wdio/table-spec.js
```diff
@@ -157,17 +157,29 @@ Terra.describeViewports('Table', ['medium', 'large'], () => {
...
     });
 
-    it('Validates second scrollable table receives focus', () => {
+    it('Validates table is scrollable horizontally via arrow keys', () => {
       browser.keys(['Tab', 'Tab', 'ArrowRight', 'ArrowRight']);
       Terra.validates.element('scrollable-table-second-scroll', { selector: scrollableTableSelector });
     });
   });
 
+  describe('Table with Sections', () => {
```
smason0, line 171: Should we also include tests for collapsible sections?

packages/terra-table/src/subcomponents/Section.jsx
```diff
@@ -0,0 +1,176 @@
...
+
+  const theme = useContext(ThemeContext);
+
+  const handleMouseDown = (event) => {
+    onSectionSelect(id);
+    event.stopPropagation();
+  };
+
+  return (
+    <>
```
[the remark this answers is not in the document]
cm9361: We were using the id to provide the information to the consumers in the callback functions.  I don't think we have a need to add the id to the DOM at the moment.  Maybe this changes in the future.

packages/terra-framework-docs/src/terra-dev-site/doc/table/About.1.doc.mdx
cm9361: Good catch.  I will get this updated.

smason0 (approved):
+1 with a few comments

packages/terra-table/tests/wdio/table-spec.js
cm9361: The collapsible sections exist in the test file so they should be covered.

packages/terra-table/CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 
 * Added
   * Added the ability to toggle zebra striping for table rows.
+  * Added sections support to the table component.
```
sdadn, line 23: Is this in the right section?

packages/terra-table/tests/jest/Table.test.jsx
```diff
@@ -56,9 +56,11 @@ const tableData = {
   ],
 };
 
+let mockSpyUuid;
 beforeAll(() => {
   jest.spyOn(console, 'error').mockImplementation();
   jest.spyOn(console, 'warn').mockImplementation();
+  mockSpyUuid = jest.spyOn(uuidv4, 'v4').mockReturnValue('00000000-0000-0000-0000-000000000000');
```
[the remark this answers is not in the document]
cm9361: Yes, the section receives a generated id.  If we do not mock the value, it will change for every snapshot.

eawww (commented):
Works great in JAWS, but VoiceOver has a lot of trouble with the secondary header cells. Unless traversing the first column, full span headers are skipped entirely. Secondary headers aren't read as part of a cell's context. The headers attribute is ignored.

Since it works well in JAWS, @scottwilmarth and I think we should go ahead with this, but could you add a notice to the Table page saying something like "Due to poor support in VoiceOver for spanned table headings, Table sections are not currently recommended for applications intended to be used on Apple devices."

eawww (approved):

PR #1855, opened 2023-10-24
[terra-folder-tree] Initial component setup

sycombs:
### Summary

This PR sets up the base terra-folder-tree component to prepare for development on terra-folder-tree.

These changes are based on the Terra File Path changes from https://github.com/cerner/terra-framework/pull/1771.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9679 

---

@cerner/terra


sycombs:
> Fails
> * terra-framework-docs
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [28efb84](https://github.com/cerner/terra-framework/commit/28efb847d6ad5951abb6c6a4e5ea4a52b5ec2d8a)

The only changes to terra-framework-docs is placeholder documentation, so omitting the Changelog entry.

artpark (approved):

packages/terra-folder-tree/translations/en-US.json
```diff
@@ -0,0 +1,3 @@
+{
+  "Terra.folder-tree.placeholder": "Placeholder"
+}
```
sdadn, line 3: nit: no newline.

sdadn (approved):

packages/terra-folder-tree/translations/en-US.json
sycombs: [9e7aea2](https://github.com/cerner/terra-framework/pull/1855/commits/9e7aea20e72da4f2d9928bf17fd119ecae72f65b)

packages/terra-folder-tree/src/FolderTree.module.scss
```diff
@@ -0,0 +1,7 @@
+@import '~terra-mixins/lib/Mixins';
```
cm9361, line 1: Do we need Terra mixins?

packages/terra-folder-tree/.eslintrc.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
```
adoroshk, line 1: I have noticed that in that and FilePath component we have `.eslintrc.js` and `jest.config.js` on package level rather than use them from terra-framework level and then use as in older components:
`"jest": "jest --config ../../jest.config.js"`
`"lint:js": "eslint --ext .js,.jsx . --ignore-path ../../.eslintignore"`
I was wondering why we switched to that?

cm9361 (approved):

packages/terra-menu/tests/jest/__snapshots__/Menu.test.jsx.snap
```diff
@@ -524,6 +524,7 @@ exports[`Menu correctly applies the theme context className 1`] = `
                                                   "timeZone": null,
                                                 }
                                               }
+                                              isTabFocusDisabled={false}
```
adoroshk, line 527: Thank you for taking care of that! Came across the same problem in other PR, glad to see it's taken care of!

packages/terra-folder-tree/src/FolderTree.module.scss
sycombs: Not sure, this is just a placeholder file to get started on FolderTree development and the content of this file was copied from the equivalent setup for file path: https://github.com/cerner/terra-framework/pull/1771/files#diff-213a5eefceecea61e9ec5bbdb472d503d5193b21a7f7e655af877752445bd241R1

packages/terra-folder-tree/.eslintrc.js
adoroshk: Either way works for me. I am just creating a new component set up for Compact Interactive List and was wondering which approach to use, and if separate config - in what cases, hence asked.

sycombs: I tried removing the package-level `.eslintrc.js` and running eslint from the package level, and it errors due to not having an eslint config. I think this is because the root eslint config is defined in the root package.json, so it can't be resolved from the package level?

We could probably define a root-level `.eslintrc.js` to fix this; for now I'm going to leave the files as they are so that the Folder Tree package is available in main, but I'm open to changing this in a future PR.

PR #1867, opened 2023-10-30
[terra-table] Make Buttons in Selectable Header Cells Focusable

cm9361:
### Summary

**What was changed:**
The Terra Table was updated such that the selectable column header cell buttons are focusable instead of the column header cells.

**Why it was changed:**
The change was made to be compliant with accessibility standards.

### Testing

This change was tested using:

- [x] WDIO
- [X] Jest
- [ ] No tests are needed

### Reviews

- [X] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details
N/A

**This PR resolves:**

UXPLATFORM-9802 

---

@cerner/terra


adoroshk (approved):
It works as expected, but I noticed that the focus outline got 1px thicker than it was before the change. I assume it is expected as it is visible on the test snapshots, so I am approving the PR, but decided to mention just to be sure it is not something we missed by accident.

cm9361:
> It works as expected, but I noticed that the focus outline got 1px thicker than it was before the change. I assume it is expected as it is visible on the test screenshots, so I am approving the PR, but decided to mention just to be sure it is not something we missed by accident.

Yes, this is the style called out by Eric for all of our focusable elements when in focus.  The table cell layout was just negatively impacting the style.

nikhitasharma (approved):

smason0 (approved):

PR #1869, opened 2023-10-31
Compact interactive list component layout

adoroshk:
### Summary
This PR creates the layout for the Compact Interactive List component, adds layout usage examples and tests.

**What was changed:**
The Compact Interactive List component layout has following features:
1. The layout is div-based with roles "grid", "row", "gridCell" assigned to corresponding semantic element.
2. The Items (semantic rows) are arrange-able in columns (the number of columns is provided via prop).
3. Semantic rows are located at fixed points, not flowed via CSS (this should be added in the future).
4. Semantic rows direction is top to bottom, but with a prop can be changed to left to right.
5. Customers can provide the cells’ width, but can also set the props for cells to grow.
6. The supported width units are px, em, rem. 

NOTE: Semantic Rows headers and column headers should be out of scope of this Jira and added once the key navigation is added. As they need to be hidden and the only purpose of them is to provide a row/column context with keyboard navigation, there is no way to test if they work before the keyboard navigation exists.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [ ] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9557


adoroshk:
Doesn't look like terra-compact-interactive-list changelog needs changes at that point, but correct me if I am wrong.

mjpalazzo:
@adoroshk - great work on CIL! I have a couple of observations. I cannot get the focus indicator (dashed black border) to display on the first item in the list. Also I could not get focus on or interact with the favorite icon.

adoroshk:
> @adoroshk - great work on CIL! I have a couple of observations. I cannot get the focus indicator (dashed black border) to display on the first item in the list. Also I could not get focus on or interact with the favorite icon.

@mjpalazzo that part is not implemented yet. For now the html layout is not intractable. The next step is to manage focus, keyboard navigation and assistive technologies announcement.

packages/terra-compact-interactive-list/src/proptypes/columnShape.js
```diff
@@ -5,22 +5,44 @@ const columnShape = PropTypes.shape({
...
+   * Whether the column width can grow (a flex growing column).
+   * Defaults to false if the width is set and true if the width is not set.
    */
-  width: PropTypes.number,
+  flexGrow: PropTypes.bool,
+
+  /**
+   * Whether the column content should be aligned to the center. If not set to true, the content will be aligned to the left.
+   */
+  alignToCenter: PropTypes.bool,
```
cm9361, line 44: Maybe we should just allow the properties of the underlying style.  You could left, center, or right align, most likely.  I would expect the properly to be something like "align" and allow me to specify.

packages/terra-compact-interactive-list/src/proptypes/columnShape.js
```diff
@@ -5,22 +5,44 @@ const columnShape = PropTypes.shape({
...
    */
   maximumWidth: PropTypes.number,
+
   /**
-   * A number (in px) specifying the width of the column. If not provided, the Table's default column width will be used.
+   * Whether the column width can grow (a flex growing column).
+   * Defaults to false if the width is set and true if the width is not set.
    */
-  width: PropTypes.number,
+  flexGrow: PropTypes.bool,
```
cm9361, line 39: What is the use case for the flex growing column?  I thought the consumers were specifying the widths?

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * Columns maximum width in units set by widthUnit prop, such as `px`, `em`, or `rem`.
+   */
+  columnMaximumWidth: PropTypes.number,
+
+  /**
+   * The type of width value. One of `px`, `em`, `rem`. Defaults to 'px'.
+   */
+  widthUnit: PropTypes.oneOf([
```
cm9361, line 41: If you are going to allow column resizing, I think you must force consumers to use pixels.  Even if we can allow multiple units, I would not expect it to be separate from the width definition.  Usually there are not two separate props in CSS.

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
```diff
@@ -0,0 +1,95 @@
...
+    flex: isResponsive ? `1 1 ${width || columnMinimumWidth}${widthUnit}` : null,
+    width: isResponsive ? null : `${width}${widthUnit}`,
+    justifyContent: `${alignToCenter ? 'center' : 'left'}`,
+    maxWidth: isResponsive && `${Math.min(maximumWidth, columnMaximumWidth)}${widthUnit}`,
+    minWidth: isResponsive && `${Math.max(minimumWidth, columnMinimumWidth)}${widthUnit}`,
+  };
+
+  return (
+    <div
+      id={id}
```
cm9361, line 81: Does your Cell component really need to have an id attribute?  Your current implementation does not guarantee uniqueness and could lead to an invalid DOM.

cm9361:
I am not clear on the design or requirements for responsiveness for this control.  Usually, a control is responsive based on the size of the container element.

eawww:
The main thing I'm seeing is the source order of the items goes left-right, top-bottom. It should go top-bottom, left-right.

packages/terra-compact-interactive-list/src/proptypes/columnShape.js
adoroshk: renamed the prop to `align` with `center`, `left`, `right` options: https://github.com/cerner/terra-framework/pull/1869/commits/ae408c7c30b997ee75cf380978e4fdc763c5a240

packages/terra-compact-interactive-list/src/proptypes/columnShape.js
adoroshk: We have both, the `width` for cases when the consumers want to specify the exact width and `flexGrow` for cases when the column can grow. I understand it's confusing as here it looks like `width` was removed, but it is actually still there on line 19. Not sure why it's shown as removed here.

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
adoroshk: Per what we discussed earlier with @eawww  there are no plans to implement the resizing in this component, and that `em` and `rem` are required.

As for `I would not expect it to be separate from the width definition. Usually there are not two separate props in CSS` - totally agree on that, however, due to unique needs of this component, its layout is heavily based on width calculations. My concern is that user might put "100px" for one column and "1.2em" for another column, which would break all calculations. Introducing that prop I was trying to ensure the way all width units are the same across the component, but open for better ways to do so.

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
adoroshk: Removed in https://github.com/cerner/terra-framework/commit/ae408c7c30b997ee75cf380978e4fdc763c5a240

adoroshk:
> I am not clear on the design or requirements for responsiveness for this control. Usually, a control is responsive based on the size of the container element.

Currently there are 2 major options:
1. The customers set width to all columns, the component is not responsive, it it's bigger than its container the scroll shows up.
2. The customers omitted width or set the flexGrow prop for at least one column. That column will be responsive and grow to fill the container width or shrink to fit in it.

In option 2, providing the uniqueness of the list component, I can see the situations when the customers might want the columns to grow (or shrink), but not exceed some specific width. In that case the customers can add maximumWidth (minimumWidth) to that column. That would make the list responsive, but not taking 100% of the width. That would allow the list to be responsive, but not limited to the container width.

eawww:
@adoroshk that's correct but they're appearing in a different order in the dom. Going over it with a screen reader, without using any extra controls, it reads in the wrong order.
image

adoroshk:
> @adoroshk that's correct but they're appearing in a different order in the dom. Going over it with a screen reader, without using any extra controls, it reads in the wrong order. image

Per discussion with @eawww adding aria-rowindex to the rows would not solve the problem and the DOM order of items should be changed to reflect the column (top to bottom, then right to left) order of items inside the list. In order to break into columns with such DOM order without wrapping elements, the items have to be of known height.

It has been also discussed that that the placeholder rows (empty spaces at the end of the list) are allowed, but must not have a role (already implemented) and aria-hidden=true (will be added in the next commit).

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples/FixedWidthColumns.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import CompactInteractiveList, { alignTypes } from 'terra-compact-interactive-list';
+import rows from './rowsData';
```
eawww, line 3: In order to make the examples more useful, all relevant implementation information should be included in the example's jsx file. For components like this, the data fed into it is important so it should be included, even if that means repeating it across examples.

In this instance, the focus of the example is the columns so the row data may not be terribly important, but there should be at least one base example that includes the row data inline.

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples/rowsData.jsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { IconDocuments, IconFeaturedOutline, IconImage } from 'terra-icon';
+
+// Source rows data for tests
+const rows = [
+  {
+    id: 'row_1',
+    cells: [
+      { content: <IconDocuments height="1.5em" width="1.5em" /> },
```
eawww, line 9: The icons in this file are all meaningful, so they should make use of the `a11yLabel` prop.

eawww (commented):
Just a couple more things.

packages/terra-compact-interactive-list/src/subcomponents/Row.jsx
```diff
@@ -0,0 +1,153 @@
...
+
+  /**
+   * Row's width in units set by widthUnit prop, such as `px`, `em`, or `rem`.
+   * Disregarder if flexGrow is set to true.
+   */
+  rowWidth: PropTypes.number,
+
+  /**
+   * Row's maximum width for flex growing rows in units set by widthUnit prop, such as `px`, `em`, or `rem`.
+   * Will have no effect if the width is set and flexGrow is omitted or false.
```
nikhitasharma, line 57: Should this comment say 'if row width is set'?
Also there should be more information on where the flexGrow property reference is from.

packages/terra-compact-interactive-list/src/subcomponents/Row.jsx
```diff
@@ -0,0 +1,153 @@
...
+   */
+  rowMaximumWidth: PropTypes.number,
+
+  /**
+   * Row's maximum width for flex growing rows in units set by widthUnit prop, such as `px`, `em`, or `rem`.
+   * Will have no effect if the width is set and flexGrow is omitted or false.
```
nikhitasharma, line 63: Same here as above

packages/terra-compact-interactive-list/src/subcomponents/Row.jsx
```diff
@@ -0,0 +1,153 @@
...
+  /**
+   * The type of width value. One of `px`, `em`, `rem`. Defaults to 'px'.
+   */
+  widthUnit: PropTypes.oneOf([
+    widthUnitTypes.PX,
+    widthUnitTypes.EM,
+    widthUnitTypes.REM,
+  ]).isRequired,
+
+  flowHorizontally: PropTypes.bool,
```
nikhitasharma, line 76: This could use a description.

packages/terra-compact-interactive-list/tests/jest/CompactInteractiveList.test.jsx
```diff
@@ -13,12 +15,235 @@ afterAll(() => {
...
+      },
+      {
+        id: 'Column-2',
+        displayName: 'Col_3',
+        width: 40,
+      },
+    ];
 
-    expect(wrapper).toMatchSnapshot();
+    it('should apply default minimunWidth correctly', () => {
```
nikhitasharma, line 38:
```suggestion
    it('should apply default minimumWidth correctly', () => {
```

packages/terra-compact-interactive-list/tests/jest/CompactInteractiveList.test.jsx
```diff
@@ -13,12 +15,235 @@ afterAll(() => {
...
+          id="compact-interactive-list-id"
+          rows={rows}
+          columns={cols}
+        />,
+      );
+      const list = wrapper.find('.compact-interactive-list');
+      expect(list.props().style.minWidth).toEqual('500px');
+    });
+
+    it('should apply minimunWidth correctly', () => {
```
nikhitasharma, line 50:
```suggestion
    it('should apply minimumWidth correctly', () => {
```

packages/terra-compact-interactive-list/tests/jest/CompactInteractiveList.test.jsx
```diff
@@ -13,12 +15,235 @@ afterAll(() => {
...
+          id="compact-interactive-list-id"
+          rows={rows}
+          columns={cols}
+          numberOfColumns={numberOfColumns}
+          widthUnit="em"
+        />,
+      );
+      const list = wrapper.find('.compact-interactive-list');
+      // the width should be equal to the sum of all columns width multiplied by numberOfColumns
+      // the width units in this example should default to 'em'
```
nikhitasharma, line 100:
```suggestion
      // the width units in this example should be the passed in widthUnit prop value 'em'
```

packages/terra-compact-interactive-list/tests/jest/CompactInteractiveList.test.jsx
```diff
@@ -13,12 +15,235 @@ afterAll(() => {
...
+        width: 200,
+      },
+      {
+        id: 'Column-2',
+        displayName: 'Col_3',
+        width: 40,
+      },
+    ];
 
-    expect(wrapper).toMatchSnapshot();
```
nikhitasharma, line 22: I thought we are still matching snapshots for the default component render once in our tests.

packages/terra-compact-interactive-list/src/subcomponents/Cell.module.scss
```diff
@@ -0,0 +1,18 @@
+// Themes
+@import '../orion-fusion-theme/Cell.module';
+@import '../clinical-lowlight-theme/Cell.module';
+
+:local {
+  .cell {
+    align-items: center;
+    display: flex;
+    font-weight: normal;
```
nikhitasharma, line 9: Is the font-size of cell contents supposed to be same for all themes. They are usually smaller for fusion theme.

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
adoroshk: I removed widthUnit prop in https://github.com/cerner/terra-framework/pull/1869/commits/6f635dff9b8c8ee90cd59dfd9f5d3d61f8e42947. The customer will be allowed to pass all width props for columns (such as width, maxWidth, minWidth) with valid css string. I expect the customer pass all width props using the same width unit, the failed prop will be disregarded, which will not break the layout, just make the column flex growing. There will be also a console warning for that.

adoroshk:
I removed widthUnit prop in https://github.com/cerner/terra-framework/pull/1869/commits/6f635dff9b8c8ee90cd59dfd9f5d3d61f8e42947. The customer will be allowed to pass all width props for columns (such as width, maxWidth, minWidth) with valid css string. I expect the customer pass all width props using the same width unit, the failed prop will be disregarded, which will not break the layout, just make the column flex growing. There will be also a console warning for that.
This new functionality will need some more tests added.

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples/FixedWidthColumns.jsx
adoroshk: Fixed in https://github.com/cerner/terra-framework/pull/1869/commits/a60c1c52389d252309f3aeaed4b1de21ace9f470

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples/rowsData.jsx
adoroshk: Added in https://github.com/cerner/terra-framework/pull/1869/commits/a60c1c52389d252309f3aeaed4b1de21ace9f470

packages/terra-compact-interactive-list/src/utils/utils.js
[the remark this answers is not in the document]
adoroshk: I was following the pattern in DataGrid where it was called utils: https://github.com/cerner/terra-framework/tree/main/packages/terra-data-grid/src/utils. I also observed it in terrra-tables: https://github.com/cerner/terra-framework/tree/main/packages/terra-table/src/utils. I don't mind renaming, just curious if that already became another established pattern?

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
```diff
@@ -23,23 +32,146 @@ const propTypes = {
...
+    columnMaximumWidth,
   } = props;
 
   const theme = useContext(ThemeContext);
 
+  const columnMinWidth = columnMinimumWidth || DefaultListValues.columnMinimumWidth[widthUnit];
+  const columnMaxWidth = columnMaximumWidth;
+  // check if list has responsive columns
+  const isResponsive = checkIfRowHasResponsiveColumns(columns);
+  // if there are responsive columns, the items will need maxWidth and minWidth
```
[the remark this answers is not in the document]
adoroshk: Fixed in https://github.com/cerner/terra-framework/pull/1869/commits/6d4771b2bc54be722bb4c4efeca3d218192a205b

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
```diff
@@ -23,23 +32,146 @@ const propTypes = {
...
+  const rowMaxWidth = isResponsive ? getRowMaximumWidth(columns, columnMaxWidth) : null;
+  const rowMinWidth = isResponsive ? getRowMinimumWidth(columns, columnMinWidth) : null;
+  // calculate row width based on the width of its columns
+  const getRowWidthSum = (total, column) => total + column.width;
+  const rowWidth = isResponsive ? 0 : columns.reduce(getRowWidthSum, 0);
+  const rowsPerColumn = Math.ceil(rows.length / numberOfColumns);
+  const calculatedRowHeight = flowHorizontally ? null : rowHeight || DefaultListValues.rowDefaultHeight[widthUnit];
+  // calculate list width based on the item width and number of columns
```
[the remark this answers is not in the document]
adoroshk: Fixed in https://github.com/cerner/terra-framework/pull/1869/commits/6d4771b2bc54be722bb4c4efeca3d218192a205b

packages/terra-compact-interactive-list/src/subcomponents/Row.jsx
```diff
@@ -0,0 +1,153 @@
...
+   */
+  columns: PropTypes.arrayOf(columnShape).isRequired,
+
+  /**
+   * A number for column minimum width. Defaults to 60.
+   */
+  columnMinimumWidth: PropTypes.number,
+
+  /**
+   * A number for column minimum width.
```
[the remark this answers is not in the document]
adoroshk: Fixed in https://github.com/cerner/terra-framework/pull/1869/commits/6d4771b2bc54be722bb4c4efeca3d218192a205b

packages/terra-compact-interactive-list/src/subcomponents/Row.jsx
```diff
@@ -0,0 +1,153 @@
...
+   */
+  columnMaximumWidth: PropTypes.number,
+
+  /**
+   * A number of visual columns.
```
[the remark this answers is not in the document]
adoroshk: Fixed in https://github.com/cerner/terra-framework/pull/1869/commits/6d4771b2bc54be722bb4c4efeca3d218192a205b

packages/terra-compact-interactive-list/src/subcomponents/Row.jsx
```diff
@@ -0,0 +1,153 @@
...
+  numberOfColumns: PropTypes.number.isRequired,
+
+  /**
+   * Indicates if the column is flex growing column (can grow wider or shrink).
+   */
+  isResponsive: PropTypes.bool.isRequired,
+
+  /**
+   * Row's width in units set by widthUnit prop, such as `px`, `em`, or `rem`.
+   * Disregarder if flexGrow is set to true.
```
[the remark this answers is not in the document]
adoroshk: Fixed in https://github.com/cerner/terra-framework/pull/1869/commits/6d4771b2bc54be722bb4c4efeca3d218192a205b

packages/terra-compact-interactive-list/src/subcomponents/Row.jsx
adoroshk: Fixed in https://github.com/cerner/terra-framework/pull/1869/commits/6d4771b2bc54be722bb4c4efeca3d218192a205b

packages/terra-compact-interactive-list/src/subcomponents/Row.jsx
adoroshk: Fixed in https://github.com/cerner/terra-framework/pull/1869/commits/6d4771b2bc54be722bb4c4efeca3d218192a205b

packages/terra-compact-interactive-list/src/subcomponents/Row.jsx
```diff
@@ -0,0 +1,153 @@
...
+  rowWidth: PropTypes.number,
+
+  /**
+   * Row's maximum width for flex growing rows in units set by widthUnit prop, such as `px`, `em`, or `rem`.
+   * Will have no effect if the width is set and flexGrow is omitted or false.
+   */
+  rowMaximumWidth: PropTypes.number,
+
+  /**
+   * Row's maximum width for flex growing rows in units set by widthUnit prop, such as `px`, `em`, or `rem`.
```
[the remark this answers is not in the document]
adoroshk: Fixed in https://github.com/cerner/terra-framework/pull/1869/commits/6d4771b2bc54be722bb4c4efeca3d218192a205b

packages/terra-compact-interactive-list/src/subcomponents/Row.jsx
adoroshk: Fixed in https://github.com/cerner/terra-framework/pull/1869/commits/6d4771b2bc54be722bb4c4efeca3d218192a205b

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
```diff
@@ -23,23 +32,146 @@ const propTypes = {
...
+      for (let i = placeholdersNumber; i > 0; i -= 1) {
+        result.splice(position, 0, { id: `placeholder-row-${i}` });
+        position -= rowsPerColumn - 1;
+      }
+    }
+    return result;
+  };
+
+  const mappedRows = mapRows();
+  const checkIfLeftmost = (index) => (flowHorizontally ? index % numberOfColumns === 0 : index < rowsPerColumn);
```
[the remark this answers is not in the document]
adoroshk: Fixed in https://github.com/cerner/terra-framework/pull/1869/commits/6d4771b2bc54be722bb4c4efeca3d218192a205b

packages/terra-compact-interactive-list/tests/jest/CompactInteractiveList.test.jsx
adoroshk: No longer exists

packages/terra-compact-interactive-list/tests/jest/CompactInteractiveList.test.jsx
adoroshk: No longer exists

packages/terra-compact-interactive-list/tests/jest/CompactInteractiveList.test.jsx
adoroshk: No longer exists

packages/terra-compact-interactive-list/src/subcomponents/Cell.module.scss
adoroshk: Great catch, I am not sure how `font-weight` it appeared to be there.

packages/terra-compact-interactive-list/tests/jest/CompactInteractiveList.test.jsx
adoroshk: I still need to add some tests, including an initial snapshot. Will get back to this comment once I have an update.

eawww (approved):
Component is laid out as designed. Meets meaningful sequence requirements. Not evaluated for interactivity as this is only to serve as a base for further work.

packages/terra-compact-interactive-list/src/utils/utils.js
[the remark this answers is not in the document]
nikhitasharma: Sure, its fine then.

nikhitasharma (approved):
Looks good to me given some minor test updates are done.

JessieRandle (approved):

PR #1874, opened 2023-11-03
[terra-embedded-content-consumer] - Show focus indicator on iframe of the embedded content when content can scroll, and has no interactable element

kolkheang:
### Summary

- Show a visual focus indicator on the frame when the embedded content:

  - Does not have any interactable element such as link, button, form field, etc. inside
  - The iframe is scrollable
  - The content is scrollable (content height is bigger than client height)
- Also show a visual focus indicator on the frame when the frame has focus
- No visual focus indicator on the frame when there is any interactable element in the content in the frame.

This change allows keyboard only users to see where focus is, and use keyboard to scroll the embedded content when content is not fully visible within the view port.

**What was changed:**
- Show outline style on the iframe when it has focus, and remove the outline style when it's not in focus (blur)
- Set focus on the embedded content with `tabIndex=0` when the content doesn't have any interactable element, iframe is scrollable, and content is scrollable

**Why it was changed:**
It's necessary for keyboard only users to see where the focus is, and give users indication that while focus is on the frame, they can use arrow keys to scroll the content inside the frame when needed

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

Related PR (dependency of this PR):
https://github.com/cerner/xfc/pull/96

Screen Shots:

Screenshot 2023-11-03 at 8 49 54 PM
Screenshot 2023-11-03 at 8 49 44 PM

Screen recording of the new functionality using the examples on the dev site:
https://github.com/cerner/terra-framework/assets/3170314/e5c46b11-88c2-47f7-83ed-bf5b1668cd49

**This PR resolves:**
UXPLATFORM-9772

---

@cerner/terra


kolkheang:
Related PR https://github.com/cerner/xfc/pull/96

Pending change in the above PR, I will trigger a rebuild of this PR and flip this PR to review

kolkheang:
I'm still trying to figure out why the examples in the site doc of the PR build isn't working as expected.

kolkheang:
I'm still trying to figure out why my change isn't showing up in the PR build. Trying different things now to troubleshoot.

packages/terra-embedded-content-consumer/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Added
   * Added screenreader support to announce context of embedded iframe content.
+  * Added visual focus indicator on the iframe when the content doesn't have any inteactable element, iframe is scrollable, and content is scrollable for keyboard only users
```
RayGunY, line 7: nit: needs a period at the end

kolkheang: Fixed with https://github.com/cerner/terra-framework/pull/1874/commits/10ee5fc7493b71f8c90e1e097bc16610a5f9d3b2

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.module.scss
```diff
@@ -1,5 +1,13 @@
...
+@import './clinical-lowlight-theme/EmbeddedContentConsumer.module';
+@import './orion-fusion-theme/EmbeddedContentConsumer.module';
+
 :local {
   .visually-hidden-text {
     height: 0;
   }
+
+  .iframe-focus-style {
+    outline: var(--terra-embedded-content-consumer-focus-outline, 2px dashed #000);
```
cm9361, line 11: Don't you also need an outline-offset?  Did UX not specify one in this case?

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -123,6 +128,70 @@ class EmbeddedContentConsumer extends React.Component {
...
 
+  handleFrameVisualFocusIndicator() {
+    // Selectors for interactable elements
+    const interactableElementSelector = 'a[href]:not([tabindex=\'-1\']), area[href]:not([tabindex=\'-1\']), input:not([disabled]):not([tabindex=\'-1\']), '
+      + "select:not([disabled]):not([tabindex='-1']), textarea:not([disabled]):not([tabindex='-1']), button:not([disabled]):not([tabindex='-1']), "
+      + "[contentEditable=true]:not([tabindex='-1'])";
+
+    // Check if the content is scrollable
+    const isContentScrollable = () => {
+      const doc = this.xfcFrame?.iframe?.contentWindow?.document;
```
cm9361, line 139: Maybe we rename the variable to frameDocument?

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -123,6 +128,70 @@ class EmbeddedContentConsumer extends React.Component {
...
+      return (doc.documentElement.scrollHeight > doc.documentElement.clientHeight
+        || doc.body.scrollHeight > doc.body.clientHeight
+        || doc.documentElement.scrollWidth > doc.documentElement.clientWidth
+        || doc.body.scrollWidth > doc.body.clientWidth);
```
cm9361, line 143: It would probably be helpful to add a comment to explain this logic.  It looks like you are checking both the HTML element and body elements for scrolling.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -123,6 +128,70 @@ class EmbeddedContentConsumer extends React.Component {
...
+    };
+
+    // Check if the iframe has `scrolling` attribute set or not
+    // The default `scrolling` attribute is `auto`.
+    const scrollingEnabled = () => (this.xfcFrame?.iframe?.getAttribute('scrolling') !== 'no');
```
cm9361, line 148: Is there a reason that this should not be part of the isContentScrollable method?  Also, do we need this additional check.  If there is nothing to scroll, I would wonder if the setting matters.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -123,6 +128,70 @@ class EmbeddedContentConsumer extends React.Component {
...
+
+      if (scrollingEnabled() && isContentScrollable() && this.hasInteractableElement === false) {
+        // Set tabIndex="0" so focus can go into the document when
+        // using tab key when scrolling is enabled
+        this.xfcFrame.iframe.contentWindow.document.body.tabIndex = 0;
+      }
+    });
+
+    // Event listener and callback function for `resize` event of the iframe
+    this.xfcFrame?.iframe?.contentWindow?.addEventListener('resize', () => {
```
cm9361, line 168: We should store the content window reference in a local variable.  It is used quite a bit in the file and would make the logic look cleaner.  We also would not have to look up the reference every time.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -123,6 +128,70 @@ class EmbeddedContentConsumer extends React.Component {
...
+      } else if (this.xfcFrame?.iframe?.contentWindow?.document.body.getAttribute('tabIndex') === '0') {
+        this.xfcFrame.iframe.contentWindow.document.body.removeAttribute('tabIndex');
+      }
+    });
+
+    // Event listener and callback function for `focus` event is in the iframe
+    this.xfcFrame?.iframe?.contentWindow?.addEventListener('focus', () => {
+      if (this.hasInteractableElement === true || !isContentScrollable()) {
+        return;
+      }
```
cm9361, line 182: I don't think this block is necessary.  You will only apply the styles if the condition below it qualifies.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -123,6 +128,70 @@ class EmbeddedContentConsumer extends React.Component {
...
+
+      if (scrollingEnabled() && isContentScrollable() && this.hasInteractableElement === false) {
+        this.xfcFrame.iframe.className = cx('iframe-focus-style');
+      }
+    }, true);
+
+    // Event Listener and callback function for `blur` event in the iframe
+    this.xfcFrame?.iframe?.contentWindow?.addEventListener('blur', () => {
+      this.xfcFrame.iframe.removeAttribute('class');
```
cm9361, line 191: You wouldn't removed the entire class attribute.  You would only remove the class that you added.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -123,6 +128,70 @@ class EmbeddedContentConsumer extends React.Component {
...
+
+    // Event listener and callback function for `resize` event of the iframe
+    this.xfcFrame?.iframe?.contentWindow?.addEventListener('resize', () => {
+      if (scrollingEnabled() && isContentScrollable() && this.hasInteractableElement === false) {
+        // Set tabIndex="0" so focus can go into the document when
+        // using tab key when scrolling is enabled
+        this.xfcFrame.iframe.contentWindow.document.body.tabIndex = 0;
+      } else if (this.xfcFrame?.iframe?.contentWindow?.document.body.getAttribute('tabIndex') === '0') {
+        this.xfcFrame.iframe.contentWindow.document.body.removeAttribute('tabIndex');
+      }
```
cm9361, line 175: If the tabIndex has been set by the page owner, do we want to remove it?

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
kolkheang: Yeah, I was using that `classList.remove()` but that only removes the value and not the attribute `class`.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
kolkheang: Yes, we need this additional check because if this is set to 'no', then scrolling isn't available. The `isContentScrollable` changes based on the screen size and available display.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.module.scss
kolkheang: I wasn't aware of that `outline-offset`. I was not working with UX on this. @cm9361 do you know who I can reach out to? 

cc: @scottwilmarth

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
kolkheang: Let me update to use this method https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/remove instead. That seems safer than accidentally removing other keys.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
kolkheang: Good question... probably not. But, I'm not sure how to distinguish the value between us setting and removing and page owner setting it. If we don't remove it, the focus would till be there when the content is no longer scrollable due to bigger view port after the resize event.

cc: @scottwilmarth

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.module.scss
cm9361: We have been working with @eawww (Eric Wilson)

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
kolkheang: I've re-worked the logic to re-apply the previous value (or null when not set initially).

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
cm9361: So the scrollWidth is greater than the clientWidth even when scrolling is not available?

packages/terra-framework-docs/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 
 * Updated
   * Updated `terra-table` example to show providing custom masked cell message label.
+  * Updated examples, and tests to show the visual focus indicator on the iframe of the embedded content
```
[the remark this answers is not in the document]
kolkheang: Added

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
kolkheang: @cm9361 - I looked at `scrollHeight` and `clientHeight` instead since I've vertical scrolling enabled/disabled to check. And yes, even if scrolling is disabled for the iframe, the `scrollHeight` is larger than `clientHeight`.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -123,6 +128,70 @@ class EmbeddedContentConsumer extends React.Component {
     this.addEventListeners(this.props.eventHandlers);
   }
 
+  handleFrameVisualFocusIndicator() {
+    // Selectors for interactable elements
+    const interactableElementSelector = 'a[href]:not([tabindex=\'-1\']), area[href]:not([tabindex=\'-1\']), input:not([disabled]):not([tabindex=\'-1\']), '
```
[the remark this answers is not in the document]
kolkheang: I moved it down.

https://github.com/cerner/terra-framework/pull/1874/commits/b53056bf58a6fd60d158c1ff8ec4609a75d99f4a#diff-77b1c1b39ad95249bd3956ef509a0995b4687f8c94ba87a3110f3e6b1ce8ccfeR157-R160

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
kolkheang: https://github.com/cerner/terra-framework/pull/1874/commits/b53056bf58a6fd60d158c1ff8ec4609a75d99f4a#diff-77b1c1b39ad95249bd3956ef509a0995b4687f8c94ba87a3110f3e6b1ce8ccfeR180-R189

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
kolkheang: https://github.com/cerner/terra-framework/pull/1874/commits/b53056bf58a6fd60d158c1ff8ec4609a75d99f4a#diff-77b1c1b39ad95249bd3956ef509a0995b4687f8c94ba87a3110f3e6b1ce8ccfeR193-R203

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -123,6 +128,70 @@ class EmbeddedContentConsumer extends React.Component {
...
+      }
+    });
+
+    // Event listener and callback function for `focus` event is in the iframe
+    this.xfcFrame?.iframe?.contentWindow?.addEventListener('focus', () => {
+      if (this.hasInteractableElement === true || !isContentScrollable()) {
+        return;
+      }
+
+      if (scrollingEnabled() && isContentScrollable() && this.hasInteractableElement === false) {
```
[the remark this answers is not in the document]
kolkheang: Updated https://github.com/cerner/terra-framework/pull/1874/commits/b53056bf58a6fd60d158c1ff8ec4609a75d99f4a#diff-77b1c1b39ad95249bd3956ef509a0995b4687f8c94ba87a3110f3e6b1ce8ccfeR181

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
kolkheang: Removed https://github.com/cerner/terra-framework/pull/1874/commits/b53056bf58a6fd60d158c1ff8ec4609a75d99f4a#diff-77b1c1b39ad95249bd3956ef509a0995b4687f8c94ba87a3110f3e6b1ce8ccfeL180-L182

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -123,6 +128,70 @@ class EmbeddedContentConsumer extends React.Component {
...
+        // using tab key when scrolling is enabled
+        this.xfcFrame.iframe.contentWindow.document.body.tabIndex = 0;
+      } else if (this.xfcFrame?.iframe?.contentWindow?.document.body.getAttribute('tabIndex') === '0') {
+        this.xfcFrame.iframe.contentWindow.document.body.removeAttribute('tabIndex');
+      }
+    });
+
+    // Event listener and callback function for `focus` event is in the iframe
+    this.xfcFrame?.iframe?.contentWindow?.addEventListener('focus', () => {
+      if (this.hasInteractableElement === true || !isContentScrollable()) {
```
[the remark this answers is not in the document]
kolkheang: I removed that check

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
kolkheang: https://github.com/cerner/terra-framework/pull/1874/files#diff-77b1c1b39ad95249bd3956ef509a0995b4687f8c94ba87a3110f3e6b1ce8ccfeR136-R148

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
kolkheang: https://github.com/cerner/terra-framework/pull/1874/files#diff-77b1c1b39ad95249bd3956ef509a0995b4687f8c94ba87a3110f3e6b1ce8ccfeR143

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
cm9361: I think the final question then is if this logic should be part of the isContentScrollable method.  It seems like this would all be part of one method.  Is there a scenario where you would not call both?

kolkheang: Good question. Yes, we can combine the logic into one function.

eawww:
I noticed on the first example on the main doc page of the component that the content can receive keyboard focus even when the content doesn't scroll but does not show a focus indicator in Microsoft Edge. 

When a component can receive keyboard focus, it should always have a visual indicator. So this would either need logic to make non-scrolling iframes unable to be keyboard focused or provide a focus indicator even when content doesn't scroll.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.module.scss
kolkheang: I worked with @eawww on this and we've come up with something that would work.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -112,6 +112,9 @@ class EmbeddedContentConsumer extends React.Component {
     // Mount the provided source as the application into the content wrapper.
     this.xfcFrame = Consumer.mount(this.embeddedContentWrapper, this.props.src, frameOptions);
 
+    // Set additional style on xfcFrame
+    this.xfcFrame.iframe.classList.add(cx('xfc-iframe-margin'));
```
cm9361, line 116: Do you need a separate class just for the margin?  Or is this the class that you want for your iFrame?  You normally don't name your class based on styles.

kolkheang: This is to set the `margin-top` and `margin-bottom` for the default case (before focus) so that when we apply the style during focus, the content doesn't move. I think we need it. I'll try to rename it.

eawww (approved):
Component allows keyboard focus on iframe in browsers that do not provide that functionality by default, even when no interactable elements are present. Focus indicator is not shown when an element inside of it is focused. Screen reader still announces embedded content properly.

vinaybhargavar (approved):

RayGunY (approved):

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.module.scss
```diff
@@ -1,4 +1,20 @@
...
+
 :local {
+  .iframe-style {
+    margin-bottom: 4px;
+    margin-top: 4px;
+  }
+
+  .iframe-focus-style {
+    margin-bottom: 4px;
+    margin-top: 4px;
```
cm9361, line 13: If the margin is already part of the iframe-style, does it need to be duplicated in the focus style?

kolkheang: I just tested this and, to answer your question, no, we don't need this redundant style since the margin is already set as a default. Good catch!

cm9361 (approved):

kolkheang:
xfc version 1.12.0 is released here https://www.npmjs.com/package/xfc/v/1.12.0

kolkheang:
@sdadn - this PR is ready to be merged. 

After it is merged, I will issue another PR to update the path from relative to absolute for the pages embedded in the iframes used in the examples, and tests so that the examples will work when deployed under https://engineering.cerner.com/terra-framework/ and https://engineering.cerner.com/terra-ui sites.

sdadn (approved):

kolkheang:
The examples and tests are able to load fine with the relative paths:
- https://engineering.cerner.com/terra-framework/components/cerner-terra-framework-docs/embedded-content-consumer/embedded-content-consumer
- https://engineering.cerner.com/terra-framework/tests/cerner-terra-framework-docs/embedded-content-consumer/consumers/basic-consumer

I won't issue another PR to update the paths.

PR #1878, opened 2023-11-07
[terra-data=grid] Resize handle communication fix

cm9361:
### Summary

**What was changed:**
The resize handle was updated so that its ARIA label was not communicated when a non-header cell is focused.

**Why it was changed:**
The resize handle was being communicated when a non-header cell was focused.

### Testing

This change was tested using:

- [ ] WDIO
- [X] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [X] Accessibility review
- [X] Functional review

### Additional Details
N/A

**This PR resolves:**

UXPLATFORM-9803 

---

@cerner/terra


adoroshk (approved):

packages/terra-data-grid/src/subcomponents/ColumnResizeHandle.jsx
sdadn, line 1: Now that #1851 is merged, I think these changes should be done on the table subcomponents.

chrismichalewicz:
@cm9361  I tested this both for Functional and A11y, since the issue requires JAWS to test, with the scope strictly to just the issue you are addressing. 

Tested with Edge + JAWS

Keyboard navigation:
Navigated the column headers with arrow keys,
•	Resize handles only were communicated when they have focus
•	Column headers when in focus communicated without mention of the resize handle
•	Navigating to a non-header cell is communicated appropriately
 
Mouse:
•	Unable to recreate the issue, selecting a non header cell does not communicate the resize handle as it previously was. 
•	Noticed with a mouse that the new column widths aren't communicated, but this is also occuring on the Prod doc site so is not related to this work.  Will follow up outside this PR on this issue. 

cc: @mjpalazzo

sdadn (approved):

PR #1882, opened 2023-11-08
[flowsheet-data-grid] Hide columns headers

sdadn:
### Summary

This PR hides the column headers in the flowsheet-data-grid to meet functional requirements.

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9771 

---

@cerner/terra


packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -295,6 +295,7 @@ function FlowsheetDataGrid(props) {
         onCellSelect={handleCellSelection}
         onClearSelection={handleClearSelectedCells}
         onCellRangeSelect={handleCellRangeSelection}
+        hasColumnHeaders={false}
```
cm9361, line 298: The Flowsheet should have column headers by default.  The prop should be able to be set by consumers.

sdadn: Oh I misunderstood the requirements then. I'll change it.

chrismichalewicz:
@sdadn  Based on Charles's comment, and your response, looks like you are addressing this. But from a documentation/examples perspective I think we should have a specific example without column headers and then keep the others to have headers.

packages/terra-data-grid/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
 * Added
   * Added ability to select a range of cells to FlowsheetDataGrid.
 
+* Changed
+  * Updated FlowsheetDataGrid to no longer have column headers.
```
cm9361, line 14: I think this will now go under the Added section.  Also, you are allowing the column headers to be hidden.  However, it is still possible.  Maybe this update was missed based on your other updates in the PR.

packages/terra-framework-docs/src/terra-dev-site/test/data-grid/flowsheet-data-grid/ColumnHeadersHidden.jsx
```diff
@@ -0,0 +1,134 @@
...
+      ariaLabel="Flowsheet Data Grid"
+      onCellSelect={onCellSelect}
+      onClearSelectedCells={onClearSelectedCells}
+      onCellRangeSelect={onCellRangeSelect}
+      hasColumnHeaders={false}
+    />
+    );
+};
+
+export default DefaultFlowsheetDataGrid;
```
cm9361, line 134: The default example would not have the headers hidden, correct?

packages/terra-framework-docs/src/terra-dev-site/test/data-grid/flowsheet-data-grid/ColumnHeadersHidden.jsx
```diff
@@ -0,0 +1,134 @@
+import React, { useState, useCallback } from 'react';
+import { FlowsheetDataGrid } from 'terra-data-grid';
+
+const DefaultFlowsheetDataGrid = () => {
```
cm9361, line 4: I believe this should be ColumnHeadersHidden instead of DefaultFlowsheetDataGrid.

sdadn:
@eawww , @chrismichalewicz  I added the no header example, so it should now be ready for review.

chrismichalewicz:
> @eawww , @chrismichalewicz I added the no header example, so it should now be ready for review.
@sdadn  Did you update the right example?  
The "Columns Header Hidden" example is showing headers, it looks like the "No Results Cell" might have been updated by mistake as it shows with no headers.  

Also can "Columns Header Hidden" be updated to "Column Headers Hidden"?

packages/terra-framework-docs/src/terra-dev-site/doc/data-grid/FlowsheetDataGrid.4/Examples.6/ColumnHeadersHidden.jsx
```diff
@@ -0,0 +1,134 @@
+import React, { useState, useCallback } from 'react';
+import { FlowsheetDataGrid } from 'terra-data-grid';
+
+const DefaultFlowsheetDataGrid = () => {
```
cm9361, line 4: Shouldn't this be ColumnHeadersHidden instead of DefaultFlowsheetDataGrid?

sdadn:
@chrismichalewicz thanks for pointing that out, I duplicated the files and updated the wrong one 🤦‍♂️. It should be fixed now.

packages/terra-framework-docs/src/terra-dev-site/doc/data-grid/FlowsheetDataGrid.4/Examples.6/ColumnHeadersHidden.jsx
sdadn: Fixed with 4e96fbb

cm9361 (approved):

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -126,6 +126,11 @@ const propTypes = {
    * rendered to allow for row selection to occur.
    */
   hasSelectableRows: PropTypes.bool,
+
+  /**
+   * Boolean indicating whether or not the DataGrid should hide the column headers.
+   */
+  hasColumnHeaders: PropTypes.bool,
```
eawww, line 133: It has column headers. They're just invisible 🫥
```suggestion
  hasVisibleColumnHeaders: PropTypes.bool,
```

packages/terra-framework-docs/src/terra-dev-site/doc/data-grid/FlowsheetDataGrid.4/Examples.6/ColumnHeadersHidden.4.doc.mdx
```diff
@@ -0,0 +1,6 @@
+import ColumnHeadersHidden from './ColumnHeadersHidden?dev-site-example';
+
```
eawww, line 2: This should include a description of what this prop is used for and make sure consumers know they still have to provide column header labels, render their own column headers at the same widths as the columns, and ensure the two match.

The example itself should include that as well and there should be mention of it in the prop description.

packages/terra-framework-docs/src/terra-dev-site/doc/data-grid/FlowsheetDataGrid.4/Examples.6/ColumnHeadersHidden.jsx
```diff
@@ -0,0 +1,134 @@
+import React, { useState, useCallback } from 'react';
+import { FlowsheetDataGrid } from 'terra-data-grid';
+
+const ColumnHeadersHidden = () => {
+  const gridDataJSON = {
+    cols: [
+      { id: 'Column-0', displayName: 'Vitals', isSelectable: false },
```
eawww, line 7: Column headers shouldn't be selectable when hidden.
That could be done here by setting isSelectable to false on all the headers, but it would be better if it were a hard constraint of the component.

eawww (commented):
Functionally, it's great! 

Some documentation requests though.

packages/terra-framework-docs/src/terra-dev-site/doc/data-grid/FlowsheetDataGrid.4/Examples.6/ColumnHeadersHidden.jsx
chrismichalewicz: Agreed, in testing the example when in the first cell has focus I can up arrow once, right arrow, and then down arrow and the focus is in the cell to the right (however many time I hit the right arrow) I wouldn't expect you'd be able to navigate to and through them.

chrismichalewicz:
@sdadn  Not sure if it's related to the isSelectable observation @eawww  made but in testing with JAWS in Edge the column headers are also being communicated when tabbing into the table.

sdadn:
> @sdadn Not sure if it's related to the isSelectable observation @eawww made but in testing with JAWS in Edge the column headers are also being communicated when tabbing into the table.

@chrismichalewicz I believe that is intended because the implementation to hide column headers only makes them invisible but doesn't remove them from the DOM completely. @cm9361 / @eawww can correct me if I'm wrong.

chrismichalewicz:
> > @sdadn Not sure if it's related to the isSelectable observation @eawww made but in testing with JAWS in Edge the column headers are also being communicated when tabbing into the table.
> 
> @chrismichalewicz I believe that is intended because the implementation to hide column headers only makes them invisible but doesn't remove them from the DOM completely. @cm9361 / @eawww can correct me if I'm wrong.

We discussed this and this is working as expected so there are no updates needed for this specific comment. +1

packages/terra-framework-docs/src/terra-dev-site/doc/data-grid/FlowsheetDataGrid.4/Examples.6/ColumnHeadersHidden.jsx
sdadn: Fixed here: 3dfb79bf7ca0dbd76593a7dcd3cea5889a98b113

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -376,6 +385,7 @@ const DataGrid = injectIntl((props) => {
     const key = event.keyCode;
     switch (key) {
       case KeyCode.KEY_UP:
+        if (!hasVisibleColumnHeaders && nextRow - 1 === 0) break;
```
cm9361, line 388: I would think you could move this logic to this block of code
    if (nextCol < 0 || nextRow < 0) {
      event.preventDefault(); // prevent the page from moving with the arrow keys.
      return;
    }

Your current implementation does not account for preventing the default behavior so that the page does not scroll.  The basis of that logic is already lower in the file and checking for the top boundary.  You would just have to modify the condition to also handle the hidden column header use case.

packages/terra-framework-docs/src/terra-dev-site/doc/data-grid/FlowsheetDataGrid.4/Examples.6/ColumnHeadersHidden.4.doc.mdx
```diff
@@ -0,0 +1,21 @@
...
+### Usage
+The `hasVisibleColumns` is a boolean. By default, it is set to `true`.
+When it is set to `false`, the column headers are hidden, but still exist in the DOM and can be read by screenreaders when providing context for the table.
+This is useful when needing to use custom alternative to column headers, such as Stella Timeline.
+
+**Note:**
+- Column header labels should still be provided in the dataset to ensure that sceenreaders will still read out the appropriate context for the table.
+- When using custom column headers, then logic must be added to ensure that the column widths match are in sync if resizing columns are enabled.
+The [columns.width](/components/cerner-terra-framework-docs/data-grid/flowsheet-data-grid/about#column) property for columns can be used to set the column widths programatically.
```
cm9361, line 16: It seems like this should be on the About page instead of an example page.  You would not want consumers to miss this information because they did not select one of the examples.

sdadn: For now, I'm just adding the necessary information. It can be reorganized in the implementation guide story which will be updating the documentation as a whole and making sure all pieces fit together.

packages/terra-data-grid/src/DataGrid.jsx
sdadn: Updated here: 766fbf53f940f41b8d382040e42f2f83feccf791

eawww (approved):

cm9361 (approved):

PR #1885, opened 2023-11-09
Terra Table Resize Fixes

kenk2:
### Summary

**What was changed:**
Added tests for column resizing in table. Addressed an issue of column resizing not announcing aria values when in non-grid contexts.

**Why it was changed:**
To fix some potential aria issues and to test the behavior of resizing columns for tables.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-XXXX 

---

@cerner/terra


packages/terra-table/src/subcomponents/ColumnResizeHandle.jsx
```diff
@@ -138,6 +138,11 @@ const ColumnResizeHandle = (props) => {
     resizeHandleRef.current.style.height = `${height}px`;
   };
 
+  const handleFocus = () => {
+    setIsActive(true);
```
cm9361, line 142: Why do we need add this change now?

packages/terra-table/tests/jest/ColumnResizeHandle.test.jsx
```diff
@@ -105,7 +105,7 @@ describe('ColumnResizeHandle', () => {
   it('sets the appropriate prop values on space keydown', () => {
     const wrapper = mountWithIntl(
       <ColumnContext.Provider value={{ setColumnHeaderAriaLiveMessage: jest.fn() }}>
-        <ColumnResizeHandle />
+        <ColumnResizeHandle setIsActive={jest.fn()} />
```
cm9361, line 108: We are adding a new prop to pass a function, but it does not seem like we are testing to ensure it was called.  What are we attempting to validate?

cm9361:
You don't have a JIRA in your PR template.

packages/terra-framework-docs/CHANGELOG.md
sdadn, line 1: There have been recent changes in main, so make sure that the CHANGELOG entries are in the right section.

eawww (approved):

packages/terra-framework-docs/src/terra-dev-site/doc/table/Examples/ResizableColumnsTable.jsx
cm9361, line 1: Don't we already have example for column resizing?  We should not need a new example to show this functionality.

packages/terra-framework-docs/src/terra-dev-site/test/table/ResizableColumnsTable.test.jsx
cm9361, line 1: We already have tests for column resizing.  We should not need a new test file.

packages/terra-table/tests/jest/ColumnResizeHandle.test.jsx
```diff
@@ -15,6 +15,11 @@ afterAll(() => {
 });
 
 describe('ColumnResizeHandle', () => {
+  const mockSetIsActive = jest.fn();
```
cm9361, line 18: Why is this not just defined in the tests that require it?  This does not seem like something we should have to validate every test.

packages/terra-table/src/subcomponents/ColumnResizeHandle.jsx
```diff
@@ -108,6 +108,8 @@ const ColumnResizeHandle = (props) => {
 
   // Ref variable for native resize handle element
   const resizeHandleRef = useRef();
+  // Ref variable to prevent active state on mouse drag
+  const handleFocus = useRef(true);
```
cm9361, line 112: Can't we use the isActive prop?  It seems like that would work without this extra logic.

cm9361: In that scenario, we don't have the isActive prop that we can leverage.  Also, focus is being given to the DataGrid externally.  Why is the isActive prop not sufficient would be my question.

mjpalazzo:
@kenk2 - I tried validating the resizable columns example with JAWS, but the keyboard navigation was not working for me. I could Tab to the column resizing tool. JAWS announced "Press Enter to use. Patient column header, column 1 row" I pressed Enter and I no longer had focus on the resizing tool. Focus seemed to go to the frame around the table.

sdadn (approved):

PR #1888, opened 2023-11-14
[terra-folder-tree] Base Component Creation

sycombs:
### Summary

The PR adds the terra-folder-tree base component, which displays items in a hierarchical folder structure. Keyboard navigation and expand/collapse functionality are not included in this PR and will be implemented as part of a future Jira.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9679 

---

@cerner/terra


packages/terra-folder-tree/src/FolderTree.jsx
```diff
@@ -1,26 +1,50 @@
...
+      {children}
+    </ul>
+  </>
 );
 
 FolderTree.propTypes = propTypes;
 FolderTree.defaultProps = defaultProps;
 
-export default injectIntl(FolderTree);
+FolderTree.Item = FolderTreeItem;
```
sdadn, line 49: Let's create an index.js file to do the exports.

packages/terra-folder-tree/tests/jest/FolderTree.test.jsx
sdadn, line 1: +1 on the unit tests

packages/terra-folder-tree/src/FolderTree.jsx
```diff
@@ -1,26 +1,50 @@
...
+
+const cx = classNames.bind(styles);
 
 const propTypes = {
   /**
-   * @private
-   * The intl object containing translations. This is retrieved from the context automatically by injectIntl.
+   * List of FolderTree.Items to be displayed as content within the FolderTree.
+   */
+  children: PropTypes.node.isRequired,
```
sdadn, line 16: Will there be any issues to making this required? What if a consumer does `<FolderTree><FolderTree/>` and dynamically adds the children later?

JessieRandle (approved):

packages/terra-folder-tree/src/FolderTree.jsx
```diff
@@ -1,26 +1,50 @@
...
+  /**
+   * The title of the folder tree.
    */
-  intl: PropTypes.shape({ formatMessage: PropTypes.func }).isRequired,
+  title: PropTypes.string.isRequired,
+  /**
+   * The heading level for the title of the folder tree.
+   */
+  headerLevel: PropTypes.number,
```
cm9361, line 24: Is this a value that we want to be configurable or should it be a constant value for all FolderTree components.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -0,0 +1,122 @@
...
+        role="treeitem"
+        aria-selected={isSelected}
+      >
+        <input
+          type="radio"
+          checked={isSelected}
+          onClick={onClick}
+        />
+        {/* eslint-disable-next-line react/forbid-dom-props */}
+        <span style={{ paddingLeft: `${level}rem` }}>
```
cm9361, line 102: Why are we adding a span element around the Terra Arrange?  Are we trying to use this like the Spacer component?

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.module.scss
```diff
@@ -0,0 +1,30 @@
...
+
+:local {
+  .folder-tree-item {
+    align-items: center;
+    background-color: var(--terra-folder-tree-item-background-color, #fff);
+    border-bottom: var(--terra-folder-tree-item-border-bottom, 1px solid #dedfe0);
+    display: flex;
+    margin: 0;
+    min-height: var(--terra-folder-tree-item-min-height, 2.92857rem);
+    overflow-wrap: anywhere;
```
cm9361, line 12: We would want this to be "break-word".  I don't think we want to use anywhere as our option.

https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.module.scss
```diff
@@ -0,0 +1,30 @@
+@import '../clinical-lowlight-theme/FolderTree.module';
+@import '../orion-fusion-theme/FolderTree.module';
+
+:local {
+  .folder-tree-item {
+    align-items: center;
+    background-color: var(--terra-folder-tree-item-background-color, #fff);
```
cm9361, line 7: I think it is better to use background instead of background-color.  The background style would allow for both colors and gradients without a future style change.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.module.scss
```diff
@@ -0,0 +1,30 @@
...
+  
+    &:hover {
+      background-color: var(--terra-folder-tree-item-hover-background-color, #f4fafe);
+    }
+  }
+  
+  .selected {
+    background-color: var(--terra-folder-tree-item-selected-background-color, darken(#ebf6fd, 2%));
+  }
+
```
cm9361, line 24: What about the focus state?

packages/terra-folder-tree/src/orion-fusion-theme/FolderTree.module.scss
```diff
@@ -1,3 +1,6 @@
-// :local {
-//   .orion-fusion-theme {}
-// }
+:local {
+  .orion-fusion-theme {
+    --terra-folder-tree-item-background-color-hover: #f4fafe;
+    // --terra-folder-tree-item-background-color-selected: ;
```
cm9361, line 4: Looks like you might need to remove this commented code.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -0,0 +1,122 @@
...
+          <Arrange
+            fitStart={(
+              <Spacer paddingLeft="medium" paddingRight="medium" isInlineBlock>
+                {itemIcon}
+              </Spacer>
+            )}
+            fill={label}
+            alignFitStart="center"
+          />
+        </span>
```
cm9361, line 112: Instead of adding this complexity.  Can't we just use the listStyleImage property to set the image?  Also, are we able to leverage list-style-position to assist with the indentation?  Can we maybe limit the custom styling necessary?

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -0,0 +1,122 @@
...
+
+  return (
+    <>
+      <li
+        className={itemClassNames}
+        role="treeitem"
+        aria-selected={isSelected}
+      >
+        <input
+          type="radio"
```
eawww, line 97: These shouldn't be radio buttons since they can't be properly associated with each other and we'll be using aria-selected instead.

packages/terra-folder-tree/src/FolderTree.jsx
```diff
@@ -1,26 +1,50 @@
...
 
-const FolderTree = () => (
-  <div>test</div>
+const FolderTree = ({ children, title, headerLevel }) => (
+  <>
+    <ActionHeader
+      text={title}
+      level={headerLevel}
+    />
+    <ul
```
eawww, line 37: This needs an aria-labeledby pointing to the header above it or an aria-label with the same text.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * The label for the folder tree item.
+   */
+  label: PropTypes.string.isRequired,
+  /**
+   * The icon to display to the left of the name.
+   */
+  icon: PropTypes.element,
+  /**
+   * List of FolderTree.Items to display in a subfolder when this FolderTreeItem is clicked.
```
eawww, line 25:
```suggestion
   * List of FolderTree.Items to display in a subfolder when this FolderTreeItem is clicked. This makes the item expandable and collapsible. It also makes the item's icon a Folder, overriding anything given as the icon prop.
```

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -0,0 +1,122 @@
...
+   */
+  subfolderItems: PropTypes.arrayOf(PropTypes.element),
+  /**
+   * Whether or not the item is selected.
```
eawww, line 29:
```suggestion
   * Whether or not the item is selected. Since this component has the appearance of a radio button group, only one item should be selected at a time.
```

nikhitasharma (approved):
Approving. The remaining review comments will be addressed in a follow-up PR.

packages/terra-folder-tree/src/FolderTree.jsx
sycombs: My thought in marking this as required was that without any FolderTreeItems, it's really not a folder tree at all but a floating header. But it doesn't really hurt to make this optional since it's up to consumers what content to put in the tree.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.module.scss
sycombs: With `break-word`, long strings without spaces do not wrap which is why I used `anywhere`, do you know of a way to handle that with `break-word`?

image

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.module.scss
sycombs: The focus state should be added with keyboard navigation (UXPLATFORM-9776).

packages/terra-folder-tree/src/FolderTree.jsx
sycombs: Talked with @eawww, we want to keep the header level configurable since we don't know where on a page FolderTree will be used so consumers will need the option to set custom heading levels.

PR #1890, opened 2023-11-14
[data-grid] Retain Row and Column ID's when rows and columns are removed

kenk2:
### Summary

**What was changed:**
The Row ID is now saved whenever a user interacts with the data grid. 

**Why it was changed:**
To fix errors with accessing indexes that are no longer valid when filtering or deleting rows from the data grid.

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9815 

---

@cerner/terra

### DIFF
--- a/packages/terra-table/src/subcomponents/ColumnHeader.jsx
+++ b/packages/terra-table/src/subcomponents/ColumnHeader.jsx
@@ -84,6 +84,7 @@ const ColumnHeader = (props) => {
   return (
     <thead>
       <tr
+        aria-rowindex={1}
         data-row-id={`${tableId}-header-row`}
         className={cx('column-header-row', { hidden: !hasVisibleColumnHeaders })}
         height={hasVisibleColumnHeaders ? headerHeight : undefined}
--- a/packages/terra-table/tests/jest/ColumnHeader.test.jsx
+++ b/packages/terra-table/tests/jest/ColumnHeader.test.jsx
@@ -27,6 +27,7 @@ describe('ColumnHeader', () => {
     expect(columnHeader).toHaveLength(1);
     expect(columnHeader.props().height).toBe('3rem');
     expect(columnHeader.props()['data-row-id']).toBe('test-table-header-row');
+    expect(columnHeader.props()['aria-rowindex']).toBe(1);
 
     //  Validate ColumnHeaderCell React component
     const columnHeaderCells = columnHeader.find(ColumnHeaderCell);
@@ -59,6 +60,7 @@ describe('ColumnHeader', () => {
     expect(columnHeader).toHaveLength(1);
     expect(columnHeader.props().height).toBe('3rem');
     expect(columnHeader.props()['data-row-id']).toBe('test-table-header-row');
+    expect(columnHeader.props()['aria-rowindex']).toBe(1);
 
     // Validate ColumnHeaderCell React component
     const columnHeaderCells = columnHeader.find(ColumnHeaderCell);
@@ -89,6 +91,8 @@ describe('ColumnHeader', () => {
     const columnHeader = wrapper.find('.column-header-row');
     expect(columnHeader).toHaveLength(1);
     expect(columnHeader.props().height).toBe('3rem');
+    expect(columnHeader.props()['data-row-id']).toBe('test-table-header-row');
+    expect(columnHeader.props()['aria-rowindex']).toBe(1);
 
     // Validate ColumnHeaderCell React component
     const columnHeaderCells = columnHeader.find(ColumnHeaderCell);
@@ -126,6 +130,7 @@ describe('ColumnHeader', () => {
     expect(columnHeader).toHaveLength(1);
     expect(columnHeader.props().height).toBe('3rem');
     expect(columnHeader.props()['data-row-id']).toBe('test-table-header-row');
+    expect(columnHeader.props()['aria-rowindex']).toBe(1);
 
     // Validate ColumnHeaderCell React component
     const columnHeaderCell = columnHeader.find(ColumnHeaderCell).first();
@@ -159,6 +164,7 @@ describe('ColumnHeader', () => {
     const columnHeader = wrapper.find('.hidden');
     expect(columnHeader).toHaveLength(1);
     expect(columnHeader.props()['data-row-id']).toBe('test-table-header-row');
+    expect(columnHeader.props()['aria-rowindex']).toBe(1);
 
     expect(wrapper).toMatchSnapshot();
   });
--- a/packages/terra-data-grid/tests/jest/__snapshots__/DataGrid.test.jsx.snap
+++ b/packages/terra-data-grid/tests/jest/__snapshots__/DataGrid.test.jsx.snap
@@ -505,6 +505,7 @@ exports[`DataGrid verifies onCellSelect callback is triggered when space is pres
             >
               <thead>
                 <tr
+                  aria-rowindex={1}
                   className="column-header-row"
                   data-row-id="test-terra-data-grid-table-header-row"
                   height="2.5rem"
@@ -3152,6 +3153,7 @@ exports[`DataGrid verifies onCellSelect callback is triggered when space is pres
             >
               <thead>
                 <tr
+                  aria-rowindex={1}
                   className="column-header-row"
                   data-row-id="test-terra-data-grid-table-header-row"
                   height="2.5rem"
@@ -5810,6 +5812,7 @@ exports[`DataGrid verifies row selection column header selection 1`] = `
             >
               <thead>
                 <tr
+                  aria-rowindex={1}
                   className="column-header-row"
                   data-row-id="test-terra-data-grid-table-header-row"
                   height="2.5rem"
@@ -9193,6 +9196,7 @@ exports[`DataGrid verifies that the grid created is consistent with the rows and
             >
               <thead>
                 <tr
+                  aria-rowindex={1}
                   className="column-header-row"
                   data-row-id="test-terra-data-grid-table-header-row"
                   height="2.5rem"
@@ -11808,6 +11812,7 @@ exports[`DataGrid verifies the rows are created with the right props 1`] = `
             >
               <thead>
                 <tr
+                  aria-rowindex={1}
                   className="column-header-row"
                   data-row-id="test-terra-data-grid-table-header-row"
                   height="2.5rem"
--- a/packages/terra-data-grid/tests/jest/__snapshots__/WorklistDataGrid.test.jsx.snap
+++ b/packages/terra-data-grid/tests/jest/__snapshots__/WorklistDataGrid.test.jsx.snap
@@ -613,6 +613,7 @@ exports[`Row selection verifies Shift+Click enables row selection when row selec
                 >
                   <thead>
                     <tr
+                      aria-rowindex={1}
                       className="column-header-row"
                       data-row-id="test-terra-worklist-data-grid-table-header-row"
                       height="2.5rem"
@@ -3400,6 +3401,7 @@ exports[`Row selection verifies Shift+Click establishes anchor and selects row w
                 >
                   <thead>
                     <tr
+                      aria-rowindex={1}
                       className="column-header-row"
                       data-row-id="test-terra-worklist-data-grid-table-header-row"
                       height="2.5rem"
@@ -6946,6 +6948,7 @@ exports[`Row selection verifies Shift+Click honors the anchor established by row
                 >
                   <thead>
                     <tr
+                      aria-rowindex={1}
                       className="column-header-row"
                       data-row-id="test-terra-worklist-data-grid-table-header-row"
                       height="2.5rem"
@@ -10462,6 +10465,7 @@ exports[`Row selection verifies Shift+Space enables row selection when row selec
                 >
                   <thead>
                     <tr
+                      aria-rowindex={1}
                       className="column-header-row"
                       data-row-id="test-terra-worklist-data-grid-table-header-row"
                       height="2.5rem"
@@ -13249,6 +13253,7 @@ exports[`Row selection verifies Shift+Space establishes anchor and selects row i
                 >
                   <thead>
                     <tr
+                      aria-rowindex={1}
                       className="column-header-row"
                       data-row-id="test-terra-worklist-data-grid-table-header-row"
                       height="2.5rem"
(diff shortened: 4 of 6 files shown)


packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -279,6 +280,7 @@ const DataGrid = injectIntl((props) => {
     setFocusedCol(columnIndex);
 
     if (onColumnSelect) {
+      setLastActiveRowId('');
```
cm9361, line 283: The column header is a valid row as well. I would not expect us to have an empty value because we selected a column header.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -476,7 +484,19 @@ const DataGrid = injectIntl((props) => {
     if (!event.currentTarget.contains(event.relatedTarget)) {
       // Not triggered when swapping focus between children
       if (handleFocus.current) {
-        setFocusedRowCol(focusedRow, focusedCol, true);
+        // When modifying the number of rows, we set the focused cell to the last selected RowId and columnId
+        if (lastActiveRowId) {
```
cm9361, line 488: You have logic for the rows, but there does not seem to be logic to ensure that the column is valid.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -459,6 +463,10 @@ const DataGrid = injectIntl((props) => {
       event.preventDefault(); // prevent the page from moving with the arrow keys.
       return;
     }
+
+    const newRowId = nextRow === 0 ? undefined : dataGridRows[nextRow - 1]?.id;
```
cm9361, line 467: I don't think this logic will work with sections.  You might have to get the row identifier from the table row element.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -459,6 +463,10 @@ const DataGrid = injectIntl((props) => {
...
+    setLastActiveRowId(newRowId);
```
cm9361, line 468: Shouldn't we be setting this value where we set the focused row and column?  I wouldn't expect them to be separated.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -181,6 +181,7 @@ const DataGrid = injectIntl((props) => {
   const [checkResizable, setCheckResizable] = useState(false);
   const [focusedRow, setFocusedRow] = useState(0);
   const [focusedCol, setFocusedCol] = useState(0);
+  const [lastActiveRowId, setLastActiveRowId] = useState();
```
cm9361, line 184: I think that this could be a reference variable.

packages/terra-framework-docs/src/terra-dev-site/test/data-grid/worklist-data-grid/WorklistDataGridDeleteRows.test.jsx
```diff
@@ -0,0 +1,49 @@
...
+  const onRowSelectAll = useCallback(() => {
+    const newRowData = rowData.map(row => ({ ...row, isSelected: true }));
+    setRowData(newRowData);
+  }, [rowData]);
+
+  return (
+    <div>
+      <div>
+        <button type="button" id="delete-rows-button" onClick={() => setRowData(rowData.filter(row => !row.isSelected))}>Delete Rows</button>
+        <button type="button" id="filter-rows-button" onClick={() => setRowData(rowData.filter(row => row.isSelected))}>Filter Rows</button>
```
cm9361, line 33: I am not clear on how this is a "Filter" button.  It seems like you should only need the one delete button.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -178,6 +178,9 @@ const DataGrid = injectIntl((props) => {
   const tableContainerRef = useRef();
   const handleFocus = useRef(true);
 
+  const lastActiveRowId = useRef();
+  const lastActiveColumnId = useRef();
```
cm9361, line 182: Since the is a reference variable, I think you can just create one single object with the two properties.  Also, I still think we should use "focused" instead of "active".  They have two different meanings in CSS.  It would also be easier to follow if the naming conventions were the same.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -196,15 +199,18 @@ const DataGrid = injectIntl((props) => {
...
   // functions
 
-  const isRowSelectionCell = (columnIndex) => (
+  const isRowSelectionCell = useCallback((columnIndex) => (
     hasSelectableRows && columnIndex < displayedColumns.length && displayedColumns[columnIndex].id === WorklistDataGridUtils.ROW_SELECTION_COLUMN.id
-  );
+  ), [displayedColumns, hasSelectableRows]);
 
-  const setFocusedRowCol = (newRowIndex, newColIndex, makeActiveElement) => {
+  const setFocusedRowCol = useCallback((newRowIndex, newColIndex, makeActiveElement) => {
```
cm9361, line 206: Why do we need to add a useCallback hook here?

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -196,15 +199,18 @@ const DataGrid = injectIntl((props) => {
   // -------------------------------------
   // functions
 
-  const isRowSelectionCell = (columnIndex) => (
+  const isRowSelectionCell = useCallback((columnIndex) => (
```
cm9361, line 202: I would not expect us to need a useCallback for such a simple function.

packages/terra-data-grid/tests/wdio/worklist-data-grid-spec.js
```diff
@@ -396,6 +396,45 @@ Terra.describeViewports('WorklistDataGrid', ['medium', 'large'], () => {
     });
   });
 
+  describe('row deletion', () => {
```
cm9361, line 399: I would think that you could test most of these scenarios with jest.  Do we really need as many WDIO tests?

packages/terra-table/src/subcomponents/ColumnHeader.jsx
```diff
@@ -86,6 +86,7 @@ const ColumnHeader = (props) => {
       <tr
         aria-rowindex={1}
         className={cx('column-header-row', { hidden: !hasColumnHeaders })}
+        data-row-id="table-header-row"
```
cm9361, line 89: Would not be a correct row id.  If you had two Tables on the page, you would have the same id and the DOM would be broken.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -476,7 +482,23 @@ const DataGrid = injectIntl((props) => {
     if (!event.currentTarget.contains(event.relatedTarget)) {
       // Not triggered when swapping focus between children
       if (handleFocus.current) {
-        setFocusedRowCol(focusedRow, focusedCol, true);
+        // When modifying the number of rows, we set the focused cell to the last selected RowId and columnId
+        if (lastActiveRowId.current && lastActiveColumnId.current) {
```
cm9361, line 486: I would think that you would deal with the row and column separately.  I am not sure we need both to exist to operate on one.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -476,7 +482,23 @@ const DataGrid = injectIntl((props) => {
...
+          let newRowIndex = dataGridRows.findIndex(row => row.id === lastActiveRowId.current);
+          let newColumnIndex = displayedColumns.findIndex(col => col.id === lastActiveColumnId.current);
```
cm9361, line 488: Can we spell out "column" to avoid unnecessary abbreviations?

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -276,28 +282,27 @@ const DataGrid = injectIntl((props) => {
 
   const handleColumnSelect = useCallback((columnId) => {
     const columnIndex = displayedColumns.findIndex(column => column.id === columnId);
-    setFocusedCol(columnIndex);
+    setFocusedRowCol(focusedRow, columnIndex);
```
cm9361, line 285: For a column selection, the row would be 0, right?

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -476,7 +482,23 @@ const DataGrid = injectIntl((props) => {
     if (!event.currentTarget.contains(event.relatedTarget)) {
       // Not triggered when swapping focus between children
       if (handleFocus.current) {
-        setFocusedRowCol(focusedRow, focusedCol, true);
+        // When modifying the number of rows, we set the focused cell to the last selected RowId and columnId
+        if (lastActiveRowId.current && lastActiveColumnId.current) {
+          let newRowIndex = dataGridRows.findIndex(row => row.id === lastActiveRowId.current);
+          let newColumnIndex = displayedColumns.findIndex(col => col.id === lastActiveColumnId.current);
+
+          newRowIndex = newRowIndex === -1
```
cm9361, line 490: You should add small comments to your logical blocks of code.

cm9361:
How are you testing your column boundary conditions?

packages/terra-data-grid/tests/wdio/worklist-data-grid-spec.js
```diff
@@ -396,6 +396,54 @@ Terra.describeViewports('WorklistDataGrid', ['medium', 'large'], () => {
...
+
+    it('retains the last selected row index when filtering rows', () => {
+      clickCell(2, 1, deleteRowsSelector);
+      clickCell(8, 1, deleteRowsSelector);
+      clickCell(5, 1, deleteRowsSelector);
+
+      browser.$('#filter-rows-button').click();
+      browser.keys(['Tab']);
+
+      expect(browser.$('tbody tr:nth-child(2) th').isFocused()).toBe(true);
```
nikhitasharma, line 413: @kenk2  Are we intentionally not doing the snapshot matches for these tests?

nikhitasharma: That is fine, since visual validation may not be important in this case.

nikhitasharma (approved):

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -195,6 +195,8 @@ const DataGrid = forwardRef((props, ref) => {
   const tableContainerRef = useRef();
   const handleFocus = useRef(true);
 
+  const lastFocusedIds = useRef({ rowId: '', columnId: '' });
```
cm9361, line 198: I would just name this variable, focusedCell.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -495,7 +502,27 @@ const DataGrid = forwardRef((props, ref) => {
     if (!event.currentTarget.contains(event.relatedTarget)) {
       // Not triggered when swapping focus between children
       if (handleFocus.current) {
-        setFocusedRowCol(focusedRow, focusedCol, true);
+        let newRowIndex = focusedRow;
+        let newColumnIndex = focusedCol;
+        const headerOffset = hasVisibleColumnHeaders ? 1 : 0;
```
cm9361, line 507: I would not expect us to need a headerOffset.  Even though the headers are not visible, they are still part of the table.  Do you have a unit test for this scenario?

packages/terra-table/src/subcomponents/ColumnHeader.jsx
```diff
@@ -84,6 +84,8 @@ const ColumnHeader = (props) => {
   return (
     <thead>
       <tr
+        aria-rowindex={1}
+        data-row-id={`${tableId}-header-row`}
```
cm9361, line 88: There should be Jest updates based on this change.  You should be testing that these value exist, as expected.

packages/terra-framework-docs/src/terra-dev-site/test/data-grid/worklist-data-grid/WorklistDataGridDeleteRowsAndColumns.test.jsx
```diff
@@ -0,0 +1,71 @@
...
+      setColumnData(newColumns);
+      setRowData(newRows);
+    }
+  };
+
+  return (
+    <div>
+      <div>
+        <button type="button" id="delete-rows-button" onClick={() => setRowData(rowData.filter(row => !row.isSelected))}>Delete Rows</button>
+        <button type="button" id="filter-rows-button" onClick={() => setRowData(rowData.filter(row => row.isSelected))}>Filter Rows</button>
```
cm9361, line 54: How are you testing removing the column with the selection?

chrismichalewicz:
@kenk2  This has been functionally reviewed.

packages/terra-data-grid/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Fixed
   * Fixed issue where focus was given to the column header instead of its button element.
   * Fixed issue where row selection was being announced twice in Worklist Data Grid.
+  * Fixed an issue when data accesses invalid indexes when deleting rows and columns.
```
cm9361, line 8: The issue is not specific to rows or columns being "deleted".  The data for the table is being updated and the row and/or column index no longer exists.

packages/terra-table/src/subcomponents/ColumnHeader.jsx
cm9361: It appears you missed validation for the aria-rowindex.

packages/terra-table/tests/jest/ColumnHeader.test.jsx
```diff
@@ -178,8 +182,11 @@ describe('ColumnHeader', () => {
     );
 
     // Verify that column headers are present
-    const columnHeader = wrapper.find('.hidden');
-    expect(columnHeader).toHaveLength(0);
+    const hiddenSelector = wrapper.find('.hidden');
+    expect(hiddenSelector).toHaveLength(0);
+
+    const columnHeader = wrapper.find('.column-header-row');
+    expect(columnHeader.props()['data-row-id']).toBe('test-table-header-row');
```
cm9361, line 189: You probably don't need to add the check to this test.  The purpose of this test is just to ensure the column headers are not hidden.

cm9361 (approved):

PR #1891, opened 2023-11-14
[worklist-data-grid] Delete duplicate row selection messages

kenk2:
### Summary

**What was changed:**
Row Selection messages are deleted from WorklistDataGrid component.

**Why it was changed:**
Row selection messages are now handled in the Table component.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9863 

---

@cerner/terra


cm9361:
If you are removing the injectIntl, you should update the JEST tests accordingly.  It is currently using "WithIntl" variants of the enzyme functions.

cm9361:
Shouldn't you also remove the row selection unit tests?

packages/terra-data-grid/src/WorklistDataGrid.jsx
```diff
@@ -165,14 +157,10 @@ function WorklistDataGrid(props) {
     onDisableSelectableRows,
     onEnableRowSelection,
     hasSelectableRows,
-    intl,
     rowHeaderIndex,
   } = props;
 
-  const rowSelectionEffectTriggered = useRef(false);
   const selectedRows = useRef([]);
```
cm9361, line 163: Shouldn't this variable have also been removed?  It was created to manage the row selection state for the aria live region logic.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -219,7 +218,7 @@ const DataGrid = injectIntl((props) => {
   // cells. In order to do so, these datagrid methods will be exposed to
   // allow those components to request focus change.
   useImperativeHandle(
```
cm9361, line 220: Do we still need the imperative handle?  With a reference to the DataGrid component, couldn't you just called the methods directly without needing to forward refs?

cm9361:
> Shouldn't you also remove the row selection unit tests? At least the jest ones since that should be covered by the Table. Did we ensure that all the proper tests have been moved to the Table?

I think you might have missed this comment.

mjpalazzo:
@kenk2 - Please provide some details of what Accessibility Review you need me to do and on which component(s). It looks like I should validate row selection with JAWS for both WDG and Table, but I would appreciate your guidance.

mjpalazzo:
I validated WDG Row Selection example with Edge/JAWS. Interactions and announcements are working as expected. Nice work @kenk2!

sdadn (approved):

cm9361 (approved):

PR #1893, opened 2023-11-16
[terra-slider] Fix wrapping issue for min/max labels

nikhitasharma:
### Summary

**What was changed:**
Modified css for min/max labels to avoid wrapping which is currently occurring on Fusion integration page. 

### Testing

Responsiveness after these updates works as expected in callouts:
image

On Fusion page, the issue is fixed-
Before:
Before_updates

After: 
After_updates

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9871
MPAGESCORE-87696

---

@cerner/terra


cm9361:
Shouldn't we have a unit test to cover this?  Also, why was the text wrapping?  The way that this is coded, the text can now exceed the boundaries of the container.  I am not sure that is correct.

nikhitasharma:
> Shouldn't we have a unit test to cover this? Also, why was the text wrapping? The way that this is coded, the text can now exceed the boundaries of the container. I am not sure that is correct.
> 
> https://developer.mozilla.org/en-US/docs/Web/CSS/white-space

This is not happening on the Terra side and was not reproducible even on small screens, it is only happening on the Fusion page and was tested on Fusion side. 
Based on conversation with Eric, we are expecting the min/max labels to be pretty small and not elaborate descriptions (we have the main field label for it and that wraps fine). So these labels need not wrap and callouts also don't require it.

cm9361:
> > Shouldn't we have a unit test to cover this? Also, why was the text wrapping? The way that this is coded, the text can now exceed the boundaries of the container. I am not sure that is correct.
> > https://developer.mozilla.org/en-US/docs/Web/CSS/white-space
> 
> This is not happening on the Terra side and was not reproducible even on small screens, it is only happening on the Fusion page and was tested on Fusion side. Based on conversation with Eric, we are expecting the min/max labels to be pretty small and not elaborate descriptions (we have the main field label for it and that wraps fine). So these labels need not wrap and callouts also don't require it.

Yes, but your content should not be allowed to exceed your container, correct?  You would then be overlaying other content because of your overflow.  There must be something that the Fusion team is doing in the integration that is causing the issue.  Have we found the true root cause of the problem?

eawww:
@nikhitasharma yeah I agree with Charles. I didn't realize this change would allow the label to exceed the container bounds. I'd expect that the slider portion would flex as the container shrinks (because growing the container and causing horizontal scroll at small widths would be really clunky on mobile)

cm9361:
> usion page also does not ha

Do you maybe need to do a troubleshooting session with the Fusion team.  Did you analyze all the styles of their parent container?  Also, looking at Eric's comment, can't you use your flexbox to shrink the size of the slider when there is not enough room?

nikhitasharma:
> > usion page also does not ha
> 
> Do you maybe need to do a troubleshooting session with the Fusion team. Did you analyze all the styles of their parent container? Also, looking at Eric's comment, can't you use your flexbox to shrink the size of the slider when there is not enough room?

I had a meeting with the developer working on the fusion side. Looks like having the widths as percentages was causing the unexpected wrapping. Removing those, the issue did not occur on fusion side. 

- I have added new wdio tests to test on tiny screen to make sure the main field label wrapping works as expected when long field names are given. 

- I have added a screenshot of how the slider looks on responsive screen sizes in the PR description above. 

- I have also added align left for main field label as that seems to be on the callouts and I missed it last time:
image

- @eawww  The horizontal scrolling will be introduced at screen widths smaller than 365px because of the min-width requirement of the input range. But on screens larger than that, the horizontal scrolling won't be needed. 
image

cm9361:
Do these style updates mean we should have used the Terra Arrange component?

nikhitasharma:
> Do these style updates mean we should have used the Terra Arrange component?

Maybe we should have.

cm9361:
> > Do these style updates mean we should have used the Terra Arrange component?
> 
> Maybe we should have.

Should we try implementing it here then?  If we already have a Terra component that does what is required, it seems like it would be best to reuse it.

nikhitasharma:
> > > Do these style updates mean we should have used the Terra Arrange component?
> > 
> > 
> > Maybe we should have.
> 
> Should we try implementing it here then? If we already have a Terra component that does what is required, it seems like it would be best to reuse it.

I am trying to get this fix in as soon as possible because of a release deadline of Monday Nov 20th. I don't think there is enough time for me to implement using terra arrange now. We have this Jira logged for improving this in future: https://jira2.cerner.com/browse/UXPLATFORM-9778

cm9361:
Last question.  Should the label text be allowed to wrap unbound or should we truncate and use an ellipsis at some point?

nikhitasharma:
> Last question. Should the label text be allowed to wrap unbound or should we truncate and use an ellipsis at some point?

The callouts mention wrapping of the label but nothing about truncation. I don't think the label text is supposed to be too long either that we might need to truncate. But I could be wrong in assuming that though.  @eawww Do you see this as something we might require with labels for Slider?

sdadn (approved):

cm9361 (approved):

eawww:
@nikhitasharma @cm9361 our design standards say that labels shouldn't be truncated so I don't think we should build it into a component. It's the responsibility of consumers to ensure their labels are concise.

eawww:
@nikhitasharma I'm still getting overflow with no scrolling in all of the examples.
image

eawww (approved):
Update: It appears the undesirable overflow behavior is a result of the example template. Using the raw test page, overflowed slider on very narrow screens is reachable via horizontal scroll.
Further consideration should be made on how this can be refined. I'll circle back with the designer to find improvements we can make in the future.

PR #1896, opened 2023-11-17
[terra-data-grid] Add section support to Flowsheet

cm9361:
### Summary

**What was changed:**
The Flowsheet Data Grid was updated to allow sections for the data displayed.

**Why it was changed:**
The change was made as this is required functionality for the component.

### Testing

This change was tested using:

- [X] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [X] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details
This PR does not cover keyboard interactions or selection. There are currently jiras logged for adding those features for sections.

**This PR resolves:**

UXPLATFORM-9708 

---

@cerner/terra


packages/terra-table/src/index.js
```diff
@@ -1,5 +1,12 @@
...
+import cellShape from './proptypes/cellShape';
+import columnShape from './proptypes/columnShape';
+import rowShape from './proptypes/rowShape';
+import sectionShape from './proptypes/sectionShape';
+import validateRowHeaderIndex from './proptypes/validators';
 
 export default Table;
-export { GridContext, GridConstants };
+export {
+  GridContext, GridConstants, cellShape, columnShape, rowShape, sectionShape, validateRowHeaderIndex,
```
sdadn, line 11: Do we want to give consumers access to these exports? If these are only used internally within Terra then maybe we can import them using the full path as an exception.

packages/terra-table/src/proptypes/columnShape.js
```diff
@@ -44,4 +44,5 @@ const columnShape = PropTypes.shape({
   sortIndicator: PropTypes.oneOf(Object.values(SortIndicators)),
 });
 
+export default columnShape;
 export { columnShape, SortIndicators };
```
sdadn, line 48: I think we can remove the second `columnShape` export.

packages/terra-table/src/index.js
cm9361: Would it hurt for them to be exported?  I was trying to keep them encapsulated but also reuse the logic as we previously had duplicates.

sycombs (approved):

eawww (approved):
Approving this only as a base for multiple people to work off of. Confirmed that JAWS in Edge reads section headers while navigating across columns within respective sections and a basic semantic table structure is used. This does not constitute a UX or a11y review. 

Some observations to be covered in future work, most of which already have jiras:
- Grid keyboard controls inoperable.
- When scrolling overflow columns, Section headers move out of viewable area.

sdadn (approved):

packages/terra-table/src/index.js
sdadn: Will table (🥁) this topic for now so approving the PR.

PR #1897, opened 2023-11-17
[terra-data-grid] FlowsheetDataGrid - Remove unsupported props from docs

smason0:
### Summary

**What was changed:**
FlowsheetDataGrid API docs were updated to remove `isSelectable` from both Column and Cell, and `isMasked` from Cell. All columns and cells in FlowsheetDataGrid now set this prop in the underlying component to always be true.

**Why it was changed:**
Flowsheet should not allow columns and cells to be unselectable or masked. These concepts are needed by the sibling component WorklistDataGrid, but should not be included as configurable options for flowsheet.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9882 

---

@cerner/terra


cm9361:
I understand the change for isSelectable.  I am curious if there is never a use case for masked cells in the flowsheet.  I also don't see the harm in it being available.  If this is a hard requirement, should we also have logic to set isMasked to false for every cell similar to the logic you added for isSelectable?

smason0:
> I understand the change for isSelectable. I am curious if there is never a use case for masked cells in the flowsheet. I also don't see the harm in it being available. If this is a hard requirement, should we also have logic to set isMasked to false for every cell similar to the logic you added for isSelectable?

It is a requirement of Flowsheet, and isMasked sets a cell to be unselectable.  The reason I don't set it for every cell is because it is falsy by default, unlike `isSelectable

cm9361 (approved):

nikhitasharma (approved):

sdadn (approved):

artpark (approved):

PR #1903, opened 2023-11-20
[terra-menu] Allow Menu to dynamically set height

sugan2416:
### Summary

**What was changed:**

Terra-Menu now dynamically adjusts height to fit according to the number of menu items.

**Why it was changed:**

- Avoids additional white space below last item when submenu contains fewer items than main-menu
- Avoids scrollbar appearing when submenu has more items than main-menu.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] No tests are needed

### Reviews

- [x] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9791 

---

@cerner/terra

image


packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -181,6 +183,22 @@ class MenuContent extends React.Component {
     }
   }
 
+  getSubmenuHeight() {
+    if (this.props.index > 0 && this.listNode) {
+      const bufHeight = (this.context.name || this.context.className) === 'orion-fusion-theme' ? 20 : 10;
```
supreethmr, line 188: instead of hardcoding the bufHeight can we use `childHeight` itself as bufHeight.

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -181,6 +183,22 @@ class MenuContent extends React.Component {
...
+      const submenuHeight = this.listNode.clientHeight + this.listNode.parentNode.parentNode.parentNode.firstChild.clientHeight + bufHeight;
```
supreethmr, line 189: Curious, listNode.clientHeight + childHeight (if header is available) should be enough to find the height of the menu !! why do we need childHeight (for when header was not present )+ buffHeight.

sugan2416: `bufHeight` would be necessary here to add for the divider space between header(firstChild.clientHeight) and the first menu item. Header will always be there since we’re setting height only for submenus.

packages/terra-menu/src/_MenuContent.jsx
sugan2416: https://github.com/cerner/terra-framework/pull/1903#discussion_r1415276136

supreethmr:
@sugan2416  Please resolve the conflicts and rebase with Main

sugan2416:
> @sugan2416 Please resolve the conflicts and rebase with Main

@supreethmr Resolved.

supreethmr (approved):

rahulkumar8cs:
+1 for accessibility.
Few Axe issue that we need to fix.
Screenshot 2023-12-08 at 6 17 13 PM

PR #1906, opened 2023-11-20
Table row selection at

cm9361:
### Summary

**What was changed:**
The AT messaging for the row selection feature of the Table was corrected to speak the proper indices.

**Why it was changed:**
The announcement of the wrong row indices would break the user experience for screenreader users.

### Testing

This change was tested using:

- [X] WDIO
- [X] Jest
- [ ] Visual testing (please attach a screenshot or recording)
- [ ] Other (please describe below)
- [ ] No tests are needed

### Reviews

In addition to engineering reviews, this PR needs:

- [X] UX review
- [X] Accessibility review
- [ ] Functional review

### Additional Details
N/A

**This PR resolves:**

UXPLATFORM-9887 

---

Thank you for contributing to Terra.
@cerner/terra

PR #1901, opened 2023-11-18
[terra-table] Table single row select

cm9361:
### Summary

**What was changed:**
The Terra Table was updated to support a single row selection mode.

**Why it was changed:**
The change was made to maintain passivity with existing functionality in our clinical applications.  Product management deemed that the lack of this feature would be a major dissatisfier for clients. 

### Testing

This change was tested using:

- [X] WDIO
- [ ] Jest
- [ ] Visual testing (please attach a screenshot or recording)
- [ ] Other (please describe below)
- [ ] No tests are needed

### Reviews

In addition to engineering reviews, this PR needs:

- [] UX review
- [ ] Accessibility review
- [X] Functional review

### Additional Details
N/A

**This PR resolves:**

UXPLATFORM-9879 

---

Thank you for contributing to Terra.
@cerner/terra

PR #1907, opened 2023-11-20
[terra-data-grid] Flowsheet Data Grid - Row Selection

smason0:
### Summary

**What was changed:**
Added new prop `onRowSelect` to FlowsheetDataGrid. This allows consumers to add behavior for when a cell in the row header column is selected. An example has also been added to the docs.

**Why it was changed:**
This change is needed to remain passive with the flowsheet functionality expected by some consumers. The example of using the `onRowSelect` callback gives consumers a reference to use when implementing their own callback function.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9735 

---

@cerner/terra


packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -247,18 +258,18 @@ function FlowsheetDataGrid(props) {
...
   const handleCellSelection = useCallback((selectionDetails) => {
-    // Exclude the row header column.
+    // Call onRowSelect for row header column
     if (selectionDetails.columnIndex === 0) {
-      return;
-    }
-
-    if (selectionDetails.isShiftPressed && anchorCell.current !== null) {
+      if (onRowSelect) {
+        onRowSelect(selectionDetails.rowId);
```
cm9361, line 264: Now that we have section support (understanding it is a work in progress), shouldn't we return both the section id and row id?

smason0: I can add the section id as a param here. Would this also apply to onCellSelect and onCellRangeSelect?

mjpalazzo:
@smason0 - Overall the Flowsheet row selection example works well, but I observed one issue when testing with Edge/JAWS. I used the Down Arrow key to navigate to a row and pressed Space to select the row. JAWS announced, "Space, Row 2 selected" as expected. I pressed Down Arrow to navigate to row 3. AWS announced, "Space, Row 3 selected, **row 1** unselected" JAWS announced the wrong index for the unselected row. I tested with Safari/VO as well and observed the same issue.

eawww:
@mjpalazzo this sounds like what was fixed by https://github.com/cerner/terra-framework/pull/1906 . Should be good once those changes make it into this branch.

smason0:
@mjpalazzo As @eawww said, that might have been because I was on an older revision of main. I have now pulled in the fix #1906 into the base of my branch.

smason0:
@mjpalazzo As @eawww said, that might have been because I was on an older revision of main. I have now pulled in the fix #1906 into the base of my branch.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
cm9361: I think we have a separate JIRA logged for the onCellSelect and onCellRangeSelect work.  That will be a little tricky as we will have to determine if we want to break the existing API and make the section id the first parameter.  We probably should have been returning objects instead of a parameter list for this reason.

cm9361: Maybe onRowSelect should "return" an object as well?  I am open to opinions.

smason0: I would be fine with making it an object. That way it might be a bit easier on consumers if we were to add anything to it, rather than introducing a new parameter.

smason0: Changed to an object

eawww:
This will require the same a11y work detailed for [table single select](https://github.com/cerner/terra-framework/pull/1901#pullrequestreview-1742724128) but it works correctly aside from that.

eawww (approved):

cm9361 (approved):

nikhitasharma (approved):

smason0:
> @smason0 - Overall the Flowsheet row selection example works well, but I observed one issue when testing with Edge/JAWS. I used the Down Arrow key to navigate to a row and pressed Space to select the row. JAWS announced, "Space, Row 2 selected" as expected. I pressed Down Arrow to navigate to row 3. AWS announced, "Space, Row 3 selected, **row 1** unselected" JAWS announced the wrong index for the unselected row. I tested with Safari/VO as well and observed the same issue.

During testing of the most recent deployment, we've confirmed the described issue is now fixed.

PR #1915, opened 2023-11-27
[terra-folder-tree] Add the ability to expand/collapse folders

sycombs:
### Summary

This PR adds the ability to expand and collapse folders so that users of `terra-folder-tree` can show and hide items in the folder tree.

<details>
<summary>Screenshots</summary>

Collapsed:
image

Expanded:
image

</details>

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9680 

---

@cerner/terra


packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -61,10 +61,14 @@ const FolderTreeItem = ({
 }) => {
   const theme = useContext(ThemeContext);
 
-  const subfolder = subfolderItems?.length > 0 ? (
+  const [isExpanded, setIsExpanded] = useState(false);
```
cm9361, line 64: Shouldn't this be a prop that is set by consumers?  It would match the pattern of our other Terra components.  Also, do we need an isCollapsible prop in the API as well?

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -91,7 +103,10 @@ const FolderTreeItem = ({
       <li
         className={itemClassNames}
         role="treeitem"
+        aria-expanded={isFolder ? isExpanded : null}
         aria-selected={isSelected}
+        onClick={handleExpandCollapse}
+        onKeyDown={handleExpandCollapse}
```
cm9361, line 109: Aren't there specific keys that trigger this behavior?  The current logic appears to expand/collapse when any key is pressed.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -77,11 +81,19 @@ const FolderTreeItem = ({
   ) : null;
 
   const itemIcon = subfolder ? <IconFolder a11yLabel={intl.formatMessage({ id: 'Terra.folder-tree.folder-icon' })} /> : icon;
+  const expandCollapseIcon = isExpanded
+    ? <IconCaretDown height="8px" width="8px" />
+    : <IconCaretRight height="8px" width="8px" />;
```
cm9361, line 86: Is there a requirement to reduce the icon size?

packages/terra-folder-tree/tests/jest/FolderTree.test.jsx
cm9361, line 1: Your jest tests don't appear to be covering the key press scenarios.

packages/terra-folder-tree/tests/wdio/folder-tree-spec.js
cm9361, line 1: Can't we cover the click and key down events in jest and eliminate the need for WDIO tests?

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -102,9 +117,18 @@ const FolderTreeItem = ({
         <span style={{ paddingLeft: `${level}rem` }}>
           <Arrange
             fitStart={(
-              <Spacer paddingLeft="medium" paddingRight="medium" isInlineBlock>
-                {itemIcon}
-              </Spacer>
+              <>
```
cm9361, line 120: You should not need the React fragment here since everything is contained within a single Spacer element at the moment.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -102,9 +117,18 @@ const FolderTreeItem = ({
...
+                <Spacer paddingLeft="medium" paddingRight="medium" isInlineBlock>
+                  {
+                    isFolder ? (
+                      <Spacer paddingRight="small" isInlineBlock>
+                        {expandCollapseIcon}
```
cm9361, line 125: Should this icon be center aligned vertically?

packages/terra-folder-tree/tests/wdio/folder-tree-spec.js
```diff
@@ -1,19 +1,34 @@
-Terra.describeViewports('FolderTree', ['medium', 'large'], () => {
+Terra.describeViewports('FolderTree', ['small', 'medium'], () => {
```
sdadn, line 1: Do we need both screen sizes for this?

cm9361:
The background color for non-selected elements does not appear to be correct in the lowlight theme

packages/terra-folder-tree/tests/wdio/folder-tree-spec.js
sycombs: I'm not sure which would be more appropriate here. Maybe just medium?

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
sycombs: Yes, the expand/collapse icon should be 8px x 8px per UX requirements.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
sycombs: `onKeyDown` was added here to fix an eslint error, the actual keyboard logic will be added in UXPLATFORM-9776.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
sycombs: Looking ahead at Expand All/Collapse All, I think you're right. Will go ahead and make that change.

> Also, do we need an isCollapsible prop in the API as well?

Based on the AC for this Jira, all items with children (AKA folders) should be collapsible, so I don't think we necessarily need a prop for that.

packages/terra-folder-tree/tests/wdio/folder-tree-spec.js
sycombs: Just wondering why we don't want WDIO tests for the expand/collapse cases? I know we've had some discussion around eliminating unnecessary WDIO tests, but wouldn't we want to keep track of possible visual changes here?

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.module.scss
```diff
@@ -18,6 +18,12 @@
     }
   }
   
+  .folder {
+    &:hover {
```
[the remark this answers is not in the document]
sycombs: [6436a2b](https://github.com/cerner/terra-framework/pull/1915/commits/6436a2bcf9ca2528f3156c2386b74ba7dd35ed02)

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
cm9361: The key down event would be part of the expand/collapse logic though.  You can handle the specific key for just this functionality without accounting for the keyboard navigation.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
sycombs: Oddly the default [vertical-align property](https://github.com/cerner/terra-core/blob/main/packages/terra-icon/src/Icon.module.scss#L7) from `terra-icon` looks fine for default-sized icons, just not the 8px x 8px caret icons. Overriding the style and setting it to `baseline` seems to center them properly.

[a45606e](https://github.com/cerner/terra-framework/pull/1915/commits/a45606ee7c783e477d67e17b25c62af7ac9d7678)

packages/terra-folder-tree/tests/wdio/folder-tree-spec.js
cm9361: I just knew we had recently created Jest tests that could cover validating such updates.  However, maybe we do want the visual confirmation as well.

packages/terra-folder-tree/tests/wdio/folder-tree-spec.js
sdadn: Seems like you're testing basic functionality that doesn't correspond to screen size. I would suggest just doing medium or large so we don't have redundant tests.

If you want to test layout at various widths, then we can have specific tests for those screen sizes.

cm9361:
The background color for non-selected elements does not appear to be correct in the lowlight theme

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.module.scss
cm9361, line 1: The overflow-wrap style still needs to be corrected.  You wouldn't want to use anywhere.

packages/terra-folder-tree/tests/jest/FolderTree.test.jsx
```diff
@@ -1,66 +1,85 @@
 import React from 'react';
 import ActionHeader from 'terra-action-header';
 /* eslint-disable-next-line import/no-extraneous-dependencies */
-import { shallowWithIntl } from 'terra-enzyme-intl';
+import { mountWithIntl, shallowWithIntl } from 'terra-enzyme-intl';
 import FolderTree from '../../src/FolderTree';
+import FolderTreeItem from '../../src/subcomponents/FolderTreeItem';
```
sdadn, line 6: Can we import these from index.js? This way, the usage would be more in line with how a consumer would utilize it.

sdadn (approved):
Added a minor comment but approving since it doesn't impact functionality.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.module.scss
cm9361: yes, they do.  Break word will break anywhere when you run out of room.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -91,22 +108,34 @@ const FolderTreeItem = ({
...
+        onClick={onToggle}
+        onKeyDown={onToggle}
       >
         <input
           type="radio"
           checked={isSelected}
-          onClick={onClick}
+          onChange={onClick}
+          aria-hidden // Hiding the radio button from assistive technology since they cannot be grouped correctly
+          tabIndex={-1} // Prevent tabbing to the button since it should not be read or acknowledged by assistive technology
```
cm9361, line 121: This will allow the control to get focus when you click though.  Is that what you want?

JessieRandle (approved):

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
sycombs: We want the radio button to be clickable (per the previous story's AC) but we need to hide it from screenreaders so that they aren't reading radio buttons out of the tree hierarchy.

PR #1919, opened 2023-11-30
Compact interactive list keyboard navigation

adoroshk:
### Summary
This PR adds keyboard navigation to compact interactive list component.

**What was changed:**
Following functionality has been added:
* The list is a single roving tab stop.
* Right/Left Arrows navigate across visual rows. When focus hits the end of the visual row navigation stops.
* Up/Down Arrows navigate across visual columns. Navigating down from the bottom of a visual column takes you to the corresponding cell at the top of the next visual column and vice versa (except for the last and first items).

* Moves to the **first cell** in the **focused visual row**
  * Windows
    * Home
  * Mac
    * Home
    * Fn + Left Arrow (equivalent to Home)
    * Cmd + Left Arrow

* Moves to the **last cell** in the **focused visual row**
  * Windows
    * End
  * Mac
    * End
    * Fn + Right Arrow (equivalent to End)
    * Cmd + Right Arrow

* Moves to the **first cell** in the **first item** (top left)
  * Windows
    * Ctrl + Home
  * Mac
    * Cmd or Ctrl + Home
    * Cmd or Ctrl + Fn + Left Arrow (equivalent to Home)
    * Cmd + Ctrl + Left Arrow

* Moves to the **last cell** in the **last item** (bottom item in rightmost visual column)
  * Windows
    * Ctrl + End
  * Mac
    * Cmd or Ctrl + End
    * Cmd or Ctrl + Fn + Right Arrow (equivalent to End)
    * Cmd + Ctrl + Right Arrow

* If cell has an interactive element (only one per cell allowed), focus is set to that element, not the cell (only one interactive element per cell allowed) and the arrow keys behavior is default to that element (NOTE: currently it causes problems with input elements that won't allow to navigate away from that with keyboard, a decision needs to be maid by UX team on what is the best way around that).
* Cells are select-able by Space key (it triggers onClick for cell) unless the cell has an interactable element within, in that case its not selectable neither via mouse click nor per space key.

### Testing

Note: at that moment tests are work in progress.
This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9788


packages/terra-compact-interactive-list/src/proptypes/columnShape.js
```diff
@@ -10,7 +10,12 @@ const columnShape = PropTypes.shape({
   /**
    * String of text which will be used to provide a column context and to serve as a column header if needed.
    */
-  displayName: PropTypes.string,
+  displayName: PropTypes.string.isRequired,
+
+  /**
+   * Undicates whether the column is a column header (provides the context for screen readers).
+   */
+  rowHeader: PropTypes.bool,
```
cm9361, line 18: If this is a column shape, wouldn't this be a column header?

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
```diff
@@ -99,9 +124,108 @@ const CompactInteractiveList = (props) => {
...
+    }
+
+    setFocusedRowCol(moveFocusTo);
+    event.preventDefault(); // prevent the page from moving with the arrow keys.
+  };
+
+  const onFocus = (event) => {
+    if (listRef.current.contains(event.target)) {
+      setFocusOnCell(focusedCell.current);
+      event.preventDefault();
```
cm9361, line 226: I would not expect us to prevent the default focus behavior.

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
```diff
@@ -99,9 +124,108 @@ const CompactInteractiveList = (props) => {
...
+        return;
+      default:
+        return;
+    }
+
+    setFocusedRowCol(moveFocusTo);
+    event.preventDefault(); // prevent the page from moving with the arrow keys.
+  };
+
+  const onFocus = (event) => {
```
cm9361, line 223: Do we need the focus event listener for all cases here?  This event is triggered event when focus is moving inside the control because the focus event bubbles in React.

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
```diff
@@ -99,9 +124,108 @@ const CompactInteractiveList = (props) => {
...
+    const elementToFocus = listRef.current.children[row + 1].children[cell];
+    const interactiveChildren = getFocusableElements(elementToFocus);
+    // if no interactive elements
+    if (onCellSelect && interactiveChildren?.length === 0) {
+      onCellSelect(rows[row].id, columns[cell].id);
+    }
+  }, [onCellSelect, rows, columns]);
+
+  const handleKeyDown = (event) => {
+    let moveFocusTo = focusedCell.current;
```
cm9361, line 168: Maybe we could rename this variable to something like, nextCell, or something similar.  I think it would be helpful to specify that this is a cell.  The current name looks like a method.

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
```diff
@@ -70,12 +83,24 @@ const propTypes = {
   width: PropTypes.string,
 
   /**
-   * Columns minimum width should be a valid css string in value in px, em, or rem units..
+   * Callback function that is called when a selectable cell is selected. Parameters:
+   * @param {string} rowId rowId
+   * @param {string} columnId columnId
```
cm9361, line 88: You should return an object with named parameters for rowId and columnId instead of two parameters.

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
```diff
@@ -99,9 +124,108 @@ const CompactInteractiveList = (props) => {
...
+  const handleOnCellSelect = useCallback(({ row, cell }) => {
+    focusedCell.current = { row, cell };
+    // check for interactive elements in that cell
+    // add 1 to the row number to accomodate for hidden header
+    const elementToFocus = listRef.current.children[row + 1].children[cell];
+    const interactiveChildren = getFocusableElements(elementToFocus);
+    // if no interactive elements
+    if (onCellSelect && interactiveChildren?.length === 0) {
+      onCellSelect(rows[row].id, columns[cell].id);
+    }
```
cm9361, line 164: Why would the presence of interactive elements prevent us from triggering the callback for cell selection?  What happens if the user clicks on the cell but not the interactive element?

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
```diff
@@ -99,9 +124,108 @@ const CompactInteractiveList = (props) => {
...
+  const setFocusedRowCol = ({ row, cell }) => {
+    focusedCell.current = { row, cell };
+    setFocusOnCell({ row, cell });
+  };
+
+  const handleOnCellSelect = useCallback(({ row, cell }) => {
+    focusedCell.current = { row, cell };
+    // check for interactive elements in that cell
+    // add 1 to the row number to accomodate for hidden header
+    const elementToFocus = listRef.current.children[row + 1].children[cell];
```
cm9361, line 159: If you are selecting the cell, it should already be getting focus.

packages/terra-compact-interactive-list/src/subcomponents/Row.jsx
```diff
@@ -153,6 +159,7 @@ const Row = (props) => {
           columnMinimumWidth={columnMinimumWidth}
           columnMaximumWidth={columnMaximumWidth}
           widthUnit={widthUnit}
+          onCellSelect={() => onCellSelect(index)}
```
cm9361, line 162: This will cause an unnecessary rerender every time.  You should create a method and wrap with the useCallback hook to prevent this.

packages/terra-compact-interactive-list/src/utils/keyHandlerUtils.js
```diff
@@ -0,0 +1,279 @@
...
+    return validTypes.indexOf(inputType) >= 0;
+  }
+
+  return false;
+};
+
+/**
+ * Move focus to next focusable element outside the container element
+ */
+export const moveFocusFromElement = (containerRef, id, moveForward) => {
```
cm9361, line 31: This is being duplicated from the data-grid package.  Do we need to put this logic in a common file somehow?

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
eawww: The intention is for the interactive element to be the only interactive thing in the cell.

mjpalazzo:
@adoroshk - Great work on this component, Alla! There is a lot nice functionality in your implementation. I validated keyboard navigation using the Edge browser, and most interactions worked as expected. Visible focus indicator displays as I navigate to each cell.
Here are some minor observations:
Cell Content and Cell Selection examples - I used the arrow key to navigate to the cell with an text input field. I entered "test" and then got trapped. I could not use the arrow keys to navigate to another cell. I was able to press Tab and leave the list completely.
All examples - the favorites icon is not interactable. Should it be?

mjpalazzo:
@adoroshk The title of the PR indicates that it is for keyboard navigation so I did do any screen reader testing.

packages/terra-compact-interactive-list/src/proptypes/formattedColumnShape.js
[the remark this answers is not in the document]
adoroshk: Good catch, fixed in https://github.com/cerner/terra-framework/pull/1919/commits/fc32888e860735982a59eadfd00346f3b976688d

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
```diff
@@ -80,12 +87,14 @@ const Cell = (props) => {
...
   return (
+    // eslint-disable-next-line jsx-a11y/no-static-element-interactions
     <div
-      role="gridcell"
+      role={rowHeader ? 'rowheader' : 'gridcell'}
       className={className}
       tabIndex={-1}
       // eslint-disable-next-line react/forbid-dom-props
       style={style}
+      onMouseDown={onCellSelect}
```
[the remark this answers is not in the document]
adoroshk: Added in https://github.com/cerner/terra-framework/pull/1919/commits/fc32888e860735982a59eadfd00346f3b976688d

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
adoroshk: I fixed the code to return the object in https://github.com/cerner/terra-framework/pull/1919/commits/fc32888e860735982a59eadfd00346f3b976688d

packages/terra-compact-interactive-list/src/subcomponents/Row.jsx
adoroshk: Fixed in https://github.com/cerner/terra-framework/pull/1919/commits/fc32888e860735982a59eadfd00346f3b976688d

packages/terra-compact-interactive-list/src/utils/keyHandlerUtils.js
adoroshk: That might be a good idea. I was thinking of it, but wasn't sure where exactly to put it so that data grid and compact interactive list both can consume it.

adoroshk:
> Cell Content and Cell Selection examples - I used the arrow key to navigate to the cell with an text input field. I entered "test" and then got trapped. I could not use the arrow keys to navigate to another cell. I was able to press Tab and leave the list completely.
> All examples - the favorites icon is not interactable. Should it be?

1. As for Cell Content and Cell Selection examples trapping keyboard navigation - @eawww and I are waiting for the answer from the UX team on how to avoid that. The idea was that the compact intercative list would not have a focus trap as grid has, but an interactive element within a cell would keep its native keyboard navigation, which leads to that case. 

2. The favorites icon is not interactable. Should it be? - That is a great suggestion, it totally should be intercatable. I'll see if I can make it in this PR or a part of the upcoming documentation JIRA.

adoroshk:
> @adoroshk The title of the PR indicates that it is for keyboard navigation so I did do any screen reader testing.

@mjpalazzo , As for the screen reader testing, I guess the part that needs screen reading testing is when the keyboard navigation moves between cells the column and the row context should be clear to the user. Here is what I meant:
1. Each semantic column, such as icon, text, and featured icon, has a hidden column header, which comes from the columns displayedName prop and even though it's hidden, it needs to be announced by JAWS when focus moves between semantic columns.
2. Each semantic row (which consists of 3 columns, such as icon, text, in majority examples, or 4 columns in cell content and cell selection examples) has a row header, identified by zero-based rowHeaderIndex prop, set to 1 (meaning in all examples it'll be a text column, such as 'Discern Care Set (1)'). That row header needs to be anounced by screen readers when user navigates between semantic rows.

Is this part working as expected? I apologize for such a long explanation, feel free to reach out to me for details.

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
adoroshk: I was trying to be consistent with the same method for similar functionality in DataGrid, thought it would be less confusing for developers if the naming is the same: https://github.com/cerner/terra-framework/blob/ba1d238e55bfa51b74723029baa89b34f6dea65d/packages/terra-data-grid/src/DataGrid.jsx#L374

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
adoroshk: Great catch, fixed focus to not be triggered by children swap in https://github.com/cerner/terra-framework/pull/1919/commits/092e8379aaec6fda0ec2087e958f6f7dcdd8ae37

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
adoroshk: removed in https://github.com/cerner/terra-framework/pull/1919/commits/092e8379aaec6fda0ec2087e958f6f7dcdd8ae37

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
adoroshk: You are right, that cell already has focus. The logic behind this line has different purpose and I changed the naming from `elementToFocus` to be `focusedElement`.
This line is here as a stepping stone for detecting if currently focused cell has an interactive element, as per my understanding if the cell has an interactive element, the cell should not be selectable, hence onCellSelect method should not be called. 

@eawww, I wanted to double-check with you the logic in case the cell has interactive element in it. Does it make the cell non selectable, or did I get it wrong?

eawww:
Synced up with @adoroshk for clarification on the keyboard shortcuts and a slight discrepancy with the layout which will be addressed later.

Original notes said:
* Ctrl + Cmd + Left (on Mac) / Ctrl + Home (on Windows) moves focus to the first cell in the first item in the list.
* Ctrl + Cmd + Right (on Mac) / Ctrl + End (on Windows) moves focus to the last cell in the last item in the list.
* Cmd + Left (on Mac) / Home (on Windows) moves focus to the first cell in the first item in the visual row.
* Cmd + Right (on Mac) / End (on Windows) moves focus to the first cell in the first item in the visual row.

Corrected a few things and laid out in a more digestible format:
* Moves to the **first cell** in the **focused visual row**
  * Windows
    * Home
  * Mac
    * Home
    * Fn + Left Arrow (equivalent to Home)
    * Cmd + Left Arrow

* Moves to the **last cell** in the **focused visual row**
  * Windows
    * End
  * Mac
    * End
    * Fn + Right Arrow (equivalent to End)
    * Cmd + Right Arrow

* Moves to the **first cell** in the **first item** (top left)
  * Windows
    * Ctrl + Home
  * Mac
    * Cmd or Ctrl + Home
    * Cmd or Ctrl + Fn + Left Arrow (equivalent to Home)
    * Cmd + Ctrl + Left Arrow

* Moves to the **last cell** in the **last item** (bottom item in rightmost visual column)
  * Windows
    * Ctrl + End
  * Mac
    * Cmd or Ctrl + End
    * Cmd or Ctrl + Fn + Right Arrow (equivalent to End)
    * Cmd + Ctrl + Right Arrow

adoroshk:
> Synced up with @adoroshk for clarification on the keyboard shortcuts and a slight discrepancy with the layout which will be addressed later.
> 
> Original notes said:
> 
>     * Ctrl + Cmd + Left (on Mac) / Ctrl + Home (on Windows) moves focus to the first cell in the first item in the list.
> 
>     * Ctrl + Cmd + Right (on Mac) / Ctrl + End (on Windows) moves focus to the last cell in the last item in the list.
> 
>     * Cmd + Left (on Mac) / Home (on Windows) moves focus to the first cell in the first item in the visual row.
> 
>     * Cmd + Right (on Mac) / End (on Windows) moves focus to the first cell in the first item in the visual row.
> 
> 
> Corrected a few things and laid out in a more digestible format:
> 
>     * Moves to the **first cell** in the **focused visual row**
>       
>       * Windows
>         
>         * Home
>       * Mac
>         
>         * Home
>         * Fn + Left Arrow (equivalent to Home)
>         * Cmd + Left Arrow
> 
>     * Moves to the **last cell** in the **focused visual row**
>       
>       * Windows
>         
>         * End
>       * Mac
>         
>         * End
>         * Fn + Right Arrow (equivalent to End)
>         * Cmd + Right Arrow
> 
>     * Moves to the **first cell** in the **first item** (top left)
>       
>       * Windows
>         
>         * Ctrl + Home
>       * Mac
>         
>         * Cmd or Ctrl + Home
>         * Cmd or Ctrl + Fn + Left Arrow (equivalent to Home)
>         * Cmd + Ctrl + Left Arrow
> 
>     * Moves to the **last cell** in the **last item** (bottom item in rightmost visual column)
>       
>       * Windows
>         
>         * Ctrl + End
>       * Mac
>         
>         * Cmd or Ctrl + End
>         * Cmd or Ctrl + Fn + Right Arrow (equivalent to End)
>         * Cmd + Ctrl + Right Arrow

Thank you for a detailed clarification! The key navigation should work now as described per change made in https://github.com/cerner/terra-framework/pull/1919/commits/466e8c8925bbcc558f73fbe0983801049ef68aa2

mjpalazzo:
> > > > @adoroshk The title of the PR indicates that it is for keyboard navigation so I did do any screen reader testing.
> > > 
> > > 
> > > @mjpalazzo , As for the screen reader testing, I guess the part that needs screen reading testing is when the keyboard navigation moves between cells the column and the row context should be clear to the user. Here is what I meant:
> > > 
> > > 1. Each semantic column, such as icon, text, and featured icon, has a hidden column header, which comes from the columns displayedName prop and even though it's hidden, it needs to be announced by JAWS when focus moves between semantic columns.
> > > 2. Each semantic row (which consists of 3 columns, such as icon, text, in majority examples, or 4 columns in cell content and cell selection examples) has a row header, identified by zero-based rowHeaderIndex prop, set to 1 (meaning in all examples it'll be a text column, such as 'Discern Care Set (1)'). That row header needs to be anounced by screen readers when user navigates between semantic rows.
> > > 
> > > Is this part working as expected? I apologize for such a long explanation, feel free to reach out to me for details.
> > 
> > 
> > @adoroshk - Thanks for your reply. There was a typo in my comment regarding screen reader. What I meant to say is I did **not** do any screen reader testing. Sorry about that! If the component is ready for screen reader testing, let me know and I will get to work on it. What is expected is that JAWS announces the name, role, and value of the control, its state if it can change state (e.g., expand/collapse, selected/unselected), helper text (visually hidden text instructions for the screen reader user on how to interact with the control), and the name, role, value, state, helper text of any interactive components inside the control. Also, any JAWS specific keyboard navigation shortcuts work as expected.
> 
> @mjpalazzo, thank you for such a detailed response! Per my understanding what's missing from the list above is a helper text (visually hidden text instructions for the screen reader user on how to interact with the control). Can I by any chance ask to provide the copy of those instructions so that I can add them to the component?

@eawww - Is helper text something you would provide? If so, please post some for @adoroshk. If not, who typically authors helper text?

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
cm9361: I see you updated the code.  However, I think you also need to update the documentation.

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
```diff
@@ -70,20 +85,38 @@ const propTypes = {
...
+  /**
+   * Columns minimum width should be a valid css string in value in px, em, or rem units.
    */
   columnMinimumWidth: PropTypes.string,
 
   /**
-   * Columns maximum width should be a valid css string in value in px, em, or rem units..
+   * Columns maximum width should be a valid css string in value in px, em, or rem units.
    */
   columnMaximumWidth: PropTypes.string,
```
cm9361, line 107: Wouldn't the minimum and maximum widths be part of the columnShape?  I would expect to be able to control the values for each column.

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
cm9361: I don't see a variable named, "moveFocusTo", in that file.  Maybe I am not understanding properly.

eawww:
@mjpalazzo @adoroshk I wouldn't want to give overly verbose instructions detailing every possible interaction, especially since that also needs to be available to non screen reader users, but something like "Use arrow keys to navigate and space or enter to activate a cell" would probably be helpful.
We still need to make the external instructions components for all these grids.

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
adoroshk: There are 2 ways to control column max and min width:
Option 1: specify min and max width for a specific column as a part of column shape prop: https://github.com/cerner/terra-framework/blob/466e8c8925bbcc558f73fbe0983801049ef68aa2/packages/terra-compact-interactive-list/src/proptypes/columnShape.js#L27
Option 2: (the one you are looking at) columnMaximumWidth and olumnMinimumWidth prop of the whole component ,this way they'll be applied to each column.

Option 1 takes precedence on Option 2, so if there is columnMaximumWidth set to the whole component and hence applied to all columns, but then some columns have minimumWidth (or maximumWidth) props set for these columns directly, the last ones would be applied to those columns.

packages/terra-compact-interactive-list/src/utils/keyHandlerUtils.js
```diff
@@ -0,0 +1,310 @@
...
+
+/* NOTE: This method doesn't include check for element.closest('[inert]') === null in purpose
+ * The method is caled on initial render to ensure the cell has no interactable elements, and set the cell tabindex based on that.
+ * on render one of the parent elements has inert="" which made it always return no interactible elements for every cell.
+*/
+export const getFocusableElements = (parentElement) => [...parentElement.querySelectorAll(`${focusableElementSelector}`)].filter(
+  element => !element.hasAttribute('disabled')
+      && !element.getAttribute('aria-hidden')
+      && window.getComputedStyle(element).visibility !== 'hidden',
+);
```
cm9361, line 13: We have a similar method in focusManagement.js in the Table component.  I would think we would not want to duplicate it all.  Same comment for the other methods.

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
```diff
@@ -1,15 +1,37 @@
...
+
+  /**
+   * The cell's row position in the table. This is zero based.
+   */
+  rowIndex: PropTypes.number,
+
+  /**
+   * The cell's column position in the table. This is zero based.
+   */
+  columnIndex: PropTypes.number,
```
cm9361, line 28: Shouldn't the cell also have a column id?

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
```diff
@@ -62,6 +103,34 @@ const Cell = (props) => {
...
+      if (interactiveChildren?.length === 0) {
+        setIsSelectableCell(true);
+      }
+    }
+  }, [cellRef]);
+
+  const handleKeyDown = (event) => {
+    // does not need setFocusedCell call as the cell focused by key is tracked in CompactInteractiveList component
+    const key = event.keyCode;
+    if (key === KeyCode.KEY_SPACE && isSelectableCell) {
```
cm9361, line 119: So a cell cannot be selected if it doesn't have interactable elements?

packages/terra-compact-interactive-list/src/utils/keyHandlerUtils.js
```diff
@@ -0,0 +1,310 @@
...
+  }
+  // vertical flow
+  const rowsPerColumn = Math.ceil(rowsLength / numberOfColumns);
+  const numberOfPlaceholders = (rowsPerColumn * numberOfColumns) - rowsLength;
+  if ((numberOfPlaceholders === 0 && row + 1 === rowsLength) || (numberOfPlaceholders !== 0 && row + 1 === rowsLength + numberOfPlaceholders - 1)) {
+    // the last semantic row in the last visual column, focus shouldn't go anywhere
+    return focusedCell;
+  }
+  // focus should go to the next semantic row unless the last row
+  return { row: row + 1 < rowsLength ? row + 1 : row, cell };
```
cm9361, line 89: You probably could simplify this by using the Math.min function.

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
adoroshk: Per @eawww at that moment (unless it changes in the future) the cell can NOT be selected If there is an interactable element in it (and there can be only one intreactive element per cell).

eawww (approved):
Thank you @adoroshk! It works fantastically.

One last request

Could you add to the prop description for `numberOfColumns`: "Number of rows is calculated as the number of items divided by the number of columns, rounded up." or if there's a more accurate wording, that'd be cool.

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
adoroshk: It does, 2 lines above, line 18 is  `rowId: PropTypes.string.isRequired`

packages/terra-compact-interactive-list/src/utils/keyHandlerUtils.js
adoroshk: Not sure I got how to do that. An example will be appreciated.

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
```diff
@@ -99,9 +132,96 @@ const CompactInteractiveList = (props) => {
...
+    }
+    focusedCell.current = moveFocusTo;
+    focusCell(moveFocusTo);
+    event.preventDefault(); // prevent the page from moving with the arrow keys.
+  };
+
+  const onFocus = (event) => {
+    if (!event.currentTarget.contains(event.relatedTarget)) {
+      // Not triggered when swapping focus between children
+      focusCell(focusedCell.current);
```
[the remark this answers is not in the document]
adoroshk: Good catch. In this commit I changed the focused cell data to row and column ids instead of indexes: https://github.com/cerner/terra-framework/pull/1919/commits/396ed597aeb86c4aad0c6304fa5c4f8f0e6b4f

adoroshk:
> Thank you @adoroshk! It works fantastically.
> 
> One last request
> 
> Could you add to the prop description for `numberOfColumns`: "Number of rows is calculated as the number of items divided by the number of columns, rounded up." or if there's a more accurate wording, that'd be cool.

@eawww added the description here: https://github.com/cerner/terra-framework/blob/396ed597aeb86c4aad0c6304fa5c4f8f0e6b4f10/packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx#L73

packages/terra-compact-interactive-list/src/utils/keyHandlerUtils.js
cm9361: You should not need to change it.  You just need to update your jest tests appropriately.  If you check the other jest tests, there is code to handle providing an offset value.  The logic should work for your component.

adoroshk: @cm9361 huge thank you for the help with tests to accommodate the full original logic from terra-table. I fixed the tests and currently added getFocusableElements is the original one that is implemented in terra-table.

In order to reuse the same logic rather than copy-paste it, I added the export for the `getFocusableElements` method from the `terra-table`. Once the terra-table is released, I will remove copied getFocusableElements method from the compact interactive list and attach the one exported from the terra-table. Would that work?
https://github.com/cerner/terra-framework/pull/1919/commits/cdbf56a32d01b0831f45994d9fef08cdce5c9d12

cm9361: You can use a relative path instead of having to add a package dependency.  That logic was released in terra-table quite a while ago though, if I am not mistaken.

adoroshk: Correct, imported from terra-table: https://github.com/cerner/terra-framework/pull/1919/commits/e7343db5f21961d67aaf2ef57d4d2577e7bb1ec4

cm9361 (approved):

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
```diff
@@ -99,9 +133,103 @@ const CompactInteractiveList = (props) => {
...
+      case KeyCode.KEY_UP:
+        moveFocusTo = handleUpKey(moveFocusTo, numberOfColumns, flowHorizontally, rows.length);
+        event.preventDefault();
+        break;
+      case KeyCode.KEY_DOWN:
+        moveFocusTo = handleDownKey(moveFocusTo, numberOfColumns, flowHorizontally, rows.length);
+        event.preventDefault();
+        break;
+      case KeyCode.KEY_LEFT: {
+        moveFocusTo = handleLeftKey(event, moveFocusTo, numberOfColumns, flowHorizontally, columns.length, rows.length);
```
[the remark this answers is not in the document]
adoroshk: Good catch on `preventDefault` inconsistency. Actually, we don't need the `preventDefault` on Up and Down keys, as there is one preventDefault for all cases after the switch. Escape and Tab bave return at the end, so they need their own preventDefaults, but arrow keys have `break`, not `return` and go to the end of the method to reach the `preventDefault` there and I just missed that. I removed extra in this commit: https://github.com/cerner/terra-framework/pull/1919/commits/a6bcd2f4f4f1e60dca4694a8e1c018ed21b3e639

packages/terra-compact-interactive-list/src/utils/keyHandlerUtils.js
```diff
@@ -0,0 +1,328 @@
...
+};
+
+/**
+ * @param {object} focusedCell - a { row, cell } pair, where row and cell are indexes of currently focused semantic row and a cell within that row.
+ * @param {number} numberOfColumns - a number of visual columns.
+ * @param {boolean} flowHorizontally - sematic rows horizontal flow direction
+ * @param {number} rowsLength - a total number of seamntic rows in the list.
+ * @returns - a { row, cell } indexes pair to focus on.
+ */
+export const handleDownKey = (focusedCell, numberOfColumns, flowHorizontally, rowsLength) => {
```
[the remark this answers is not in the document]
adoroshk: I understand your concern, I'll look into shared logic between these methods and see if it can be improved in my next PR, this one gets too complex to monitor the changes.

PR #1925, opened 2023-12-04
[terra-menu]Menu Issue with contentWidth as auto fixed

ashishkcerner:
### Summary
When setting the value of the prop `contentWidth` for a Menu as "auto", the width of the dropdown menu is squished to 0 and does not display anything.(Same scenario is occurring on collapsible-menu-view as well which is consuming terra-menu .)

**What was changed:**
Added the fix in menu.jsx file for the above issue.

**Why it was changed:**
When setting the value of the prop `contentWidth` for a Menu as "auto", the width of the dropdown menu is squished to 0 and does not display anything.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9880 

Before Implementation Terra-menu:
Screenshot 2023-12-04 at 5 57 43 PM

After Implementation Terra-Menu
Screenshot 2023-12-04 at 5 59 28 PM

Before Implementation Terra-CollapsibleMenuView:
Screenshot 2023-12-06 at 11 51 03 AM

After implementation Terra-CollapsibleMenuView: 
Screenshot 2023-12-06 at 11 54 21 AM

---

@cerner/terra


packages/terra-framework-docs/src/terra-dev-site/test/menu/menu/LargeMenuAutoWidth.test.jsx
```diff
@@ -0,0 +1,96 @@
...
+    return (
+      <div>
+        <div>
+          This menu should have a large height. And all items should be visible without scrolling.
+        </div>
+        <Menu
+          isOpen={this.state.open}
+          targetRef={this.getButtonNode}
+          onRequestClose={this.handleRequestClose}
+          contentWidth="auto"
```
sugan2416, line 47: I'm seeing that this example works fine when added to main branch as well. Are we sure this is the right example to test the change?
image

ashishkcerner: Yes ,To recreate the scenario we can revert the changes, the given example wont work and instead will throw an broken menu .

sugan2416: The above screenshot is from 'main'.  Example is not broken here

ashishkcerner: Ok In this particular case you can increase the size of menu by adding the more menuitems .(When the menu size is smaller than screen size in that scenario it is working fine .But when screen size is smaller than menu size in that case we can recreate the broken menu)

sugan2416 (approved):

packages/terra-framework-docs/src/terra-dev-site/test/menu/menu/LargeMenuAutoWidth.test.jsx
supreethmr: Does the reported issue happens when height is set auto as well ?

ashishkcerner: No on height as auto it is working fine

supreethmr (approved):

PR #1931, opened 2023-12-06
[terra-pills]Fixed focus issue on pressing Delete Key.

saket2403:
### Summary

The focus was getting lost when Delete Key (Windows - Delete Key, Mac- Fn+Delete) was pressed. 

**What was changed:**

The OR operator was used in the switch case condition which does not work as expected. So separated the conditions in different case blocks.

**Why it was changed:**

It was changed because the combination of || operator and switch case does not work and it was causing unexpcted behaviors like loss of focus. 

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9910 

---

@cerner/terra


supreethmr (approved):

sugan2416 (approved):

rahulkumar8cs:
+1

PR #1933, opened 2023-12-06
[terra-flowsheet-data-grid] Flowsheet Sections Multi-Cell/Range Select

kenk2:
### Summary

**What was changed:**
Implements multi-select in flowsheet with sections.

**Why it was changed:**
To implement multi-selection in flowsheet with sections. It was previously only working with rows and not sections.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9877 

---

@cerner/terra


packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -66,11 +66,13 @@ const propTypes = {
    * Callback function that is called when a selectable cell is selected. Parameters:
    * @param {string} rowId rowId
    * @param {string} columnId columnId
+   * @param {string} sectionId sectionId
```
cm9361, line 69: We will want to return an object with named parameters for the section id, row id, and column id.

packages/terra-data-grid/tests/wdio/__snapshots__/reference/clinical-lowlight-theme/en/chrome_large/flowsheet-data-grid-spec/flowsheet-with-sections-multi-cell-select-bottom-anchor.png
cm9361, line 1: Shouldn't the focused field also be highlighted?

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -229,33 +231,56 @@ function FlowsheetDataGrid(props) {
...
+    let rowIndexTopBound = Math.min(anchorRowIndex, rowIndex);
+    let rowIndexBottomBound = Math.max(anchorRowIndex, rowIndex);
     const columnIndexLeftBound = Math.min(anchorColumnIndex, columnIndex);
     const columnIndexRightBound = Math.max(anchorColumnIndex, columnIndex);
 
+    if (flowsheetSections) {
+      const anchorSection = flowsheetSections.find(section => section.id === anchorCell.current.sectionId);
+      const selectedSection = flowsheetSections.find(section => section.id === sectionId);
+
+      if (anchorSection.id !== selectedSection.id) {
```
cm9361, line 256: The sectionId is being passed as an argument to the method.  Why do we need to use the selectedSection object here?

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -229,33 +231,56 @@ function FlowsheetDataGrid(props) {
...
+        rowIndexTopBound = newSectionOnTop ? 0 : rowIndex;
+        rowIndexBottomBound = newSectionOnTop ? rowIndex : (rowsToSelect.length - 1);
+
+        // Change the anchor cell when selecting into a new section
+        anchorCell.current = {
+          rowId: newSectionOnTop ? rowsToSelect[rowIndexTopBound].id : rowsToSelect[rowIndexBottomBound].id,
+          columnId: anchorCell.current.columnId,
+          sectionId,
+        };
+      }
```
cm9361, line 268: I am not following this logic well.  What is newSectionOnTop?  Also, it seems like there should be a simpler way to handle this implementation.  What specifically are you trying to accomplish?

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
cm9361, line 1: I would expect that the signatures for callbacks like onRowSelect, onCellRangeSelect, and onClearSelectedCells would also have to be updated to return the section properly.

eawww:
Something I hadn't considered is when a selection exists in a section and then that section is collapsed. Since there's no indication on the header that the section contains selected cells, cells in a section should be unselected when that section is collapsed.

eawww:
> Should the anchor be reset when sections or column headers are selected?

I'd say yes. I think we also missed this case in Worklist.

eawww (commented):
Not sure if it was doing it before, but the screen reader is no longer saying the number of cells selected when selecting a range. When the selection state changes to include multiple cells selected, the aria live region should be updated to announce the number of cells currently selected.

eawww (approved):

chrismichalewicz:
Functionally Reviewed

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -268,6 +268,7 @@ const DataGrid = forwardRef((props, ref) => {
     () => ({
       setFocusedRowCol,
       getFocusedCell() { return { row: focusedRow, col: focusedCol }; },
+      getGridRef() { return grid.current; },
```
cm9361, line 271: We could just use a refcallback pattern similar to what we have with the Terra button.  I don't think we need to leverage the imperative handle hook here if we truly need access to the table.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -206,10 +207,17 @@ function FlowsheetDataGrid(props) {
     return newSections;
   }, [intl, sections]);
 
+  const flattenSections = (sectionsToFlatten) => sectionsToFlatten.flatMap(section => section.rows.map(row => ({
+    ...row,
+    sectionId: section.id,
+  })));
```
cm9361, line 213: If you needed to flatten the sections, wouldn't you want to memoize the value?  It looks like you are calling this method in multiple places.  However, the same value should be returned.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -227,35 +235,51 @@ function FlowsheetDataGrid(props) {
...
-  const selectCellRange = useCallback((rowIndex, columnIndex) => {
+  const selectCellRange = useCallback((rowId, columnId, sectionId) => {
     if (anchorCell.current === null) {
       return;
     }
 
-    const anchorRowIndex = rows.findIndex(row => row.id === anchorCell.current.rowId);
+    const rowsToSearch = flowsheetSections
+      ? flattenSections(flowsheetSections)
+      : rows;
```
cm9361, line 247: Instead of having to flatten the sections and try to figure out the position in the DOM, can't you just use a querySelector since you have a reference to the grid?  You know that the section id is set as an attribute on the tr element (I believe).  You can use that knowledge to find the bounds of the section.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -277,11 +305,28 @@ function FlowsheetDataGrid(props) {
...
+    let anchorSectionId = '';
+    if (flowsheetSections) {
+      for (let rowIdx = rowIndex; rowIdx > 0; rowIdx -= 1) {
+        const currentSectionId = gridRef.rows[rowIdx].getAttribute('data-section-id');
+        if (currentSectionId) {
+          anchorSectionId = currentSectionId;
+          break;
+        }
+      }
+    }
```
cm9361, line 319: Why can't the handleCellRangeSelection method provide the section id?  Wouldn't that already be available without you having to query the grid?

artpark (approved):

sycombs (approved):

cm9361 (approved):

PR #1937, opened 2023-12-07
[terra-folder-tree] Update cursor style for interactable regions

sycombs:
### Summary

This change updates the cursor style for folder tree rows to be a pointer so that there is a visual cue that the regions are clickable. This also updates the selection behavior so that folder rows can only be selected by the radio button and clicking anywhere else on the row will toggle expand/collapse, and non-folder rows can be selected by clicking anywhere in the row.

### Testing

This change was tested using:

- [ ] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9978 

---

@cerner/terra

### DIFF
--- a/packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
+++ b/packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
@@ -98,10 +98,16 @@ const FolderTreeItem = ({
     cx(
       'folder-tree-item',
       { selected: isSelected },
-      { folder: isFolder },
       theme.className,
     ),
   );
+
+  const handleRadioButtonClick = (event) => {
+    // Stop click propagation to prevent triggering expand/collapse when selecting folders
+    event.stopPropagation();
+
+    onClick();
+  };
 
   return (
     <>
@@ -110,15 +116,17 @@ const FolderTreeItem = ({
         role="treeitem"
         aria-expanded={isFolder ? isExpanded : null}
         aria-selected={isSelected}
-        onClick={onToggle}
-        onKeyDown={onToggle}
+        onClick={isFolder ? onToggle : onClick}
+        onKeyDown={isFolder ? onToggle : onClick}
       >
         <input
           type="radio"
           checked={isSelected}
           onChange={onClick}
+          onClick={handleRadioButtonClick}
           aria-hidden // Hiding the radio button from assistive technology since they cannot be grouped correctly
           tabIndex={-1} // Prevent tabbing to the button since it should not be read or acknowledged by assistive technology
+          className={cx('radio')}
         />
         {/* eslint-disable-next-line react/forbid-dom-props */}
         <span style={{ paddingLeft: `${level}rem` }}>
--- a/package-lock.json
+++ b/package-lock.json
@@ -44234,7 +44234,7 @@
       }
     },
     "packages/terra-table": {
-      "version": "5.2.1",
+      "version": "5.3.0",
       "license": "Apache-2.0",
       "dependencies": {
         "classnames": "^2.2.5",
--- a/packages/terra-folder-tree/CHANGELOG.md
+++ b/packages/terra-folder-tree/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Updated cursor style to pointer for clickable regions.
+
 ## 1.0.0-alpha.2 - (December 5, 2023)
 
 * Added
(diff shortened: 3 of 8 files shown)


packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -98,27 +98,35 @@ const FolderTreeItem = ({
...
-      { folder: isFolder },
       theme.className,
     ),
   );
 
+  const handleRadioButtonClick = (event) => {
+    // Stop click propagation to prevent triggering expand/collapse when selecting folders
+    event.stopPropagation();
+
+    onClick();
```
cm9361, line 109: Since onClick is not required, we either need to check here or when the event handler is defined to prevent an exception.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -98,27 +98,35 @@ const FolderTreeItem = ({
...
     <>
       <li
         className={itemClassNames}
         role="treeitem"
         aria-expanded={isFolder ? isExpanded : null}
         aria-selected={isSelected}
-        onClick={onToggle}
-        onKeyDown={onToggle}
+        onClick={isFolder ? onToggle : onClick}
+        onKeyDown={isFolder ? onToggle : onClick}
```
cm9361, line 120: Isn't there a specific key or keys that should trigger this action?  I would not expect this to happen for any key pressed.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -98,27 +98,35 @@ const FolderTreeItem = ({
...
       >
         <input
           type="radio"
           checked={isSelected}
           onChange={onClick}
+          onClick={handleRadioButtonClick}
```
cm9361, line 126: What would be the difference between onChange and onClick?  Do we need both event handlers?

eawww (approved):
Just to document but out of scope for this story: 
There's one example where we get the pointer cursor on an item that can't be selected or expanded, but this is an issue with the example. When consumed properly, all items would be interactable so pointer cursor everywhere is fine.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
sycombs: The `onChange` handler was originally used instead of `onClick` to resolve this [React failed proptype warning](https://stackoverflow.com/questions/36715901/reactjs-error-warning). Unfortunately, stopPropagation will not prevent the row click event if the radio button uses onChange since they are different event types, so I had to use an `onClick` instead.

An alternative to addressing the React warning would be to set `readOnly` on the radio, but that seems incorrect to me as well since it's not really a readonly element, so I've left it as `onChange` for now, but I'm open to other solutions if we don't want to have an extra event handler here.

sdadn (approved):

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
sycombs: Yes, I took a look at the keyboard Jira (UXPLATFORM-9776) and expand/collapse should be triggered by the right and left arrows and selection should be triggered via the enter key. I went ahead and removed the onKeyDown handler and disabled the eslint error for now, these should be re-enabled as part of the keyboard interaction story.
[ccdc63e](https://github.com/cerner/terra-framework/pull/1937/commits/ccdc63e3b8594b45b7e4d111070b6707dac7135d)

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
sycombs: Good catch, [deebcdc](https://github.com/cerner/terra-framework/pull/1937/commits/deebcdce1dfbe6dbdd44fd9e553aa3076b58383a)

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
sycombs: Yep, and that's much nicer, updated: [460c1f2](https://github.com/cerner/terra-framework/pull/1937/commits/460c1f253685252c571ae1926cf60d947cb24e8e)

cm9361 (approved):

PR #1950, opened 2023-12-15
Compact interactive list implementation guide added

adoroshk:
### Summary
This PR:
1. Adds the implementation guide to the terra-compact-interactive-list.
2. Update the examples descriptions and makes the Featured column in examples interactable.
3. Fixes focus defect for interactable cells which was discovered after the **featured** cells became interactable.
2. Per adding 

### Testing

This change was tested using:

- [ ] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

**This PR resolves:**

UXPLATFORM-9789


mjpalazzo:
@adoroshk - Great jog on the documentation. I have a few minor comments:
In the Design Decisions section, the following sentence is difficult to understand and has a typo. It needs to be simplified.
"For business reasons, the consumer may decide not to honor that cell selection and instead does nothing resulting in the the cell not getting selected in the Worklist Data Grid."
What is the impact to the user if the cell selection is not honored? Does the cell retain visual focus? Does the cell selection style change?
In the Example section, I have been recommending that if the example demonstrates a particular prop, then the Description should reference the prop and how it is set.

I have a separate question: is it still an issue of the cell contains a text input box?

mjpalazzo:
One more comment for something I encountered while testing keyboard interaction. 
The documentation says, "Selects the focused cell. Cell cannot be selected if it contains focusable elements in it."
 If I press Space on a cell that does not have an interactive component, nothing happens. What should happen? If the cell contains an interactive component Space key activates it.

mjpalazzo:
The CIL is working well but in testing with JAWS I noticed the following:
JAWS announces the role "grid" but does not announce the name of the component or any navigation helper text.
On first item of the CIL JAWS announces 3 columns and 6 rows and the first row as row2. There is no visible header so this is confusing.
Arrow keys work as required. JAWS announces the row header when I changed rows and the row index (although the index is incorrect).
With the exception of the Featured controls, JAWS announces the name role and state of the interactive elements and helper text. For the Featured control, what kind of control is it and how does the user interact with it?
Tab and Shift Tab work as required.
Space does not select the cell.
Using a MacBook on a Window cloud desktop made testing some of key combinations impossible to test. I validated them with Safari VoiceOver.
Escape works as required.

mjpalazzo:
Safari/VoiceOver worked similarly to JAWS with similar issues. I was able to test the Control Option Arrow and Control Command Arrow interactions and these worked as required. MacBook equivalents of Home and End keys (Command Right Arrow and Command Left Arrow did not work as documented.

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/About.1.doc.mdx
```diff
@@ -1,23 +1,105 @@
...
+
+The Compact Interactive List:
+  * Supports only one tab stop. When tabbing into the Compact Interactive List, focus will either go to the last visited cell, or the first cell upon initial entry.
+  While on the Compact Interactive List, `Tab` or `Shift+Tab` will exit the list.
+  * Does not maintain consumer data state. The state must be maintained by the consumer.
+  * Uses callbacks to communicate the actions taken by the user.
+  The callbacks are provided by the consumer and therefore the implementation is consumer specific.
+  It is up to the consumer to determine if and what actions should be taken.
+  For example, the user may click on a selectable cell in the Compact Interactive List which calls the _onCellSelect_ callback.
+  For business reasons, the consumer may decide not to honor that cell selection and instead does nothing resulting in the the cell not getting selected in the Worklist Data Grid.
```
sycombs, line 50: Can we reword this slightly for clarity? Also, should this say `Compact Interactive List`?

```suggestion
  For business reasons, the consumer may decide not to honor that cell selection and instead do nothing, resulting in the the cell not getting selected in the Compact Interactive List.
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/About.1.doc.mdx
```diff
@@ -1,23 +1,105 @@
...
+  * Supports single interactive element per cell.
```
sycombs, line 51: Nit:

```suggestion
  * Supports a single interactive element per cell.
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/About.1.doc.mdx
```diff
@@ -1,23 +1,105 @@
...
+|[ResponsiveColumns](/components/cerner-terra-framework-docs/compact-interactive-list/examples/responsive-columns)|An example of the Compact Interactive List with responsive text columns.|
+|[ScalableUnits](/components/cerner-terra-framework-docs/compact-interactive-list/examples/scalable-units)|An example of the Compact Interactive List with width props set in units relative to the font size.|
+|[FlowDirection](/components/cerner-terra-framework-docs/compact-interactive-list/examples/flow-direction)|The examples of the Compact Interactive List with vertical and horizontal flow directions.|
+|[NumberOfColumns](/components/cerner-terra-framework-docs/compact-interactive-list/examples/number-of-columns)|An example of the Compact Interactive List placed inside of a parent element that uses width breakpoints to dynamically change the number of columns.|
+|[CellContent](/components/cerner-terra-framework-docs/compact-interactive-list/examples/cell-content)|Compact Interactive List cell content examples.|
+
+## Accessibility
+### Keyboard Interactions
+
+List of keyboard interactions supported by Compact Interactive List.
```
sycombs, line 66: Nit: wondering if a colon is more appropriate here:

```suggestion
List of keyboard interactions supported by Compact Interactive List:
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/About.1.doc.mdx
```diff
@@ -27,14 +109,60 @@ This component requires the following peer dependencies be installed in your app
...
- * [Internationalization Support](https://engineering.cerner.com/terra-ui/about/terra-ui/component-standards#internationalization-i18n)
- * [LTR/RTL Support](https://engineering.cerner.com/terra-ui/about/terra-ui/component-standards#ltr--rtl)
+The supported units for width props are `px`, `em`, and `rem`. 
+The width units should be consistent across all the props in the same component. That includes:
+[_columnMinimumWidth_](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table), 
+[_columnMaximumWidth_](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table), 
+[_width_](/components/cerner-terra-framework-docs/compact-interactive-list/about#column), 
+[_minimumWidth_](/components/cerner-terra-framework-docs/compact-interactive-list/about#column), and 
+[_maximumWidth_](/components/cerner-terra-framework-docs/compact-interactive-list/about#column).
+The width units inconsistent with the rest of the component props will be disregarded.
```
sycombs, line 129: Is there an order of priority for which units will be used if there are inconsistencies?

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/FixedWidthColumns.1.doc.mdx
```diff
@@ -0,0 +1,15 @@
...
+
+# Fixed Width Columns
+
+### Description
+
+This example demonstrates a [Compact Interactive List](/components/cerner-terra-framework-docs/compact-interactive-list/about) which renders a collection of interactive data in a list format into a single tab stop. 
+
+* The list is arranged in 2 visual columns by the [_numberOfColumns_](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) prop set to `2`.
+* Each visual column consists of 3 semantic columns, such as Icon, Text Description, and an interactive Featured column.
+* All semantic columns have [_width_](/components/cerner-terra-framework-docs/compact-interactive-list/about#column) prop explicitly set, with the [_flexGrow_](/components/cerner-terra-framework-docs/compact-interactive-list/about#column) prop omitted. The columns will keep their width and not grow wider or shrink.
```
sycombs, line 11: Nit:
```suggestion
* All semantic columns have the [_width_](/components/cerner-terra-framework-docs/compact-interactive-list/about#column) prop explicitly set, with the [_flexGrow_](/components/cerner-terra-framework-docs/compact-interactive-list/about#column) prop omitted. The columns will keep their width and not grow wider or shrink.
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/FixedWidthColumns.1.doc.mdx
```diff
@@ -0,0 +1,15 @@
...
+* Horizontal scroll will show up if screen width is less than the width of the list.
```
sycombs, line 12: Does this make more sense to be "scrollbar"?
```suggestion
* Horizontal scrollbar will show up if screen width is less than the width of the list.
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/FixedWidthColumns.1.doc.mdx
```diff
@@ -0,0 +1,15 @@
...
+* The list is a one tab stop. The user can navigate between cells with the arrow keys. **Space key** press or mouse click on the Featured cell will toggle their state. **Esc key** press while on the list would unselect all featured items.
```
sycombs, line 13: Can we reword this slightly:
```suggestion
* The list is a one tab stop. The user can navigate between cells with the arrow keys. **Space key** press or mouse click on a Featured cell will toggle its state. **Esc key** press while on the list will unselect all featured items.
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/CellContent.6.doc.mdx
```diff
@@ -0,0 +1,24 @@
...
+  Since Compact Intercative List is one tab stop and arrow keys are used for navigation between cells, these keys are not available to navigate multiple focusable elements inside a cell. 
+* If a cell contains a focusable element inside of it, when navigated to the cell via arrow keys or mouse click, the focusable element will get the focus. 
+* The arrow key press when focused on a cell will navigate away from that cell to the new cell located in corresponding direction.
+
+### Cell selection
+
+This example demonstrates a [Compact Interactive List](/components/cerner-terra-framework-docs/compact-interactive-list/about) that has support for cell selection. 
+
+* A cell is selected by clicking on the cell or pressing the Spacebar on the focused cell. 
+* A cell is selectable unless it contains an interactable element, in that case the interactive element default controls apply.
```
sycombs, line 17: Nit:
```suggestion
* A cell is selectable unless it contains an interactable element, in which case the interactive element default controls apply.
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/ResponsiveColumns.2.doc.mdx
```diff
@@ -0,0 +1,51 @@
...
+* The minimum width would take the default value as [_minimumWidth_](/components/cerner-terra-framework-docs/compact-interactive-list/about#column) property is not set.
+* Once the list reaches its minimum width, defined by fixed width columns and the default minimum width of the flex growing column, the scroll shows up.
+
+<ResponsiveColumns />
+
+### Usage
+
+#### Creating a Responsive Column
+
+There are two ways to make a sematic column responsive:
```
sycombs, line 22: Nit: typo
```suggestion
There are two ways to make a semantic column responsive:
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/ResponsiveColumns.2.doc.mdx
```diff
@@ -0,0 +1,51 @@
...
+[_columnMinimumWidth_](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table), 
+[_columnMaximumWidth_](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table), 
+[_width_](/components/cerner-terra-framework-docs/compact-interactive-list/about#column), 
+[_minimumWidth_](/components/cerner-terra-framework-docs/compact-interactive-list/about#column), and 
+[_maximumWidth_](/components/cerner-terra-framework-docs/compact-interactive-list/about#column).
+The width units inconsistent with the rest of the component props will be disregarded.
+
+</Notice>
+
+In following example the text column has `maximumWidth: "400px"` and `minimumWidth: "200px"`:
```
sycombs, line 49: Nit:
```suggestion
In the following example, the text column has `maximumWidth: "400px"` and `minimumWidth: "200px"`:
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/FlowDirection.4.doc.mdx
```diff
@@ -0,0 +1,27 @@
...
+### Vertical Flow
+
+By default the items in the [Compact Interactive List](/components/cerner-terra-framework-docs/compact-interactive-list/about) go from top to bottom before breaking into a new visual column.
+
+<VerticalFlow />
+
+#### Usage
+
+* The list items needs the [_rowHeight_](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) property to be explicitly set for the vertical flow. If not set by user, the default value will be used.
+* The items will be distributed evenly between all the columns but the last one. The last column will accommodate the reminder after the number of items was divided by number of columns.
```
sycombs, line 15: Should this be "remainder"?
```suggestion
* The items will be distributed evenly between all the columns but the last one. The last column will accommodate the remainder after the number of items was divided by number of columns.
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/FlowDirection.4.doc.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+### Vertical Flow
+
+By default the items in the [Compact Interactive List](/components/cerner-terra-framework-docs/compact-interactive-list/about) go from top to bottom before breaking into a new visual column.
+
+<VerticalFlow />
+
+#### Usage
+
+* The list items needs the [_rowHeight_](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) property to be explicitly set for the vertical flow. If not set by user, the default value will be used.
```
sycombs, line 14: Nit:
```suggestion
* The list items need the [_rowHeight_](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) property to be explicitly set for the vertical flow. If not set by user, the default value will be used.
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/FlowDirection.4.doc.mdx
```diff
@@ -0,0 +1,27 @@
...
+* The items will be distributed evenly between all the columns but the last one. The last column will accommodate the reminder after the number of items was divided by number of columns.
+* The user must monitor that the number of rows is not divisible by [_numberOfColumns_](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) - 1 without a reminder, as it will lead to an empty last column.
```
nikhitasharma, line 16:
```suggestion
* The user must monitor that the number of rows is not divisible by [_numberOfColumns_](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) - 1 without a remainder, as it will lead to an empty last column.
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/NumberOfColumns.5.doc.mdx
```diff
@@ -0,0 +1,23 @@
...
+* If not explicitly set by user, the [_numberOfColumns_](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) defaults to 1.
+
+### Change the Number of Columns dynamically
+
+This example demonstrates a [Compact Interactive List](/components/cerner-terra-framework-docs/compact-interactive-list/about) being placed in the parent container, which uses the width breakpoints to change the [_numberOfColumns_](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) prop depending on the screen width. 
+Change the browser window width to see the number of columns change.
+
+<WidthBreakpoints />
+
+<Notice variant="caution" ariaLevel="2">
```
nikhitasharma, line 18:
```suggestion
<Notice variant="important" ariaLevel="2">
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/NumberOfColumns.5.doc.mdx
```diff
@@ -0,0 +1,23 @@
...
+
+With vertical flow the list items are distributed evenly between all the columns but the last one. The last column accommodates the reminder after the number of items is divided by number of columns.
```
nikhitasharma, line 20:
```suggestion
With vertical flow,  the list items are distributed evenly between all the columns but the last one. The last column accommodates the remainder after the number of items is divided by number of columns.
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/NumberOfColumns.5.doc.mdx
```diff
@@ -0,0 +1,23 @@
...
+The user must monitor that the number of rows is not divisible by `numberOfColumns - 1` without a reminder, as it will lead to an empty last column.
```
nikhitasharma, line 21:
```suggestion
The user must monitor that the number of rows is not divisible by `numberOfColumns - 1` without a remainder, as it will lead to an empty last column.
```

cldeitrich (approved):
[OHAIUAD-205125](https://jira3.cerner.com/browse/OHAIUAD-205125): Approved for style and formatting.

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/About.1.doc.mdx
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/About.1.doc.mdx
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/About.1.doc.mdx
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/About.1.doc.mdx
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/FixedWidthColumns.1.doc.mdx
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/FixedWidthColumns.1.doc.mdx
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/FixedWidthColumns.1.doc.mdx
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/CellContent.6.doc.mdx
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/ResponsiveColumns.2.doc.mdx
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/ResponsiveColumns.2.doc.mdx
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/FlowDirection.4.doc.mdx
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/FlowDirection.4.doc.mdx
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/About.1.doc.mdx
```diff
@@ -1,23 +1,105 @@
...
+Supports:
+* Arranging items in columns.
+* Fixed and responsive width columns.
+* `px`, `em`, and `rem` width units.
+* Vertical and horizontal flow.
+* Minimum and maximum column width for responsive columns set for all columns, or to each column individually.
+* Selectable cells.
+* Interactable elements in cells (one per cell).
+
+Refer to the [examples](/components/cerner-terra-framework-docs/compact-interactive-list/about#examples) section for detailed demonstration of the features supported by the Search Field.
```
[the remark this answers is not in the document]
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/About.1.doc.mdx
```diff
@@ -1,23 +1,105 @@
...
+|Down Arrow     | Moves focus one cell down. If focus is on the bottom cell of the visual column, it will move to the first cell of the next visual column. If focus is on the bottom cell of the last visual column, focus does not move.|
+|Right Arrow    | Moves focus one cell to the right. If focus is on the right-most cell in the visual row, focus does not move.|
+|Left Arrow     | Moves focus one cell to the left. If focus is on the left-most cell in the visual row, focus does not move.|
+|Home           | Moves focus to the first cell in the visual row that contains focus.|
+|End            | Moves focus to the last cell in the visual row that contains focus.|
+|Control + Home (Windows) <br/> or <br/> Ctrl + Command + Left Arrow (Mac) | Moves focus to the first cell in the first visual row.|
+|Control + End (Windows) <br/> or <br/> Ctrl + Command + Right Arrow (Mac) | Moves focus to the last cell in the last visual row.|
+|Tab            | Moves focus out of the list to the next focusable element in the tab sequence.|
+|Shift + Tab    | Moves focus out of the grid to the previous focusable element in the tab sequence.|
```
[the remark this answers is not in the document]
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/About.1.doc.mdx
```diff
@@ -1,23 +1,105 @@
...
+* [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby)
+  * **Compact Interactive List**: When the [_ariaLabelledBy_](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) prop is specified, the screen reader will use the _aria-labelledby_ element to determine the [accessible name](https://w3c.github.io/accname/#dfn-accessible-name) for the list.
+* [rowheader](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/rowheader_role)
+  * **Compact Interactive List**: When the [_rowHeaderIndex_](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) is provided, the screen reader will use the corresponding cell as the header cell for a row, establishing a relationship between it and the other cells in the same row.
```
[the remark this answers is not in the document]
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/About.1.doc.mdx
```diff
@@ -27,14 +109,60 @@ This component requires the following peer dependencies be installed in your app
...
+The width units inconsistent with the rest of the component props will be disregarded.
 
-## Compact Interactive List Props Table
+</Notice>
+
+### Compact Interactive List Props Table
 <CompactInteractiveListPropsTable />
+
+### Column
+A column specifies the data to render a cell in the header row of the Worklist Data Grid.
```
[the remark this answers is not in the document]
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/FlowDirection.4.doc.mdx
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/NumberOfColumns.5.doc.mdx
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/NumberOfColumns.5.doc.mdx
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/NumberOfColumns.5.doc.mdx
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/About.1.doc.mdx
```diff
@@ -27,14 +109,60 @@ This component requires the following peer dependencies be installed in your app
...
+
+|Name|Type|Is Required|Default Value|Description|
+|---|---|---|---|---|
+|**id**|string|<span style={{color:'#d53040'}}>required</span>|none|An identifier to uniquely identify the column within the grid.|
```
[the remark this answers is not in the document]
adoroshk: Great catch! Fixed: https://github.com/cerner/terra-framework/pull/1950/commits/7f4e33181e5c3a0660da70b7f7718f7ecb3ff576

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/About.1.doc.mdx
```diff
@@ -1,23 +1,105 @@
...
+  * **Compact Interactive List**: When the [_ariaLabelledBy_](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) prop is specified, the screen reader will use the _aria-labelledby_ element to determine the [accessible name](https://w3c.github.io/accname/#dfn-accessible-name) for the list.
+* [rowheader](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/rowheader_role)
+  * **Compact Interactive List**: When the [_rowHeaderIndex_](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) is provided, the screen reader will use the corresponding cell as the header cell for a row, establishing a relationship between it and the other cells in the same row.
+* [columnheader](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/columnheader_role)
+  * **Column**: When the [_displayedName_](/components/cerner-terra-framework-docs/compact-interactive-list/about#column) is specified, the screen reader will use it to define as the header information for the semantic column.
```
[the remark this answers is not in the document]
adoroshk: Great suggestion, and much appreciated, but it looks like all wording and styling comment were outdated per this commit by @cldeitrich: https://github.com/cerner/terra-framework/pull/1950/commits/62c72c0098e6edb05ec923b22cf997650ac247bf

adoroshk:
> Safari/VoiceOver worked similarly to JAWS with similar issues. I was able to test the Control Option Arrow and Control Command Arrow interactions and these worked as required. MacBook equivalents of Home and End keys (Command Right Arrow and Command Left Arrow did not work as documented.

@mjpalazzo - can you please give some more details when Command Right Arrow and Command Left Arrow did not work as documented. I test on Macbook and they work for me, so I am trying to figure out how to reproduce the error.

adoroshk:
> The CIL is working well but in testing with JAWS I noticed the following: JAWS announces the role "grid" but does not announce the name of the component or any navigation helper text. On first item of the CIL JAWS announces 3 columns and 6 rows and the first row as row2. There is no visible header so this is confusing. Arrow keys work as required. JAWS announces the row header when I changed rows and the row index (although the index is incorrect). With the exception of the Featured controls, JAWS announces the name role and state of the interactive elements and helper text. For the Featured control, what kind of control is it and how does the user interact with it? Tab and Shift Tab work as required. Space does not select the cell. Using a MacBook on a Window cloud desktop made testing some of key combinations impossible to test. I validated them with Safari VoiceOver. Escape works as required.

@mjpalazzo - thank you for such detailed response, much appreciated. I made some changes and have some questions below.

**JAWS announces the role "grid" but does not announce the name of the component or any navigation helper text.**  - I have added the labels with the name of the component. That made it consistant with the Worklist Data Grid and Flowsheet, as they announce their names only. But per this comment https://github.com/cerner/terra-framework/pull/1919#issuecomment-1850863713 I can also add "Use arrow keys to navigate and space or enter to activate a cell" to the arrow label. Should I add that text or just leave it as is?

**On first item of the CIL JAWS announces 3 columns and 6 rows and the first row as row2. There is no visible header so this is confusing.** You are right that it is happening because of the header. My implementation is consistent with Flowsheet with hidden headers, which also has a similar problem with JAWS announcing the number of rows including headers.

I investigated the issue and haven't found a solution. The only way I found to exclude the headers from the row count is by making them hidden from assistive technologies, which basically removes the column context from keyboard navigation. Here is what I tried and failed:
1. Changing `aria-rowcount` to number smaller than the actual number of rows + header to exclude the header didn't work. The `aria-rowcount` attribute set works only for the number of rows which exceeds the real number of row displayed, which is mentioned in the documentation.
2. Changing `aria-rowidex` - the same problem, works only if the `rowindex` exceeds the real `rowindex`, otherwise ignored.
Currently I have no solution for that and if such exists it needs to be implemented for the `terra-table` and `terra-flowsheet` as well.

**Space does not select the cell.** - fixed the cell selection example to change the color of selected cell.

adoroshk:
> @adoroshk - Great jog on the documentation. I have a few minor comments: In the Design Decisions section, the following sentence is difficult to understand and has a typo. It needs to be simplified. "For business reasons, the consumer may decide not to honor that cell selection and instead does nothing resulting in the the cell not getting selected in the Worklist Data Grid." What is the impact to the user if the cell selection is not honored? Does the cell retain visual focus? Does the cell selection style change? In the Example section, I have been recommending that if the example demonstrates a particular prop, then the Description should reference the prop and how it is set.
> 
> I have a separate question: is it still an issue of the cell contains a text input box? The last example includes a cell with a text box and it is currently a keyboard trap. Should we note this until it is fixed?

@mjpalazzo - thank you for such a detailed review. Here is my response to the questions:
1.  **"For business reasons, the consumer may decide not to honor that cell selection and instead does nothing resulting in the the cell not getting selected in the Worklist Data Grid."** This part has been changed in the documentation.
2. **In the Example section, I have been recommending that if the example demonstrates a particular prop, then the Description should reference the prop and how it is set.** - agreed on that, the props were edited by @cldeitrich and me to reference themselves in prop tables.
3. **I have a separate question: is it still an issue of the cell contains a text input box? The last example includes a cell with a text box and it is currently a keyboard trap. Should we note this until it is fixed?** - added a note that reads:

- Only one interactive element per cell is allowed. Because the Compact Intercative List component is one tab stop and the arrow keys are used to move between cells, these keys are not available to move to multiple interactive elements in a cell.
- Focusable elements in the cells are limited to buttons and anchors. In current implementation the cell does not trap focus, all focusable elements keep their native arrow key events. Placing text input, text area or select elements in a cell will make it impossible to navigate away from a cell via arrow keys.

adoroshk:
@mjpalazzo, **For the Featured control, what kind of control is it and how does the user interact with it?**

The feature column contains an icon, per space or click it changes from Featured ON to OFF. Per your comment I realized that might be not the best example, should I change it?

mjpalazzo:
> @mjpalazzo, **For the Featured control, what kind of control is it and how does the user interact with it?**
> 
> The feature column contains an icon, per space or click it changes from Featured ON to OFF. Per your comment I realized that might be not the best example, should I change it?

My recollection is the "featured" icon was used because it is one commonly used by consumers. If true, we should keep it. @eawww - what are your thoughts? Should we keep the "featured" icon example or get rid of it?

mjpalazzo:
> > @adoroshk - Great jog on the documentation. I have a few minor comments: In the Design Decisions section, the following sentence is difficult to understand and has a typo. It needs to be simplified. "For business reasons, the consumer may decide not to honor that cell selection and instead does nothing resulting in the the cell not getting selected in the Worklist Data Grid." What is the impact to the user if the cell selection is not honored? Does the cell retain visual focus? Does the cell selection style change? In the Example section, I have been recommending that if the example demonstrates a particular prop, then the Description should reference the prop and how it is set.
> > I have a separate question: is it still an issue of the cell contains a text input box? The last example includes a cell with a text box and it is currently a keyboard trap. Should we note this until it is fixed?
> 
> @mjpalazzo - thank you for such a detailed review. Here is my response to the questions:
> 
> 1. **"For business reasons, the consumer may decide not to honor that cell selection and instead does nothing resulting in the the cell not getting selected in the Worklist Data Grid."** This part has been changed in the documentation.
> 2. **In the Example section, I have been recommending that if the example demonstrates a particular prop, then the Description should reference the prop and how it is set.** - agreed on that, the props were edited by @cldeitrich and me to reference themselves in prop tables.
> 3. **I have a separate question: is it still an issue of the cell contains a text input box? The last example includes a cell with a text box and it is currently a keyboard trap. Should we note this until it is fixed?** - added a note that reads:
> 
> * Only one interactive element per cell is **allowed** perhaps use "supported" instead]. Because the Compact **Intercative** [spelling issue, sb **Interactive**] List component is one tab stop and the arrow keys are used to move between cells, these keys are not available to move to multiple interactive elements in a cell.
> * Focusable elements in the cells are limited to buttons and anchors. **In current implementation the cell does not trap focus, all focusable elements keep their native arrow key events. Placing text input, text area or select elements in a cell will make it impossible to navigate away from a cell via arrow keys.** [I suggest simplifying to "Text input, text area, and select elements are not supported and cannot be used." I think explaining why makes it harder for the reader to get the message.]

@adoroshk I have a couple of wording suggestions and minor edits.

packages/terra-compact-interactive-list/src/proptypes/cellShape.js
```diff
@@ -5,6 +5,11 @@ const cellShape = PropTypes.shape({
    * Content that will be rendered within the Cell.
    */
   content: PropTypes.node,
+
+  /**
+   * Boolean value indicating whether or not the cell should render as selected.
+   */
+  isSelected: PropTypes.bool,
```
cm9361, line 12: Shouldn't the cell shape have an isSelectable prop?  Wouldn't the cell be selected?

packages/terra-compact-interactive-list/src/proptypes/columnShape.js
```diff
@@ -48,6 +48,10 @@ const columnShape = PropTypes.shape({
     alignTypes.LEFT,
   ]),
 
+  /**
+   * Boolean value indicating whether or not the cells in this column are selectable.
+   */
+  isSelectable: PropTypes.bool,
```
cm9361, line 54: I would expect the isSelectable prop to be on the cells themselves.  What if the consumer doesn't want every cell in the column to behave the same way?

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
```diff
@@ -112,10 +122,14 @@ const Cell = (props) => {
     }
   };
 
-  const handleMouseDown = () => {
+  const handleMouseDown = (event) => {
     setFocusedCell({ rowId, columnId: id });
     if (isSelectableCell && onCellSelect) {
-      onCellSelect({ rowId, columnId: id });
+      event.preventDefault();
```
cm9361, line 128: Why do we need to prevent the default behavior of a mouse down event?

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
```diff
@@ -112,10 +122,14 @@ const Cell = (props) => {
...
+      if (onCellSelect) {
+        onCellSelect({ rowId, columnId: id });
+        cellRef?.current.focus();
```
cm9361, line 131: The component should get focus just because it was clicked.   I would not expect you to need this logic as long as the tab index is set properly.

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
```diff
@@ -112,10 +122,14 @@ const Cell = (props) => {
...
   };
 
-  const handleMouseDown = () => {
+  const handleMouseDown = (event) => {
     setFocusedCell({ rowId, columnId: id });
     if (isSelectableCell && onCellSelect) {
-      onCellSelect({ rowId, columnId: id });
+      event.preventDefault();
+      if (onCellSelect) {
+        onCellSelect({ rowId, columnId: id });
```
cm9361, line 130: If a cell is not selectable, we should not trigger the callback, correct?

cm9361:
The focus indicator styles do not seem correct for the Orion Fusion theme.

cm9361:
Question (Not for this PR): Does this component need to support zebra striping since the data will be tabular in nature?

cm9361:
Should we default the isSelectable prop to true?  I am not sure what the primary use case would be.

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
```diff
@@ -71,14 +76,19 @@ const Cell = (props) => {
...
     isRowHeader,
   } = props;
 
   const theme = useContext(ThemeContext);
   const cellRef = React.useRef();
   const [isSelectableCell, setIsSelectableCell] = useState(false);
-  const className = cx('cell', theme.className, { selectable: isSelectableCell });
+  const selectable = column.isSelectable && isSelectableCell;
+  const className = cx('cell', theme.className, {
+    selected: selectable && isSelected,
```
cm9361, line 89: I don't think the selected state should depend on the cell being selectable.  I could programmatically select a cell without allow a user to do so.

packages/terra-compact-interactive-list/tests/jest/__snapshots__/CompactInteractiveList.test.jsx.snap
cm9361, line 1: Do we need this many snapshots?  If the validation can occur with the expect statements, we can remove the snapshots.

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/CellContent.jsx
```diff
@@ -0,0 +1,143 @@
...
+const CellContent = () => {
+  const [displayedRows, setDisplayedRows] = useState(updateRows(rows, cols));
+
+  const onCellSelect = ({ rowId, columnId }) => {
+    setDisplayedRows(updateRows(displayedRows, cols, { rowId, columnId }));
+  };
+
+  const onClearSelection = () => {
+    setDisplayedRows(updateRows(rows, cols));
+  };
```
cm9361, line 126: You should be using the useCallback hook for these methods.  Same comment for any method passed in your other examples.  Your current implementation will cause any component accepting the method as a prop to rerender even if memoized.

packages/terra-compact-interactive-list/tests/jest/__snapshots__/CompactInteractiveList.test.jsx.snap
adoroshk: This are basically 2 snapshots per initial rendering for 2 major cases - fixed width columns and flex columns, added per request of @nikhitasharma. I don't mind removing them, just wanted to double-check what is our policy on that - do we need initial snapshots or we are moving away from that practice?

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
adoroshk: In Compact Interactive list the cell should become non-selectable if after the component rendered it was detected that cell has intractable elements in it (like buttons). As that becomes known only after the cell was rendered, I am not sure how to change that, an example would be appreciated.

packages/terra-compact-interactive-list/src/proptypes/cellShape.js
adoroshk: Currently it is determined for the whole column rather than each cell and included in the column shape. The cell selection mechanism might be a subject to change in the future, current implementation might change in the future as it wasn't completely clear what is the best way for it.

packages/terra-compact-interactive-list/src/proptypes/columnShape.js
adoroshk: Currently it is determined for the whole column rather than each cell and included in the column shape. The cell selection mechanism might be a subject to change in the future, current implementation might change in the future as it wasn't completely clear what is the best way for it. Improving that is out of scope of this JIRA and rather a subject for the future improvements, as it needs discussion with @eawww

adoroshk:
> Question (Not for this PR): Does this component need to support zebra striping since the data will be tabular in nature?

I asked that question before, the answer was no, no zebra striping.

adoroshk:
> Should we default the isSelectable prop to true? I am not sure what the primary use case would be.

It might be a good idea, but as I mentioned before - the cell select-ability here is very basic and is a subject to be changed, however, it's out of scope for this JIRA.

adoroshk:
> * Focusable elements in the cells are limited to buttons and anchors. **In current implementation the cell does not trap focus, all focusable elements keep their native arrow key events. Placing text input, text area or select elements in a cell will make it impossible to navigate away from a cell via arrow keys.** [I suggest simplifying to "Text input, text area, and select elements are not supported and cannot be used." I think explaining why makes it harder for the reader to get the message.]

@mjpalazzo, thank you for that suggestion, I do think it would improve it. It looks like the text was changed by @cldeitrich in this commit though: https://github.com/cerner/terra-framework/commit/9e44ae7a02292120708fc7ff17075649f3807b10

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
cm9361: If the consumer sets the isSelected prop, won't the cell be selected regardless?

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
adoroshk: I added that line after encountering an issue when the featured icon was placed inside of a cell. Even though the icon wasn't an interactive element, per click on the icon the cell would lose the focus. By adding `event.preventDefault()` that problem resolved. As this `event.preventDefault()` is called only for sells that are select-able (which means they do not have interactive elements in them), that should basically change nothing for the cell.

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
adoroshk: This line was added to fix the same issue I described above. When the featured icon was placed inside of a cell, even though the icon wasn't an interactive element, per click on the icon the cell would lose the focus. Adding `event.preventDefault()` and `cellRef?.current.focus()` resolved the problem.

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/CellContent.6.doc.mdx
```diff
@@ -0,0 +1,31 @@
...
+
+* The [content](/components/cerner-terra-framework-docs/compact-interactive-list/about#cell) of a cell in the [Compact Interactive List](/components/cerner-terra-framework-docs/compact-interactive-list/about) component may be empty, text, numerical, or any combination of widgets, but only one interactive element per cell is allowed.
+* If a cell contains an interactive element, the interactive element becomes active when you select the cell. 
+* If you press an arrow key when a cell is active, the next or previous cell becomes active.
+
+### Cell Selection
+
+The example below demonstrates a [Compact Interactive List](/components/cerner-terra-framework-docs/compact-interactive-list/about) component that supports cell selection. 
+
+* You can select a cell or press the **SPACERBAR** on the active cell to select it.
```
eawww, line 23:
```suggestion
* You can click a cell or press the **SPACERBAR** on the active cell to select it. 
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/FixedWidthColumns.1.doc.mdx
```diff
@@ -0,0 +1,15 @@
+import FixedWidthColumns from './FixedWidthColumns?dev-site-examples';
+
+# Fixed Width Columns
+
+### Description
+
+This example demonstrates a [Compact Interactive List](/components/cerner-terra-framework-docs/compact-interactive-list/about) component that renders a collection of interactive information in a list format that has a single tab stop. 
+
+* The [numberOfColumns](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) property is set to **2** to arrange the list in two visual columns.
+* Each visual column has three semantic columns, such as an icon, a text description, and an interactive element column.
```
eawww, line 10:
```suggestion
* Each visual column has three semantic columns: an icon, a text description, and an interactive element column.
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/FixedWidthColumns.1.doc.mdx
```diff
@@ -0,0 +1,15 @@
...
+* All semantic columns have the [width](/components/cerner-terra-framework-docs/compact-interactive-list/about#column) property explicitly set, with the [flexGrow](/components/cerner-terra-framework-docs/compact-interactive-list/about#column) property omitted. The columns keep their width and do not expand or shrink.
+* A horizontal scrollbar is displayed if the screen width is less than the width of the list.
+* The list is a one-tab stop. You can move between cells with the arrow keys. Select the interactive element or use the arrow keys and press the **SPACEBAR** to change the element's state. Press the **ESC** key while in the list to clear all featured items.
```
eawww, line 13:
```suggestion
* The list is one-tab stop. You can move focus between cells with the arrow keys or by clicking cells. Click the interactive element or use the arrow keys and press the **SPACEBAR** to change the element's state. Press the **ESC** key while in the list to clear all selected items.
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/FlowDirection.4.doc.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+By default, the items in the [Compact Interactive List](/components/cerner-terra-framework-docs/compact-interactive-list/about) component are read from top to bottom before moving to a new visual column.
+
+<VerticalFlow />
+
+#### Usage
+
+* The [rowHeight](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) property is required for the list items to use vertical flow. If not set, the default value is used.
+* The items are distributed evenly between all columns but the last one. The last column accommodates the remainder after the number of items is divided by the number of columns.
+* Ensure that the number of rows is not divisible by the [numberOfColumns](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) value minus 1 without a remainder, which leads to an empty last column.
```
eawww, line 16: This isn't necessarily something a consumer _shouldn't_ do. It's just a kind of weird side effect of how it works.
```suggestion
* Note: If the number of rows is divisible by the [numberOfColumns](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) value minus 1 without a remainder, the last column will be empty.
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/NumberOfColumns.5.doc.mdx
```diff
@@ -0,0 +1,23 @@
+import { Notice } from "@cerner/terra-docs";
+import WidthBreakpoints from './WidthBreakpoints?dev-site-examples';
+
+# Number of Columns
+
+<Notice variant="important" ariaLevel="2">
+
+With vertical flow, the list items are distributed evenly between all columns but the last one. The last column accommodates the remainder after the number of items is divided by the number of columns.
+Ensure that the number of rows is not divisible by the numberOfColumns value minus 1 without a remainder, which leads to an empty last column.
```
eawww, line 9:
```suggestion
Note: If the number of rows is divisible by the [numberOfColumns](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table) value minus 1 without a remainder, the last column will be empty.
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/ResponsiveColumns.2.doc.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+<ResponsiveColumns />
+
+### Usage
+
+#### Create a Responsive Column
+
+Use one of the following methods to make a sematic column responsive:
+
+* Omitt the [width](/components/cerner-terra-framework-docs/compact-interactive-list/about#column) property to make a semantic column expand or shrink to fill all available space.
```
eawww, line 24:
```suggestion
* Omit the [width](/components/cerner-terra-framework-docs/compact-interactive-list/about#column) property to make a semantic column expand or shrink to fill all available space.
```

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/ScalableUnits.3.doc.mdx
```diff
@@ -0,0 +1,23 @@
...
+<Notice variant="important" ariaLevel="3">
+
+The supported units for width properties are `px`, `em`, and `rem`. 
+Ensure that the width units are consistent across all properties in the same component. The width properties are
+[columnMinimumWidth](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table), 
+[columnMaximumWidth](/components/cerner-terra-framework-docs/compact-interactive-list/about#compact-interactive-list-props-table), 
+[width](/components/cerner-terra-framework-docs/compact-interactive-list/about#column), 
+[minimumWidth](/components/cerner-terra-framework-docs/compact-interactive-list/about#column), and 
+[maximumWidth](/components/cerner-terra-framework-docs/compact-interactive-list/about#column).
+Any width units inconsistent with the rest of the component properties are disregarded.
```
eawww, line 15: Does something weird happen if there's no most common one?

adoroshk: > Does something weird happen if there's no most common one?

No, it'll work. But if first cells have width in `px`, then second's cell width is in `em`, the width in `em` will be disregarded and a console warning will show up. That will make the second column a column without width. Thise columns just grow and shrink instead of keeping their width.

eawww:
> > @mjpalazzo, **For the Featured control, what kind of control is it and how does the user interact with it?**
> > The feature column contains an icon, per space or click it changes from Featured ON to OFF. Per your comment I realized that might be not the best example, should I change it?
> 
> My recollection is the "featured" icon was used because it is one commonly used by consumers. If true, we should keep it. @eawww - what are your thoughts? Should we keep the "featured" icon example or get rid of it?

I think that's a fine thing to have but it could be better as a utility button that looks the same as this. Dunno if that would be too much extra logic for the example.

mjpalazzo:
> > > @adoroshk - Great job on the documentation. I have a few minor comments: In the Design Decisions section, the following sentence is difficult to understand and has a typo. It needs to be simplified. "For business reasons, the consumer may decide not to honor that cell selection and instead does nothing resulting in the the cell not getting selected in the Worklist Data Grid." What is the impact to the user if the cell selection is not honored? Does the cell retain visual focus? Does the cell selection style change? In the Example section, I have been recommending that if the example demonstrates a particular prop, then the Description should reference the prop and how it is set.
> > > I have a separate question: is it still an issue of the cell contains a text input box? The last example includes a cell with a text box and it is currently a keyboard trap. Should we note this until it is fixed?
> > 
> > 
> > @mjpalazzo - thank you for such a detailed review. Here is my response to the questions:
> > 
> > 1. **"For business reasons, the consumer may decide not to honor that cell selection and instead does nothing resulting in the the cell not getting selected in the Worklist Data Grid."** This part has been changed in the documentation.
> > 2. **In the Example section, I have been recommending that if the example demonstrates a particular prop, then the Description should reference the prop and how it is set.** - agreed on that, the props were edited by @cldeitrich and me to reference themselves in prop tables.
> > 3. **I have a separate question: is it still an issue of the cell contains a text input box? The last example includes a cell with a text box and it is currently a keyboard trap. Should we note this until it is fixed?** - added a note that reads:
> > 
> > 
> > * Only one interactive element per cell is **allowed** perhaps use "supported" instead]. Because the Compact **Intercative** [spelling issue, sb **Interactive**] List component is one tab stop and the arrow keys are used to move between cells, these keys are not available to move to multiple interactive elements in a cell.
> > * Focusable elements in the cells are limited to buttons and anchors. **In current implementation the cell does not trap focus, all focusable elements keep their native arrow key events. Placing text input, text area or select elements in a cell will make it impossible to navigate away from a cell via arrow keys.** [I suggest simplifying to "Text input, text area, and select elements are not supported and cannot be used." I think explaining why makes it harder for the reader to get the message.]
> 
> @adoroshk I have a couple of wording suggestions and minor edits.

Safari/VoiceOver Keyboard interactions are working as required. Thanks @adoroshk for helping with the correct key combinations.

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
adoroshk: UPD: I changed that logic in this commit: https://github.com/cerner/terra-framework/pull/1950/commits/75f1f241355ea225b2b4cfb4ef22ac221df279d3

The logic is needed to prevent the click on the cell that has a button in it (the button should be clicked, but not the cell that contains it)

Here is how it works now:
`if (isSelectableCell) {
      if (onCellSelect) {
        onCellSelect({ rowId, columnId: id });
      }
    } else {
      // allows clickable elements inside non-selactable cells to be clicked, but not the cell itself
      event.preventDefault();
    }
  };`

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
adoroshk: UPD: update explained here: https://github.com/cerner/terra-framework/pull/1950#discussion_r1433007127

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
adoroshk: That is correct, there is if statement to check if a cell is selectable above:https://github.com/cerner/terra-framework/blob/75f1f241355ea225b2b4cfb4ef22ac221df279d3/packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx#L127-L130

packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx
adoroshk: No, if the cell has an interactable element (like button) inside, it will not be selectable per design from @eawww

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/CellContent.jsx
```diff
@@ -0,0 +1,143 @@
...
+    id: 'Column-3',
+    displayName: 'Featured',
+    width: '50px',
+    align: alignTypes.CENTER,
+    isSelectable: true,
+  },
+];
+
+const CellContent = () => {
+  const [displayedRows, setDisplayedRows] = useState(updateRows(rows, cols));
```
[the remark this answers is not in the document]
adoroshk: Removed in all examples https://github.com/cerner/terra-framework/pull/1950/commits/12cd0f75a670a84f53dab11ed679c9878c741526

adoroshk:
> The focus indicator styles do not seem correct for the Orion Fusion theme.

@cm9361  Good catch, updated here: https://github.com/cerner/terra-framework/pull/1950/commits/bfcc3e3dde83392ad18180b0ae6cd9cea79341a7

adoroshk:
> > > @mjpalazzo, **For the Featured control, what kind of control is it and how does the user interact with it?**
> > > The feature column contains an icon, per space or click it changes from Featured ON to OFF. Per your comment I realized that might be not the best example, should I change it?
> > 
> > 
> > My recollection is the "featured" icon was used because it is one commonly used by consumers. If true, we should keep it. @eawww - what are your thoughts? Should we keep the "featured" icon example or get rid of it?
> 
> I think that's a fine thing to have but it could be better as a utility button that looks the same as this. Dunno if that would be too much extra logic for the example.

Updated the examples, replaced the featured icon with Featured button with utility icon

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/CellContent.jsx
```diff
@@ -0,0 +1,147 @@
...
+    };
+    return newRow;
+  })];
+};
+
+const FeaturedIcon = () => {
+  const [isFeatured, setIsFeatured] = useState(false);
+  const onButtonClick = () => setIsFeatured(!isFeatured);
+  return (
+    isFeatured ? <Button variant="utility" text="Featured button" icon={<IconFeatured />} onClick={onButtonClick} /> : <Button variant="utility" text="Featured off button" icon={<IconFeaturedOff />} onClick={onButtonClick} />
```
eawww, line 35: Making the name of the button the action that it performs when activated.
```suggestion
    isFeatured ? <Button variant="utility" text="Unfavorite item" icon={<IconFeatured />} onClick={onButtonClick} /> : <Button variant="utility" text="Favorite item" icon={<IconFeaturedOff />} onClick={onButtonClick} />
```

packages/terra-compact-interactive-list/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Updated `handleMouseDown` method for a cell.
```
nikhitasharma, line 6: This changelog needs to be updated to reflect other updates done.

eawww:
Looking great! Some styling notes:
Background color in lowlight theme to #2D3539 and the borders from white to black
Outline offset on the focus indicator from -2 to -4 for lowlight and default themes so there's a little teeny 1px gap between the outline and the border of the cell.

packages/terra-compact-interactive-list/src/orion-fusion-theme/Cell.module.scss
```diff
@@ -1,5 +1,11 @@
 :local {
   .orion-fusion-theme {
     --terra-compact-interactive-list-hover-background-color: #e0f2fb;
+    --terra-compact-interactive-list-cell-selected-background-color: #cbe7fa;
+
+    &.cell:focus {
+      --terra-compact-interactive-list-cell-focus-outline: none;
+      box-shadow: inset 0 0 1px 3px rgba(76, 178, 233, 0.5), inset 0 0 7px 4px rgba(76, 178, 233, 0.35);
```
nikhitasharma, line 8: Should there be wdio tests to check for focus indicator?

packages/terra-framework-docs/src/terra-dev-site/doc/compact-interactive-list/Examples.2/CellContent.jsx
adoroshk: Good catch, added in https://github.com/cerner/terra-framework/pull/1950/commits/a05da56ab53470a876d0d2de33585afa8fa78a50

The rest of examples don't have methods, as they don't have onCellSelect method passed

packages/terra-compact-interactive-list/src/orion-fusion-theme/Cell.module.scss
adoroshk: Yes, you are absolutely right. This PR grew to be much bigger than it was supposed to be and at this moment following wdio tests needs to be added:
1. Cell focus indicator.
2. Buttons and links inside of a cell focus indicator (should be their native focus styling). 
3. Selected cell background change.
I'll raise that question with the team, whether it should be done as a separate PR (as this one grew up to be a monstrosity and hard to maintain) or needs to be included in this PR.

packages/terra-compact-interactive-list/CHANGELOG.md
adoroshk: Good catch! Updated here: https://github.com/cerner/terra-framework/pull/1950/commits/bcd1cd027dbd6870c571197300632e6f6eb42950

adoroshk:
> Looking great! Some styling notes: Background color in lowlight theme to #2D3539 and the borders from white to black Outline offset on the focus indicator from -2 to -4 for lowlight and default themes so there's a little teeny 1px gap between the outline and the border of the cell.

Fixed styling in https://github.com/cerner/terra-framework/pull/1950/commits/f736c5125d0dd4383d101009e9f058f02ff28474
I have renamed  `CompactInteractiveList.module.css` to be `Row.module.css`, as it has only row styles for better readability, but the style changes are only related to the mentioned change, everything else stayed.

nikhitasharma (approved):

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
```diff
@@ -223,8 +223,9 @@ const CompactInteractiveList = (props) => {
   };
 
   const onFocus = (event) => {
-    if (!event.currentTarget.contains(event.relatedTarget)) {
+    if (!event.currentTarget.contains(event.relatedTarget) || event.target === listRef?.current) {
```
cm9361, line 226: Why do we need the OR condition.  Won't this cause the event to trigger every time focus is changed within the list?  If that is what you truly need, then there is no purpose for the first part of this condition.

cm9361:
This does not seem like the correct focus indicator for the icon cell for the default theme
Screenshot 2023-12-21 at 8 12 38 AM

cm9361:
The hyperlink in your example does not have a focus indicator when it has focus.  Is it just an issue with the control being used or styling?

adoroshk:
> The hyperlink in your example does not have a focus indicator when it has focus. Is it just an issue with the control being used or styling? I did more testing. The first time I enter the grid, the hyperlinks do not show the indicator. If I tab out of the control and back in, the focus indicator begins to appear.

Per documentation purposes, there is a known bug for element.focus() not always working in Chrome and Safari. In current implementation it affects native anchors created with <a> tag in both, Safari and Chrome and <button> (Safari only) obtaining focus styling per arrow key navigation on first time after page was loaded. 

The `terra-button` and `terra-hyperlink` works at any cases, so the examples and documentation have been updated to use them as supported elements.

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
adoroshk: reverted adding that as I can't reproduce the bug it was meant to fix anymore

packages/terra-compact-interactive-list/src/orion-fusion-theme/Cell.module.scss
adoroshk: Here is a JIRA for tracking: UXPLATFORM-10113

eawww (approved):
Great job, Alla!!!

packages/terra-compact-interactive-list/src/subcomponents/Row.module.scss
```diff
@@ -1,6 +1,6 @@
 // Themes
-@import '../orion-fusion-theme/CompactInteractiveList.module';
-@import '../clinical-lowlight-theme/CompactInteractiveList.module';
+@import '../clinical-lowlight-theme/Row.module';
+@import '../clinical-lowlight-theme/Row.module';
```
cm9361, line 3: Could this be a typo?  It looks like there are two of the same imports.

adoroshk: Good catch, fixed: https://github.com/cerner/terra-framework/pull/1950/commits/4bd839ac7e43b2c77fb8d914072df0cf1ad3b8ee

cm9361 (approved):

artpark (approved):

PR #1955, opened 2023-12-18
[Terra-Pills] Resolves issue with screen reader not identifying pill status after removing the last pill

supreethmr:
### Summary

Resolves issue with screen reader not identifying pill removed and add status of the last pill

**What was changed:**
Added back tabIndex for group so that focus remains on group instead of body element when group is empty. 

**Why it was changed:**
Issue 1:
When there is only one pill present and that pill it removed usig backspace/delete key, JAWS doesn't read out the removed item.
After that when a new pill it added, JAWS starts reading information related to the deleted pill
 
Issue 2:
Earlier when a pill is added, JAWS would read out the name of the added pill.
However in 1.15.1, when pills are added to an empty pillbox, JAWS doesn't read the name of the added pills. It just reads the total number of pills.
If the pillbox is not empty, then JAWS is reading the added pill.

### Testing

This change was tested using:

- [ ] WDIO
- [x] Jest
- [X] Other (please describe below)
- [ ] No tests are needed
Verified with Screen reader tools JAWS and VO

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

[UXPLATFORM-10095]() 

---

@cerner/terra


saket2403 (approved):

rahulkumar8cs:
+1

PR #1956, opened 2023-12-18
[terra-folder-tree] Keyboard navigation + focus indicator

artpark:
### Summary

**What was changed:**
Added keyboard navigation and updated the focus indicator to match other components. ***Note that the action header's focus indicator is unchanged and needs to be updated in terra-core***.

Keyboard Navigation:

Right Arrow: 
When focus is on a closed node, opens the node, and focus does not move
When focus is on an open node, moves focus to the first child node
When focus is on an end node (node with no children), does nothing.

Left Arrow
When focus is on an open node, closes the node.
When focus is on a child node that is also either an end node or a closed node, moves focus to the parent node. 
When focus is on a closed node with children, does nothing.

Down Arrow:
Moves focus to the next node that is focusable without opening or closing the node. 

Up Arrow: 
Moves focus to the previous node that is focusable without opening or closing a node

Home:
Moves focus to the first node in the tree without opening or closing it.

End:
Moves focus to the last node in the tree that is focusable without opening the node. 

Enter: 
Performs the default action of the currently focused node. 
For parent nodes, it opens or closes the node. 
In single select trees, if the node has no children the current node is select if not already selected (which is the default action)

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9776 

---

@cerner/terra


artpark:
CHANGELOG for terra-framework-docs not needed. The message from the last changelog entry is enough since I updated the same test cases.

"* Updated examples and tests for expand all/collapse all for `terra-folder-tree`."

packages/terra-folder-tree/src/FolderTree.module.scss
```diff
@@ -7,6 +7,13 @@
     list-style: none;
     margin: 0;
     padding: 0;
+
+    :focus {
+      box-shadow: var(--terra-folder-tree-focus-box-shadow, none);
+      outline: var(--terra-folder-tree-focus-outline, 2px dashed #000);
+      outline-offset: var(--terra-folder-tree-focus-outline-offset, -4px);
+      z-index: 1;
```
cm9361, line 15: Why do we need to set the z-index to 1?

artpark: When the tree items are layered on top of each other, the next tree item will be rendered on top of the outline, so this z-index is here to make sure the outline is always visible and on top

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.module.scss
```diff
@@ -29,7 +29,13 @@
     padding: 0;
   }
 
-  .radio:hover {
-    cursor: pointer;
+  .radio {
+    height: 14px;
+    margin: 17px;
+    width: 14px;
```
cm9361, line 35: Shouldn't these values have theme variables?

artpark: I checked with @eawww, these values are the same for all themes

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -112,6 +123,48 @@ const FolderTreeItem = ({
...
+  const handleKeyDown = event => {
+    switch (event.nativeEvent.keyCode) {
+      case KeyCode.KEY_RETURN:
+      case KeyCode.KEY_SPACE:
+        event.preventDefault();
+
+        if (isFolder) {
+          handleToggle(event);
+        } else {
+          onClick(event);
```
cm9361, line 135: Shouldn't we return more meaningful information in the onClick event such as information about the item selected?  The person providing the callback function should not have to use the native event to figure that out with our React implementation.

cm9361, line 135: Seems like this could be simplified to use a ternary operator.

packages/terra-folder-tree/src/FolderTree.jsx
```diff
@@ -54,6 +55,72 @@ const FolderTree = ({
...
+    });
+    const currentIndex = Array.from(visibleListItems).indexOf(event.target);
+    const lastIndex = visibleListItems.length - 1;
+
+    const handleHomeKey = () => visibleListItems[0].focus();
+    const handleEndKey = () => visibleListItems[lastIndex].focus();
+
+    switch (event.nativeEvent.keyCode) {
+      case KeyCode.KEY_END:
+        event.preventDefault();
```
sycombs, line 81: I'm probably missing the reason why, can we just call `event.preventDefault()` at the beginning of the `handleKeyDown` function to avoid calling it for every case?

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -122,6 +175,10 @@ const FolderTreeItem = ({
         aria-expanded={isFolder ? isExpanded : null}
         aria-selected={isSelected}
         onClick={isFolder ? handleToggle : onClick}
+        onKeyDown={handleKeyDown}
+        data-item-show-focus
```
cm9361, line 179: Just a question.  What is the purpose of this data attribute?

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -112,6 +123,48 @@ const FolderTreeItem = ({
...
+        break;
+      }
+      default:
+        break;
+    }
+  };
+
   return (
     <>
       {/* TODO: Re-enable this eslint rule once keyboard handling is implemented */}
```
sycombs, line 170: I think we can reenable this rule and remove this TODO now.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
artpark: While I agree, I'm using the existing implementation of how click is handled. This seems like it could be its own story

https://github.com/cerner/terra-framework/pull/1956/files/ea0913c1bbd0b9a3e889d21308d922ff24573ae9#diff-3d436b479e88c6cb5fa660bfe58c6842c902adf62dc5273b82e5bf7b95353ec7R177

packages/terra-folder-tree/src/FolderTree.jsx
cm9361: If you call it at the beginning, you won't be able to use any keys at all.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
artpark: I don't think you can use the ternary operator without assigning the end value to something

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -112,6 +123,48 @@ const FolderTreeItem = ({
     }
   };
 
+  const handleKeyDown = event => {
+    switch (event.nativeEvent.keyCode) {
+      case KeyCode.KEY_RETURN:
```
cm9361, line 128: I don't think the Return key is supposed to select an input field.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
artpark: This is used here https://github.com/cerner/terra-framework/pull/1956/files/ea0913c1bbd0b9a3e889d21308d922ff24573ae9#diff-622b2c4e339604395a6822a8aa1db5c37de2c6b45a8b6cd9878fb644c680ec1cR61

This gets all the list items that can be interacted with

packages/terra-folder-tree/src/FolderTree.jsx
artpark: I tried :( it expects a case statement to do so. And if I did something like
```
case true:
  event.preventDefault();
case KeyCode.KEY_RETURN:
case KeyCode.KEY_SPACE:
...
```

Our linter won't allow a fall-through to occur. It expects a break statement at the end of every case.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -68,22 +74,27 @@ const FolderTreeItem = ({
...
   subfolderItems,
+  parentRef,
   intl,
 }) => {
   const theme = useContext(ThemeContext);
   const isFolder = subfolderItems?.length > 0;
+  const itemNode = useRef();
+  const subFolderNode = useRef();
 
   const subfolder = isFolder ? (
```
cm9361, line 85: It seems like we should have separate React components for the Folder vs the Items.  Maybe it is a future story, but I think the logic would be cleaner with that architecture.

packages/terra-folder-tree/src/FolderTree.jsx
sycombs: Thank you for the explanation!

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
sycombs, line 135: This isn't something we need to do right now, but I think `onClick` would be better renamed to `onSelect` to be more reflective of its actual usage.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -112,6 +123,48 @@ const FolderTreeItem = ({
...
+        }
+        break;
+      case KeyCode.KEY_LEFT: {
+        event.preventDefault();
+        if (event.metaKey) { break; }
+
+        if (isFolder && isExpanded) {
+          handleToggle(event);
+        } else {
+          parentRef?.current.focus();
```
cm9361, line 145: Do we need a parent reference object?  Can't we just use the parentElement of the itemNode?

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -112,6 +123,48 @@ const FolderTreeItem = ({
...
+        break;
+      }
+      case KeyCode.KEY_RIGHT: {
+        event.preventDefault();
+        if (event.metaKey) { break; }
+
+        if (!isExpanded) {
+          handleToggle(event);
+        } else if (isFolder) {
+          const firstFolderChild = subFolderNode.current.querySelector('[data-item-show-focus=true]');
```
cm9361, line 157: If we are querying the DOM, can't we just get the first child li element?

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
artpark: No that's a requirement of the story. It needs to perform the default action on the current item, which should be to select it

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
artpark: The parent element is of the current node is actually a <ul>. What we need is that ul's sibling which is a <li>. And since there can be multiple siblings, I found this to be the easiest way to get the right li element we need.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
artpark: We could, but this accomplishes the same thing, and this ensures that we only grab one with the data-item-show-focus, an attribute exclusive to the folder tree component

packages/terra-folder-tree/src/FolderTree.jsx
```diff
@@ -54,6 +55,72 @@ const FolderTree = ({
...
+      return true;
+    });
+    const currentIndex = Array.from(visibleListItems).indexOf(event.target);
+    const lastIndex = visibleListItems.length - 1;
+
+    const handleHomeKey = () => visibleListItems[0].focus();
+    const handleEndKey = () => visibleListItems[lastIndex].focus();
+
+    switch (event.nativeEvent.keyCode) {
+      case KeyCode.KEY_END:
```
cm9361, line 80: Do we need to prevent the default behavior for every key we support?  What undesired actions are we preventing?

packages/terra-folder-tree/src/FolderTree.jsx
```diff
@@ -54,6 +55,72 @@ const FolderTree = ({
...
+    // Remove all hidden list items
+    const visibleListItems = Array.prototype.slice.call(listItems).filter((item) => {
+      let parent = item.parentNode;
+      while (parent && parent !== listNode.current) {
+        if (parent.hasAttribute('hidden')) {
+          return false;
+        }
+        parent = parent.parentNode;
+      }
+      return true;
```
cm9361, line 71: Wouldn't it be easier to just remove hidden items from the DOM?  Then you wouldn't need this logic, right?

packages/terra-folder-tree/src/FolderTree.jsx
artpark: I'm not sure actually, but I believe it's pretty standard for custom navigation elements like this to have them. Let me run some tests to see if removing them is fine

sycombs (approved):

packages/terra-folder-tree/src/FolderTree.jsx
```diff
@@ -54,6 +55,72 @@ const FolderTree = ({
...
+        const previousIndex = currentIndex - 1;
+        visibleListItems[previousIndex]?.focus();
+        break;
+      }
+      case KeyCode.KEY_DOWN: {
+        event.preventDefault();
+        const nextIndex = currentIndex + 1;
+        visibleListItems[nextIndex]?.focus();
+        break;
+      }
```
cm9361, line 99: Shouldn't we be able to use previousElementSibling and nextElementSibling to simplify the logic?

packages/terra-folder-tree/src/FolderTree.module.scss
cm9361: Why would the tree items be layered on top of each other?  They would only be visible if the folders are expanded.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.module.scss
cm9361: Will they be the same for the Redwood theme and its 3 data density variants?

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -112,6 +123,48 @@ const FolderTreeItem = ({
     }
   };
 
+  const handleKeyDown = event => {
+    switch (event.nativeEvent.keyCode) {
```
cm9361, line 127: You should not need to access the nativeEvent object.  The React event has the keyCode.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
cm9361: Should we remove the elements that can't be interacted with from the DOM?

packages/terra-folder-tree/src/FolderTree.module.scss
artpark: Sorry you're correct, with further testing they do not overlap with one another. This was a remnant from the previous focus indicator story I was working on

cm9361:
Shouldn't I be able to select a folder radio button using the Spacebar?  It seems like that would be the expected behavior for a keyboard user.

artpark:
> Shouldn't I be able to select a folder radio button using the Spacebar? It seems like that would be the expected behavior for a keyboard user.

I also thought that as well, but I double checked with Eric, here's his response
Screenshot 2023-12-18 at 1 31 37 PM

cm9361:
> > Shouldn't I be able to select a folder radio button using the Spacebar? It seems like that would be the expected behavior for a keyboard user.
> 
> I also thought that as well, but I double checked with Eric, here's his response Screenshot 2023-12-18 at 1 31 37 PM
> 
> You have to use Enter

Enter does an expand/collapse action.  It does not select the folder radio button.  If we are going to do it later, should a JIRA at least be logged?

eawww (commented):
Focus indicators look great in all themes. (Thanks for tweaking the padding on the radio button!) Screen reader experience works as expected.

For keyboard interaction, there's a discrepancy between the jira requirements and the design callouts. The jira says Enter should expand/collapse parent nodes but the design callouts say it should select them. Since the first intended use case for this component is folder selection, the interaction to select folders is a core feature so that should be fixed here.

[link to callouts](https://www.figma.com/file/x2wzCa9alMC7gqKoRO6irz/UXF-DS%3A-Outline-View%3A-List%2FDrill-in?type=design&node-id=805%3A740&mode=dev&t=jO9tBqJU7PUsD9kq-1)

Precise wording in design: 
> Enter
Performs the default action of the currently focused node. In single-select trees, the default action is typically to select the focused node.

mjpalazzo:
Great work on the keyboard navigation, @artpark! I validated all the keyboard interactions documented in the PR with Windows 11 and Edge. All worked as required. I did some screen readers testing with JAWS as well even though that was not requested in this PR. JAWS worked well. My main observation for JAWS is helper text that tells the user how to interact with the interactive elements in the tree.

eawww:
> Enter does an expand/collapse action. It does not select the folder radio button. If we are going to do it later, should a JIRA at least be logged?

@cm9361 @artpark Folder selection (with radio button) is a core feature for the first release. This is an issue that would need to be fixed prior to v1 release and ideally before any consumers get the impression that keyboard nav is complete.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
cm9361: I am confused.  Wouldn't the list items be li elements within a folder that is a ul element?  Why wouldn't the parent of the list item be its own folder?

JessieRandle:
> > Enter does an expand/collapse action. It does not select the folder radio button. If we are going to do it later, should a JIRA at least be logged?
> 
> @cm9361 @artpark Folder selection (with radio button) is a core feature for the first release. This is an issue that would need to be fixed prior to v1 release and ideally before any consumers get the impression that keyboard nav is complete.

I vote that this is a new jira since it was a problem with the requirements and it shouldn't delay our v1 release.  Also, if not the enter key, what should be used for expand/collapse?

mjpalazzo:
@artpark - I did some testing with Safari and VoiceOver. Keyboard navigation works as required with Safari. With VoiceOver on the navigation and screen reader announcements did not work as required. Since screen reader support was not the focus of this CR, I expect this will be addressed at a later time.

artpark:
> > > Enter does an expand/collapse action. It does not select the folder radio button. If we are going to do it later, should a JIRA at least be logged?
> > 
> > 
> > @cm9361 @artpark Folder selection (with radio button) is a core feature for the first release. This is an issue that would need to be fixed prior to v1 release and ideally before any consumers get the impression that keyboard nav is complete.
> 
> I vote that this is a new jira since it was a problem with the requirements and it shouldn't delay our v1 release. Also, if not the enter key, what should be used for expand/collapse?

It's all good, I just fixed it to select folders on Enter. And the only way to expand a folder is with the right arrow key

packages/terra-folder-tree/src/FolderTree.jsx
artpark: I tried making these changes now, but it seems like it's not quite the easy fix that I thought it would be. The click logic that was already implemented needs all items to be present in the dom. We could address this in a later story.

packages/terra-folder-tree/src/FolderTree.jsx
artpark: These can't be used since subFolders are grouped as `<ul>`s. And subFolders and items can be sibling with each other

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.module.scss
artpark: @eawww Do you have any ideas?

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
artpark: Continued discussion in https://github.com/cerner/terra-framework/pull/1956#discussion_r1430685528

sycombs:
Is the radio button intended to have this focus style on mouse click?
image

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.module.scss
eawww: I'm not sure. @jeremyfuksa 👆?

artpark:
> Is the radio button intended to have this focus style on mouse click? image

Fixed it, good catch!

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.module.scss
cm9361: I won't hold up the review for this.  Just asking as it seem like something that could change.  Honestly, I am surprised the value isn't different between the default theme and the fusion theme.  It might just be another JIRA to log.

cm9361 (approved):

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
artpark: Not in this folder tree, the folder list item is a separate li element. For example:
Screenshot 2023-12-18 at 2 10 02 PM

Document 4 & 5 are both grouped in a ul but the folder it's in, Projects - Level 4, is its own li item

Screenshot 2023-12-18 at 2 10 40 PM

eawww (approved):

PR #1967, opened 2024-01-05
[terra-folder-tree] Added tabIndex for only last focused item

artpark:
### Summary

**What was changed:**

All folder tree children no longer has a tabIndex of 0. When moving through the elements with the arrow keys, the currently focused item is the only item with a tabIndex of 0.

**Why it was changed:**

Tabbing in and out of the component will not make the user tab through all the items to get to the intended item.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10109 

---

@cerner/terra


packages/terra-folder-tree/tests/wdio/folder-tree-spec.js
```diff
@@ -91,6 +91,23 @@ Terra.describeViewports('FolderTree', ['medium'], () => {
...
+      browser.keys('Tab');
+      browser.keys('ArrowDown');
+      browser.keys('ArrowDown');
+      Terra.validates.screenshot('level two folder focused keyboard', { selector: '#expand-collapse-folder-tree' });
+
+      browser.keys('Tab');
+      Terra.validates.screenshot('tabbed out of focus', { selector: '#root' });
+
+      browser.keys(['Shift', 'Tab']);
+      Terra.validates.screenshot('level two folder focused keyboard', { selector: '#expand-collapse-folder-tree' });
```
sycombs, line 108: I like that we're avoiding adding extra screenshots here, just wondering if there are any potential issues with validating the same screenshot twice in this test.

packages/terra-folder-tree/src/FolderTreeUtils.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable no-param-reassign */
+const handleMoveFocus = (elementA, elementB) => {
```
sycombs, line 2: Can we add a comment to document the purpose/usage of this function, including what `elementA` and `elementB` are respectively?

packages/terra-folder-tree/tests/wdio/folder-tree-spec.js
artpark: I've done this in the past before, and both screenshots should result in the same identical image, so I believe it should be alright

packages/terra-folder-tree/src/FolderTreeUtils.js
artpark: Updated [6c7a037](https://github.com/cerner/terra-framework/pull/1967/commits/6c7a03711ef1d99939b2d8222d8eacee3aa61d63)

packages/terra-folder-tree/tests/wdio/folder-tree-spec.js
sdadn: Is there a need to validate teh same screenshot twice? If we're only checking for focus, we can use the `isFocused()` validator since that's more efficient than doing a screenshot comparison.

artpark: Oh that's right! I keep forgetting we have that, I'll update it to use that

packages/terra-folder-tree/tests/wdio/folder-tree-spec.js
sdadn, line 1: Looks good, +1!

sdadn (approved):

eawww (approved):

PR #2005, opened 2024-01-25
[terra-folder-tree] Announce when all items are expanded or collapsed

sycombs:
### Summary

This PR adds an aria-live region in Folder Tree so that screenreaders announce when all folder tree items have been expanded or collapsed (via expand all/collapse all button click).

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

Note: Unit testing done in WDIO only due to limitations in updating component state/props with Enzyme.

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10089 

---

@cerner/terra


artpark (approved):

packages/terra-folder-tree/src/FolderTree.jsx
```diff
@@ -128,6 +139,20 @@ const FolderTree = ({
...
+    const visibleListItems = getVisibleListItems(listItems);
+
+    if (visibleListItems.length === children.length) {
+      setAriaLiveMessage(intl.formatMessage({ id: 'Terra.folder-tree.button.collapse-all-announcement' }));
+    }
+
+    if (visibleListItems.length === listItems.length) {
+      setAriaLiveMessage(intl.formatMessage({ id: 'Terra.folder-tree.button.expand-all-announcement' }));
+    }
+  });
```
adoroshk, line 154: I was wondering if making those calculations on every re-render can become a problem in case of a big folder tree? I was thinking that moving the functionality implemented in this useEffect to a method, then passing a callback to that method to `FolderItem` to call on collaps/expand events might improve the performance.

sycombs: I was trying to think of a way to only get the new `visibleListItems` if `onExpandAll` or `onCollapseAll` had been called. The problem is that `visibleListItems` isn't changed until after the component renders, so we can't simply make a check on the expand all/collapse all events.

> I was thinking that moving the functionality implemented in this useEffect to a method, then passing a callback to that method to FolderItem to call on collaps/expand events might improve the performance.

The full folder tree expand/collapse does not trigger the row-level expand collapse so unfortunately this doesn't work.

cm9361, line 154: It looks like the dependency array is missing here.  This did not trigger a linter error?

sycombs: It did trigger a linter error, which is disabled above. The reason it's being disabled is to ensure the useEffect runs after `onExpandAll`/`onCollapseAll` trigger a re-render. Including the required dependency array prevents this.

Obviously the side effect is that this hook has to run on every render, but I haven't been able to find a way around that so far. I suppose a workaround could be to include a dummy state variable that updates when `onExpandAll`/`onCollapseAll` are called, thoughts?

sycombs: On second thought, I'm not sure a dummy state variable will work either, since we only want this to trigger when the items have actually been expanded, not just if the button was clicked.

sycombs: Still need to fix tests, but updated to add back the dependency array and trigger the useEffect only on `visibleListItems` change:

[c093e25](https://github.com/cerner/terra-framework/pull/2005/commits/c093e25686488d26a43dd4b2d16f53a6f0d1f7a2)

Thoughts?

sycombs: This is the best solution I've found so far to prevent the useEffect from running on every render: [c093e25](https://github.com/cerner/terra-framework/pull/2005/commits/c093e25686488d26a43dd4b2d16f53a6f0d1f7a2)

mjpalazzo:
@sycombs - Great work on Folder Tree! JAWS announces expand all and collapsed all states.  One observation has to do with index announcements. For the Basics Example, when I expand the Projects 2023 folder and use the down arrow to navigate the documents, JAWS announces 1 of 4, but as a down arrow to the next document, JAWS announces the name of the document, but not the index value, 2 of 4.
I tested the keyboard interactions as well. Here are some observations.
The documentation says Command Left Arrow and Command Right Arrow on a Mac navigate to home and end. These did not work for me. I had to use Fn Left Arrow and Fn Right arrow instead. Note that I am using a cloud desktop running Windows 11 2023 H2 from a MacBook Pro.

packages/terra-folder-tree/src/FolderTree.jsx
adoroshk: It's a nice solution for expand all/collapse all, but it looks like it is not working with the tree items being opened one by, or am I missing something?

sycombs: Based on the Jira's AC the announcement is for when the expand all/collapse all buttons are pressed, not just when the whole tree is expanded item by item. Checked with @eawww who confirmed that having the announcement on just the buttons (as long as the items are actually expanded/collapsed) is acceptable.

sdadn (approved):

adoroshk (approved):

PR #2010, opened 2024-01-29
Added terra-table to the dependencies

adoroshk:
### Summary

`terra-compact-list` use `getFocusableElements` method from `terra-table`, hence needs `terra-table` to be added to its dependencies.
This PR adds missing `terra-table` dependency.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review


sdadn (approved):

sycombs:
Should we also update the imports? Should probably be `import getFocusableElements from 'terra-table/src/utils/focusManagement';`, right?

https://github.com/cerner/terra-framework/blob/compact-interactive-list-add-terra-table-to-depenencies/packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx#L36
https://github.com/cerner/terra-framework/blob/compact-interactive-list-add-terra-table-to-depenencies/packages/terra-compact-interactive-list/src/subcomponents/Cell.jsx#L7

adoroshk:
> import getFocusableElements

@sycombs good catch, updated imports to be `import getFocusableElements from 'terra-table/lib/utils/focusManagement';`

package.json
```diff
@@ -142,6 +145,7 @@
     "terra-disclosure-manager": "^4.9.0",
     "terra-enzyme-intl": "^3.4.0",
     "terra-icon": "^3.58.0",
+    "terra-table": "^5.4.0",
```
sycombs, line 148: Just curious why this needs to be added to the root package.json?

sycombs (approved):

package.json
sdadn: Not sure why terra-icon is at the root level as well but I'm trying to move away from this pattern as these packages aren't used at the root level. I would rather have a linter exception than an extra package at the root level.

But this change can also be done in the monorepo.

+1 for now.

PR #2014, opened 2024-02-01
[terra-modal manager] Implemented focus lock to confine focus within the modal popup.

AH106586Harika:
### Summary
Fixed- Focus is continuing to go outside of the Modal Manager when navigating through the component.

**What was changed:**
Added FocusLock to confine the focus within the modal popup

**Why it was changed:**
Pressing the TAB key causes the focus to shift away from the modal popup.

### Testing

This change was tested using:

- [X] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**
[UXPLATFORM-10090](https://jira2.cerner.com/browse/UXPLATFORM-10090)
[UXPLATFORM-10169](https://jira2.cerner.com/browse/UXPLATFORM-10169)

---

@cerner/terra


JessieRandle:
We had added focus trap to this element and UX had us remove it.  @scottwilmarth  Can you chime in with why we don't want focus trap in modal manager?

scottwilmarth:
> We had added focus trap to this element and UX had us remove it. @scottwilmarth Can you chime in with why we don't want focus trap in modal manager?

@JessieRandle I am not familiar with popup any why UX would have requested it removed. My assumption is that we would want focus trapped. I think the UX SMEs need to to review this and work together to ensure everyone is on the same page. I'll have them take a look.

packages/terra-abstract-modal/src/_ModalContent.jsx
```diff
@@ -153,18 +154,19 @@ const ModalContent = forwardRef((props, ref) => {
...
-        aria-label={ariaLabel}
-        aria-labelledby={ariaLabelledBy}
-        aria-describedby={ariaDescribedBy}
-        className={modalClassName}
-        role={role}
-        ref={ref}
-      >
-        <div className={modalContainerClassName} ref={setModalFocusElementRef} data-terra-abstract-modal-begin tabIndex="-1">
-          {(!isCalledFromNotificationDialog) && (
+      <FocusLock>
```
supreethmr, line 157: We need a prop to enable focusLock for passivity concerns and prop can be removed on next major version.

supreethmr:
We also need the test case for new prop to ensure focus is trapped inside modal. Existing tests are written to ensure focus is moved out of modal on Tab key press : https://github.com/cerner/terra-framework/pull/2014/files#diff-c0d6bd78f4352973058098506546cb78288d1fbdd322f34f821c0ead0c296f61

AH106586Harika:
> We also need the test case for new prop to ensure focus is trapped inside modal. Existing tests are written to ensure focus is moved out of modal on Tab key press : https://github.com/cerner/terra-framework/pull/2014/files#diff-c0d6bd78f4352973058098506546cb78288d1fbdd322f34f821c0ead0c296f61

Added new jest test cases for new enableFocusLock  prop.

packages/terra-abstract-modal/src/_ModalContent.jsx
AH106586Harika: Implemented as suggested please review.

packages/terra-abstract-modal/src/_ModalContent.jsx
```diff
@@ -5,6 +5,7 @@ import classNames from 'classnames';
 import classNamesBind from 'classnames/bind';
 import ThemeContext from 'terra-theme-context';
 import VisuallyHiddenText from 'terra-visually-hidden-text';
+import FocusLock from 'react-focus-lock';
```
cm9361, line 8: What is the difference between react-focus-trap and react-focus-lock?  Don't we usually use the former?

packages/terra-abstract-modal/src/_ModalContent.jsx
```diff
@@ -76,6 +77,11 @@ const propTypes = {
    * If set to true, the AbstractModal is rendered inside a NotificationDialog.
    */
   isCalledFromNotificationDialog: PropTypes.bool,
+  /**
+   * @private
+   * If set to true, then the focus lock will get enabled.
+   */
+  enableFocusLock: PropTypes.bool,
```
cm9361, line 84: In what scenario would we want to disable the focus lock?  Should this be an option?

packages/terra-abstract-modal/src/_ModalContent.jsx
```diff
@@ -153,48 +206,18 @@ const ModalContent = forwardRef((props, ref) => {
...
-          )}
+      {enableFocusLock
+        && (
+        <FocusLock>
+          {modalContent}
+        </FocusLock>
+        )}
+      {!enableFocusLock
+        && (
+        <div>
```
cm9361, line 217: We should not require this extra div, correct?  The modal content is already in a div element.  There does not appear to be additional styling provided by this additional div.

packages/terra-abstract-modal/src/_ModalContent.jsx
```diff
@@ -153,48 +206,18 @@ const ModalContent = forwardRef((props, ref) => {
...
+          {modalContent}
         </div>
-      </div>
+        )}
```
cm9361, line 220: Seems like a ternary operator would be better in this scenario.  Or the modalContent variable could be updated if enableFocus lock is true and then rendered.

packages/terra-abstract-modal/src/AbstractModal.jsx
```diff
@@ -78,6 +78,11 @@ const propTypes = {
    * If set to true, the AbstractModal is rendered inside a NotificationDialog.
    */
   isCalledFromNotificationDialog: PropTypes.bool,
+  /**
+   * @private
+   * If set to true, then the focus lock will get enabled.
+   */
+  enableFocusLock: PropTypes.bool,
```
cm9361, line 85: In what scenario would we not want the focus lock to be enabled?  Should this really be a configurable property?

supreethmr: This was not suppose to be private prop @AH106586Harika , We are suppose to create this as class prop on modal manager as well. Fusion team is consuming modal manager for pass-through and fusion component has a prop called `shouldTrapFocus` to enable focus trap ( by default focusTrap is not enabled ). In order to address this GAP modal manager should also have similar prop to enable focus trap when it is required.

packages/terra-notification-dialog/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added
+  * Added FocusLock to confine focus within the modal popup.
```
supreethmr, line 6: this is not correct !! we need to add logs about the changes made for notification-dialog which has got JEST SNAPSHOT update

AH106586Harika: updated.

packages/terra-abstract-modal/src/AbstractModal.jsx
AH106586Harika: Added as suggested.

packages/terra-abstract-modal/src/_ModalContent.jsx
AH106586Harika: Done.

packages/terra-abstract-modal/src/_ModalContent.jsx
AH106586Harika: Done.

packages/terra-abstract-modal/src/_ModalContent.jsx
AH106586Harika: As FocusTrap is not working in few of the scenarios like single select dropdown inside modal manger popup. Hence Instead of using FocusTrap now we are using FocusLock which is working in all the scenarios.

packages/terra-abstract-modal/src/_ModalContent.jsx
```diff
@@ -140,6 +146,72 @@ const ModalContent = forwardRef((props, ref) => {
...
+            When an aria-label is set and tabIndex is set to 0, VoiceOver will read
+            the aria-label value when the modal is opened
+          */
+        }
+        <FocusLock>
+          {modalContent}
+        </FocusLock>
+      </React.Fragment>
+    );
+  }
```
cm9361, line 213: This isn't the correct implementation.  You would not create two return paths.  You could just set the modalContent variable equal to <FocusLock>{modalContent}</FocusLock>

packages/terra-abstract-modal/src/AbstractModal.jsx
```diff
@@ -90,6 +94,7 @@ const defaultProps = {
   rootSelector: '#root',
   zIndex: '6000',
   isCalledFromNotificationDialog: false,
+  shouldTrapFocus: true,
```
supreethmr, line 97:
```suggestion
  shouldTrapFocus: false,
```
this should be false by default

AH106586Harika: Updated.

packages/terra-date-input/tests/wdio/date-input-spec.js
MadanKumarGovindaswamy, line 1: Hi @supreethmr . we have removed this line of code because of WDIO failure. With the latest change worked upon date-input component selecting current month on the select field will generate the current date, because of this build will fail every time so we made the selection to January.

packages/terra-abstract-modal/src/_ModalContent.jsx
```diff
@@ -153,48 +204,11 @@ const ModalContent = forwardRef((props, ref) => {
...
-                );
-              }}
-            </FormattedMessage>
-          )}
-        </div>
-      </div>
+      {
+        shouldTrapFocus
+          ? <FocusOn onClickOutside={closeOnOutsideClick ? onRequestClose : null}>{modalContent}</FocusOn>
+          : <>{modalContent}</>
```
cm9361, line 210: You should not need a React fragment around a div element.

packages/terra-abstract-modal/package.json
```diff
@@ -32,6 +32,7 @@
     "keycode-js": "^3.1.0",
     "mutationobserver-shim": "<=0.3.3",
     "prop-types": "^15.5.8",
+    "react-focus-on": "^3.9.1",
```
cm9361, line 35: Are we now introducing a new third-party library?  If so, how was this vetted?  Why do the other two lock libraries not work?

AH106586Harika: Hi @cm9361 
We found issues with react-focusTrap in certain scenarios, like form-single select inside the modal popup, As a solution, we attempted to use react-focusLock. However, we found that the closeOnOutsideClick functionality wasn't functioning properly. We then explored **focusOn**, which provided support for all the required functionalities.

packages/terra-abstract-modal/src/_ModalContent.jsx
AH106586Harika: If I remove this fragment it is throwing the below error.

Screenshot 2024-02-08 at 9 39 23 PM

cm9361:
Terra already has logic to close a modal dialog when you click outside of it.  The problem is that the ModalManager is just not exposing it at the moment.  I think we should just be leveraging that logic instead of creating a custom implementation that must be maintained separately.

packages/terra-abstract-modal/package.json
cm9361: Do we need to also expose this closeOnOutsideClick prop in ModalManager.  I don't think users of that component, such as MPages Fusion, can set that value as-is.

cm9361 (approved):

AH106586Harika:
> As suggested in the PR, the user focus when navigating using the keyboard is confined inside the modal dialog and the focus is to be navigated in cyclic order within the dialog. However, the issue is that tabindex="0" added to the modal dialog container, causing it to receive focus every time the user navigates within the modal dialog in cyclic order.
> 
> It is recommended to add tabindex="-1" to the modal dialog container
> 
> , to set the initial focus on the modal dialog container when it is opened.

Implemented. Thanks.

packages/terra-abstract-modal/src/_ModalContent.jsx
```diff
@@ -144,12 +144,9 @@ const ModalContent = forwardRef((props, ref) => {
   delete customProps.closePortal;
   delete customProps.fallbackFocus;
 
-  const platformIsiOS = !!navigator.platform && /iPad|iPhone|iPod/.test(navigator.platform);
-
   const modalContent = (
     <div
       {...customProps}
-      tabIndex={platformIsiOS || isCalledFromNotificationDialog ? '-1' : '0'}
```
supreethmr: Does container receive initial focus  without tabIndex ?

packages/terra-abstract-modal/src/_ModalContent.jsx
```diff
@@ -138,7 +144,49 @@ const ModalContent = forwardRef((props, ref) => {
   delete customProps.closePortal;
   delete customProps.fallbackFocus;
 
-  const platformIsiOS = !!navigator.platform && /iPad|iPhone|iPod/.test(navigator.platform);
+  const modalContent = (
+    <div
+      {...customProps}
+      aria-label={ariaLabel}
+      aria-labelledby={ariaLabelledBy}
+      aria-describedby={ariaDescribedBy}
```
supreethmr, line 152: We need to validate with screen readers to ensure values provided to above aria attributes are read when the focus is on modal container. Since there is no tabIndex on this div, it may get ignored by screen readers

packages/terra-abstract-modal/src/_ModalContent.jsx
AH106586Harika: Yes, the primary focus is on the container.PFA

https://github.com/cerner/terra-framework/assets/126655256/88bc9479-2c09-4dcc-be88-05b2309e550a

packages/terra-abstract-modal/src/_ModalContent.jsx
AH106586Harika: Hi Supreeth, when the focus is on model container all the aria attributes are working as expected. Please find the attached video for reference.

https://github.com/cerner/terra-framework/assets/126655256/a525f425-680b-4254-8818-08cd568d131f

supreethmr: Please verify on JAWS as well

AH106586Harika: In JAWS as well It is announcing all the aria attributes PFA video


https://github.com/cerner/terra-framework/assets/126655256/b8ca364b-948f-4b05-b1b3-a8c67cee2e90

rahulkumar8cs:
+1

PR #2017, opened 2024-02-05
[terra-time-input] - Mode changing issue on JAWS and SR fixes

PK106552:
### Summary

**What was changed:**

1. Changed Up/Down key functionality same as +/- functionality to make uniform.
2. Changed input type to `number` instead of text.
3. Fixed label reading twice issue in time input.

**Why it was changed:**

1. Up/Down key functionality was not same as +/- functionality.
2. Input type was text.
3. Label value was reading twice in JAWS.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-6328 

---

@cerner/terra

**Before fix :** 

Screenshot 2024-02-02 at 10 15 42 AM

**After Fix :** 

Screenshot 2024-02-02 at 10 15 20 AM


packages/terra-time-input/src/TimeUtil.js
```diff
@@ -436,6 +436,8 @@ TimeUtil.DoubleZeroDigit = '00';
 
 TimeUtil.initialValue = ['0', '00'];
 
+TimeUtil.letterKeycode = [38, 40, 65, 66, 67, 68, 69, 70, 80, 81, 82, 83, 84, 85, 86, 87, 88, 89, 90];
```
saket2403, line 439: What is this array for?

saket2403:
Left/Right Arrow key not moving the focus between the input fields. Please compare with previous functionality.

packages/terra-time-input/src/TimeUtil.js
PK106552: Safari browser allows letters when we have type as number, to prevent this we have to check all the letters keyCode

PK106552:
> Left/Right Arrow key not moving the focus between the input fields. Please compare with previous functionality.

The input element's type ('number') does not support setSelectionRange, that's why left and right arrow key is not shifting focus, i'm investigating it.

packages/terra-time-input/src/TimeUtil.js
saket2403: Can you use regex like this.
```
        let inputValue = e.target.value;
        let numericValue = inputValue.replace(/[^0-9]/g, '');
        e.target.value = numericValue;
```

PK106552: handleKeyDown event is triggering when we press any letter, so i have added regex check in keyDown event and removed keyCode array value from TimeUtil file

rbsree:
+1 for Accessibility Review.
The issues mentioned in the PR are fixed now. 
- The input field for the Terra Time Input is implemented as type="number" field. 
The value of the hour or minute field can be changed using the +/- or UP/DOWN arrow key as expected. 
- Hotkey implemented for the Time Input field "N" is also supported as expected, setting the current time as the input.

Issue:
- When users are provided an input in the time field, the user is not able to navigate between the hour and minute fields using the right/left arrow key. When the field is empty the arrow key navigation is working as expected.
- The arrow navigation is not a default behavior for <input type="number">, in this case, it has been implemented through the property "setSelectionRange" which works when the input field is empty, but the same is not supported when the "Terra Time Input" field has input defined.
- This is not a major blocker for users, as user can still navigate through the input field using TAB/SHIFT+TAB key.

packages/terra-time-input/tests/wdio/time-input-twelve-hour-spec.js
```diff
@@ -124,70 +64,6 @@ Terra.describeViewports('Time Input Twelve Hour', ['medium'], () => {
...
-      browser.keys('11');
-      browser.keys('59');
-      $('#timeInput input[name="terra-time-minute-time-input"]').click();
-      browser.keys('+');
-      expect($('#timeInput input[name="terra-time-hour-time-input"]')).toHaveValue('12');
-      expect($('#timeInput input[name="terra-time-minute-time-input"]')).toHaveValue('00');
-      validateElement('+ causes rollover to evening');
-    });
-  });
-
```
saket2403, line 190: Why are these tests removed? Instead the click events before pressing +/- should be updated so that they match the expected output.

PK106552: Already we have added +/- hotkey test cases in time-input-spec file.

PK106552: Updated

packages/terra-time-input/tests/wdio/time-input-twelve-hour-spec.js
```diff
@@ -27,66 +27,6 @@ Terra.describeViewports('Time Input Twelve Hour', ['medium'], () => {
...
-    browser.refresh();
-    Terra.hideInputCaret('#timeInput input[name="terra-time-hour-time-input"]');
-
-    $('#timeInput input[name="terra-time-hour-time-input"]').click();
-    browser.keys('12');
-    $('#timeInput input[name="terra-time-hour-time-input"]').click();
-    browser.keys(['ArrowDown']);
-    validateRoot('switch meridiem - down arrow');
-  });
-
```
saket2403, line 89: Please update these tests so that they test the newly added functionality of the Up/Down arrow keys instead of removing them.

packages/terra-time-input/tests/wdio/time-input-twelve-hour-spec.js
```diff
@@ -126,29 +66,28 @@ Terra.describeViewports('Time Input Twelve Hour', ['medium'], () => {
 
   describe('pressing -', () => {
     it('should roll over to 11 AM if hour and minute is at 12:00 PM', () => {
-      browser.url('/raw/tests/cerner-terra-framework-docs/time-input/twelve-hour/filled-evening');
+      browser.url('/raw/tests/cerner-terra-framework-docs/time-input/twelve-hour/default');
```
saket2403, line 69: Why are these URLs changed what was the issue with the existing url?

PK106552: Typed input is not rendering If we use filled input field, that's why i have used default empty field test example

saket2403: So that means the existing functionality is breaking. In previous implementation you could type without clearing the field. Please try to fix.

PK106552: I have verified it, previous implementation is working fine, we can able to edit filled time when we navigate using tab key, but when we click on field using mouse that time only we could not able to edit the filled time.

saket2403: So please look into it if that can be fixed

PK106552: Updated

packages/terra-time-input/tests/wdio/time-input-twelve-hour-spec.js
PK106552: Updated

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -488,13 +472,22 @@ class TimeInput extends React.Component {
     } = this.state;
     const variant = TimeUtil.getVariantFromLocale(this.props);
 
+    if (event.currentTarget.value.length === 2 && event.key.match(/^[0-9]/g)) {
+      event.currentTarget.value = ''; // eslint-disable-line no-param-reassign
+    }
+
+    if (!event.key.match(/^[0-9]/g) && !(event.keyCode === KeyCode.KEY_BACK_SPACE || event.keyCode === KeyCode.KEY_DELETE || event.keyCode === KeyCode.KEY_TAB || event.keyCode === KeyCode.KEY_RIGHT || event.keyCode === KeyCode.KEY_LEFT)) {
```
sugan2416, line 479: should we not include Up/Down as well here ?

PK106552: No need, because automatically Up/Down key will be increase/decrease value when we use type as number in input field

packages/terra-time-input/src/TimeInput.module.scss
```diff
@@ -311,4 +311,10 @@
       }
     }
   }
+  // stylelint-disable property-no-vendor-prefix
+  input::-webkit-outer-spin-button,
```
sugan2416, line 315: why was this added? Was it to suppress warning?

PK106552: This style was included to hide arrow icons in input field when we use type as number, it was to suppress property-no-vendor-prefix warning

sugan2416 (approved):

sugan2416 (approved):

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -769,10 +769,6 @@ Terra.describeViewports('DateTimePicker', ['large'], () => {
           expect($('input[name="terra-date-day-input"]')).toHaveValue((`0${today.date()}`).slice(-2));
           expect($('input[name="terra-time-hour-input"]')).toHaveValue((`0${today.hour()}`).slice(-2));
           expect($('input[name="terra-time-minute-input"]')).toHaveValue((`0${today.minute()}`).slice(-2));
-          const secondsValue = Number($('input[name="terra-time-second-input"]').getValue());
-          const secondsTimeValue = today.seconds();
-          const secondsInRange = (secondsTimeValue === secondsValue + 1 || secondsTimeValue === secondsValue);
-          expect(secondsInRange).toBe(true);
```
saket2403, line 775: Why has this been removed?

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -1186,22 +1178,18 @@ Terra.describeViewports('DateTimePicker', ['large'], () => {
...
+          const today = moment.utc().add(1, 'seconds');
           expect($('input[name="terra-date-year-input"]')).toHaveValue(today.year().toString());
           expect($('input[name="terra-date-month-input"]')).toHaveValue((`0${(today.month() + 1)}`).slice(-2));
           expect($('input[name="terra-date-day-input"]')).toHaveValue((`0${today.date()}`).slice(-2));
           expect($('input[name="terra-time-hour-input"]')).toHaveValue((`0${today.hour()}`).slice(-2));
           expect($('input[name="terra-time-minute-input"]')).toHaveValue((`0${today.minute()}`).slice(-2));
-          const secondsValue = Number($('input[name="terra-time-second-input"]').getValue());
-          const secondsTimeValue = today.seconds();
-          const secondsInRange = (secondsTimeValue === secondsValue + 1 || secondsTimeValue === secondsValue);
-          expect(secondsInRange).toBe(true);
```
saket2403, line 1204: This test says that you are testing +1 sec to the seconds but I cannot see the validation for seconds field here.

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -1256,16 +1244,12 @@ Terra.describeViewports('DateTimePicker', ['large'], () => {
...
```
saket2403, line 1268: This test says that you are testing +1 sec to the seconds but you have removed seconds field validation

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
PK106552: Updated

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -817,17 +813,13 @@ Terra.describeViewports('DateTimePicker', ['large'], () => {
...
```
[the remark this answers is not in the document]
PK106552: Updated

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
PK106552: Updated

packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
PK106552: Updated

saket2403 (approved):

PR #2019, opened 2024-02-07
[terra-date-input] Validation issues

KV106606Viswanath:
### Summary
1. Text is not visible for month select list options in edge browser - Fixed
2. Screen reader announcing wrong date when we press T key, was getting the correct date but it was announcing the previous date - Fixed

**What was changed:**

**Why it was changed:**

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-6327

---

@cerner/terra


packages/terra-date-input/tests/wdio/date-input-spec.js
saket2403, line 1: Why are changes required in this file?

KV106606Viswanath: When we select the current month from the month select, we expect to get today's date. This functionality has been implemented as part of the hot keys feature. However, a specific test case is failing because it selects January as the month, and at that time, the current date is not reflected in the snapshot. To address this issue, I updated the test case to select February as the month, allowing the test to pass. However, now the test is failing again because it expects the date corresponding to February. To resolve this inconsistency, I reverted the test case back to selecting January as before.

KV106606Viswanath: Selecting current month to get current date logic has been removed.

saket2403 (approved):

packages/terra-date-input/tests/wdio/date-input-spec.js
```diff
@@ -75,21 +75,19 @@ Terra.describeViewports('Date Input', ['medium'], () => {
...
 
   describe('Month Select placeholder option clears value', () => {
-    it('Selects February in month select', () => {
+    it('Selects January in month select', () => {
       browser.url('/raw/tests/cerner-terra-framework-docs/date-input/day-month-year-date-input');
       Terra.hideInputCaret('input[name="terra-date-day-date-input"]');
       Terra.hideInputCaret('input[name="terra-date-year-date-input"]');
       $('select[name="terra-date-month-date-input"]').click();
       browser.keys(['ArrowDown']);
-      browser.keys(['ArrowDown']);
```
supreethmr, line 84: why do we removing the arrowDown key event here !!

packages/terra-date-input/src/DateInput.jsx
```diff
@@ -827,7 +819,7 @@ class DateInput extends React.Component {
     const format = DateInputUtil.getDateFormat(this.props);
     const label = a11yLabel || this.props.intl.formatMessage({ id: 'Terra.date.input.labelDefault' });
     const dateFormatLabel = this.props.intl.formatMessage({ id: 'Terra.date.input.dateFormatLabel' });
-    const inputDate = moment(new Date(this.state.dateString)).format('dddd MMMM D YYYY');
+    const inputDate = moment(this.state.dateString).format('dddd MMMM D YYYY');
```
ShettyAkarsh, line 822: what is the reason for removing the new operator here ?

packages/terra-date-input/src/DateInput.module.scss
```diff
@@ -173,7 +173,7 @@
     }
 
     option:not([value='']) {
-      color: transparent;
+      color: #1c1f21;
```
ShettyAkarsh, line 176: is this value provided by UX?

packages/terra-date-input/src/DateInput.jsx
KV106606Viswanath: I have removed the 'new Date' instantiation because in PST timezone it is converting to one day less than the current date. So that it has been removed to fix this issue and now it is working fine.

packages/terra-date-input/tests/wdio/date-input-spec.js
KV106606Viswanath: The justification for removing arrowDown key event has been provided in above comment.
https://github.com/cerner/terra-framework/pull/2019#discussion_r1482512865
And after the discussion with saket its not required to get current by selecting the current month. So that i have removed the logic as well.

packages/terra-date-input/src/DateInput.module.scss
supreethmr: Will this remain same for all the themes. Because base color is different for low-light theme : ` #b2b5b6 ` : https://github.com/cerner/terra-core/blob/c55faac00f280f8345ec0d4756530377b520446c/packages/terra-base/src/clinical-lowlight-theme/Base.scss#L13

KV106606Viswanath: Yes, the color remains same because the color "#b2b5b6" is not clearly visible on white background color, for the low-light theme also we are having the white background color which is same as the default theme for the list options.

ShettyAkarsh (approved):

packages/terra-date-input/src/DateInput.module.scss
supreethmr: how is it showing white here in UI site it looks different : 
Screenshot 2024-02-09 at 11 54 20 AM

PR #2021, opened 2024-02-07
[terra-folder-tree] Add isSelectable prop

sycombs:
### Summary

This PR adds an `isSelectable` prop to FolderTreeItem so that consumers can define whether or not an item can be selected.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10086 

---

@cerner/terra


packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -30,14 +30,18 @@ const propTypes = {
...
+  isSelectable: PropTypes.bool,
   /**
    * Indicates whether the item is selected. Because this component has the appearance of a radio button group, only one item should be selected at a time.
    */
   isSelected: PropTypes.bool,
   /**
    * The callback function for a click event.
    */
-  onClick: PropTypes.func,
+  onSelect: PropTypes.func,
```
sycombs, line 44: Renamed this prop from `onClick` to `onSelect` to better reflect its actual intended usage. If there are concerns about passivity this can be reverted.

cm9361: I think the passivity concern would be if MPages is already using this callback.  If they have coded to use onClick, this minor version update would break them, correct?

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -108,10 +114,23 @@ const FolderTreeItem = ({
...
+    <input
+      type="radio"
+      checked={isSelected}
+      onChange={onSelect}
+      aria-hidden // Hiding the radio button from assistive technology since they cannot be grouped correctly
+      tabIndex={-1} // Prevent tabbing to the button since it should not be read or acknowledged by assistive technology
+      className={cx('radio')}
+    />
+  ) : (
+    <div className={cx('radio-container')} />
```
cm9361, line 127: A div is a block element.  Do you want an actual div?  Can we not just add the necessary padding or margin to the sibling element?

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
sycombs: Yes, this will break consumers who are already using onClick.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -169,23 +196,16 @@ const FolderTreeItem = ({
         className={itemClassNames}
         role="treeitem"
         aria-expanded={isFolder ? isExpanded : null}
-        aria-selected={isSelected}
-        onClick={isFolder ? handleToggle : onClick}
+        aria-selected={isSelectable ? isSelected : null}
```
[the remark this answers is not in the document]
sycombs: According to the [mdn docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected#values), the states for `aria-selected` are **true**, **false**, and **undefined** (the element is not selectable). I'd like to avoid setting `aria-selected={false}` on elements that are not selectable (although it does seem to behave correctly in Voiceover at least) since it implies that the element is selectable.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
sdadn: The component is still in alpha, so I think we can get away with making a non-passive change. The prop rename makes sense to me. I would just add a "breaking change" entry in the CHANGELOG.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -169,23 +196,16 @@ const FolderTreeItem = ({
...
-          checked={isSelected}
-          onChange={onClick}
-          aria-hidden // Hiding the radio button from assistive technology since they cannot be grouped correctly
-          tabIndex={-1} // Prevent tabbing to the button since it should not be read or acknowledged by assistive technology
-          className={cx('radio')}
-        />
+        {radioButton}
         {/* eslint-disable-next-line react/forbid-dom-props */}
-        <span style={{ paddingLeft: `${level}rem` }}>
+        <span style={{ paddingLeft: `${level * 14}px` }}>
```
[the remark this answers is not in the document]
sycombs: 14px is the correct value as specified by UX, the original value was incorrect.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
sycombs: Good point, added the breaking change entry here: [7555293](https://github.com/cerner/terra-framework/pull/2021/commits/75552939d325bc7006aca75820c0d60b958639f8)

sdadn: +1

adoroshk (approved):

eawww:
In testing with JAWS, I'm getting "Selected" on all items that are not selectable. Adding `aria-selected="false"` to these `<li>` elements seems to remedy it.

Otherwise it looks good.

chrismichalewicz:
Sorry for the delay in getting to this @sycombs  

Tested the following from an a11y standpoint

**Keyboard**
Was able to navigate and select and expand the items as expected with the keyboard hotkeys
Confirmed I was unable to select items that weren't set to Selectable (no radio button) 

**Screenreader**
In testing with JAWS it is not communicating if an Item is Selectable or not, as well as once an Item is selected the SR just says "Enter".  @eawww  Should we be communicating if an item is selectable or not and then if it selected/unselected?  Similar to what it does when an item is expanded/collapsed.

eawww:
@mjpalazzo yes, that's a great point. Especially since we're deviating from the standard treeview pattern, communicating whether an item is selectable would be ideal.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -203,6 +224,7 @@ const FolderTreeItem = ({
             alignFitStart="center"
           />
         </span>
+        <VisuallyHiddenText text={isSelectable ? 'Selectable' : 'Non-selectable'} />
```
sycombs, line 227: Will add translations once we confirm the text we want read here.

chrismichalewicz:
@sycombs  I have reviewed for a11y and it sounds good, great work!  I have updated the a11y label. 

Tested on a PC with Microsoft Edge Version 121.0.2277.128, JAWS 2022 version 2022.2202.38

- JAWS communicated if an item is selectable and instructions for how to select it (Enter)
- Pressing Enter selects an item
- All other keyboard navigation continued to work as expected.

PR #2025, opened 2024-02-08
[terra-date-time-picker]Single field label SR announce

ashishkcerner:
### Summary
Field label was not reading on date-time-picker component .

**What was changed:**
Added the inputAttribute handling condition for reading label .

**Why it was changed:**

### Testing

This change was tested using:

- [ ] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10072 

---

@cerner/terra


packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -227,6 +228,10 @@ const DatePickerInput = (props) => {
     delete inputAttributes.id;
   }
 
+  if (inputAttributes && inputAttributes.fieldLabel) {
+    labelString = inputAttributes.fieldLabel;
+  }
+
```
saket2403, line 234: These code changes will not work if the consumer passes the prop name as `label` or anything else instead of `fieldLabel`. How will the consumer know that they have to pass fieldLabel always to the inputAttributes?

packages/terra-time-input/src/_AccessibleInput.jsx
```diff
@@ -66,9 +66,15 @@ function AccessibleInput(props) {
   const uuid = uuidv4();
   const labelId = `${uuid}-label`;
   const descriptionId = `${uuid}-description`;
+  let labelString;
+
+  if (inputProps && inputProps.fieldLabel) {
+    labelString = inputProps.fieldLabel;
+  }
+
```
saket2403, line 74: These code changes will not work if the consumer passes the prop name as label or anything else instead of fieldLabel. How will the consumer know that they have to pass fieldLabel always to the inputAttributes?

packages/terra-date-picker/src/DateInput.jsx
ashishkcerner: This is added because of there is no field component in dateTimePicker and to handle a reading of single label in both the component require to use it through fieldLabel .

packages/terra-time-input/src/_AccessibleInput.jsx
ashishkcerner: This is added because of there is no field component in dateTimePicker and to handle a reading of single label in both the component require to use it through fieldLabel .

ashishkcerner: Have updated the description regarding timeInputAttribute for this scenario.

ashishkcerner:
https://github.com/cerner/terra-framework/pull/2031 Have raised a new pR with all the changes

PR #2028, opened 2024-02-09
[terra-table] Added setTimeout to resizeObserver

saket2403:
### Summary

**What was changed:**

Added setTimeout to resizeObserver

**Why it was changed:**
To fix **ResizeObserver loop completed with undelivered notifications** error

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10206 

---

@cerner/terra


packages/terra-table/src/Table.jsx
```diff
@@ -398,22 +400,26 @@ function Table(props) {
...
 
+  const handleResize = () => {
+    setTableHeight(tableRef.current.offsetHeight - 1);
+
+    const tableContainer = tableContainerRef.current;
+    setTableScrollable(tableContainer.scrollWidth > tableContainer.clientWidth
+                      || tableContainer.scrollHeight > tableContainer.clientHeight);
+  };
+
+  const debouncedHandleResize = LodashDebounce(handleResize, 300);
```
ShettyAkarsh, line 411: what is this 300 hardcoded value ?

saket2403: It is a 300ms delay before invoking the function.

packages/terra-table/src/Table.jsx
```diff
@@ -35,6 +37,8 @@ const TableConstants = {
 
 const ROW_SELECTION_COLUMN_ID = 'table-rowSelectionColumn';
 
+const DEBOUNCE_TIMER = 300;
```
cm9361, line 40: Do we want to debounce or throttle the resize logic?  With a debounce, you won't see any updates until you finish resizing.  Also, 300ms seems pretty high for a delay after the resize is completed.

packages/terra-table/src/Table.jsx
```diff
@@ -398,22 +402,26 @@ function Table(props) {
...
-    });
+    const resizeObserver = new ResizeObserver(debouncedHandleResize);
 
     resizeObserver.observe(tableRef.current);
 
     return () => {
       resizeObserver.disconnect();
     };
-  }, [tableRef]);
+  }, [debouncedHandleResize]);
```
cm9361, line 424: This logic is incorrect.  You are creating a new instance of debouncedHandleResize on every render.  So your useEffect will not trigger on every render and nullify the memoization.

sugan2416 (approved):

ShettyAkarsh (approved):

PR #2030, opened 2024-02-12
[terra-date-picker] - +/- functionality changes and Changed type to number

PK106552:
### Summary

**What was changed:**

1. `+ / -` keys should change the value of the respective field on which the focus is there rather than always updating the day field.
2. Changed `Up/Down` key functionality same as` +/-` functionality to make uniform.
3. Changed input type to `number` instead of text.

**Why it was changed:**

1. Day field is incrementing/decrementing instead of focused field when we press` + / -` keys.
2. Up/Down key functionality was not same as +/- functionality.
3. Input type was text.

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10138 

---

@cerner/terra


packages/terra-date-picker/tests/wdio/date-picker-spec.js
```diff
@@ -970,15 +974,15 @@ Terra.describeViewports('Date Picker', ['medium'], () => {
           expect($('input')).toHaveValue(tomorrow.toISOString().split('T')[0]);
         });
 
-        it('sets Date to tomorrow if on year input', () => {
+        it('sets yesr to last year if on year input', () => {
```
saket2403, line 977:
```suggestion
        it('sets year to next year if on year input', () => {
```

PK106552: Updated

packages/terra-date-picker/tests/wdio/date-picker-spec.js
```diff
@@ -478,11 +479,8 @@ Terra.describeViewports('Date Picker', ['medium'], () => {
   });
 
   describe('Initial Focus', () => {
-    before(() => {
+    beforeEach(() => {
       browser.url('/#/raw/tests/cerner-terra-framework-docs/date-picker/date-picker-default');
-      $('input[name="terra-date-month-date-input"]').setValue('12');
-      $('input[name="terra-date-day-date-input"]').setValue('18');
-      $('input[name="terra-date-year-date-input"]').setValue('1994');
```
saket2403, line 485: You need to remove all the tests in this describe block because the date picker no longer supports clearing the field with a single delete press.

PK106552: Updated - https://github.com/cerner/terra-framework/pull/2030/commits/752e707c0d3076d1cd22f145b5b362976cd3cc5e

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -614,6 +596,10 @@ const DatePickerInput = (props) => {
     } else if (inputType === DateUtil.inputType.DAY) {
       handleDayInputKeydown(event);
     }
+
+    if (DateUtil.isMac() && !event.key.match(/^[0-9]/g) && !(event.keyCode === KEY_BACK_SPACE || event.keyCode === KEY_DELETE || event.keyCode === KEY_TAB || event.keyCode === KEY_RIGHT || event.keyCode === KEY_LEFT)) {
```
sugan2416, line 600: why mac specific here ? Did we miss adding here as well: https://github.com/cerner/terra-framework/pull/2017/files#diff-37c08a2b0d74d4e14dd7fc4e7e160bb6ac190dd983aebaaf30397c71621b735dR479

sugan2416:
Now since the +/-/Up/Down keys are changing dates of the focused field instead of the next and previous dates. Translations may need to be updated to account for this
Screenshot 2024-02-15 at 1 51 21 PM
Currently it mentions next and previous dates but this functionality has been changed. Also it may need to mention about Up/Down arrows as well.
This needs to be discussed with UX. May need to be addressed in a later PR.

SwethaM03:
We can use the below instruction,
Date Format: MM/DD/YYYY. Thursday, February 15, 2024, Press T to set current date, plus or up arrow key for next date and minus or down arrow key for previous date

sugan2416 (approved):

PK106552:
> We can use the below instruction, Date Format: MM/DD/YYYY. Thursday, February 15, 2024, Press T to set current date, plus or up arrow key for next date and minus or down arrow key for previous date

Created JIRA to update instruction - https://jira2.cerner.com/browse/UXPLATFORM-10232

SwethaM03:
+1

PR #2043, opened 2024-02-20
[terra-flowsheet-data-grid] Allow flowsheet sections to have sticky title

sugan2416:
### Summary

**What was changed:**

Changed flowsheet sections to have sticky titles.

**Why it was changed:**

When flowsheet with sections are height restricted and have more cells in row, the section header is still absolute in its position. Due to this title of `terra-section-header` is made sticky during horizontal scroll.
https://github.com/cerner/terra-core/pull/4037 
https://github.com/cerner/terra-core/pull/4043

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] No tests are needed

### Reviews

- [x] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10215 

---

@cerner/terra

### DIFF
--- a/packages/terra-table/src/subcomponents/Section.jsx
+++ b/packages/terra-table/src/subcomponents/Section.jsx
@@ -164,7 +164,7 @@ function Section(props) {
               className={cx('section-header')}
               text={text}
               isOpen={hasSectionButton ? !isCollapsed : undefined}
-              isTitleFixed
+              isTitleSticky
               onClick={hasSectionButton ? handleClick : undefined}
             />
           </th>
--- a/packages/terra-table/src/subcomponents/Section.jsx
+++ b/packages/terra-table/src/subcomponents/Section.jsx
@@ -101,6 +101,10 @@ const propTypes = {
    * Function that is called when a collapsible section is selected. Parameters: `onSectionSelect(sectionId)`
    */
   onSectionSelect: PropTypes.func,
+  /**
+   * Bounding container for the table, will use window if no value provided.
+   */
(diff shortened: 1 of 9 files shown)


eawww (approved):
@sugan2416 I didn't see an example or test example that demonstrated the behavior this is supposed to fix, but I was able to recreate it by manually changing styles on the existing examples. That means extra steps for each reviewer, and most UX reviewers aren't going to poke around in code. To make reviewing easier in the future, could you be sure a test example demonstrating the fix is included. Instructions in the PR on how to navigate to and recreate the conditions would be great too.

If something was available and I just missed it, please disregard most of the above.

Anyway, it looks good from a UX perspective!

saket2403 (approved):

ShettyAkarsh:
added do not  merge label as the section header changes has to be released and consumed here

packages/terra-table/src/subcomponents/Section.jsx
```diff
@@ -136,6 +141,8 @@ function Section(props) {
   const isGridContext = gridContext.role === GridConstants.GRID;
 
   const hasSectionButton = isCollapsible && onSectionSelect;
+  const boundedWidth = isCollapsible && boundingRef && boundingRef.current ? boundingRef.current.clientWidth - 50 : null;
```
sugan2416, line 144: The 50px offset used here is to reduce button width (under header) further from the bounding width to maintain sticky header. It fails to stick when its exactly set to container width.

PR #2066, opened 2024-03-05
[terra-data-grid] Update to automatically focus on single buttons or hyperlinks in cells

sdadn:
### Summary

This PR updates the focus behavior to negate the need for dive-in mode in cells with only a single button or hyperlink and focus on them directly.

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10225

---

@cerner/terra


kenk2 (approved):

packages/terra-table/src/utils/focusableButton.js
```diff
@@ -0,0 +1,17 @@
+// This is a temporary file for the purpose of adding a new feature without need for refactoring across multiple components.
+// This will be removed is a separate refactor effort.
+
+const getFocusableButtonsOrHyperlinks = (parentElement) => {
+  const focusableElementSelector = 'a[href]:not([tabindex=\'-1\']), button:not([disabled]):not([tabindex=\'-1\'])';
```
adoroshk, line 5: I am wondering if we need to include buttons that are created like <div> tags with a role="button", or that is out of scope?

sdadn: I believe custom elements are out of scope for this feature.

adoroshk (approved):

packages/terra-framework-docs/src/terra-dev-site/doc/data-grid/WorklistDataGrid.1/Examples.3/WorklistDataGridFocusableCell.doc.mdx
```diff
@@ -10,10 +10,13 @@ To make these keys available for cell navigation as part of its basic functional
 Enabling single cell navigation will disable grid navigation and disabling single cell navigation will restore grid navigation.
 
 <span style={{color: 'blue'}}>
-<em>Single Cell navigation is built-in and does not require any additional setup or implementation.</em>
+<b>Notes</b>
+<ul>
+  <li><em>Single Cell navigation is built-in and does not require any additional setup or implementation.</em></li>
+  <li><em>Cells which contain only a focusable button or hyperlink will focus on those components directly without needing to diving into the cell.</em></li>
```
nikhitasharma, line 16:
```suggestion
  <li><em>Cells which contain only a focusable button or hyperlink will focus on those components directly without needing to dive into the cell.</em></li>
```

packages/terra-data-grid/tests/wdio/data-grid-spec.js
```diff
@@ -280,6 +264,26 @@ Terra.describeViewports('DataGrid', ['medium', 'large'], () => {
       Terra.validates.element('data-grid-focusable-select-retains-focus', { columnResizeSelector });
       expect(browser.$('#specialties').isFocused()).toBe(true);
     });
+
+    it('focuses on a button if it is the only component in a cell', () => {
```
nikhitasharma, line 268: Should we have a test with more than one focusable component in a cell?

packages/terra-table/src/subcomponents/Cell.jsx
```diff
@@ -213,7 +227,15 @@ function Cell(props) {
...
   }, [isGridContext]);
 
-  const onMouseDown = (event) => {
+  const handleFocus = () => {
+    const element = hasSingleButtonOrHyperlinkElement();
+
+    if (element) {
+      element.focus();
+    }
+  };
```
cm9361, line 236: Is this correct?  We don't want this behavior for the Terra Table.  This functionality should only be possible for a DataGrid component.

packages/terra-table/src/utils/focusableButton.js
cm9361: I think you would actually have to account for this.  You might need to exclude certain types rather than being so specific.  This selector seems too restrictive.

packages/terra-table/src/subcomponents/Cell.jsx
```diff
@@ -213,7 +227,15 @@ function Cell(props) {
     }
   }, [isGridContext]);
 
-  const onMouseDown = (event) => {
+  const handleFocus = () => {
```
cm9361, line 230: I don't think we want to add this logic to the onFocus synthetic event.

packages/terra-data-grid/tests/wdio/data-grid-spec.js
sdadn: I think that should be covered here:

https://github.com/cerner/terra-framework/blob/b4b9f88804cb4484fc16269a650c55c56f299bdb/packages/terra-data-grid/tests/wdio/data-grid-spec.js#L234-L239

packages/terra-table/src/subcomponents/Cell.jsx
cm9361, line 1: You did not account for the isInteractable state variable.  The voice reader commands will not be correct in the case that there is a button element based on the current changes.

packages/terra-table/src/subcomponents/Cell.jsx
```diff
@@ -266,7 +266,13 @@ function Cell(props) {
         case KeyCode.KEY_RETURN:
           // Lock focus into component
           if (isGridContext && hasFocusableElements()) {
```
cm9361, line 268: It would probably be better to create a method that checks for elements that are not input based fields.  You need a way to keep your conditions in sync.  Doing it this way won't allow that to happen.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -598,11 +616,11 @@ const DataGrid = forwardRef((props, ref) => {
           columnHeaderHeight={columnHeaderHeight}
           rowHeight={rowHeight}
           rowHeaderIndex={rowHeaderIndex}
+          onCellSelect={onCellSelect}
```
cm9361, line 619: Did we need to move this?  It was previously with the other selection callbacks.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -252,31 +255,46 @@ const DataGrid = forwardRef((props, ref) => {
...
-      } else {
-        // Set focus on input field (checkbox) of row selection cells.
-        focusedCell = grid.current.rows[newRowIndex].cells[newColIndex];
-        if (isRowSelectionCell(newColIndex) && focusedCell.getElementsByTagName('input').length > 0) {
-          [focusedCell] = focusedCell.getElementsByTagName('input');
-        }
+    let focusedCell;
 
-        // Set focus to column header button, if it exists
-        const isHeaderRow = (newRowIndex === 0 || (hasColumnHeaderActions && newRowIndex === 1));
```
cm9361, line 271: What happened to the header row logic?

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -252,31 +255,46 @@ const DataGrid = forwardRef((props, ref) => {
...
+
+    // Set focus on input field (checkbox) of row selection cells.
+    if (isRowSelectionCell(newColIndex) && focusedCell.getElementsByTagName('input').length === 1) {
+      [focusedCell] = focusedCell.getElementsByTagName('input');
+      focusedCell?.focus();
+      return;
+    }
+
+    // Set focus to a single header button or hyperlink if they are the only content in cell
+    if (!focusedCell.hasAttribute('tabindex') || focusedCell.getElementsByTagName('button').length === 1 || focusedCell.getElementsByTagName('a').length === 1) {
```
cm9361, line 290: Are we really limiting this to specifically button and anchor tags?  What if I had a div with a role of button, for example?  This logic seems a little too restrictive.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -252,31 +255,46 @@ const DataGrid = forwardRef((props, ref) => {
...
+      focusedCell = focusedCell.querySelector('[role="button"]') || focusedCell.querySelector('button') || focusedCell.querySelector('a');
```
cm9361, line 291: If you are using a querySelector, you could use a single selector instead of having 3 separate conditions to query the DOM.  You can use commas in your selector to separate.

packages/terra-data-grid/src/DataGrid.jsx
sdadn: This was an abandoned attempt at alphabetization, reverted.

packages/terra-data-grid/src/DataGrid.jsx
sdadn: I believe custom components were excluded, which is why that wasn't considered.  Unless if I misunderstood the requirements. @mjpalazzo are divs with button roles included?

packages/terra-data-grid/src/DataGrid.jsx
sdadn: Since header rows with action buttons also only have a single button, this overlaps with the current story to add focus to single buttons in any row. The isHeaderRow condition was removed, the rest of the logic was tweaked to accommodate single buttons/hyperlinks in any row.

packages/terra-data-grid/src/DataGrid.jsx
sdadn: Updated here: d86a31695501e5e103ba9f36adff2a349fedc074

packages/terra-table/src/subcomponents/Cell.jsx
sdadn: Moved this to the data grid component.

packages/terra-table/src/subcomponents/Cell.jsx
sdadn: Added.

packages/terra-data-grid/src/DataGrid.jsx
adoroshk: I tested the action button behavior and confirm that it works as expected after that change. As action button is a `<button>` element and the behavior for it is now a default behavior, it makes sense that the `isHeaderRow` condition is no longer needed.

mjpalazzo:
@sdadn - Overall this enhancement is working well. I have a couple of minor findings. 

1. When focus is on the hyperlink, the black solid line focus indicator is cut off on the left side. See attached screen shots. Focus indicator on button and cells look fine. See screenshots for these as well.
2. This is a nit but the column header text in the example is incorrect. For example the header that has "Column 1" in the header is actually column 2. This causes the screen reader to announce "Column 1" some helper text and then "Column 2"

WDG interactive cell 3 focus issue
WDG interactive cell 2
WDG interactive cell 1

sdadn:
The WDIO failure is caused by an another component. Terra-data-grid tests are all passing:

CleanShot 2024-03-14 at 09 27 36

CleanShot 2024-03-14 at 09 27 00

CleanShot 2024-03-14 at 09 27 14

nikhitasharma (approved):

adoroshk (approved):

packages/terra-table/src/subcomponents/Cell.jsx
sdadn: The voicereader output seems to be correct based on the accessibility review.

PR #2088, opened 2024-03-19
[terra-table] Prevent right click from selecting row in selection mode

sugan2416:
### Summary

**What was changed:**

- Prevents right click from selecting rows in row selection mode.
- Prevent click on interactive element in cell from selecting rows in row selection mode. 

**Why it was changed:**

- Unintended selection of row through right click.
- Unintended selection of row during interactive element click

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10312 

---

@cerner/terra


packages/terra-table/src/subcomponents/Cell.jsx
```diff
@@ -136,6 +136,12 @@ const propTypes = {
    * The color to be used for highlighting a column.
    */
   columnHighlightColor: PropTypes.oneOf(Object.values(ColumnHighlightColor)),
+
+  /**
+   * Enables row selection capabilities for the table.
+   * Use 'single' for single row selection and 'multiple' for multi-row selection.
+   */
+  rowSelectionMode: PropTypes.string,
```
saket2403, line 144: Should this be private? Since the rowSelectionMode is set by user on the Row and you are just passing the value down here internally.

sugan2416: Not needed I guess since there are other non-private props as well (like height) similar to this passed down internally to cell. But anyways marked private: https://github.com/cerner/terra-framework/pull/2088/commits/6ecfb18835da3125978259937907002382fac3cf

allanxavierm (approved):

packages/terra-table/tests/wdio/__snapshots__/reference/terra-default-theme/en/chrome_medium/table-spec/row-single-selection-cell-right-click.png
saket2403, line 1: Why is the row selected if right click I not supposed to select the row?

saket2403: Nvm this looks like the hover style and not the selected styles.

sugan2416:
@saket2403 @allanxavierm Added DNM label, as there was new requirement along with right click fix. Will include new fix in same PR

saket2403 (approved):

PR #2095, opened 2024-03-20
[terra-slide-panel] Disclosing node capturing method update

adoroshk:
This PR changes slide panel disclosing node capturing method that allows to use slide panel without **mainContent** prop.

### Summary

In order to return the focus to the disclosing node upon closing the Slide panel, the disclosing node needs to be captured upon panel opening. The existing implementation requires a **mainContent** prop to be passed, which would allow the Slide panel to create a main content node with **onClick** and **onKeyUp** event listeners being in charge of capturing disclosing node. With that approach in cases when **mainContent** prop is omitted, the disclosing node is not captured upon panel opening and there is no node saved to return focus upon panel closing.

**What was changed:**
1. The disclosing node capturing was moved from onClick and onKeyUp event listeners attached to the main content, to **componentDidUpdate** method. The **document.activeElement** is captured and saved as a panel disclosing node before Slide panel grabs focus on **isOpen** prop changing to true.
2. Setting **aria-expanded** attribute to the disclosing node has been updated to accommodate for the node role. As setting **aria-expanded** attribute to true with unsupported roles would cause accessibility violations, a method to check whether the role is supported has been added. More info on [roles associated with aria-expanded attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded#associated_roles).
3. The main content node is rendered only if **mainContent** prop was passed.
4. A test example and wdio tests for slide panel with no main content prop  were added to **terra-framework-docs**.
5. WDIO screenshots for **terra-agregator** test which uses slide panel were updated, as with new functionality added, the disclosing nodes in terra-agregator received the focus.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10319


packages/terra-aggregator/tests/wdio/__snapshots__/reference/clinical-lowlight-theme/en/chrome_large/aggregator-spec/close_from_explicit_close.png
kenk2, line 1: Is this right? This seems like it would result in an accessibility violation because there's not enough contrast.

kenk2 (commented):
Just 1 question, otherwise LGTM

packages/terra-aggregator/tests/wdio/__snapshots__/reference/clinical-lowlight-theme/en/chrome_large/aggregator-spec/close_from_explicit_close.png
sdadn: It looks like the highlighted style was updated? May need UX to confirm if this is the correct styling.

packages/terra-slide-panel/src/SlidePanel.jsx
```diff
@@ -85,33 +85,65 @@ const defaultProps = {
   panelSize: 'small',
 };
 
+/**
+ * This method ensures that aria-expanded attribute is allowed with the node role to avoid accessibility violations.
+ * More info on associated roles can be found here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded#associated_roles
+ * @param {HTMLElement} node - a node to check
+ * @returns true if node's role is compatible with aria-expanded attribute
+ */
+const roleIsCompartible = (node) => {
```
cm9361, line 94: What does compartible mean?

packages/terra-slide-panel/src/SlidePanel.jsx
```diff
@@ -85,33 +85,65 @@ const defaultProps = {
...
+    || role === 'link'
+    || role === 'listbox'
+    || role === 'menuitem'
+    || role === 'row'
+    || role === 'rowheader'
+    || role === 'tab'
+    || role === 'treeitem'
+    || role === null
+    || role === undefined
+  );
```
sdadn, line 111: This looks like a lot of conditions that will always be sequentially checked. Can we optimize it using a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) and the [Set.has method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/has)?

packages/terra-slide-panel/src/SlidePanel.jsx
```diff
@@ -85,33 +85,65 @@ const defaultProps = {
...
       return;
     }
 
     if (!this.props.isOpen && this.props.isOpen !== prevProps.isOpen) {
       if (this.disclosingNode?.focus) {
         // Return focus to the disclosing element
-        this.disclosingNode.setAttribute('aria-expanded', 'false');
+        if (this.disclosingNode.getAttribute('aria-expanded')) {
+          // Set aria-expanded attribute to false for nodes that had that attribute only
+          this.disclosingNode.setAttribute('aria-expanded', 'false');
```
cm9361, line 136: I am not clear on why we need to set the aria-expanded attribute on the disclosing node.  That element is not being expanded.

packages/terra-slide-panel/src/SlidePanel.jsx
```diff
@@ -85,33 +85,65 @@ const defaultProps = {
...
+    || role === 'listbox'
+    || role === 'menuitem'
+    || role === 'row'
+    || role === 'rowheader'
+    || role === 'tab'
+    || role === 'treeitem'
+    || role === null
+    || role === undefined
+  );
+};
```
cm9361, line 112: We should not need a method like this.  It makes me feel like we might be doing something incorrectly.

packages/terra-slide-panel/src/SlidePanel.jsx
```diff
@@ -122,13 +154,12 @@ class SlidePanel extends React.Component {
...
-    this.lastClicked = event.target;
-  }
-
   setDisclosingNode(node) {
     if (node) {
-      node.setAttribute('aria-expanded', 'true');
+      if (roleIsCompartible(node)) {
+        // Set aria-expanded attribute to true for nodes with associated roles or no role only
+        node.setAttribute('aria-expanded', 'true');
+      }
```
cm9361, line 162: This node is not being expanded.  I don't think we should be adding this role.

packages/terra-slide-panel/src/SlidePanel.jsx
sdadn: And I'm confused why are we testing for all the possible values if we don't care that role is null/undefined?

packages/terra-slide-panel/src/SlidePanel.jsx
adoroshk: aria-expanded attribute wasn't added by me, it was there before the change, I just added a role check, as it can be added [only with specific roles](aria-expanded attribute), otherwise it causes accessibility violation. I'll double check with @mjpalazzo if aria-expanded attribute needs to be removed.

packages/terra-slide-panel/src/SlidePanel.jsx
adoroshk: @sdadn aria-expanded attribute should be used [only with specific roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded#associated_roles), otherwise it causes accessibility violation. This attribute was not introduced in this change, but it was causing accessibility violations with specific roles only, so I added a check. I'll double check with @mjpalazzo what would be the best way to address that

packages/terra-slide-panel/src/SlidePanel.jsx
adoroshk: aria-expanded attribute was not added at this change, it was added before. I'll check with @mjpalazzo if it can be removed.

packages/terra-slide-panel/src/SlidePanel.jsx
adoroshk: It checks if node's role is one compatible (sorry for typo, will fix if we won't remove it at the end) with aria-expanded attribute. Here is the documentation for [roles associated with aria-expanded attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded#associated_roles).

packages/terra-aggregator/tests/wdio/__snapshots__/reference/clinical-lowlight-theme/en/chrome_large/aggregator-spec/close_from_explicit_close.png
adoroshk: WDIO screenshots for terra-agregator test example which uses slide panel were updated, as with new functionality added, the disclosing nodes in terra-agregator received the focus. However, the focus style itself hasn't changed, it just wasn't receiving focus before as Slide Panel had no means to return the focus to that node. The affected example is a test example for some other functionality, it is not presented on the doc page and is not relevant to anything. I don't think we should fix it or log a bug.

adoroshk:
@mjpalazzo, it looks like we need your help.
1. Slide panel has [`aria-expanded` prop](https://github.com/cerner/terra-framework/blob/9a2277de28e890121aed672dc1cb62ecbb2f77f1/packages/terra-slide-panel/src/SlidePanel.jsx#L131) added to the side panel disclosing element (a button that user clicked to open the panel). That was added 11 month ago and my understanding is it was added as a part of accessibility update. As I received engineers questions whether it needs to be removed, please clarify if that should stay.
2. Per the documentation the `aria-expanded` attribute can be added only with [associated roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded#associated_roles), so I added a check it disclosing node's role is compatible with aria-expanded attribute or there is no role at all. Please let me know if that is a correct way to address the issue, or needs to be changed.
cc: @cm9361, @sdadn

packages/terra-slide-panel/src/SlidePanel.jsx
cm9361: Oh, it is just a typo :).  I still don't think we need a method like this though.

adoroshk:
1. The check for the role was removed.
2. I met with @mjpalazzo, demonstrated existing workflow and discussed a possibility of adding `aria-controls` attribute. It has been decided to keep existing functionality without `aria-controls` attribute for now.

packages/terra-slide-panel/src/SlidePanel.jsx
adoroshk: Removed

packages/terra-slide-panel/src/SlidePanel.jsx
adoroshk: Removed, the role in question changed in the example

cm9361 (approved):

sdadn (approved):

PR #2096, opened 2024-03-20
[terra-table] [terra-data-grid] Add column span functionality to FlowsheetDataGrid and Table

nikhitasharma:
### Summary

**What was changed:**
Added column span functionality to Terra Table to support it in FlowsheetDataGrid

Notes:
- The column highlighting functionality is disabled for columns with multiple column spans. 
The below example has column highlighting for Illness Severity but will not be supported since it has column span of 2. 
image

**Why it was changed:**
To support multiple column spans and avoid histories to break on page loads.

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10182 

---

@cerner/terra


adoroshk (commented):
Are column actions supposed to be unsupported with `columnSpan` prop? I check that in current implementation column actions don't work with column span, but in documentation I don't see a mentioning it's not supported, like, for example, Column Highlighting functionality.

nikhitasharma:
> Are column actions supposed to be unsupported with `columnSpan` prop? I checked that in current implementation column actions don't work with column span, but in documentation I don't see a mentioning it's not supported, like, for example, Column Highlighting functionality.

Thats a good catch, It is simple prop addition for column actions header to support column spans. I have added that in and it should work now:
image

packages/terra-framework-docs/src/terra-dev-site/doc/data-grid/FlowsheetDataGrid.4/Examples.6/FlowsheetDataGridWithColumnSpan.7.doc.mdx
```diff
@@ -0,0 +1,11 @@
...
+
+# Flowsheet Data Grid With Column Span
+
+### Description
+
+This example demonstrates a [Flowsheet Data Grid](/components/cerner-terra-framework-docs/data-grid/flowsheet-data-grid/about) with column span support.
+
+<FlowsheetDataGridWithColumnSpan title='Flowsheet Data Grid With Column Span' />
+
+Note: Column Highlighting functionality will not be supported for columns with column span.
```
sycombs, line 11: nit: slight wording change since this (i assume) is how it works currently.

```suggestion
Note: Column Highlighting is not supported for columns with column span.
```

mjpalazzo:
@nikhitasharma - Should column header actions such as sort disabled for column span?

nikhitasharma:
> @nikhitasharma - Should column header actions such as sort disabled for column span?

@mjpalazzo  If column header actions is something we do not want to support with column span, we can bypass that as well similar to column highlighting. But I don't think we can restrict what users pass in for the onClick function for actions. 
What are your thoughts on this.

mjpalazzo:
> > @nikhitasharma - Should column header actions such as sort disabled for column span?
> 
> @mjpalazzo If column header actions is something we do not want to support with column span, we can bypass that as well similar to column highlighting. But I don't think we can restrict what users pass in for the onClick function for actions. What are your thoughts on this.

@nikhitasharma - Please disregard my question. I did not mean column header actions. My concern was about column states such as sort, error, and selectable, but those are supported by Worklist Data Grid with onColumnSelect, not Flowsheet which does not have the onColumnSelect prop.

adoroshk:
I just noticed that the pinned column border style might be affected if the pinned columns have spans:
pinned_column_border

nikhitasharma:
Currently looking into both of these issues @adoroshk  @mjpalazzo.

packages/terra-framework-docs/src/terra-dev-site/doc/data-grid/FlowsheetDataGrid.4/Examples.6/FlowsheetSectionsWithColHighlight.doc.mdx
```diff
@@ -16,3 +16,5 @@ This example demonstrates a [Flowsheet Data Grid](/components/cerner-terra-frame
 - `columnHighlightDescription` can be used to provide context regarding the highlighted column to screen readers.
 
 <FlowsheetSectionsWithColHighlight />
+
+Note: Column Highlighting functionality will not be supported for columns with column span.
```
sycombs, line 20: Can we do the same as https://github.com/cerner/terra-framework/pull/2096/files#diff-f7d0eebfcee8de5c7b6734eb22dd4fbb2eacbcfd1ba69389b7ca4d7ae4e21d71R11 for all the places this line is added?

nikhitasharma: Updated [here](https://github.com/cerner/terra-framework/pull/2096/commits/087ab290d418c6c6d43eeb6be0e4edb731c5e0a1)

packages/terra-table/src/subcomponents/Cell.jsx
```diff
@@ -384,12 +390,13 @@ function Cell(props) {
...
   const cellLeftEdge = (columnIndex < columnContext.pinnedColumnOffsets.length) ? columnContext.pinnedColumnOffsets[columnIndex] : null;
   const CellTag = isRowHeader ? 'th' : 'td';
-  const columnHeaderId = `${tableId}-${columnId}-headerCell`;
+  const columnHeaderId = `${tableId}-${columnId.split('_')[0]}-headerCell`;
   const rowHeaderId = !isRowHeader && rowHeaderIndex !== -1 ? `${tableId}-rowheader-${rowId} ` : '';
   const sectionHeaderId = sectionId ? `${tableId}-${sectionId} ` : '';
 
   let columnHighlight = {};
-  if (columnHighlightColor) {
+  // Column highlighting is not supported for multiple column spans
```
cm9361, line 398: If column highlighting is not supported for column with a colspan, why not just update the attributes of the column shape such that there is no column highlighting in this scenario.  We should not have to update the cell component.  The cell component doesn't need to know how many columns are spanned.

packages/terra-table/src/subcomponents/Cell.jsx
cm9361, line 1: How will the consumer know which cell was selected in the onCellSelect callback.  I thought we were going to return a colSpanIndex property or something similar.  We cannot modify the column id value.

mjpalazzo:
@nikhitasharma - This is making good progress. Keyboard navigation is working as required. While testing with JAWS I encountered a few issues. I will recreate the steps below:
Pressed Tab to navigate to data grid, column 1 row 1. JAWS announced name role value as required.
Pressed Right Arrow to navigate the column header cells. JAWS announced the column labels and position as required.
From column 6 row 1 (March 18) pressed Down Arrow. Focus went to column 4 row 2, JAWS announced row header, cell contents, cell position correctly, but **focus should have gone to column 6, row 2**.
Pressed Right Arrow to navigate cells in row 2. JAWS announced column header, cell content, and cell position as required.
Pressed Up Arrow from column 6 row 2 and nothing happened. **I could not navigate to the column header row from column 6 row 2, column 7 row 2, or column 8 row 2.**
Pressed Left Arrow to navigate to column 5 row 2. Pressed Up Arrow and nothing happened. **I could not navigate to the column header row from column 6 row 2.
Pressed Left Arrow to navigate to column 4 row 2. Pressed Up Arrow and focus went to column 6 row 1. JAWS announced "March 18 column header column 6, row 1.

sycombs (approved):

packages/terra-table/src/subcomponents/ColumnHeader.jsx
```diff
@@ -141,13 +141,15 @@ const ColumnHeader = (props) => {
         className={cx('column-header-row', { hidden: !hasVisibleColumnHeaders })}
         height={hasVisibleColumnHeaders ? headerHeight : undefined}
       >
-        {columns.map((column, columnIndex) => (
+
+        {columns.filter((column) => (column.columnSpan > 1 || !(column.columnSpanIndex))).map((column, columnIndex) => (
```
cm9361, line 145: I would not expect the column header component to require changes.  We would continue to use our column shape information.  There should not be a need to filter the data.

packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -330,17 +336,20 @@ const ColumnHeaderCell = (props) => {
 
   // Add column highlight indicator based on color
   let columnHighlightIcon;
-  if (columnHighlightColor === ColumnHighlightColor.GREEN) {
-    columnHighlightIcon = <svg className={cx('highlight-icon-svg')} xmlns="http://www.w3.org/2000/svg"><circle className={cx('highlight-icon-circle')} r="3" cx="110%" cy="11" transform="translate(-5)" /></svg>;
-  } else if (columnHighlightColor === ColumnHighlightColor.ORANGE) {
-    columnHighlightIcon = <svg className={cx('highlight-icon-svg')} xmlns="http://www.w3.org/2000/svg"><rect className={cx('highlight-icon-square')} x="110%" y="7.5" transform="translate(-8)" /></svg>;
+  // Column highlighting is supported for single column spans only
+  if (!columnSpan || columnSpan === 1) {
```
cm9361, line 340: Why not just only pass the color highlight color when there is no column span.  Then the header cell component would not require the column span information.

packages/terra-table/src/subcomponents/Row.jsx
```diff
@@ -174,6 +174,8 @@ function Row(props) {
       {cells.map((cellData, index) => {
         const cellColumnIndex = index + columnIndexOffSet;
         const columnId = displayedColumns[cellColumnIndex].id;
+        const { columnSpan, columnSpanIndex } = displayedColumns[cellColumnIndex];
```
cm9361, line 177: Shouldn't this be combined into one line of code?

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -195,6 +195,25 @@ const DataGrid = forwardRef((props, ref) => {
...
+          newDisplayedColumns[i] = { id: `${column.id}`, columnSpanIndex: `${column.columnSpan - counter + 1}` };
+          i += 1;
+        }
+      } else {
+        newDisplayedColumns[i] = column;
+        i += 1;
+      }
+    });
+    return newDisplayedColumns;
+  }, [displayedColumns]);
```
cm9361, line 216: Can we not use a spread operator and simply add the new columnSpanIndex property?

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -157,6 +157,38 @@ function FlowsheetDataGrid(props) {
...
+          ...column,
+          isSelectable: false,
+          isSelected: false,
+          isResizable: false,
+        };
+        i += 1;
+      }
+    });
+    return newDisplayedColumns;
+  }, [columns]);
```
cm9361, line 190: It seems like we are duplicating this logic in the DataGrid and Flowsheet components.  Can it not be centralized to the DataGrid component?

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -157,6 +157,38 @@ function FlowsheetDataGrid(props) {
...
+    let i = 0;
+    const newDisplayedColumns = [];
+    columns.forEach((column) => {
+      if (column.columnSpan > 1) {
+        newDisplayedColumns[i] = {
+          ...column,
+          columnSpanIndex: 0,
+          isSelectable: false,
+          isSelected: false,
+          isResizable: false,
```
cm9361, line 170: Why ware we changing these property values?

packages/terra-table/src/subcomponents/Cell.jsx
```diff
@@ -384,12 +390,13 @@ function Cell(props) {
   // Determine table cell header attribute values
   const cellLeftEdge = (columnIndex < columnContext.pinnedColumnOffsets.length) ? columnContext.pinnedColumnOffsets[columnIndex] : null;
   const CellTag = isRowHeader ? 'th' : 'td';
-  const columnHeaderId = `${tableId}-${columnId}-headerCell`;
+  const columnHeaderId = `${tableId}-${columnId.split('_')[0]}-headerCell`;
```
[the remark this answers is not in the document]
nikhitasharma: Removed this as it was stale code from previous code logic.

packages/terra-table/src/subcomponents/Cell.jsx
nikhitasharma: Updated the logic.

packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
nikhitasharma: Updated.

packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -366,7 +375,7 @@ const ColumnHeaderCell = (props) => {
 
   // Create cell content
   let cellContent;
-  if (isActionCell) {
+  if (isActionCell && !(columnSpan > 1)) {
```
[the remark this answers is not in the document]
nikhitasharma: Removed this as it was stale code from previous code logic.

packages/terra-table/src/Table.jsx
```diff
@@ -308,7 +310,25 @@ function Table(props) {
...
+      i += 1;
+      for (let counter = column.columnSpan; counter > 1; counter -= 1) {
+        displayedColumnsWithColumnSpan[i] = { id: `${column.id}`, columnSpanIndex: (column.columnSpan - counter + 1) };
+        i += 1;
+      }
+    } else {
+      displayedColumnsWithColumnSpan[i] = column;
+      i += 1;
+    }
+  });
```
cm9361, line 329: Don't we have similar logic in the DataGrid components?  If we have the logic here why do we need it in those components.  Also, we should just be able to use a spread operator for the properties.  It seems like this can be simplified quite a bit.

packages/terra-table/src/Table.jsx
```diff
@@ -308,7 +310,25 @@ function Table(props) {
     return (hasSelectableRows ? [tableRowSelectionColumn] : []).concat(pinnedColumns).concat(overflowColumns);
   }, [hasSelectableRows, intl, onRowSelectionHeaderSelect, overflowColumns, pinnedColumns]);
 
-  const [tableColumns, setTableColumns] = useState(displayedColumns.map((column) => initializeColumn(column)));
+  // Create new displayedColumns object to pass to Section sub component to account for column spans
+  const displayedColumnsWithColumnSpan = [];
```
cm9361, line 314: Do we need both displayedColumns and displayedColumnsWithColumnSpans?  It seems like we could have one object.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -195,6 +196,25 @@ const DataGrid = forwardRef((props, ref) => {
   } = props;
 
   const displayedColumns = (hasSelectableRows ? [WorklistDataGridUtils.ROW_SELECTION_COLUMN] : []).concat(pinnedColumns).concat(overflowColumns);
+  // Create new displayedColumns object to account for column spans
+  const displayedColumnsWithColumnSpan = useMemo(() => {
```
cm9361, line 200: Do we need both of these objects?  It seems like we should be able to drive off one.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -339,7 +374,8 @@ function FlowsheetDataGrid(props) {
       if (anchorCell.current === null && !gridRef.rows[rowIndex].hasAttribute('data-section-id')) {
         anchorCell.current = {
           rowId: gridRef.rows[rowIndex].getAttribute('data-row-id'),
-          columnId: columns[columnIndex].id,
+          columnId: displayedColumnsWithColumnSpan[columnIndex].id,
+          columnIndex,
```
cm9361, line 378: Why was this change required?

packages/terra-table/src/subcomponents/ColumnHeader.jsx
nikhitasharma: Updated.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
nikhitasharma: If you are asking about why columnIndex was added, the cell selection logic won't work based off of just the column ID and row ID anymore because there is no more 1-1 mapping between cells and column IDs with column spans. We need the column Index for cell selection to work properly.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -195,6 +196,25 @@ const DataGrid = forwardRef((props, ref) => {
...
+    let i = 0;
+    const newDisplayedColumns = [];
+    displayedColumns.forEach((column) => {
```
cm9361, line 203: The forEach function provides an index as a parameter to its callback function.  Is there a reason for the "i" variable that is not covered by this available argument?

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -195,6 +196,25 @@ const DataGrid = forwardRef((props, ref) => {
...
+      if (column.columnSpan > 1) {
+        newDisplayedColumns[i] = { ...column, columnSpanIndex: 0 };
+        i += 1;
+        for (let counter = column.columnSpan; counter > 1; counter -= 1) {
+          newDisplayedColumns[i] = { id: `${column.id}`, columnSpanIndex: `${column.columnSpan - counter + 1}` };
```
cm9361, line 208: Shouldn't we be using the spread operator here?

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -195,6 +196,25 @@ const DataGrid = forwardRef((props, ref) => {
...
+          i += 1;
```
cm9361, line 209: I believe this should all be handled by a single for loop.  You should not have to manually set the first element.

cm9361:
You need logic to properly handle the focus event for the DataGrid component.

packages/terra-data-grid/src/DataGrid.jsx
cm9361, line 203: Would a reducer be a better pattern for this functionality?  It seems like the accumulator could help simplify this logic similar to how we deal with the row indexes for sections in the table.

mjpalazzo:
@nikhitasharma - This is working much better now. All the navigation issus I called out above are resolved.
I found 1 issue with the End key. When I navigate to the column 1 row 1 using the Home key all is well. If I then press the End key JAWS announces "End" but there is no change in visual focus.
When I am on other table rows, Home and End work fine. The issue occurs in just in the header row.

packages/terra-table/src/subcomponents/Cell.jsx
nikhitasharma: It should now return the columnSpanIndex on selection.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -327,8 +347,18 @@ const DataGrid = forwardRef((props, ref) => {
...
+      const columnHeaderRect = grid.current.rows[0].getBoundingClientRect();
+
+      // Set new cell column index value to be within Column Header and Column Header actions bounding indexes 
+      if(toCell.row === 0 && hasVisibleColumnHeaders && toCell.col > displayedColumns.length) {
+        toCell.col = displayedColumns.length - 1;
+      }
+      if(toCell.row === 1 && hasColumnHeaderActions && toCell.col > displayedColumns.length ) {
+        toCell.col = displayedColumns.length - 1;
+      }
+
```
cm9361, line 359: We have logic to handle the HOME and END scenarios in the handleKeyDown method.  The necessary correction should be there and not in handleMoveCellFocus.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -327,8 +347,18 @@ const DataGrid = forwardRef((props, ref) => {
     if (!isSection(toCell.row)) {
     // Obtain coordinate rectangles for grid container, column header, and new cell selection
       const gridContainerRect = tableContainerRef.current.getBoundingClientRect();
-      const columnHeaderRect = grid.current.rows[0].cells[toCell.col].getBoundingClientRect();
+      const columnHeaderRect = grid.current.rows[0].getBoundingClientRect();
+
+      // Set new cell column index value to be within Column Header and Column Header actions bounding indexes 
+      if(toCell.row === 0 && hasVisibleColumnHeaders && toCell.col > displayedColumns.length) {
```
cm9361, line 353: Why would toCell.col be greater than the number of column headers?  There seems to be a logical error in the design somewhere.

packages/terra-data-grid/src/DataGrid.jsx
nikhitasharma: Moved this to handleKeyDown

packages/terra-data-grid/src/DataGrid.jsx
nikhitasharma: The length of the new columns object we are creating is not equal to the original length of columns because of spans. So we do need the `i` variable.

packages/terra-data-grid/src/DataGrid.jsx
nikhitasharma: Discussed offline. We do need these two objects.

packages/terra-table/src/Table.jsx
nikhitasharma: Removed this from Table code and using the one created at DataGrid Level here.

packages/terra-table/src/Table.jsx
nikhitasharma: Removed this from Table and re-using object from DataGrid level

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
nikhitasharma: These values are required to be added to Column Object for things to work properly. Similar to how we have [here](https://github.com/cerner/terra-framework/blob/main/packages/terra-data-grid/src/FlowsheetDataGrid.jsx#L156). 
But this has been updated to use the flowsheetColumns to create the new object.

mjpalazzo:
@nikhitasharma - All keyboard interactions and screen reader announcements are working as required. Great work!
One minor observation about the column span with header actions example. The labels of the action buttons in the action header row do not match the label of the column header. For example, in the March 17 column, the action header says "March 16 Action"

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -195,7 +203,7 @@ const DataGrid = forwardRef((props, ref) => {
   } = props;
 
   const displayedColumns = (hasSelectableRows ? [WorklistDataGridUtils.ROW_SELECTION_COLUMN] : []).concat(pinnedColumns).concat(overflowColumns);
-
+  const tableColumns = tableBodyColumns.length > 0 ? tableBodyColumns : displayedColumns;
```
cm9361, line 206: I would expect that we would send data the same way from the WorklistDataGrid and FlowsheetDataGrid.  It seems weird to have two separate implementations in a reusable base asset.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -157,6 +157,35 @@ function FlowsheetDataGrid(props) {
...
+            columnSpanIndex: (column.columnSpan - counter + 1),
+            columnHeaderIndex,
+            columnHighlightColor: undefined,
+          };
+          i += 1;
+        }
+      } else {
+        newDisplayedColumns[i] = { ...column, columnHeaderIndex };
+        i += 1;
+      }
```
cm9361, line 184: You should be able to handle all of this in a single for loop.  If the column span is undefined, you could just initialize the value to 1 for your test condition.  Then you would not require an if/else statement or have to set a value before entering the loop.  You can also use the push function to add values to your array so that you don't have to manually track the index.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -157,6 +157,35 @@ function FlowsheetDataGrid(props) {
     isResizable: false,
   })), [columns]);
 
+  const tableBodyColumns = useMemo(() => {
```
cm9361, line 160: Why do we need both the flowsheetColumns and tableBodyColumns objects?  We should be able to combine the initialization logic and use a single object to simplify the code.

packages/terra-table/src/subcomponents/Cell.jsx
```diff
@@ -144,6 +144,17 @@ const propTypes = {
...
    * @private
+   * The column span value for a column.
+   */
+  columnSpan: PropTypes.number,
+
+  /**
+   * @private
+   * The column span index value for a column.
+   */
+  columnSpanIndex: PropTypes.number,
```
cm9361, line 155: Shouldn't the cell only need access to the columnSpanIndex?  The Cell component should not care about the column span.  It is not a Cell component concern.

packages/terra-table/src/subcomponents/ColumnHeader.jsx
```diff
@@ -141,13 +140,15 @@ const ColumnHeader = (props) => {
...
       >
+
         {columns.map((column, columnIndex) => (
           <ColumnHeaderCell
             key={`${column.id}-headerCell`}
             id={`${column.id}-headerCell`}
             tableId={tableId}
             columnId={column.id}
             columnIndex={columnIndex}
+            columnSpan={column.columnSpan}
```
cm9361, line 151: Adding the span attribute to the <col> element of the table would eliminate the need to pass this information down as a prop to the ColumnHeader and ColumnHeaderCell components.

packages/terra-table/src/subcomponents/Row.jsx
```diff
@@ -180,6 +181,8 @@ function Row(props) {
...
```
cm9361, line 184: The Row component should not care about the column span.

packages/terra-table/src/Table.jsx
```diff
@@ -104,6 +104,11 @@ const propTypes = {
    */
   overflowColumns: PropTypes.arrayOf(columnShape),
 
+  /**
+   * Table body columns information.
+   */
+  tableBodyColumns: PropTypes.arrayOf(columnShape),
```
cm9361, line 110: The Terra Table component should not have two different ways to accept its column data.  We should not make a change like this as it will be difficult to maintain.

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -578,9 +608,9 @@ const DataGrid = forwardRef((props, ref) => {
 
         // Check for last focused column ID. If found set the index. Otherwise set it to the last focused column or last index.
         if (focusedCellRef.current.columnId) {
-          newColumnIndex = displayedColumns.findIndex(column => column.id === focusedCellRef.current.columnId);
+          newColumnIndex = tableColumns.findIndex(column => (column.id === focusedCellRef.current.columnId && (column.columnSpanIndex && (column.columnSpanIndex === focusedCellRef.current.columnSpanIndex))));
```
cm9361, line 611: This condition is not correct.  You will not always have a column span index.  The condition should account for that properly.  You are just getting by due to the fail safe on the next line.

packages/terra-table/src/subcomponents/Row.jsx
```diff
@@ -203,6 +206,8 @@ function Row(props) {
             firstRowId={firstRowId}
             lastRowId={lastRowId}
             columnHighlightColor={displayedColumns[cellColumnIndex].columnHighlightColor}
+            columnSpan={columnSpan}
```
cm9361, line 209: The cell should not need the column span information.  A Cell component cannot span columns.

packages/terra-table/src/Table.jsx
```diff
@@ -427,28 +436,49 @@ function Table(props) {
...
     }
 
     if (pinnedColumns.length > 0) {
       offsetArray.push(cumulativeOffset);
-
+      cellOffsetArray.push(cellCumulativeOffset);
       lastPinnedColumnIndex = hasSelectableRows ? pinnedColumns.length : pinnedColumns.length - 1;
 
+      if (pinnedColumns[0].columnSpan > 1) {
+        for (let c = pinnedColumns[0].columnSpan; c > 1; c -= 1) {
```
cm9361, line 459: We should have a more meaningful iterator variable name than "c".  Does "c" stand for column?

packages/terra-table/src/Table.jsx
```diff
@@ -427,28 +436,49 @@ function Table(props) {
...
+          cellCumulativeOffset += tableColumns[tableColumns.findIndex((col) => col.id === pinnedColumns[0].id)].width;
```
cm9361, line 460: I don't think this logic is correct.  If there is a column span, each cell would not equal the width of the overall column.

packages/terra-table/src/Table.jsx
```diff
@@ -427,28 +436,49 @@ function Table(props) {
...
+          }
+        }
+        cellOffsetArray.push(cellCumulativeOffset);
+
         offsetArray.push(cumulativeOffset);
       });
     }
-    setPinnedColumnOffsets(offsetArray);
+    setPinnedColumnHeaderOffsets(offsetArray);
+    setPinnedColumnOffsets(cellOffsetArray);
```
cm9361, line 480: These two objects could not have the same source array if there is a column span.  That would mean that some of the logic is incorrect.

adoroshk (commented):
Some observations:
1. I just noticed that table with col span example is not working.
2. I was wondering how table with resizable columns work with col span. As I didn't see examples for that, I assume that colspan is currently not supported with column resizing (at least for now, as I understand it might be out of scope of this PR). In that case we might want to mention that in documentation (unless there is a plan to change that soon)

packages/terra-table/src/subcomponents/Cell.jsx
nikhitasharma: Updated.

packages/terra-table/src/subcomponents/ColumnHeader.jsx
nikhitasharma: We are passing all column parameters as props the same way as here. Why do we need to do it different for columnSpan?

packages/terra-table/src/subcomponents/Row.jsx
nikhitasharma: Updated.

packages/terra-data-grid/src/utils/WorklistDataGridUtils.js
```diff
@@ -2,7 +2,7 @@
  * The Object representation of the row selection column.
  */
 const ROW_SELECTION_COLUMN = {
-  id: 'WorklistDataGrid-rowSelectionColumn',
+  id: 'table-rowSelectionColumn',
```
cm9361, line 5: This line of code should not have been changed.

packages/terra-data-grid/src/DataGrid.jsx
nikhitasharma: Removed

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
nikhitasharma: Updated.

packages/terra-table/src/Table.jsx
nikhitasharma: Updated.

packages/terra-data-grid/src/DataGrid.jsx
nikhitasharma: Updated.

packages/terra-table/src/Table.jsx
nikhitasharma: Updated.

packages/terra-table/src/Table.jsx
nikhitasharma: Updated.

nikhitasharma:
> @nikhitasharma - All keyboard interactions and screen reader announcements are working as required. Great work! One minor observation about the column span with header actions example. The labels of the action buttons in the action header row do not match the label of the column header. For example, in the March 17 column, the action header says "March 16 Action"

@mjpalazzo  This should be fixed now.

packages/terra-data-grid/src/utils/WorklistDataGridUtils.js
nikhitasharma: Updated.

nikhitasharma:
> Some observations:
> 
> 1. I just noticed that table with col span example is not working.
> 2. I was wondering how table with resizable columns work with col span. As I didn't see examples for that, I assume that colspan is currently not supported with column resizing (at least for now, as I understand it might be out of scope of this PR). In that case we might want to mention that in documentation (unless there is a plan to change that soon)

@adoroshk  The table with column spans should work now. I have added resizable columns to the table example with column spans.

adoroshk (approved):

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -528,7 +546,7 @@ const DataGrid = forwardRef((props, ref) => {
       onCellRangeSelect(cellCoordinates.row, cellCoordinates.col, event.keyCode);
     }
 
-    if (nextRow >= gridRowCount || nextCol >= displayedColumns.length) {
+    if (nextRow >= gridRowCount || (!isSection(nextRow) && nextCol >= grid.current.rows[nextRow].cells.length)) {
```
cm9361, line 549: The change to add the isSection check does not seem correct.

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -326,15 +325,17 @@ function FlowsheetDataGrid(props) {
     for (let rowIdx = rowIndexTopBound; rowIdx <= rowIndexBottomBound; rowIdx += 1) {
       const rowIdToSelect = rowsToSearch[rowIdx].id;
       for (let colIdx = columnIndexLeftBound; colIdx <= columnIndexRightBound; colIdx += 1) {
-        const columnIdToSelect = columns[colIdx].id;
-        cellsToSelect.push({ rowId: rowIdToSelect, columnId: columnIdToSelect, sectionId });
+        const columnIdToSelect = tableBodyColumnsRef[colIdx].id;
+        cellsToSelect.push({
+          rowId: rowIdToSelect, columnId: columnIdToSelect, columnIndex: colIdx, sectionId,
```
cm9361, line 330: Shouldn't we be returning the column span index instead of a column index parameter?

packages/terra-data-grid/src/FlowsheetDataGrid.jsx
```diff
@@ -343,12 +344,15 @@ function FlowsheetDataGrid(props) {
         onRowSelect({ rowId: selectionDetails.rowId, sectionId: selectionDetails.sectionId, isMetaPressed: selectionDetails.isMetaPressed });
       }
     } else if (selectionDetails.isShiftPressed && anchorCell.current !== null) {
-      selectCellRange(selectionDetails.rowId, selectionDetails.columnId, selectionDetails.sectionId, selectionDetails.subsectionId);
+      selectCellRange(selectionDetails.rowId, selectionDetails.columnId, selectionDetails.columnIndex, selectionDetails.sectionId, selectionDetails.subsectionId);
```
cm9361, line 347: Shouldn't we be using the column span index instead of the column index?  Consumers would not be set up to drive off a column index.  They are using the column id to find columns.

packages/terra-data-grid/src/DataGrid.jsx
nikhitasharma: Adding this is necessary to allow for navigation between sections as the nextRow cells length in the grid ref for it would be 1 and navigation won't be allowed to the right sections and columns across sections.

mjpalazzo:
@nikhitasharma - I validated keyboard navigation and JAWS using the Edge browser. Everything is working as required. The issue in the column actions example where the button label does not match the column header is still there, but that is a minor issue with the example not the component.
Great job!

nikhitasharma:
> @nikhitasharma - I validated keyboard navigation and JAWS using the Edge browser. Everything is working as required. The issue in the column actions example where the button label does not match the column header is still there, but that is a minor issue with the example not the component. Great job!

Thanks for pointing that out clearly. I fixed it.

PR #2106, opened 2024-03-27
[terra-folder-tree] Fix incorrect screenreader announcement of indexes

sycombs:
### Summary

This PR fixes an issue where screenreaders would announce subfolder items with sibling parents as being part of the same group. This has been done by manually setting the values for `aria-setsize` and `aria-posinset` instead of relying on browser calculations.

### Testing

This change was tested using:

- [ ] WDIO
- [x] Jest
- [ ] No tests are needed
Tested with Voiceover and JAWS.

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10283 

---

@cerner/terra

### DIFF
--- a/packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
+++ b/packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
@@ -54,6 +54,16 @@ const propTypes = {
   level: PropTypes.number,
   /**
    * @private
+   * The position of the item among its sibling items in the same group (subfolder).
+   */
+  posInSet: PropTypes.number,
+  /**
+   * @private
+   * Number of sibling items in the same group (subfolder).
+   */
+  setSize: PropTypes.number,
+  /**
+   * @private
    * Ref to the parent folder of the current item.
    */
   parentRef: PropTypes.oneOfType([
@@ -83,6 +93,8 @@ const FolderTreeItem = ({
   level,
   onSelect,
   onToggle,
+  posInSet,
+  setSize,
   subfolderItems,
   parentRef,
   intl,
@@ -103,11 +115,13 @@ const FolderTreeItem = ({
       hidden={!isExpanded}
       ref={subFolderNode}
     >
-      {subfolderItems.map((item) => (
+      {subfolderItems.map((item, index) => (
         <FolderTreeItem
           {...item.props}
           intl={intl}
           level={level + 1}
+          setSize={subfolderItems.length}
+          posInSet={index + 1}
           parentRef={itemNode}
         />
       ))}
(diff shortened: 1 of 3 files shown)


adoroshk (approved):
I verified that while walking up/down the folder tree, it announces the level and item index in subset correctly. Looks like expected behavior.

artpark (approved):

mjpalazzo:
@sycombs - I did some testing with JAWS with the basic folder tree example and it sounds like the issue is fixed. The levels and item index are announced correctly. I will do more testing, but I found an issue with how JAWS announces the index of the items. When a folder is expanded and I navigate with Down Arrow, JAWS announces the label, selectable, state, helper text, and index 1 of 4. When I press Down Arrow again to navigate to the next item, JAWS announces the label, selectable, state, helper text, but not the index.  I checked the prod instance and it behaves the same way. JAWS only announces the index of the first item I navigate to. 
@eawww - Should JAWS announce the index of each item at the same level as the user navigates the tree?

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -52,6 +52,16 @@ const propTypes = {
    * Level of nesting for this item.
    */
   level: PropTypes.number,
+  /**
+   * @private
+   * The position of the item among its sibling items in the same group (subfolder).
+   */
+  posInSet: PropTypes.number,
```
sdadn, line 59: does this mean `position in set`? Can we rename this to `positionInSet` or `itemPosition` for clarity?

sycombs: I was torn between naming it `posInSet`, `ariaPosInSet`, or `positionInSet` as you suggested since this is a direct passthrough to `aria-posinset`. I'm not sure which is most appropriate - maybe `ariaPosInSet` since it matches the aria attribute?

mjpalazzo:
@sycombs - here are some observations:
Folder Tree Selection example -  I am having a difficult time verifying the index announcements in the . The index on the last item in the tree does not sound right, but I may be getting lost in the nesting levels. I will schedule time to do a screenshare so I can walk through it with you.
Folder Tree Expand Collapse example - the level in the label and level JAWS announces do not match.
Folder Tree Expand Collapse All example - the level in the label and level JAWS announces do not match.

packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
sdadn: Oh if it's a direct passthrough, then yeah, `ariaPosInSet` sounds good!

sycombs: Updated both prop names: [eccff30](https://github.com/cerner/terra-framework/pull/2106/commits/eccff30d49e9d7cc6eae52c235f11dfb90dc086a)

mjpalazzo:
> @sycombs - here are some observations: Folder Tree Selection example - The indexes are working correctly. Please regard my previous comment. I was in the wrong version. Folder Tree Expand Collapse example - the level in the label and level JAWS announces do not match. The indexes start with 0 but the labels start with 1. Folder Tree Expand Collapse All example - same comment.

Great work, @sycombs! Those examples are much better and the component is working perfectly! I approved.

PR #2110, opened 2024-03-29
[compact interactive list ] No row headers if list has only one semantic column

adoroshk:
### Summary
Compact interactive list allows users to set an index of a cell in the row that needs to be a row header via rowHeaderIndex prop. If rowHeaderIndex prop is omitted, it defaults to 0, which makes the first row cell a row header. However, if compact interactive list has only one semantic column, that would make each row to have only one cell with role = rowheader, which is redundant. 
This change checks if the list has a single column, and does not assign a role = 'rowheader' to the only cell in the row. 

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10330


packages/terra-compact-interactive-list/tests/jest/CompactInteractiveList.test.jsx
```diff
@@ -1087,4 +1087,147 @@ describe('Compact Interactive List', () => {
...
+
+    it('list with one column should match a snapshot', () => {
+      const wrapper = enzymeIntl.mountWithIntl(
+        <CompactInteractiveList
+          id="compact-interactive-list-with-one-column"
+          rows={singleColRows}
+          columns={singleCol}
+        />,
+      );
+      expect(wrapper).toMatchSnapshot();
```
kenk2, line 1151: Nit: IIRC we are kind of moving away from doing these kinds of snapshot tests? These tend to be hard to verify and we just overwrite the results anyways.

kenk2 (approved):
One nit but otherwise LGTM

packages/terra-compact-interactive-list/tests/jest/CompactInteractiveList.test.jsx
adoroshk: I added the snapshot to ensure the component rendered with correct role, but jest test checks the same so I am willing to remove it. It is a bit extra.

adoroshk: Removed in https://github.com/cerner/terra-framework/pull/2110/commits/76ce35a6f621e028bcd7d31645e48727199ca632

packages/terra-compact-interactive-list/src/CompactInteractiveList.jsx
```diff
@@ -110,6 +110,7 @@ const propTypes = {
 
   /**
    * A zero-based index indicating which column represents the row header.
+   * Omitted if there is only one semantic column in the list, in this case the list will have no row headers.
```
sycombs, line 113: Can we reword to something like:
```suggestion
   * If there is only one semantic column in the list, the list will have no row headers and the rowHeaderIndex will be ignored.
```

sycombs (approved):

PR #2112, opened 2024-03-29
[terra-table] Adding subsections to sections

BenBoersma:
### Summary
 Adding subsections to sections which include a section header.

**What was changed:**

**Why it was changed:**

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**
Partially resolves.
UXPLATFORM-10224 

---

@cerner/terra


adoroshk (commented):
Some CSS observations 
1. Lowlight theme styling looks off to me:

Screenshot 2024-04-02 at 10 44 26 AM
2. NIT (if that is expect behavior - please disregard it), subsection header text alignment is not exactly aligned with section header (or anything else) on the left. It is placed much more to the right than the rest of the content.

Screenshot 2024-04-02 at 10 47 03 AM
Screenshot 2024-04-02 at 10 47 53 AM

packages/terra-table/src/subcomponents/SubSection.jsx
```diff
@@ -0,0 +1,90 @@
+// TODO: fix linter error
```
adoroshk, line 1: Is this TODO addressed, or is it out of scope of this PR?

packages/terra-table/src/proptypes/sectionShape.js
```diff
@@ -22,6 +23,10 @@ const sectionShape = PropTypes.shape({
    * An array of row objects to be rendered within the section.
    */
   rows: PropTypes.arrayOf(rowShape),
+  /**
+   * An array of subsections objects to be rendered within the section.
+   */
+  subsections: PropTypes.arrayOf(subsectionShape),
```
adoroshk, line 29: A documentation update is needed per subsection property added to the section shape here:
https://github.com/cerner/terra-framework/blob/cc25304e3fcaa751253ae37f6763f763566b1b44/packages/terra-framework-docs/src/terra-dev-site/doc/table/About.1.doc.mdx?plain=1#L148

packages/terra-table/src/proptypes/subsectionShape.js
```diff
@@ -0,0 +1,19 @@
+import PropTypes from 'prop-types';
+import rowShape from './rowShape';
+
+const sectionShape = PropTypes.shape({
```
adoroshk, line 4: A [documentation page](https://github.com/cerner/terra-framework/blob/cc25304e3fcaa751253ae37f6763f763566b1b44/packages/terra-framework-docs/src/terra-dev-site/doc/table/About.1.doc.mdx?plain=1#L155) needs to be updated with section shape info similar to the way we have for section

packages/terra-framework-docs/src/terra-dev-site/test/table/TableWithCollapsibleSectionsAndSubSections.test.jsx
```diff
@@ -0,0 +1,172 @@
...
+
+    if (sectionIndex > -1) {
+      newSections[sectionIndex].isCollapsed = !newSections[sectionIndex].isCollapsed;
+    }
+
+    setTableSections(newSections);
+  };
+
+  return (
+    <Table
```
adoroshk, line 163: I would add this example to user facing part of the documentation as well, so that users have access to the implementation code example.

adoroshk (commented):
To make sure we are on the same page I wanted to double-check that subsections are going to be supported on table only in not supported in Grids. If that's correct, that should be also mentioned in documentation. In case I am wrong and subsections are supported in grids, there should be grid examples and tests added.

sdadn:
> To make sure we are on the same page I wanted to double-check that subsections are going to be supported on table only in not supported in Grids. If that's correct, that should be also mentioned in documentation. In case I am wrong and subsections are supported in grids, there should be grid examples and tests added.

I believe subsections are only for flowsheet datagrids (@mjpalazzo is that correct?). Consuming these changes in flowsheet will be done in a separate PR.

packages/terra-table/src/subcomponents/SubSection.jsx
sdadn: Looks like it was from a previous PR but there doesn't seem to be any linter issues. Removed here: 86bd928341181f2152cd32ac5db908084fce26b1

mjpalazzo:
> > To make sure we are on the same page I wanted to double-check that subsections are going to be supported on table only in not supported in Grids. If that's correct, that should be also mentioned in documentation. In case I am wrong and subsections are supported in grids, there should be grid examples and tests added.
> 
> I believe subsections are only for flowsheet datagrids (@mjpalazzo is that correct?). Consuming these changes in flowsheet will be done in a separate PR.

That is correct. Flowsheet only

cm9361:
There doesn't appear to be a new or updated example to show how to implement subsections.

packages/terra-table/src/subcomponents/SubSection.jsx
```diff
@@ -0,0 +1,89 @@
...
+        >
+          <h3
+            className={cx('subsection')}
+          >
+            <span
+              className={cx('sticky')}
+            >
+              {text}
+            </span>
+          </h3>
```
cm9361, line 80: Shouldn't we be using the SectionHeader Terra component here?

packages/terra-table/src/subcomponents/SubSection.jsx
```diff
@@ -0,0 +1,89 @@
...
+        </th>
+      </tr>
+    </>
```
cm9361, line 83: Where is the content for the subsection?

packages/terra-table/src/subcomponents/Section.jsx
```diff
@@ -162,6 +169,87 @@ function Section(props) {
...
+                  displayedColumns={displayedColumns}
+                  rowHeaderIndex={rowHeaderIndex}
+                  onCellSelect={onCellSelect}
+                  isSelected={row.isSelected}
+                  isTableStriped={isTableStriped}
+                  rowMinimumHeight={rowMinimumHeight}
+                  firstRowId={firstRowId}
+                  lastRowId={lastRowId}
+                />
+              ))}
```
cm9361, line 245: The SubSection component should be rendering its own rows.

packages/terra-table/src/subcomponents/SubSection.module.scss
```diff
@@ -0,0 +1,50 @@
...
+
+:local {
+  .subsection {
+    background-color: #dedfe0;
+    font-size: var(--terra-table-subsection-font-size, 0.808571428571rem);
+    font-weight: var(--terra-table-subsection-font-weight, 700);
+    font-family: var(--terra-table-subsection-font-family, "Helvetica Neue", Helvetica, Arial, sans-serif);
+    border-bottom:  var(--terra-table-subsection-border-bottom, 1px solid #c8cacb);
+    padding-top: 0.3571428571rem;
+    padding-bottom: 0.3571428571rem;
```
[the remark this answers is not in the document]
sdadn: Added here: 29bba90d7

sdadn:
> There doesn't appear to be a new or updated example to show how to implement subsections.

@cm9361 This PR only adds subsection support for terra-table. Consuming that in flowsheet and adding documentation for consumers will be done in the next PR.

packages/terra-table/src/subcomponents/SubSection.jsx
sdadn: I believe @BenBoersma tried that but terra section header doesn't have the subsection styling guidelines provided by UX.

packages/terra-table/src/clinical-lowlight-theme/SubSection.module.scss
```diff
@@ -0,0 +1,8 @@
+:local {
+  .orion-fusion-theme {
+    --terra-table-subsection-font-size: 0.808571428571rem;
+    --terra-table-subsection-font-weight: 700;
+    --terra-table-subsection-font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
```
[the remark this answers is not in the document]
sdadn: Removed here: 6ee7c59c4

packages/terra-table/src/proptypes/subsectionShape.js
sdadn: Added here: 2395c42a838cd07a92ab4f7a69f793eea5b14996

packages/terra-table/src/subcomponents/Cell.jsx
```diff
@@ -387,6 +393,7 @@ function Cell(props) {
   const columnHeaderId = `${tableId}-${columnId}-headerCell`;
   const rowHeaderId = !isRowHeader && rowHeaderIndex !== -1 ? `${tableId}-rowheader-${rowId} ` : '';
   const sectionHeaderId = sectionId ? `${tableId}-${sectionId} ` : '';
+  const subsectionHeaderId = subsectionId ? `${tableId}-${sectionId}-${subsectionId} ` : '';
```
cm9361, line 396: Do we need the section id in the id for the subsection?

packages/terra-table/src/subcomponents/Cell.jsx
cm9361, line 1: It does not appear that the onCellSelect method has been updated to include the subsection id.

packages/terra-table/src/subcomponents/Section.jsx
```diff
@@ -162,37 +168,118 @@ function Section(props) {
...
+    lastRowId,
+    onCellSelect,
+    rowHeaderIndex,
+    rowMinimumHeight,
+    rowSelectionMode,
+    sectionId: !isHidden ? id : undefined,
+    tableId,
+  };
+
+  if (subsections) {
```
cm9361, line 217: What happens if the section has its own rows as well as subsections?  I don't believe that use case is covered in this design.

packages/terra-table/src/subcomponents/Section.jsx
```diff
@@ -162,37 +168,118 @@ function Section(props) {
...
+                  scope="col"
+                  tabIndex={isGridContext ? -1 : undefined}
+                >
+
+                  <SectionHeader
+                    className={cx('subsection')}
+                    text={subsection.text}
+                    isTitleSticky
+                    boundedWidth={boundingWidth}
+                    level={3}
```
cm9361, line 249: Seems like the value of 3 should be a constant so that we can give a meaningful name to the value.

packages/terra-table/src/subcomponents/Section.jsx
```diff
@@ -162,37 +168,118 @@ function Section(props) {
...
+                  />
+
+                </th>
+              </tr>
+            </tbody>
+            <tbody className={cx('subsection-rows', {
+              collapsed: isCollapsed,
+              collapsible: isCollapsible,
```
cm9361, line 257: Subsections are not collapsible.  We should not need classes on the subsection. The subsection visibility should be covered by the parent section.  We shouldn't even render subsections if the section is collapsed.

packages/terra-table/src/subcomponents/Section.jsx
```diff
@@ -162,37 +168,118 @@ function Section(props) {
...
+                  className={cx('header-cell')}
+                  align="left"
+                  colSpan={displayedColumns.length}
+                  role="columnheader"
+                  scope="col"
+                  tabIndex={isGridContext ? -1 : undefined}
+                >
+
+                  <SectionHeader
+                    className={cx('subsection')}
```
cm9361, line 245: Shouldn't this class be something like subsection-header-row to match the section classes?

packages/terra-table/src/subcomponents/Section.jsx
```diff
@@ -162,37 +168,118 @@ function Section(props) {
...
+                    text={subsection.text}
+                    isTitleSticky
+                    boundedWidth={boundingWidth}
+                    level={3}
+                  />
+
+                </th>
+              </tr>
+            </tbody>
+            <tbody className={cx('subsection-rows', {
```
cm9361, line 255: It seems like this would be the subsection class.

cm9361:
Doesn't the isSection method in the DataGrid component need to be updated to account for the new subsections?  If that is done, does that fix all your keyboard navigation issues?

packages/terra-table/src/subcomponents/Section.jsx
sdadn: This appends the `subsection` to the existing `header-row` class, so it would be `header-row subsection`, but I can rename it if it's preferred.

packages/terra-table/src/subcomponents/Section.jsx
sdadn: Subsections don't get rendered, as confirmed by this test. But I can remove these classes if they are redundant.

https://github.com/cerner/terra-framework/blob/92705576279e10198dbac9708a6ee55c676814cc/packages/terra-table/tests/wdio/table-spec.js#L193-L197

packages/terra-table/src/subcomponents/Section.jsx
sdadn: Added constant in a018196d8

packages/terra-table/src/subcomponents/Cell.jsx
sdadn: fixed here: a981a9069

packages/terra-table/src/subcomponents/Section.jsx
mjpalazzo: Subsections definitely can have rows. See screen shot.
image

mjpalazzo: Here is a link to a video that shows the full workflow: 
https://oracle-my.sharepoint.com/:v:/r/personal/sriraksha_k_oracle_com/Documents/Orders%20for%20Signature%20Accessibility/Routing%20Section%20issue%20in%20OFS.mov?csf=1&web=1&e=pJAyt6&nav=eyJwbGF5YmFja09wdGlvbnMiOnt9LCJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJTdHJlYW1XZWJBcHAiLCJyZWZlcnJhbE1vZGUiOiJtaXMiLCJyZWZlcnJhbFZpZXciOiJwb3N0cm9sbC1jb3B5bGluayIsInJlZmVycmFsUGxheWJhY2tTZXNzaW9uSWQiOiI4MmMxMWE2ZC02YjQ3LTQ5MTgtOTRhZC1lNzViYWU5YzUxYTUifX0%3D

mjpalazzo: The subsections are used to show the routing of medication orders. There is typically a row under the subsection for each medication that qualifies for that route; e.g., electronic prescription, pharmacy call, etc.

sdadn: Oh I meant if section level row and subsections are both simultaneously possible?

e.g.
- section header
  - section row
  -  section row
  - subsection
    - subsection row
    - subsection row 2

mjpalazzo: I will have to reach out the Clinical product manager for that question.

sdadn:
> Some CSS observations
> 1. Lowlight theme styling looks off to me:
> 2. NIT (if that is expect behavior - please disregard it), subsection header text alignment is not exactly aligned with section header (or anything else) on the left. It is placed much more to the right than the rest of the content.

@adoroshk 
1. I have corrected the lowlight styling, thanks for pointing that out!
2. The offset is by design from UX.

packages/terra-table/src/subcomponents/Section.jsx
```diff
@@ -162,37 +168,118 @@ function Section(props) {
...
+            <tbody className={cx('subsection', {
+              collapsed: isCollapsed,
+              collapsible: isCollapsible,
+            }, theme.className)}
+            >
+
+              <tr
+                aria-rowindex={subsection.subSectionRowIndex}
+                className={cx('header-row', theme.className)}
+                data-section-id={id}
```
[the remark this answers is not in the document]
sdadn: Fixed here: d1c988e17

packages/terra-table/src/subcomponents/Section.jsx
sdadn: Looks like this is not needed, removed here: 0fc369b63

packages/terra-table/src/subcomponents/Section.jsx
sdadn: Removed redundant css here: 0fc369b63

packages/terra-table/src/subcomponents/Cell.jsx
sdadn: I added it so we'll know which section the subsection is associated with. Do we need that or is the section id redundant?

sdadn:
> Doesn't the isSection method in the DataGrid component need to be updated to account for the new subsections? If that is done, does that fix all your keyboard navigation issues?

@cm9361 I haven't been able to work on the DataGrid component yet. So far, I've mainly been focusing on adding support for subsections to the Table component for this Jira & PR.

packages/terra-table/src/subcomponents/Cell.jsx
cm9361: I guess having the section helps to ensure uniqueness.

cm9361 (approved):

PR #2116, opened 2024-04-02
[terra-table] Fix action header row sticky behavior

cm9361:
### Summary

**What was changed:**
The action header row for the Terra Table was fixed so that it remained pinned to the top of the table during scrolling.

**Why it was changed:**
The fix was made so that the information was always visible and accessible to users.

### Testing

This change was tested using:

- [X] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [X] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10351 

---

@cerner/terra


packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -424,7 +424,7 @@ const ColumnHeaderCell = (props) => {
       onMouseDown={isSelectable && onColumnSelect ? handleMouseDown : undefined}
       onKeyDown={(isSelectable || isResizable) ? handleKeyDown : undefined}
           // eslint-disable-next-line react/forbid-component-props
-      style={{ width: `${width}px`, height: isActionCell ? 'auto' : headerHeight, left: cellLeftEdge }}
+      style={{ height: isActionCell ? 'auto' : headerHeight, left: cellLeftEdge, top: isActionCell ? headerHeight : undefined }}
```
adoroshk, line 427: Awesome job, very nice solution, very neat! 

One small suggestion. To avoid having `top: undefined` (see snapshots), I suggest adding `style.top` conditionally, only when it is an action cell. For that in line 405 create a variable like:

```
const cellStyle = {
    height: isActionCell ? 'auto' : headerHeight,
    left: cellLeftEdge,
  };

  if (isActionCell) {
    cellStyle.top = headerHeight;
  }
```
  Then in line 427 assign to that variable:
`  style={cellStyle}`

sycombs (approved):

packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
cm9361: I value of undefined or null will make the attribute be ignored.  Do we need the additional logic in that case?

adoroshk: Let me double check this one, but my understanding was that for attribute it will be ignored, but for css property it'l be populated with "undefined". But now I am double questioning myself, so let me check.

adoroshk: You were right, it is being disregarded. I guess I mixed it up with direct assigning css properties via js. Approving!

adoroshk (approved):

cm9361:
I received approval from @mjpalazzo for the functional review so I will merge the code.

PR #2117, opened 2024-04-02
[terra-table] Add Aria Sort to Column Header Cells.

kenk2:
### Summary

**What was changed:**
Adds Aria-Sort to the terra-table

**Why it was changed:**
So JAWS can ead the sortable table order.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10327 

---

@cerner/terra

### DIFF
--- a/packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
+++ b/packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
@@ -418,6 +418,7 @@ const ColumnHeaderCell = (props) => {
       tabIndex={isGridContext && !hasButtonElement ? -1 : undefined}
       role={!isActionCell ? 'columnheader' : undefined}
       scope={!isActionCell ? 'col' : undefined}
+      aria-sort={sortIndicator}
           // action Cell has to own a corresponding resize handle to avoid a double announcement on handle focus
       aria-owns={ownsResizeHandle ? resizeHandleId : undefined}
       title={!isActionCell ? displayName : action?.label}
--- a/packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
+++ b/packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
@@ -391,9 +391,15 @@ const ColumnHeaderCell = (props) => {
         className={cx('header-container')}
         {...hasButtonElement && { ref: columnHeaderCellRef, role: 'button' }}
         tabIndex={buttonTabIndex}
+        aria-sort={sortIndicator}
       >
         {errorIcon}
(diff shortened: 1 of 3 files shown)


packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -391,9 +391,15 @@ const ColumnHeaderCell = (props) => {
         className={cx('header-container')}
         {...hasButtonElement && { ref: columnHeaderCellRef, role: 'button' }}
         tabIndex={buttonTabIndex}
+        aria-sort={sortIndicator}
```
cm9361, line 394: This attribute is not fully supported by most of the screen readers.  Will this really resolve an issue?  What is not being covered by the hidden text?

packages/terra-table/tests/jest/__snapshots__/ColumnHeaderCell.test.jsx.snap
```diff
@@ -80,6 +80,7 @@ exports[`ColumnHeaderCell renders a pinned column header cell 1`] = `
     width={100}
   >
     <th
+      aria-sort="ascending"
```
cm9361, line 83: If we do have to add this attribute, there should be a larger impact to testing.

mjpalazzo:
@kenk2 Is this for the issue we discussed this morning? I thought we agreed to assign it back to UX for guidance.

mjpalazzo:
> @kenk2 Is this for the issue we discussed this morning? I thought we agreed to assign it back to UX for guidance.

kenk2:
@mjpalazzo That was a different issue than this one. I closed https://github.com/cerner/terra-framework/pull/2114 because that was the one related to removing the tab focus in that example.

mjpalazzo:
> @mjpalazzo That was a different issue than this one. I closed #2114 because that was the one related to removing the tab focus in that example.

@kenk2 - thanks for letting me know.

packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
kenk2: It was originally requested to add aria-sort to the column header so that it would read the sort order. However upon discovery that JAWS and most screen readers do not support this I have changed the approach to using `aria-live` instead on the sorting text.

packages/terra-table/tests/jest/__snapshots__/ColumnHeaderCell.test.jsx.snap
kenk2: Note: No longer doing this change after it was advised to not use aria-sort.

PR #2123, opened 2024-04-04
Cell auto focus property

cm9361:
### Summary

**What was changed:**
The isAutoFocusEnabled property was introduced to control whether cell with a single focusable element that is a button or hyperlink gets focus automatically during cell navigation.

**Why it was changed:**

### Testing

This change was tested using:

- [X] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-XXXX 

---

@cerner/terra


packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -149,6 +149,11 @@ const propTypes = {
    * With this property the height of the cell will grow to fit the cell content.
    */
   rowMinimumHeight: PropTypes.string,
+
+  /**
+   * Determines if focus is moved to the interactive element of a cell when a single button or hyperlink element is the only interactive element.
+   */
+  isAutoFocusEnabled: PropTypes.bool,
```
sdadn, line 156: would `enableAutofocus` make sense as a simpler name?

packages/terra-data-grid/src/DataGrid.jsx
```diff
@@ -282,7 +289,7 @@ const DataGrid = forwardRef((props, ref) => {
 
     // Set focus to a single header button or hyperlink if they are the only content in cell
     const cellButtonOrHyperlink = focusedCell.querySelector('a, button');
-    if ((isHeaderRow && !focusedCell.hasAttribute('tabindex')) || cellButtonOrHyperlink) {
+    if ((isHeaderRow && !focusedCell.hasAttribute('tabindex')) || (isAutoFocusEnabled && cellButtonOrHyperlink)) {
```
sdadn, line 292: will action buttons on header cells still be autofocused, regardless of the prop?

sdadn (commented):
We may need to update these wdio tests to account for the new prop.

https://github.com/cerner/terra-framework/blob/26f559fda7d80602307d499221b782f553c117e0/packages/terra-data-grid/tests/wdio/data-grid-spec.js#L283-L296

packages/terra-data-grid/src/DataGrid.jsx
cm9361: I tried to follow the naming pattern of some of the other props.  However, I can make this change if it is the desired update.

packages/terra-data-grid/src/DataGrid.jsx
cm9361: The first half of the condition is for the header row and all unit tests pass.  I will manually confirm though.

cm9361:
> We may need to update these wdio tests to account for the new prop.
> 
> https://github.com/cerner/terra-framework/blob/26f559fda7d80602307d499221b782f553c117e0/packages/terra-data-grid/tests/wdio/data-grid-spec.js#L283-L296

I already made this update in the test file to account for the new prop.

packages/terra-data-grid/src/DataGrid.jsx
sycombs: I think the `is___` pattern is preferred. Would a name like `isAutofocusable` or similar work? (Not sure if that's sufficiently descriptive of the prop's usage though)

sdadn: If we're keeping to the `is__` pattern, then I think both these names are similar enough that either works.

sdadn (approved):

sycombs (approved):

packages/terra-framework-docs/src/terra-dev-site/doc/data-grid/WorklistDataGrid.1/Examples.3/WorklistDataGridAutoFocusableCell.jsx
```diff
@@ -0,0 +1,68 @@
...
+      )}
+      <WorklistDataGrid
+        id="default-terra-worklist-data-grid-focusable-cell"
+        overflowColumns={cols}
+        defaultColumnWidth={170}
+        rows={rows}
+        rowHeaderIndex={rowHeaderIndex}
+        rowHeight="50px"
+        ariaLabel="Worklist Data Grid"
+      />
```
adoroshk, line 63:
```isAutoFocusEnabled
      />
```

adoroshk:
Is there an example where the action button autofocus can be tested with `isAutoFocusEnabled` prop set to false? I am not sure where to test this combination.

adoroshk (approved):

mjpalazzo:
@cm9361 - I tested 2 examples Worklist Data Grid Autofocusable Cell and Worklist Data Grid Focusable Cell. Keyboard navigation is working as required in both. JAWS announcements are working as required in both.

PR #2127, opened 2024-04-05
[terra-collapsible-menu-view] Removes dialog role for menu and sets aria-haspopup

sugan2416:
### Summary

**What was changed:**

- Sets aria-haspopup to boolean values
- Removed dialog role terra-menu

**Why it was changed:**

- aria-haspopup was set to dialog, changed to boolean.
- Dialog role causes axe a11y failure being an inappropriate role.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10322 

---

@cerner/terra

### DIFF
--- a/packages/terra-menu/src/Menu.jsx
+++ b/packages/terra-menu/src/Menu.jsx
@@ -199,6 +199,7 @@ class Menu extends React.Component {
         hookshotPostionFixed
         isHeaderDisabled
         isContentFocusDisabled
+        popupContentRole={null}
       >
         {slides}
       </Popup>
--- a/packages/terra-collapsible-menu-view/CHANGELOG.md
+++ b/packages/terra-collapsible-menu-view/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Fixed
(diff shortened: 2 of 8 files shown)


packages/terra-menu/src/Menu.jsx
```diff
@@ -199,6 +199,7 @@ class Menu extends React.Component {
         hookshotPostionFixed
         isHeaderDisabled
         isContentFocusDisabled
+        popupContentRole={null}
```
cm9361, line 202: Why do we have to set the property value to null?  Won't it be undefined just because the property is not specified?

sugan2416: Added since popup currently defaults to role as dialog: https://github.com/cerner/terra-framework/blob/main/packages/terra-popup/src/Popup.jsx#L123

allanxavierm (approved):

saket2403 (approved):

PR #2131, opened 2024-04-10
An example of css styling needed for text truncation in terra-table with auto layout

adoroshk:
### Summary
After an auto layout was introduced to `terra-table` there are questions coming from consumers as for how to truncate the text inside the cell. An example of additional css needed to be added by customers to the cell content that requires truncation was added to the documentation.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

Truncated text example:

Screenshot 2024-04-10 at 12 09 11 PM

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review


packages/terra-framework-docs/src/terra-dev-site/doc/table/Examples/AutoLayoutTable.jsx
```diff
@@ -1,6 +1,32 @@
...
+const TruncatedText = () => (
+  <div style={{
+    position: 'absolute',
+    left: 0,
+    right: 0,
+    top: 0,
+    bottom: 0,
+    display: 'flex',
+    alignItems: 'center',
+    padding: '0 7px',
```
cm9361, line 16: We don't want consumers manipulating the padding, right?  This value is not consistent across all themes.

packages/terra-framework-docs/src/terra-dev-site/doc/table/Examples/AutoLayoutTable.jsx
```diff
@@ -1,6 +1,32 @@
+/* eslint-disable react/forbid-dom-props */
 import React from 'react';
 import Table from 'terra-table';
 
+const longText = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.';
+
+const TruncatedText = () => (
+  <div style={{
```
cm9361, line 8: Shouldn't we be able to add styles on the td element directly?  Do we need consumers to create nested div containers?  It looks like the changes might impact our desired layout for the cell.

packages/terra-framework-docs/src/terra-dev-site/doc/table/Examples/AutoLayoutTable.jsx
adoroshk: I was concerned about the padding too, but this is an absolutely positioned div inside the cell div, so there will be no padding provided by table and without any padding the text will be aligned to the left (see attachment). 
Screenshot 2024-04-10 at 1 17 40 PM

packages/terra-framework-docs/src/terra-dev-site/doc/table/Examples/AutoLayoutTable.jsx
adoroshk: The way I was able to implement the truncation involves absolutely positioned wrapper, which can't be <td> as it would mess with its width and needs to be placed inside of <td>. Possibly, if some other css implementation exists, that would be possible to apply the styles to <td>, however in that case my concern with adding styles to <td> element is that consumers can override our styles for <td>, which is also not good.

packages/terra-framework-docs/src/terra-dev-site/doc/table/Examples/AutoLayoutTable.jsx
cm9361: I don't think we should give consumers this approach as they would have to properly account for themes as well.

packages/terra-framework-docs/src/terra-dev-site/doc/table/Examples/AutoLayoutTable.jsx
adoroshk: I updated the `<td>` element styling [here](https://github.com/cerner/terra-framework/pull/2131/commits/78c887c9fb73d23f947dea7abe14e3549fc86537), but please let me know if you know a better way to do that.

packages/terra-framework-docs/src/terra-dev-site/doc/table/Examples/AutoLayoutTable.jsx
adoroshk: Changed in this commit: https://github.com/cerner/terra-framework/pull/2131/commits/78c887c9fb73d23f947dea7abe14e3549fc86537

PR #2137, opened 2024-04-15
[terra-folder-tree] Increase radio button click target size

sycombs:
### Summary

This change increases the size of the radio button click target to meet [target size accessibility requirements](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html) so that users do not have to click inside a small radio button to select a row.

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10097 

---

@cerner/terra


packages/terra-folder-tree/src/subcomponents/FolderTreeItem.jsx
```diff
@@ -133,16 +133,23 @@ const FolderTreeItem = ({
     ? <IconCaretDown height="8px" width="8px" style={{ verticalAlign: 'baseline' }} /> // eslint-disable-line react/forbid-component-props
     : <IconCaretRight height="8px" width="8px" style={{ verticalAlign: 'baseline' }} />; // eslint-disable-line react/forbid-component-props
 
+  /* eslint-disable jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions, jsx-a11y/label-has-associated-control */
```
sycombs, line 136: Trying to see if there is any change to avoid disabling these. For now these rules are all being disabled because:
1. jsx-a11y/click-events-have-key-events - the onClick is only used to stop propagation and prevent expand/collapse, which is not an issue for the keyboard interactions
2. jsx-a11y/no-noninteractive-element-interactions - this is just unavoidable as long as there is an onClick on a label.
3. jsx-a11y/label-has-associated-control - not sure if a bug but in theory nesting label -> input the way is being done here should satisfy this rule.

adoroshk (approved):
I verified that click on the label area around the checkbox works the same way as the click on the checkbox. LGTM.
Screenshot 2024-04-18 at 11 24 26 AM

sdadn (approved):
Looks good!

artpark (approved):

PR #2158, opened 2024-05-08
[terra-popup] Enabled close button for examples

adavijit:
### Summary

**What was changed:**

**Why it was changed:**

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-XXXX 

---

@cerner/terra


packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -82,17 +82,22 @@ const propTypes = {
...
+  isMenu: PropTypes.bool,
 };
 
 const defaultProps = {
   classNameInner: null,
   contentHeightMax: -1,
   contentWidthMax: -1,
   isFocusedDisabled: false,
-  isHeaderDisabled: false,
+  isHeaderDisabled: true,
```
saket2403, line 96: Default value should be false only. Since prop name is isHeaderDisabled it should remove the header when true but as per your implementation it will add header when true

adavijit: Changed it to false

packages/terra-pills/src/subcomponents/_Pill.jsx
```diff
@@ -52,6 +52,15 @@ const propTypes = {
...
   title: PropTypes.string,
+
+  /**
+   * Adds a header with close button to the popup.
+   */
+  isPopupHeaderDisabled: PropTypes.boolean,
+};
+
+const defaultProps = {
+  isPopupHeaderDisabled: true,
```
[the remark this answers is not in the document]
adavijit: Removed this prop since Popup by default provides the header with close button

saket2403 (approved):

sugan2416 (approved):

PR #2166, opened 2024-05-16
[terra-navigation-side-menu] A11y changes

MadanKumarGovindaswamy:
### Summary

**What was changed:**

-- Added headerLevel prop for the title of the menu.
-- Added missing A11y attributes.

**Why it was changed:**

Part of A11y changes.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10270 

---

@cerner/terra


packages/terra-navigation-side-menu/src/_MenuItem.jsx
```diff
@@ -137,6 +137,14 @@ class MenuItem extends React.Component {
...
     );
 
+    const ariaAttributes = {};
+    ariaAttributes['aria-haspopup'] = hasChevron;
+    ariaAttributes['aria-disabled'] = isDisabled;
+
+    if (hasChevron) {
+      ariaAttributes['aria-expanded'] = !hasChevron;
+    }
+
```
saket2403, line 147: Why do we need if condition here? If it is to check undefined / null then we should have default values declared for all rops.

MadanKumarGovindaswamy: we don't want add aria-expanded for all the list items, so only when it has submenu then append this attribute. like if we don't add this condition then SR reads as collapsed for all list items even the item does not have submenu.

packages/terra-navigation-side-menu/translations/es-ES.json
```diff
@@ -1,3 +1,5 @@
 {
-  "Terra.navigation.side.menu.selected": "Seleccionado"
+  "Terra.navigation.side.menu.selected": "Seleccionado",
+  "Terra.navigation.side.menu.folder": "Folder",
+  "Terra.navigation.side.menu.document": "Document"
```
saket2403, line 4: I think you should log translations

packages/terra-navigation-side-menu/src/NavigationSideMenu.module.scss
```diff
@@ -84,4 +84,15 @@
     width: 100%;
     word-wrap: break-word; /* For IE 10 and IE 11 */
   }
+
+  .drill-in-title {
```
sugan2416, line 88: Can we keep the common styles between title and drill-in-title in .title and conditionally add-in the differences in .drill-in-title

MadanKumarGovindaswamy: Updated. Thanks

packages/terra-navigation-side-menu/translations/es-ES.json
MadanKumarGovindaswamy: yeah we need to log, will check with UX team once on this. Thanks

saket2403 (approved):

sugan2416 (approved):

rbsree:
+1, for accessibility review for Terra Navigation Side Menu - Drillin example.

